PR #110, opened 2017-02-17
Python travis integration and update python unit tests

fjaviersanchez:
Now after you install the C libaries and you do `python setup.py install` you can test directly the python module typing `python setup.py test`. If we want to use python in Travis CI  to a major modification of `setup.py` is required to install the C libraries and the python package when you do `python setup.py install`


damonge:
@fjaviersanchez Two things that would be great to do:
 - Add instructions regarding the new installation method on the README (and the note if relevant). Some documentation on the use of travis would also be great on CONTRIBUTE.md
 - We should make sure pip install works on a variety of platforms. I'm not sure how many it's been tested on. I'll test on my own (ubuntu) today.

PR #194, opened 2017-06-13
Cosmic emulator

elisachisari:
This branch incorporates the cosmic emulator from Lawrence et al. (2017), including the license. Some checks are pending so this is not yet ready to merge.


elisachisari:
The cosmic emulator probes the range 0<z<2. We might be able to reach z>2 with the current emulator if the emulator accuracy matches our requirement. We need to define what this requirement is. Tagging @elikrause.

tmcclintock:
Some notes:
* CLASS and the emulator disagree at the 5% level around k=1e-4 and k=1, but agree much better between there (see attached plot).
* The emulator behaved fine within z=0-2.
* The error message about the emulator not working beyond z=2 (e.g. at z=3) doesn't get passed correctly from C into Python. This might be symptomatic of a larger issue about error messages.
* I don't know how to deal with the conflicts that github is saying we need to resolve. I'll leave that to @elisachisari .

I approve this PR.

cosmicemu_test

tmcclintock:
Tagging @elikrause so it's easy to find.

tmcclintock:
The documentation in ```src/ccl_emu17.c``` is from Earl Lawrence who originally wrote the code. It does not match doxygen format and needs to be improved, but this is low priority since that file is never supposed to be accessed by users.

katrinheitmann:
Hi Tom,

Is this test for LCDM? Then the result is really surprising, in our 
paper we showed very good agreement with e.g. Takahashi et al at those 
scales. We do use CAMB not CLASS but I would guess that can't explain 
the difference. One question: what are your units? h or no h? (The 
emulator has no h, of course that would show up much more than 5% if 
that was the issue, but still wanted to double check).

For a model that involves wa, there are differences in how we 
implemented it compared to what other groups did, so I could understand 
that.

Thanks! Katrin

tmcclintock:
@katrinheitmann Here is the cosmology I used to generate the plot. I used them because they are used elsewhere in some unit tests

```python
Omega_c=0.27
Omega_b=0.022/0.67**2
h=0.67
sigma8=0.8
n_s=0.96
N_nu_rel=3.04
N_nu_mass=0.
m_nu=0.
```

katrinheitmann:
ok, thanks! I'll look into it (but have a deadline today so will not be 
able to do this immediately.)

elisachisari:
@katrinheitmann: the cosmic emulator implementation has been validated against the smoothed power spectra that Earl sent us. You might see deviations at small and large k due to our own implementation of splines and/or extrapolation. We are planning to revisit those in a separate PR, but the checks against the smoothed P(k) from the simulations are automated within CCL and pass without problems.

katrinheitmann:
Is it easy for you to send me the CLASS prediction output for the 
nonlinear power spectrum? I know I should use CCL to generate it etc. 
but since I am really under time pressure right now, that would help a 
lot. I have not used CLASS at all in the past, what did it implement for 
the nonlinear prediction? I did a lot of tests with CAMB-Takahashi and 
CAMB-Mead and that all looked good ...

katrinheitmann:
Ah, thanks! So I should not (yet) worry then?

tmcclintock:
@katrinheitmann here is a file with k and P(k) from CLASS. Units are (I believe h/Mpc, whatever the default units from CCL are, @elisachisari would know).

[k_and_Pclass.txt](https://github.com/LSSTDESC/CCL/files/1422860/k_and_Pclass.txt)

elisachisari:
if you got this straight from CCL, our units don't have h's:
https://github.com/LSSTDESC/CCL/wiki/CCL-units

elisachisari:
@katrinheitmann I wouldn't worry for now about the low/high-k behavior, but to be safe I will carry out some additional comparisons between the emulator and CLASS within the CCL implementation and varying spline parameters.

elikrause:
The code looks good to me; I checked the error handling in C and changes in ccl_power.c in detail and approve.
Regarding the differences with CLASS at high k: CLASS takes A_s as input not sigma_8, so CCL iterativly determines A_s to match the requested sigma_8 (in ccl_power.c:ccl_get_class_As()). It may be worth a try to iterate once more for this comparison. It's hard to imagine though that this would help at low k.

katrinheitmann:
So the low-k part is not provided by the emulator. The emulator stops at 
k~10^-3. I suppose the reason why you extrapolate is that you want 
everything out to 10^-4? We can do that in the next generation of 
emulator --

I assume something in the extrapolation is not working. The kink looks 
really weird at the cross-over.

salmanhabib:
I think it's relatively easy to go to lower k without the kinks (that's 
probably coming from the spline fit or a different code being used to 
get to lower k). At low k one knows the asymptotics, so that's much 
better than using blind extrapolation. As Katrin say, we can extend the 
emulator to do this.

elisachisari:
Thanks @katrinheitmann @salmanhabib. Let's check the extrapolation routines and ranges within CCL first - we knew this needed to be done and I was waiting for other two PRs (#254 and #255) to be completed before diving into this check. We'll report back on this soon.

c-d-leonard:
Hi @elisachisari. I've had a look at the branch now too and here's what I find:
- I have no problem compiling in C or in python; both make check and run_tests.py all pass fine.
- If I run Tom's script to produce the above plots, I get the same thing he does - considerable differences at low and high k, as well as an error that isn't very informative when trying to call at a scale factor below a = 1/3.
- If I try to run with N_nu_rel = 3.046 rather than 3.04, I get a Segmentation Fault. I assume this is a restriction of the emulator, but since 3.046 is our default value we should make sure if someone calls with this the error message is clear about what's going on.
- I actually also get a Segmentation Fault with N_nu_rel = 3.04, N_nu_mass = 1 and m_nu = 0.06. 
In all the above cases I've called CCL through python and attempted to calculate the power spectrum using a modified version of the same script at Tom used.

elisachisari:
@c-d-leonard @tmcclintock Regarding the error message when calling the emulator at z>2: This is implemented in ccl_power.c:
screen shot 2017-11-01 at 15 54 31
but apparently it is not being passed to pyccl. I'm not sure how to fix that, maybe @philbull can suggest a solution to this.
In C, the message is passed correctly:
screen shot 2017-11-01 at 16 00 52

elisachisari:
@c-d-leonard Regarding the failure when you call the emulator with neutrinos. I made it exit for neutrino cosmologies because I haven't implemented automated checks for these. In C, you would be getting this error:
screen shot 2017-11-01 at 16 00 24
Because the errors are not being passed to python (pinging @philbull for help here again), you get a seg fault.
I'm looking into what I need to implement the missing tests.

c-d-leonard:
Hi @elisachisari okay cool, hopefully we can get those error messages passing to python.
Another thing, just to clarify - does the emulator require Neff = 3.04 (rather than 3.046) exactly? How should we deal with the fact that 3.046 is our default value?

elisachisari:
@c-d-leonard My solution was to let the user know that they need to pass 3.04 in an error message, but other suggestions are welcome. So basically you cannot run CCL with the defaults in the emulator case.
Regarding the python errors, I think this is fixed in master now from a PR from @philbull in the last few days. So let me try merging master first and we might get a solution for that already.

damonge:
Eventually we need to make Neff a parameter, right? (e.g. whenever we want to make CCL useful beyond LSST). Would this be a good time to do that? Definitely a topic for a different PR though

c-d-leonard:
@damonge We do have an Neff parameter - it's just if the user doesn't pass a value for that it defaults to 3.046, which I think is confusing if you try to leave the default and the emulator fails. Sorry for the confusion!

c-d-leonard:
@elisachisari Okay, that solution seems fine to me as long as it's well-documented.

elisachisari:
@c-d-leonard After merging master, there are some improvements in catching errors, but I still get seg faults. So, for example, the z>2 error message from `ccl_linear_matter_power()` is now propagated to python. But it seems that the errors I'm raising within `ccl_cosmology_compute_power_emu()` are not being passed correctly. Maybe @philbull can help with this?

philbull:
I'm debugging this now. The problem is to do with catching uninitialised data structures.

When the Python wrapper is running, we make errors non-fatal so that they don't kill the Python interpreter. The wrapper is then supposed to check the error status and raise a Python-friendly exception, without quitting.

The problem is, some places in the new cosmicemu code assume that errors are fatal. If the initial call to CLASS fails because the parameters are out of bounds (e.g. because N_eff != 3.04), some data structures are left uninitialised. These aren't checked for initialisation though (it's assumed that the code would have already failed if initialisation had failed), so uninitialised variables get passed to a GSL interpolation function. Then you get a segfault.

I've pushed a couple of changes to prevent this from happening. One explicitly checks that the array being passed to GSL is initialised (which we should always do!), and the other does the cosmicemu validity check earlier in the function, so an error is raised before any lengthy computations are done. This should fix the problem.

There is another issue though, which is `exit()` statements in `ccl_emu17.c`. These will cause the Python interpreter to quit, so should be replaced. I'll try to do that later today.

philbull:
I think the emulator error handling etc. should all be fixed up now, but there's quite a big conflict with the baryon code that landed recently in `master`. I'm working on it.

elisachisari:
@philbull Yes, this conflict is expected. If you prefer, since I'm familiar with the BCM changes, I can take care of it. Just let me know.

philbull:
@elisachisari I'm mostly done with it now, but will ping you if I run into any trouble.

philbull:
OK, thanks for your patience. I did the merge, which was reasonably involved, but all of the tests are coming out OK now. Note that I made it so that the baryonic corrections can be applied to the halofit and cosmicemu nonlinear power spectra.

I probably changed enough things that someone else should review what I did, but this PR is otherwise good to go as far as I'm concerned.

c-d-leonard:
Hi @elisachisari and @philbull 

The problems with segmentation faults which were occurring when calling through python are now resolved; the test code I was using runs correctly now.

I've looked through the code changes and everything looks reasonable to me. I haven't rigorously tested every possible edge case of usage, but in general this looks fine and I'm happy for this to be merged. I will approve changes.

philbull:
I will resolve the conflicts and then merge later today.

philbull:
I merged `master` into this to solve some conflicts, but now I'm getting several tests failing (both in C and Python). I dug into it a little, and they seem to be related to the new CMB tracer. @damonge could you take a look at this please?

elisachisari:
I'm getting the following errors. It's strange that it is the baryons test and the correlation tests that are affected. It is possible this has something to do with spline parameters having changed, but I think there has to be something else that went wrong.
screen shot 2017-11-29 at 9 46 09 am

philbull:
I'm going through the diff now. One thing that did change is the minimum scale factor for the splines in `include/ccl_params.ini`:
```
-A_SPLINE_MIN=0.1
+A_SPLINE_MIN=0.0003
+A_SPLINE_MIN_PK=0.01
```
Could it be this?

elisachisari:
@philbull @damonge I think it can't be the spline parameters alone. CMBlensing has already been merged into master and the checks passed at that stage. So the fact that the checks are failing now means that something else has gone wrong. The cosmicemu branch doesn't change ccl_cls though, so I'm not sure where the problem is coming from.

elisachisari:
In this branch, I changed A_SPLINE_MIN_PK and A_SPLINE_MIN to their original value - the error persists and it has to do with an error when integrating the Cls needed to compute the correlation functions. The original Cls check passes, it is only when we try to integrate them to get correlation functions that the problem arises:
`ccl_cls.c: ccl_angular_cl(): error integrating over k`

philbull:
It's possible that the merge of the baryons stuff screwed something up. I'll go through the diff line by line to see if there are any likely suspects...

philbull:
I went back and re-did the merge. This solved a lot of problems, and now all of the C tests pass. One of the Python tests (related to the CMB Cls) fails though. Given that this required reverting several commits, I put this in a new branch, `cosmicemu_pb`.

elisachisari:
Tanks @philbull. I've made the branch visible through a PR. I'll compare commit histories later today and ping @dmonge about the CMB lensing check.

damonge:
@elisachisari can this PR be deleted, since there's a new CosmicEmu branch? (#296)

elisachisari:
Most likely - just waiting for the new branch to be merged. I'll close after that is in.

damonge:
OK, we'll ping you on slack regarding this.

elisachisari:
Closing this PR, as it is superseded by #296

PR #222, opened 2017-07-07
Updated auxiliary files of the note

elisachisari:
These files are auxiliary and should be updated in master.


elisachisari:
I am merging this branch despite travis-CI complaining. This is because it hasn't yet been fully configured.

PR #231, opened 2017-07-12
Add License Terms

brianv0:
Add license terms of libraries used by CCL, including GSL, FFTW, and FFTLog. Include a note which disclaims that CCL code is licensed as BSD, but de-facto dually-licensed as GPLv3 due to inclusion of GSL and FFTW.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -27,5 +27,51 @@ SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
 CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-Contact GitHub API Training Shop Blog About
-© 2017 GitHub, Inc. Terms Privacy Security Status Help
+
+
+
+GPLv3 Disclaimer
+As CCL uses GSL and FFTW internally, CCL is dually licensed under GPLv3. You
+can find the GPL one the GNU web site (https://www.gnu.org/licenses/gpl-3.0.html)
+
+Any derivative work which is redistributed that includes GSL and/or FFTW must also 
+be redistributed under the terms of GPLv3. Derivative work which does not
+include those libraries listed under the GPL licenses sections are not subject to
+the terms of GPLv3.
+
+
+
+============================================================================
+   CCL Subcomponents:
+
+   The CCL project contains subcomponents with separate copyright notices 
+   and license terms. Your use of the source code for the these
+   subcomponents is subject to the terms and conditions of the following
+   licenses.
+
+
+========================================================================
+GPL licenses
+========================================================================
+
+The following components are provided under the MIT License. See project 
+link for details. 
+
+    (GPLv3 License) GSL (https://www.gnu.org/software/gsl/)
+    (GPLv2 License) FFTW (http://www.fftw.org/)
+
+License information:
+
+    GPLv2: https://www.gnu.org/licenses/gpl-2.0.html
+    GPLv3: https://www.gnu.org/licenses/gpl-3.0.html
+
+
+========================================================================
+CC licenses
+========================================================================
+
+    (CC0 1.0 License) FFTLog (http://casa.colorado.edu/~ajsh/FFTLog/)
+
+License information:
+
+    CC0: https://creativecommons.org/publicdomain/zero/1.0/


villarrealas:
Should we note the citation clause on CLASS?

jablazek:
CCL does not actually include FFTW or GSL, which must be separately installed. FFTLog and CLASS are included.

villarrealas:
This license addresses situations such as distribution of a Docker image, which would necessarily include FFTW and GSL, rather than have it as mere aggregation.

brianv0:
For CLASS, there is no license, but there is a term which says that papers must cite CLASS:
https://github.com/lesgourg/class_public

> You can use CLASS freely, provided that in your publications, you cite at least the paper CLASS II: Approximation schemes <http://arxiv.org/abs/1104.2933>. Feel free to cite more CLASS papers!

I think we should add that to the LICENSE file, but I'm not real sure what that actually means (technically, it's a restriction on the use of the software). Maybe we should add a CITING file?

villarrealas:
Maybe best to send them an email? I worry that as written it would imply all users of CCL should cite CLASS, which we would want to be explicit about.

elisachisari:
@brianv0 @villarrealas Are you looking for a reviewer for this branch?

villarrealas:
So: I did some further digging and found something that might cover the case of CLASS and other codes that request citation, in GNU GPL v3:

"Notwithstanding any other provision of this License, for material you add to a covered work, you may (if authorized by the copyright holders of that material) supplement the terms of this License with terms:
[...]
b) Requiring preservation of specified reasonable legal notices or author attributions in that material or in the Appropriate Legal Notices displayed by works containing it"

I am not a lawyer, but my reading of this is that if we consider adding CLASS to our GPL compatible framework, we are allowed to preserve a notice for author attributions. It also implies that author attributions do NOT count as an additional restriction.

villarrealas:
And I'm less sure of things after finding a FAQ on the matter:

"Does the GPL allow me to add terms that would require citation or acknowledgment in research papers which use the GPL-covered software or its output?

No, this is not permitted under the terms of the GPL. While we recognize that proper citation is an important part of academic publications, citation cannot be added as an additional requirement to the GPL. Requiring citation in research papers which made use of GPL'd software goes beyond what would be an acceptable additional requirement under section 7(b) of GPLv3, and therefore would be considered an additional restriction under Section 7 of the GPL. And copyright law does not allow you to place such a requirement on the output of software, regardless of whether it is licensed under the terms of the GPL or some other license."

elisachisari:
Here is another potential clash. The testing framework we are using is released under its own license. This found in the ctest.h file under the tests directory. In particular, it says:
"Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with the License. You may obtain a copy of the License at http://www.apache.org/licenses/LICENSE-2.0". Perhaps @joezuntz has ran into this before?

joezuntz:
Apache license is apparently compatible with the GPL v3, based on the FSF web page.  So is that all we need?

jchiang87:
Hi All,

Following up on a question from @villarrealas  about CCL's licensing issues, I inquired with one of SLAC's lawyers who has intellectual property expertise, Michael Willardson.  After considering the issues, Michael was able to provide the following clarifications:

* From the DOE and SLAC's perspective, the public distribution of binary images of CCL code linked against the FFTW and GSL libraries does make CCL a derivative work which also then has to be licensed under GPL.  However, the CLASS citation requirement does conflict with the GPL terms, so if CCL continues to use CLASS, it cannot be GPL-licensed and cannot, in particular, publicly distribute binary copies linked against FFTW and GSL. (Cases like providing a script like a dockerfile to create an image are ok.)

* SLAC (and the DOE more generally) prefers all open source licensing for DOE funded efforts to be BSD-3-Clause; and though there is no stated policy, such software should not be GPL licensed if at all possible, since the GPL flow-down stipulations for derivative works makes the software less attractive for outside developers to use.  Note that the current LICENSE file text in the master branch is consistent with BSD-3-Clause.

So given these two points, I would recommend rejecting the proposed changes in this pull-request and limit distribution of CCL binary images (e.g., docker images) to be internal within the collaboration.

From my discussions with Michael, a couple more licensing issues arose that need to be addressed:

* The CLASS library has no explicit licensing text in their repository.  That means that its authors retain all copyright privileges, and so in addition to adding their citation clause as an explicit requirement for using CCL, we should also obtain explicit permission from them to copy and modify their code in our repository.

* The FFTLog code is not available under the Creative Commons license.  The https://github.com/prisae/fftlog repository is CC0, but that only pertains to the **additions** to FFTLog in that project.  So we need to request explicit permission from Andrew Hamilton to copy (and modify) his code.

Finally, regarding the CLASS library and any other software not written originally for DESC, I think it would be better to treat all of these third-party codes as external dependencies rather than copying the code directly into a DESC repository.  This helps avoid the copyright and licensing issues; and in the case of CLASS, it makes it possible for someone to drop in their own interface-compatible library and thereby avoid the citation requirement.

elisachisari:
@jchiang87: thanks for the detailed information on the licensing issues! 
@damonge: would you be up for contacting the CLASS authors?
@villarrealas: where is our current docker image hosted? should we move this to confluence so that it is only available internally? 
I can contact Andrew Hamilton regarding FFTlog.
For now, this should let us comply with the requirements listed by @jchiang87 above, but we should discuss longer-term plans.

elisachisari:
We have approval from Andrew Hamilton regarding FFTlog.

elisachisari:
The docker image is not currently distributed with CCL, so we are ok in that regard. The user can build their own docker image if they wish following our instructions.

elisachisari:
The clause to cite CLASS has been added to the README and the note in a separate PR and merged.

elisachisari:
@brianv0 @villarrealas @jchiang87 I except this PR will be closed once @damonge finishes the automated CLASS integration. Could I ask one of you to summarise in a few bullet points at least the status of the CCL license? I think this should be added to our wiki: https://github.com/LSSTDESC/CCL/wiki and we should revisit the question of whether we need to make updates to the license in the repository.

jchiang87:
Sure, I can do that.  I'd like to understand what @damonge 's changes are intended to do, however.  I see the PR, #295, but there are a lot of files being changed and I can't easily tell what's being done.  Could some post a summary comment at that PR please?

damonge:
@jchiang87 : done. The process to install class automatically is now described in the README.md of that branch.

elisachisari:
This enhancement deserves several smiley faces.

elisachisari:
@jchiang87 Could you add this summary in the wiki over the next few days? @damonge's branch will be merged shortly.

jchiang87:
@elisachisari Sure, I'll do it this week.

jchiang87:
I've updated the wiki with a summary:
https://github.com/LSSTDESC/CCL/wiki/Licensing-and-Code-Distribution
I also note that the #295 PR still mentions a built-in version of CLASS,
https://github.com/LSSTDESC/CCL/blob/class_installation_script/README.md#compiling-against-an-external-version-of-class
which I'm assuming should be updated in that PR.  I'll post a comment at that issue.

elisachisari:
CLASS authors have given us their approval after adopting the new procedure to install it. From what I see in https://github.com/LSSTDESC/CCL/wiki/Licensing-and-Code-Distribution, we don’t need to change anything there. @jchiang87 do we need to change anything in https://github.com/LSSTDESC/CCL/blob/master/LICENSE?

jchiang87:
I'd like to make a small tweak to the formatting of the copyright statement to enable GitHub's license detection code to recognize that the LICENSE file is BSD-3-clause.   If there are no objections, I'll do that on this branch (also removing the GPL stuff).  Other than that small change wrt master, I don't think we need anything else, particularly with regard to the CLASS situation.

elisachisari:
Yes, please go ahead. It might be worth opening a new PR because this one is very far behind master, and some other changes to the LICENSE were made here. Should those be discarded?

jchiang87:
Sure, good idea.  I'll make a new PR.  I guess we can just close this PR and discard the branch.

elisachisari:
Superseded by pr #310 and documented in the wiki. I'm closing this PR.

PR #237, opened 2017-07-13
Add transverse comoving distance to the global namespace

philbull:
Small tweaks to the global (Python) namespace to make more functions available.


philbull:
Looks like this is already fixed. Closing the pull request.

PR #240, opened 2017-07-14
Redefine the N_K parameter per log10 bin

wagoner47:
I switched the definition of nk in all the places I could find it (I found 3 instances) in src/ccl_power.c to be the number of decades times N_K now. I used ceil to safely cast it to an integer in the case of non-integer number of decades. I also updated the default value of N_K in the parameter file so that it would give as close to 1000 bins as possible for the default kmin and kmax.


elisachisari:
All checks pass and in principle looks good. I identified a 1e-3 difference of the output power spectra at high-h. Here is an example. We need to check more systematically whether varying N_K per bin gives different results at high k.
screen shot 2017-07-14 at 7 40 36 pm

tmcclintock:
@elisachisari you can review this very quickly and merge it if you want.

elisachisari:
@tmcclintock I looked at this some time ago. The problem is that we are not passing the 1e-4 requirement with this implementation. This is part of a bigger problem of convergence of the splines and extrapolation scheme at high-k, however. I'm happy for this to be merged,  bearing in mind that the convergence issue needs to be revisited in a wider context (issue #258). I merged master into this branch and all checks pass. The plots in the note discussing spline convergence will become outdated after this PR is merged.

Issue #78, opened 2017-02-11
Use PP constants


elisachisari:
We've noticed differences at >1e-4 between PP constants (currently implemented in CCL), CLASS constants and GSL constants. (G is one example.) We have agreed the best is to use PP constants and we need to contact the CLASS developers to ask them to incorporate these as well if possible.

matthewkirby:
I ran comparisons between the constants in CCL and some other sources. The results are attached.

<img width="347" alt="screen shot 2017-07-14 at 1 27 55 pm" src="https://user-images.githubusercontent.com/5395125/28223235-a48ad32c-6898-11e7-847e-5318cba67280.png">

I submitted a PR where Msun is fixed to be the GSL value and rho_crit is derived.

matthewkirby:
#

tmcclintock:
This is a great analysis, @matthewkirby. Thanks for doing this.

tmcclintock:
By the way, @matthewkirby make sure you cross-link the issue like so: #241 when you issue the PR, so it's slightly easier to locate the issue.

PR #241, opened 2017-07-14
Updating Particle Physics constants - Issue #78

matthewkirby:
Checked consistency of CCL constants with other software being used and outside sources. Msun was updated to be consistent with GSL and rho_crit was changed to be derived from GNEWT.


matthewkirby:
I got swig running and regenerated the wrapper. 
Checks are now passing.

elisachisari:
hi @matthewkirby. C tests pass but I had a problem with the python test. When I run run_tests.py, I got an error: "cannot import SOLAR_MASS". I'm not sure where this is coming from, but it might be that we are missing "include gsl" somewhere linked by the python wrapper and this could be why the constant cannot be found.

@tmcclintock: There is no Sigma_CRIT in CCL yet.

elisachisari:
@matthewkirby Have you checked on the issue mentioned above with the "SOLAR_MASS" parameter?

matthewkirby:
@elisachisari I took a quick look but I was unable to reproduce the error.

elisachisari:
I'm still having this problem though. Can you try installing from scratch? I.e., go to the CCL directory and then remove the "build/" directory. Then re-install CCL and its python version (run the python install command twice in a row). 
screen shot 2017-08-16 at 11 44 26

matthewkirby:
Alright, I am now able to get the error. I will play around with this and get it working.

elisachisari:
@matthewkirby: @damonge fixed the problem with the python installation. The checks now pass both in C and python. Before we merge this, could you edit the CCL note indicating which constants we are using and why. I remember you had a table with relative differences which was very useful, which I think should go into a new section of the note. You can edit directly in this branch.

damonge:
It'd also be good to merge master into this branch to check that the new setup.py also works here

matthewkirby:
Sorry for all of commits that got displayed again. I wanted to verify that my changes still worked with master so I
1. Pulled master
2. Rebased PPconstants with master
3. Rebuilt and ran the C and python tests (they all passed)
4. Pulled PPconstants, then pushed PPconstants and it put all of those commits there

Was this not the correct process? 
As for the note, I will take care of that this afternoon. Sorry again fro the delays.

elisachisari:
Thanks @matthewkirby. Usually we merge master onto the branch instead of rebasing. One can adopt different options really: https://www.atlassian.com/git/tutorials/merging-vs-rebasing But our policy has been to do `git merge master` to bring the branch up to date with master and fix any conflicts before the final merge with master. It seems your option worked too, although the commit history is a bit strange (`https://github.com/LSSTDESC/CCL/pull/241/commits`, some commits appear twice in the history?). On the other hand, there don't seem to be any conflicts and the tests pass.

tmcclintock:
@matthewkirby @elisachisari I see that there are conflicts here, but they are on SWIG files. I am not familiar with how swig works at all, but I believe they can just be rebuilt. Is this true, Elisa?

elisachisari:
@philbull @damonge I'm having some problems merging with master. After `git merge master`, I ran swig to update the files in conflict and it gives me an error like this:
`pyccl/ccl_constants.i:17: Error: Unable to find 'gsl/gsl_const_mksa.h'`
I committed the merge but the python implementation is currently not working. Would one of you be willing to take a look at this?

damonge:
@elisachisari : alright, you need to include -I<path_to_gsl> inside the swig command if GSL is in a non-standard path. I'll add a note to this in CONTRIBUTE.md

I've checked after doing this everything works for me

PR #244, opened 2017-08-07
MCMC tutorial

hsnee:
Added an MCMC jupyter notebook to the tests.


elisachisari:
I'm having trouble importing seaborn. I'm not sure this is my problem or something generic. I get the following warning message:
screen shot 2017-08-15 at 16 03 16

elisachisari:
It would be good to make the ticks and labels larger here:
screen shot 2017-08-15 at 16 29 16

elisachisari:
I think it would be good to clarify that the "data" you are using is the shear auto-spectrum without tomography (one big redshift bin) and the range of ells you are using.
screen shot 2017-08-15 at 16 45 41

elisachisari:
I would consider adding some baseline shape noise in the covariance. It would make this more realistic. The relevant numbers are in [confluence](https://confluence.slac.stanford.edu/pages/viewpage.action?pageId=211788509).

elisachisari:
I get an error at the following point, there seems to be a problem with the definition of icov. Do you know where this might be coming from?
screen shot 2017-08-15 at 17 01 53

hsnee:
> I'm having trouble importing seaborn. I'm not sure this is my problem or something generic. I get the following warning message: 

It seems like you're using an old version of ipython (4.0) the latest stable version is 6.1. Could you try to update it and try again?

hsnee:
> I'm having trouble importing seaborn. I'm not sure this is my problem or something generic. I get the following warning message

It seems you have an old version of ipywidgets
Can you try `pip install ipywidgets` and run again?

hsnee:
> It would be good to make the ticks and labels larger here:

Things will look different once seaborn is imported, can you update this when that works? (you might still have the same comments...)

elisachisari:
Have you tried changing the burn-in requirement? Is 50 sufficient?

hsnee:
> I would consider adding some baseline shape noise in the covariance. It would make this more realistic. The relevant numbers are in confluence.
I actually used the covariance from there previously, but is it all okay to make public? I thought it might be safer to use data that is not super meaningful yet good enough as a proof of concept throughout.

WeikangLin:
@hsnee @elisachisari @mishakb I got stuck when running following:
```
In [ * ] :  sampler = emcee. Ensemblesampler( nwalkers, ndim, Inprob) 
Run 50 steps as a burn-in 
```
It seems not getting through. I tried the notebook both on my Mac and on NERSC. Both stop at this step. I checked that emcee has been installed on my Mac and on NERSC. And ipywidgets is already installed in my Mac.

If I stop and run it, that step will get through, but it will get stuck again in the next step. The first step of importing the modules run successfully. Am I missing anything?

hsnee:
Hi @WeikangLin, How exactly does it get stuck? Does it just take a large amount of time to run, or does the kernel die? I would expect that step to take ~2 hours on a NERSC node so if it's just taking a lot of time, that's normal. If the kernel dies, though, the RAM is most likely not enough to run it.

WeikangLin:
Hi @hsnee, thanks! It keeps running, and the kernel is alive. So, it maybe just taking a long time. I will let it run for a while, and let you know how it goes.

WeikangLin:
Hi @hsnee , After running the notebook on NERSC for~ 2 hours, the kernel dies.

fjaviersanchez:
Hi @hsnee @WeikangLin @elisachisari I have been testing the notebook at NERSC and with the default transfer function (CLASS) the kernel dies after a while (I don't know exactly why yet but I think it's because I am using an old version of the master branch). In any case, I have a couple of suggestions:

1) I would change the transfer function type to `'eisenstein_hu'`. This way the notebook runs faster (and it doesn't die).
2) I think it would be useful to add one line with the timing for the function `lnprob`, e.g., you can add a cell with:

`%time lnprob(p0[0,:])`

This will give you the time that it takes to evaluate the likelihood for one iteration. You can also do `%timeit lnprob(p0[0,:])` and it will evaluate several iterations and give you the best/worst cases and the average time.

WeikangLin:
Hi @hsnee @fjaviersanchez @elisachisari, I changed the transfer function type from "boltzmann" to "bbks" (as @fjaviersanchez suggested, but I used bbks instead), and the notebook finishes in ~20 mins. So, it is probably not enough RAM as @hsnee suggested? 

If we copy the notebook to a python script, and submit a job to NERSC, would it help? I currently don't know how to submit a job on NERSC yet. Any tutorial about it?

fjaviersanchez:
@WeikangLin It will definitely help. You can request one or more nodes with 128 GB each and I don't think that will run out of memory. You can find some examples on how to send jobs to NERSC here: http://www.nersc.gov/users/computational-systems/cori/running-jobs/batch-jobs/

Also about the RAM, I think that might be true. All the jupyter jobs at NERSC run in certain node (that is shared with a lot of users) and running out of RAM is a possibility. I sometimes run the same program at night Pacific Time with no problem and during business hours it says memory error...

hsnee:
making a .py file out of it and running it should work, but it kind of goes against the fact that this is supposed to be a tutorial rather than a script (but at least for testing it is helpful). You can also get an allocation for an interactive job on NERSC http://www.nersc.gov/users/computational-systems/cori/running-jobs/interactive-jobs/, which you can then use to run the jupyter notebook (this is how I've run the tutorial in the past -- as @fjaviersanchez said running it on the head node might be troublesome memory-wise if a lot of other people are using a lot of memory at the same time.) 

also thanks for the suggestions on using a different transfer function and %time! I'll edit the notebook to include them.

WeikangLin:
Thanks @hsnee and @fjaviersanchez for the links.

elisachisari:
@hsnee @fjaviersanchez @WeikangLin It looks like one option concerning this PR is to use BBKS in the notebook for now and address the memory leak in a separate issue. Are you happy with this plan? Would this be ready to merge then?

fjaviersanchez:
Personally, I am happy with this plan and I think that this is ready to merge once these changes are addressed.

hsnee:
I just pushed the notebook to the mcmc branch with the suggestions edited in -- do you think it's ready?

fjaviersanchez:
@elisachisari I think this is ready to merge. @WeikangLin what do you think?

WeikangLin:
@fjaviersanchez @hsnee @elisachisari, The notebook tutorial reads good, and I have run the notebook on NERSC successfully. I think it is ready to merge too.

PR #247, opened 2017-08-30
CCL paper in ApJ format

elisachisari:
This branch starts the development of the CCL paper in ApJ format. A new folder has been added to the doc directory where the paper will be developed separately from the note. The starting point i the CCL v0.2 note.


elisachisari:
@kadrlica: There seems to be some formatting issue when an author has to affiliations in the ApJ format. The second affiliation introduces a page break. Do you know how to fix that? I looked into the cls files but couldn't find where to change this.

kadrlica:
@elisachisari I haven't seen this in DES author lists generated for ApJ with `mkauthlist` (we have many authors with multiple affiliations). I think I'll need an example.

elisachisari:
If you go to the ccl_paper folder and compile the tex (just "make" as usual), then you'll get ccl_paper.pdf which has this problem. I copy here a screenshot of what the first two pages look like in my compiled pdf. There is a page break below Tim's first affiliation.
screen shot 2017-08-30 at 14 47 12

kadrlica:
Seems to work for me using `mkauthlist v1.2.3`:
[ccl-frontpage.pdf](https://github.com/LSSTDESC/CCL/files/1263884/ccl-frontpage.pdf)

Maybe we should iterate on Slack?

elisachisari:
Problem solved. Thanks, @kadrlica. 
To compile the paper, simply go to `ccl_paper/` and `make apj`.

kadrlica:
This issue was solved offline on Slack (the root cause was a mismatch between document class and author list style). To make the paper use `make apj`, which should keep the two consistent.

elisachisari:
This PR is intended for developing the paper. Let's keep it open so we can all keep track of further issues and progress.

mishakb:
General remarks about the overall paper and not specific review of sections:
-	The paper sections look overall balanced and consistent except for a few redundancies of equations  mentioned below
-	There is no apparent structure problem. The astract may need some specifics and possibly numbers. It may sound too general at it is. We need to discuss what specifics  we may want to put in the abstract.
-	Also maybe we need to specify in the intro to which audience the paper is written
-	If we say that the accuracy is of 10^{-4} up to redshift z=2 for example, should we say what is expected beyond such redshift? (see for example, page 3)
-	Table one needs a more informative title or caption
-	Some sections need cleaning from redundancies/repetitions of text or equations probably due to the incremental and iterative previous steps in writing the note. See for example after equation (9)
-	  Is it better to itemize the second half of 2.3?
-	Would it be useful to specify what CCL call the transfer function in section 2.4 compared to the more common use of transfer function
-	Is f_K(\chi) used in sub-section 2.4.2 (to SS question whether this sub-section is needed: should we think if it can be distributed over the previous sub-section  or  leave for the sake of consistency between the structure of 2.4 and 3.4)
-	Should we add to the outlook the new additional features that came up for consideration since last fall.
-	Once a few sections are finalized, I think we should proceed to do cross-reviews of sections and that will help getting this wrapped up.

mishakb:
General remarks about the overall paper and not specific review of sections:
-	The paper sections look overall balanced and consistent except for a few redundancies of equations  mentioned below
-	There is no apparent structure problem. The astract may need some specifics and possibly numbers. It may sound too general at it is. We need to discuss what specifics  we may want to put in the abstract.
-	Also maybe we need to specify in the intro to which audience the paper is written
-	If we say that the accuracy is of 10^{-4} up to redshift z=2 for example, should we say what is expected beyond such redshift? (see for example, page 3)
-	Table one needs a more informative title or caption
-	Some sections need cleaning from redundancies/repetitions of text or equations probably due to the incremental and iterative previous steps in writing the note. See for example after equation (9)
-	  Is it better to itemize the second half of 2.3?
-	Would it be useful to specify what CCL call the transfer function in section 2.4 compared to the more common use of transfer function
-	Is f_K(\chi) used in sub-section 2.4.2 (to SS question whether this sub-section is needed: should we think if it can be distributed over the previous sub-section  or  leave for the sake of consistency between the structure of 2.4 and 3.4)
-	Should we add to the outlook the new additional features that came up for consideration since last fall.
-	Once a few sections are finalized, I think we should proceed to do cross-reviews of sections and that will help getting this wrapped up.

elisachisari:
In addition to the comments above, we need to ask the slack channel for updates on contributions/list of authors.

ellison-j:
Hack Day session -- CCL paper general comments. 

* Overall the paper reads nicely and is well structured. 

* Abstract: The sentence about the accuracy level is refers to 
"documented here" and the "Code Comparison Project". Since the abstract 
should be self-contained, it should be changed to something like: "These 
routines have been validated to the accuracy level required for the 
future large scale surveys of LSST." Also a few other minor edits: 

The Core Cosmology Library (CCL) provides routines to compute basic 
cosmological observables with validated numerical accuracy. These 
routines have been validated to the accuracy level required for the 
future large scale surveys of LSST. In the current version, predictions 
are provided for distances, background quantities, power spectra, 
angular auto-correlation and cross-correlation spectra of cosmic shear 
and galaxy clustering, and the halo mass function. Fiducial 
specifications for the expected LSST galaxy distributions and clustering 
bias are also included, together with the capability of computing 
redshift distributions for a user-defined photometric redshift model. 
CCL is written in C, with a Python interface. 

* Fig.1: Fix the missing "..." text? Also better alignment of boxes and 
lines. 

* Table 1: Caption needs more detailed explanation. 

* Splines: explain on first introduction (page 3) 

* Eq.(8), there is a prime missing from the second k. 

* In the BBKS section, `$\Delta(k) \propto T^2(k) k^{3+n}$` is 
incorrect. Should be corrected to `$P(k) = 2 \pi^2 \Delta^2(k) /k^3 
\propto T^2(k) k^{n_s}$` or similar. 

* The description of the effects of baryons on P(k) is split in two 
places: it would be clearer to move the first paragraph (right before 
the BBKS section) to the point in the text directly after the "Cosmic 
emulator" subsection. 

* The description of the implementation of massive neutrinos in section 
2.7 is probably a bit unclear for a general reader. It can be improved 
by giving a few overview sentences stating what is done: calculate the 
neutrino energy density given some assumptions and use this for 
calculating distances, etc.... Also it seems that references to specific 
routines should be removed as this is not done in general in the other 
sections. 

* Section 3.4: same comment about mentioning specific routines. 

* Section 4: We could probably make it clearer that we are validating 
the numerical accuracy of CCL compared with analytical calculations and 
other code e.g. CLASS, rather than trying to achieve 10^-4 accuracy of 
the theoretical predictions themselves (which is obviously not the case 
but should nevertheless be made clear). 

* Section 3.6: section missing 

* In general -- editing of the text for clarity and grammar needed.

Issue #245, opened 2017-08-11
documentation in pyccl.power


c-d-leonard:
The documentation for the python library gives incorrect units for the output linear and nonlinear matter power spectra.

elisachisari:
What units does it give?

c-d-leonard:
Mpc^{-3}

PR #248, opened 2017-09-01
Fix power spectrum units in power.py

elisachisari:
Change from Mpc^{-3} to Mpc^3. This relates to issue #245.


c-d-leonard:
This is a simple case of fixing units in documentation; @elisachisari has fixed the issue so I will go ahead and merge.

PR #254, opened 2017-09-12
Impact of baryons in the matter power spectrum

elisachisari:
This branch incorporates the impact of baryons on the total matter power spectrum according to the model of Schneider & Teyssier (2015). The branch is not yet ready for review, but a minimal implementation is already in place. The next steps are to incorporate checks (including the effects of power spectrum splines), edit the note and extend the implementation to pyccl.


elisachisari:
This PR is ready for review. 

- The model has been implemented in C and in python. 
- A new python test has been added. 
- There is a C example `ccl_sample_baryons.c'
- An example is shown in the Python power spectrum notebook. 
- There is documentation in the CCL note. 

Also in this PR I'm seeing problems with the splines at high k. This will be addressed in a different issue (#258) as it is a more general problem.

I tested the implementation against my own in a different code - but it would be great if someone else could provide a benchmark for our `make check'. @damonge?

doc/0000-ccl_note/main.tex
```diff
@@ -327,10 +327,22 @@ \subsubsection{Eisenstein and Hu}
...
+
+\ccl incorporates the impact of baryons on the total matter power spectrum via the ``baryonic correction model'' of \citet{Schneider15}. The main consequences of baryonic processes are: to suppress the power spectrum at intermediate scales ($k\sim$ a few $h/$Mpc) due to the ejection of gas by Active Galactic Nuclei feedaback, and to enhance it at smaller scales due to enhanced cooling. Three effective parameters govern the contribution of baryonic processes to modifying the total matter power spectrum:
+
+\begin{itemize}
+  \item $\log_{10} [M_c/($M$_\odot/h)]$: the mass of the clusters responsible for feedback, which regulates the amount of suppression of the matter power spectrum at intermediate scales;
+  \item $\eta_b$: a dimensionless parameter which determines the scale at which suppression peaks;
+  \item and $k_s$ [$h/$Mpc]: the wavenumber that determines the scale of the stellar profile.
+\end{itemize}
+
+If these parameters are not specified, \ccl assumes the default paramters of \citet{Schneider15}. The user may pass these explicitly if desired.
```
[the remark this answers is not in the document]
damonge: Note that this phrasing might imply that the default power spectrum option is "with baryons", but that's not the case. Unless you explicitly ask for the baryonic transfer function CCL will use bare HaloFit without baryons. I think this is the correct behaviour, but you should check that this is what a reader would understand.

damonge:
One other thing: there's no unit test for this. I'm not saying you should create one, but we could maybe open a permanent issue with a list of pending unit tests to be implemented (since this is not the only example).

elisachisari:
I added a test here `tests/ccl_test_pyccl_interface.py` in one of the commits. There is no accuracy check though. I checked against my own implementation in a different code and everything seemed to work, but better if you have one curve from your own implementation somewhere else.

doc/0000-ccl_note/main.tex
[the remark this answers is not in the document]
elisachisari: Addressed

src/ccl_power.c
```diff
@@ -1305,14 +1141,25 @@ double ccl_nonlin_matter_power(ccl_cosmology * cosmo, double k, double a, int *s
...
     }
     else { //Extrapolate NL regime using log derivative
       log_p_1 = ccl_power_extrapol_highk(cosmo,k,a,cosmo->data.p_nl,status);
-      return exp(log_p_1);
+      pk=exp(log_p_1);
     }
-
-  default:
+    if(cosmo->config.matter_power_spectrum_method==ccl_baryons){
+      int pwstatus;
```
damonge: You need to initialize this!

src/ccl_power.c
```diff
@@ -1281,21 +1115,23 @@ TASK: compute the nonlinear power spectrum at a given redshift
...
-    return ccl_linear_matter_power(cosmo,k,a,status);
+  
+  //If the matter PS specified was linear, then do the linear compuation
+  if(cosmo->config.matter_power_spectrum_method==ccl_linear) {
     
-  case ccl_halofit:
+    return ccl_linear_matter_power(cosmo,k,a,status);
+
+  }
+  else if (cosmo->config.matter_power_spectrum_method==ccl_halofit || cosmo->config.matter_power_spectrum_method==ccl_baryons){
```
damonge: We should think about what's gonna happen if we want to use the emulator AND baryons. Currently baryons means halofit+baryons... We could make ccl_baryons a separate flag (not related to a PS method). What do you think?

src/ccl_power.c
elisachisari: Fixed

elisachisari:
@damonge I merged master and updated the doxygen docs. Let me know if you approve once more and I will merge.

damonge:
@elisachisari : Yesyesyes. Gogogogo

PR #255, opened 2017-09-26
CMB lensing

damonge:
The branch implements CMB lensing as a possible tracer for correlation functions and power spectra. This is still work in progress.

What has been done:
- Implemented CMB lensing as an alternative tracer for power spectra.
- This implies being able to compute background quantities up to z~1100, so A_SPLINE_MIN had to be overridden.
- However CLASS is not able to compute the matter power spectrum at such high redshifts, and therefore we need an extra parameter A_SPLINE_MIN_PK>A_SPLINE_MIN. The power spectrum routines therefore spline the power spectrum up to A_SPLINE_MIN_PK, and scale it to higher redshifts using the linear growth factor. Is this good enough? For reference, the default is currently A_SPLINE_MIN=0.0003, A_SPLINE_MIN_PK=0.01.

What hasn't been done yet:
- [x] Make a unit test for this (an external benchmark would be needed)
- [x] Make sure this will work for correlation functions too
- [x] Carry everything over to python.
- [x] Document

Other problems:
 - It's not clear to me how to get z_LSS from CLASS, so currently the user has to supply a redshift for the source plane (e.g. z=1100)


damonge:
Update: I've made my own unit test. For this I've generated my own benchmark with an independent code, but this should be validated by someone else. 

Note that the unit test has a couple of hackish lines to work around issue #256 . These can be removed once the issue have been sorted.

damonge:
Second update: CMB lensing now documented in the note (and in all docstrings, both in C and python)

src/ccl_cls.c
```diff
@@ -369,6 +371,11 @@ static CCL_ClTracer *cl_tracer_new(ccl_cosmology *cosmo,int tracer_type,
       }
     }
   }
+  else if(tracer_type=CL_TRACER_CL) {
```
elisachisari: I think you need a "==" instead of "=" here. The compiler gave me a warning.

tests/ccl_test_cmblens.c
```diff
@@ -0,0 +1,91 @@
...
+  ccl_configuration config = default_config;
+  config.transfer_function_method = ccl_bbks;
+  config.matter_power_spectrum_method = ccl_linear;
+  ccl_parameters params = ccl_parameters_create_flat_lcdm(data->Omega_c,data->Omega_b,data->h,
+							  data->A_s,data->n_s, &status);
+  params.Omega_k=0;
+  params.Omega_g=0;
+  params.Omega_n_rel=0;
+  params.Omega_l = 1.0 - params.Omega_m;
+  params.sigma_8=data->sigma_8;
```
elisachisari: This is just a comment in general for several of the test functions: why do we need data->sigma_8 having provided data->A_s? Is params.sigma_8 being populated by the correct value corresponding to the A_s from the input?

tests/ccl_test_cmblens.c
```diff
@@ -0,0 +1,91 @@
...
+  ccl_cosmology * cosmo = ccl_cosmology_create(params, config);
+  ASSERT_NOT_NULL(cosmo);
+
+  //Fix spline parameters for the high-redshift needs of the CMB lensing power spectrum
+  ccl_splines->A_SPLINE_NA=10000;
+  ccl_splines->N_A=500;
+
+  char fname[256];
+  FILE *fi_cc;
+  CCL_ClTracer *tr_cl=ccl_cl_tracer_cmblens_new(cosmo,1090.,&status);
```
elisachisari: I'm not sure whether we want z_source to be an input of the user in this way. Did you consider making it a parameter in the ini file or the cosmo struct?

tests/ccl_test_cmblens.c
```diff
@@ -0,0 +1,91 @@
...
+  ASSERT_NOT_NULL(tr_cl);
+  sprintf(fname,"tests/benchmark/codecomp_step2_outputs/run_log_cl_cc.txt");
```
elisachisari: Is this benchmark really coming from the code comparison step 2? I forgot we had compared CMB lensing outputs as well. Can you confirm?

src/ccl_cls.c
damonge: Well spotted! Done.

tests/ccl_test_cmblens.c
damonge: Only for CLASS, but not for E&H or BBKS

tests/ccl_test_cmblens.c
damonge: The correct thing to do here would be to query z_LSS from CLASS or something like that, but I couldn't figure out how to do that (or how to avoid calling CLASS if we just want the BBKS or E&H power spectra).

There's one advantage to having z_source as an external number: imagine that you want to compute the power spectrum of a source plane at any z, i.e. not just CMB lensing, but 21cm lensing or something like that...

The best solution would be probably to leave z_source as a user-defined variable, but allow for ccl_cosmology to compute z_LSS from cosmological parameters, which is the most common user case (i.e. users could compute z_LSS from ccl_cosmo and then pass it as an argument to compute C_ells).

tests/ccl_test_cmblens.c
damonge: We didn't. I generated my own benchmark using the same code I used to generate the benchmarks from phase 2 of the code comparison. I mentioned this above - happy to take in any external benchmarks!

damonge:
@elisachisari I'll answer your queries in order (apparently I'm not allowed to tick them off).
1- Added example in ccl_sample_run.c
2- Doxygen done
3- I actually didn't touch the notebooks...
4- I added the new spline parameter A_SPLINE_MIN_PK
5- CMB lensing is already included in the note in section 3.8.1
6,7 - Regarding cross-correlations: CMB lensing is implemented as a tracer, and can therefore be any one of the 2 tracers needed to compute any cross-correlation
8- Regarding 3.9 and correlation functions in general: I actually don't like the current phrasing in the note, which labels the 3 possible combinations as "galaxy-galaxy", "galaxy-lensing" and "lensing-lensing". A more general and more precise nomenclature would be "spin0-spin0", "spin0-spin2" and "spin2-spin2". I say this because CMB kappa is a spin-0 quantity, and in that sense behaves just like galaxy clustering when computing a correlation function. I'd be happy to rephrase this section along these lines, but I don't know if everyone would agree.
9- Fixed typo in Eq. 30
10- Fixed typo in Correlation.ipynb.
11- I can comment on the benchmark, but first could you confirm that it's fine to use my own benckmark?

elisachisari:
I'm happy with these changes. We are just waiting for a BBKS benchmark from @sukhdeep1989 to complete this PR.

damonge:
@elisachisari : benchmark by @sukhdeep1989 included and merged with master.

damonge:
@elisachisari : ready for last review

elisachisari:
The changes I suggested have been addressed. The code compiles and passes all tests. New tests are in place for CMB lensing. I'm merging this.

PR #259, opened 2017-10-06
Fix cpp compatibility

vrastil:
When using `CCL` in external c++ code, variable length arrays in function declaration / definitions were throwing compile errors. Only in `ccl_background.h` and `ccl_background.c` files.


vrastil:
Maybe also add `__cplusplus` guards into all header files?

PR #260, opened 2017-10-10
Fix cpp compatibility

vrastil:
added `__cplusplus` guards -- tested with CCL compiled with `gcc` while external c++ program (no guards there) compiled with `g++` -- working

`make check` fails because of swig timestamp, but `swig -python -threads -o pyccl/ccl_wrap.c pyccl/ccl.i` doesn`t produce any files


philbull:
Can we add a make rule that compiles a trivial C++ program, please? This would allow any user to easily test compatibility with C++ on their system.

elisachisari:
@vrastil do you think you could add a make rule as @philbull has suggested? Then we can review and merge

vrastil:
@philbull @elisachisari I added make rule `check-cpp` and it works for me on Ubuntu with `g++`. Can you test this on different systems as well?

elisachisari:
@vrastil I tried it on Mac. It seems to work but it gives me a warning.
Will travis know of this new make check automatically or do we need to do something to add it? @fjaviersanchez 
screen shot 2017-11-06 at 10 07 32 am

vrastil:
@elisachisari I do not get this warning but I think it is only by suffix of ***tests/ccl_sample_run.c*** Can you try changing the suffix to ***.cpp*** and check again? If that works I can add simple cpp wrapper ***tests/ccl_sample_run.cpp*** containing the same code as C version to get rid of this warning

vrastil:
@elisachisari I removed the ***.cpp*** version and modified `make check-cpp` to create its simply copy by `#include "ccl_sample_run.c"`

elisachisari:
Great, thanks. Could you add the relevant documentation in the note and the README within this branch?

vrastil:
I added info about C++ into README and doc to installation procedures, not sure whether it should be moved to some other parts

compiling in travis succeed but there is a problem with dynamic library, I should probably export some `LD_LIBRARY_PATH` before this step, or we can omit running the ***ccl_sample_run***?

elisachisari:
@vrastil Does the check involve running `ccl_sample_run` or just compiling it? Does the problem arise in both cases?

vrastil:
Without runnig the program the Travis build succeeds

vrastil:
tested on Ubuntu with g++ v6 and on Debian g++ v4

elisachisari:
@vrastil Looks like we merged with master at the same time. I don't think that generates any problems, it's just redundant.
But I did run into one problem with the compiler on my Mac desktop when makeing check-cpp:
screen shot 2017-11-08 at 16 58 25

vrastil:
@elisachisari It seems that clang doesn`t support debug option `-Og`, I will look for alternatives and try to use autotools-generated variables for debug mode

elisachisari:
Ok, thanks. It looks like travis is now failing after the merge. It seems there is a conflict with ccl_sample_run due to the implementation of baryons. I'm looking into it.

vrastil:
@elisachisari OK, problem in ***ccl_sample_run*** seems fine. I couldn't find anything in autotools to pregenerate debug flags but now they should be OK for most compilers (hopefully all). Please test it on machine where it was failing.

elisachisari:
@vrastil There's a different issue this time. Now the compiler can't file gsl
screen shot 2017-11-08 at 17 46 13
Any ideas what modifications caused this?

vrastil:
I think that this error was there previously, it just fail earlier due to `-Og`. And CCL compiled fined? No errors due to `gsl`?

elisachisari:
CCL was ok, make check runs without problems.

vrastil:
When you performed the `./configure` step, did you use some extra `CFLAGS` or `LDFLAGS`?

elisachisari:
I add a prefix:
./configure --prefix=/home/chisarie

vrastil:
@elisachisari I don't know why it is failing, if you didn't specify gsl location then it should be in some default look-up location...I tried add some other flags where to look into `make check-cpp`, you can try it. I'm a little afraid that it won't work

elisachisari:
@vrastil Success! ccl_sample_run compiles and runs now in my system with cpp :) Go ahead and merge!

vrastil:
@elisachisari Glad to hear that! :)

Issue #262, opened 2017-10-20
halo_bias invariably throws errors for every halo mass and cosmology I've tried running it on.


anguta:
Every time I try to run halo_bias as in:

import pyccl as ccl
cosmo=ccl.Cosmology(Omega_c=0.26,Omega_b=0.048,h=0.6774,sigma8=.8159,n_s=0.9667)
ccl.halo_bias(cosmo,halo_mass=1e13,a=1)

I get the following errors: 

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-13-b67bac4af4af> in <module>()
      1 import pyccl as ccl
      2 cosmo=ccl.Cosmology(Omega_c=0.26,Omega_b=0.048,h=0.6774,sigma8=.8159,n_s=0.9667)
----> 3 ccl.halo_bias(cosmo,halo_mass=1e13,a=1)

/Users/ewallwic/miniconda2/envs/forest/lib/python2.7/site-packages/pyccl/massfunction.pyc in halo_bias(cosmo, halo_mass, a, odelta)
     68     """
     69     return _vectorize_fn4(lib.halo_bias, 
---> 70                           lib.halo_bias_vec, cosmo, halo_mass, a, odelta)

/Users/ewallwic/miniconda2/envs/forest/lib/python2.7/site-packages/pyccl/pyutils.pyc in _vectorize_fn4(fn, fn_vec, cosmo, x, a, d, returns_status)
    238 
    239     # Check result and return
--> 240     check(status)
    241     if scalar:
    242         return f[0]

/Users/ewallwic/miniconda2/envs/forest/lib/python2.7/site-packages/pyccl/pyutils.pyc in check(status)
     16     # Check for known error status
     17     if status in pyccl.core.error_types.keys():
---> 18         raise RuntimeError("Error %d: %s" % (status, pyccl.core.error_types[status]))
     19 
     20     # Check for unknown error

RuntimeError: Error 1034: CCL_ERROR_MF

damonge:
@anguta 
Thanks for pointing this out. The problem here is that the halo bias is only defined for the Tinker 2010 mass function model, and the default one when you create the ccl.Cosmology object is Tinker 2008.
I see two fixes needed:
a) The error message is not helpful at all.
b) We should make Tinker 2010 the default configuration.

I've fixed b) in a PR, and will ping the appropriate developers to fix a)

philbull:
[...]

PR #263, opened 2017-10-20
fix consistency with default mass function prescription

damonge:
Fixes issue #262 - the default mass function prescription is Tinker2008, but the halo bias is only defined for Tinker2010. The error function in this case is not helpful at all, so it'd be nice to change the current error handling system to output the actual error message raised at the C library level (probably in a different PR to avoid mission creep)


philbull:
I fixed the Python error handling in this branch too -- it now prints the error message string from C if one is available.

elisachisari:
@damonge are we ready to merge this and close the related issue?

PR #267, opened 2017-10-26
Incorporate CLASS citation clause

elisachisari:
This PR incorporates the request to cite the CLASS paper both in the note and in the README.


doc/0000-ccl_note/main.tex
```diff
@@ -1032,6 +1032,11 @@ \section{Feedback}
 
 If you would like to contribute to \ccl or contact the developers, please do so through the \ccl github repository located at \url{https://github.com/LSSTDESC/CCL}.
 
+\section{Citing \ccl}
+\label{sec:cite}
+
+If you use \ccl in your work, please provide a link to the repository and cite it as LSST DESC (in preparation). \ccl has a built-in version of {\tt CLASS} for convenience. We request that \ccl users cite the {\tt CLASS} paper: {\it  CLASS II: Approximation schemes}, D. Blas, J. Lesgourgues, T. Tram, arXiv:1104.2933, JCAP 1107 (2011) 034.
```
brianv0: This is good, but is it actually a request?

> As CLASS is built-in, \ccl users must cite the {\tt CLASS} paper:

elisachisari: no, it is a constraint. They have to cite CLASS, it isn't optional. I guess I should rephrase it.

jchiang87: I think a revision along these lines would be appropriate:

"We request that \ccl users cite the {\tt CLASS} paper:" -> "For free use of the CLASS library, the CLASS developers require that the {\tt CLASS} paper be cited:"

It would probably be good to include a link to their page where this requirement is stated:

http://class-code.net/

Though I don't think we need to state this, the implication of the CLASS language is that if someone is willing to pay, then citation of that paper isn't required.

elisachisari: Thanks, I rephrased it according to this suggestion

Issue #257, opened 2017-09-27
Memory leak while performing MCMC


fjaviersanchez:
While running the MCMC tutorial @WeikangLin and I found that after a while the kernel dies. I think that this might be due to be running out of memory at NERSC (where the resources are shared and maybe there's not that much memory available at "rush hour"). However, it is fishy that it runs out of memory with only 2 parameters and 10 walkers * 100 iterations. I tried it with fewer walkers and it still dies so, this probably means that the splines or CLASS aren't properly freeing the allocated memory. I have to say that I haven't done a proper profiling but I had trouble with the memory allocated by the splines before...

Any thoughts? @philbull @damonge @elisachisari

damonge:
we've been discussing adding a script that runs valgrind on CCL, and this will pick up these leaks, so hopefully we can sort this out soon.

tmcclintock:
@fjaviersanchez can you DM me the code you ran on slack? The thing I really need to get at is the kernel, or likelihood function.

My plan is to design a script w/ valgrind that looks like the tutorial on the CCL main page but substituted in with the functions you called.

elikrause:
@fjaviersanchez Would you be able to test whether the leak disappears if you use BBKS instead of CLASS?

fjaviersanchez:
Yes, when we ran `bbks` it worked with no problem!

elikrause:
Fixed in PR #268 , which was merged into master.

PR #268, opened 2017-10-26
Memory leaks

elikrause:
This PR fixes memory leaks associated with calling CLASS from CCL (#257) and comparisons to uninitialized values.
I would appreciated if the reviewers could check my logic for init_arr[i_init++] in ccl_run_class . Previously, init_arr[6], which is the condition for calling spectra_free() in ccl_free_class_structs, was never filled.
My question is whether a CLASS structure should only be freed if the function filling/allocating completed without errrors, or also if the this function did not complete (which could lead to partial allocation of structure elements). In the former case, the current code should be correct. In the latter case, the init_arr[i_init++] should be moved from line 223 to line 182.


elikrause:
Attached valgrind output with remaining issues - these won't cause a crash over normal MCMC run time, but should remain on our radar.
[memory_leaks.txt](https://github.com/LSSTDESC/CCL/files/1419896/memory_leaks.txt)

elisachisari:
@elikrause, are we ready to merge this PR?

damonge:
@elikrause : all re-tested. Merge, merge, merge!!!

PR #276, opened 2017-11-07
Store input arrays when creating new ClTracer object

philbull:
This is a trivial PR to store the input arrays (e.g. number density, bias) when a new ClTracer object is created. This allows the user to easily access the data that they originally used to specify the ClTracer object, which is useful in codes that need to keep track of many ClTracer objects (e.g. Fisher forecasting codes).


elisachisari:
@philbull: This looks good. I tested it on a python notebook and committed my edits as well, so the user knows how to retrieve these functions. There are some conflicts with master that need to be resolved.

philbull:
Thanks Elisa, I solved the merge conflicts (which involved making a few edits to the Jupyter notebook too). This should be ready to merge.

PR #172, opened 2017-04-27
Interface Angpow

jeremyneveu:
This branch proposes a link between Angpow and CCL. To install it:
- install the 'rsd' branch of Angpow from [https://gitlab.in2p3.fr/campagne/AngPow ](https://gitlab.in2p3.fr/campagne/AngPow )
- modify the temporary makefile `compile.angpow_interface.sh` according to your configuration 
- install the test executable with `source compile.angpow_interface.sh` 
- run it `./tests/ccl_test_angpow`

The new files are `include/ccl_angpow_interface.h` and `tests/ccl_test_angpow.cc`

Here is a comparison between CLASSgal and CCL for an auto-correlation with a gaussian window z=1 and sigma_z=0.02. On the top blue and green are superimposed, and the orange curve is without RSD effect included. Below the relative comparison (look at the orange curve).
figure_rsd2

But concerning the packaging of Angpow inside CCL, I don't know how to proceed. Help needed ;)

PR #283, opened 2017-11-15
Non limber speedup

jeremyneveu:
The Angpow accuracy parameters are automaticly set so that Angpow computes the C_ell quickly  with a mean relative difference with the native computation two orders of magnitude below the relative cosmic variance in most of the parameter space (see plots below for auto-correlation with gaussian windows with lmax=200, 500, 1000).
An automated check is proposed, but could not find how to include tests/ccl_test_angpow in the check loop.
The note is updated with a short paragraph on the non Limber computations.

benchmark_angpow_ccl_lmax200
benchmark_angpow_ccl_lmax500
benchmark_angpow_ccl_lmax1000


elisachisari:
@jeremyneveu What are the differences between this PR and the angpow PR #172? Does this replace that PR or has it branched off that one?

jeremyneveu:
This PR replaces angpow PR #172 as it includes now David's code and everything. Previous PR goal was to implement Angpow and compare it with CLASS. Now it is widely compared within a broad parameter space.

vrastil:
@jeremyneveu do you need some help with this PR?

jeremyneveu:
@vrastil For the moment no. I was blocked because pyccl do not install on my system (see my posted issue). But @fjaviersanchez  helped me a lot and to fix this I need to... reinstall all my python environment ! So as soon as it is repaired, I can continue the merging of my branch with master and ask for a review.

jeremyneveu:
Ok the Travis tests pass now. I think this PR is ready for review.

slosar:
Hi,

Sorry for my delayed review. In general a very nice numerical work. Here are my comments, mostly to do with implementation. I cc @damonge as his fingers in this pie too. I'm not going to block PR over this, but it would be good if you considered them:

 * In the same way CLASS was demerged, I think you should consider demerging angpow as it is a pretty big library on its own. And then write a similar `install_angpow.py` script.
 * I don't like how now we have a sole C++ file in the src directory `ccl_cls.cc`. It would be better to include a thin C wrapper inside angpow than bloating CCL with a C++ dependency. In fact, while I like angpow a lot, I think it should be made an optional dependency so that people requiring just distances or limber should not be forced to have a working C++ compiler.
 * The non-limber method is specified in `CCL_ClWorkspace`. Fine, but i) `ccl_angular_cl` doesn't take it and `ccl_angular_cls` still claims *Computes limber power spectrum for two different tracers* in description. Whatever you do, `ccl_angular_cls` should do the same as `ccl_angular_cl` but for more then one cl, otherwise very confusing.
 * I would make `ccl_cl_workspace_default_limber` that defaults to Limber, and `ccl_cl_workspace_default` that defaults to angpow if present with sensible settings.
 * It would be awfully nice if I could pass -1 as ` l_limber` and have the code put a sensible `l_limber` by itself. I.e. the user shouldn't have to think -- it should just work out of the box. (although now you the python version uses -1 to specify limber everywhere, huh).
 * `ccl_sample_angpow.c` still have some garbage leftover from copying default distance demo
 * I hate this `_new_` bullshit, there will always be something more new and APIs will change. Best to get rid of this scrounge before it spreads to too many places.  
 * the branch won't merge. I tried fixing it but failed miserably.

jeremyneveu:
@slosar @damonge Hi all. Thanks again for the review Anze. The exam period is over now and I can go through your comments. I started yesterday the merge but it is not easy as CLASS became optional meanwhile. We discussed yesterday at the CCL telecon and we propose if you agree that the "Angpow optional" point would be address in another PR, in order to have a funcitonal merged Angpow at SLAC. Unless @damonge given your experience in removing CLASS you can address this point easily before DESC ?  Maybe point 2 also about the .cc file also can be postponed. What do you think ?

jeremyneveu:
@slosar @damonge Hi. I think I finished to implement all you requirements. In particular now Angpow has an optional installation like CLASS. I created a LSSTDESC github repository for an Angpow release interfaced with CCL, and a python script to download it. I adapted the make and configure files to this major change. The ccl_cls.cc file was split in two, a ccl_cls.c file with the native non-limber computations, and an angpow_ccl.cc file that stay in the Angpow directory and is downloaded from the Angpow4CCL repository. The travis tests are ok, but they do not include the Angpow installation (I don't know how to add this, it would just need to run python `angpow_install.py`). So tell me if you are happy with these changes, and let's finish this PR.

jeremyneveu:
@slosar I modify the Travis file so that it install and make the tests with angpow. If forgot to answer the comment about the automatic l_limber so that user do not think. Jean-Eric worked on this and he developped a version of angpow that do part of this. It could be included in a future PR, as well as magnification lensing effects, etc.

jeremyneveu:
@damonge @slosar Hi. I have just added a new tests to check if the @damonge non-limber computation reaches the limber solution at high ell for a very thick redshift window. I think it is better to have this test apart from the angpow installation because 1) angpow is now optional in the installation and 2) the actual angpow version is not good for evry thick redshift shells (see plots above). As soon as this PR is approved, I will work on including the last angpow version which includes many other feautres (improved precision, multi-shells, magnification lensing effect, etc...)

Makefile.am
```diff
@@ -12,32 +11,53 @@ ccl_DATA = include/ccl_params.ini
 
 lib_LTLIBRARIES = libccl.la
 libccl_la_SOURCES = src/ccl_background.c src/ccl_core.c src/ccl_error.c src/ccl_lsst_specs.c src/ccl_placeholder.c src/ccl_power.c src/ccl_utils.c src/ccl_cls.c src/ccl_massfunc.c src/ccl_neutrinos.c src/ccl_emu17.c src/ccl_correlation.c src/fftlog.c
+
+if LOCALANGPOW
+
+ libccl_la_SOURCES += angpow/src/angpow_ccl.cc
```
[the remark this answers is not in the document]
jeremyneveu: This file is downloaded with the Angpow released I prepared for CCL but is not in the original Angpow. It makes the interface between CCL and Angpow and is only added to CCL if Angpow is downloaded.

tests/ccl_test_nonlimber.c
```diff
@@ -0,0 +1,133 @@
...
+  // Initialize cosmology object given cosmo params
+  ccl_cosmology *ccl_cosmo=ccl_cosmology_create(ccl_params,ccl_config);
+
+  // Create tracers for angular power spectra
+  double z_arr_gc[NZ],nz_arr_gc[NZ],bz_arr[NZ],sz_arr[NZ];
+  for(int i=0;i<NZ;i++)
+    {
+      z_arr_gc[i]=Z0_GC-5*SZ_GC+10*SZ_GC*(i+0.5)/NZ;
+      nz_arr_gc[i]=exp(-0.5*pow((z_arr_gc[i]-Z0_GC)/SZ_GC,2));
+      bz_arr[i]=1;//+z_arr_gc[i];
```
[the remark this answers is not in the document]
jeremyneveu: Ok

tests/ccl_test_nonlimber.c
```diff
@@ -0,0 +1,133 @@
...
+
+  double rel_precision = 0.;
+  for(int ii=2;ii<NL;ii++) {
+    int l = ells[ii];
+    double cl_gg_nl=cells_gg_native[ii];
+    double cl_gg_lim=cells_gg_limber[ii];
+    double ratio = fabs(cl_gg_nl-cl_gg_lim)/cl_gg_nl;
+    if(l>NL/2)
+      rel_precision += ratio / sqrt(2./(2*l+1));
+    //printf("%d %.3g %.3g %.3g %.3g\n",l,cl_gg_nl,cl_gg_lim,ratio,ratio / sqrt(2./(2*l+1)));
```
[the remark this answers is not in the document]
jeremyneveu: Ok

jeremyneveu:
Ok tomorrow I will update the readme and the docs, that's true I completely forgot it.
The Angpow version I plugged to CCL was not able to compute the lensing magnification term for the LSS, but with Jean-Eric we worked on it and the most recent version is able to. 
In principle Angpow could do the integrals for the lensing, but I did not make the interface between the CCL lensing and Angpow because I wanted to make the interface step by step, first LSS. Unfortunately it took a very long time to go until that point. Both items are definetely in my todo list but it would be much more confortable for me that this branch is merged first.

damonge:
OK, I've benchmarked limber against master. With this refactoring we lose a bit of speed (~30%) with respect to the previous version on a single C_ell calculation. However, because now we have a reliable scheme to sample the C_l and then interpolate, we actually gain a lot in speed as soon as we have to compute a couple hundred C_ls. I'm happy to leave it as it is and reconsider speeding it up afterwards if needed.

configure.ac
```diff
@@ -16,10 +16,13 @@ AM_MAINTAINER_MODE
 
 # Checks for programs.
 AC_PROG_CC
+AC_PROG_CXX
```
slosar: So this will now require a working C++ compiler even if angpow is not used? Can we wrap this only if angpow is downloaded?

doc/0000-ccl_note/main.tex
```diff
@@ -585,100 +586,75 @@ \subsubsection{The Limber approximation}
...
+\item  $P(k,z)$ : the matter power spectrum at redshift $z$ 
+\item $\widetilde{\Delta}_\ell(z,k)$: a function describing the physical processes such as matter density fluctuations, redshift-space distortions as described for instance in references \citet{2008cmb..book.....D,2009PhRvD..80h3514Y,2010PhRvD..82h3508Y, 2011PhRvD..84d3516C,2011PhRvD..84f3505B}. Currently, the \texttt{Angpow} version delivered with \ccl only can deal with galaxy clustering tracers (no lensing) and this without the magnification lensing term (equation \ref{eq:deltaM}). The incorporation of those transfer functions is left for future work, but in principle \texttt{Angpow} has already the capability to treat them. For now, for galaxy clustering tracers we defined $\widetilde{\Delta}_\ell(z,k)$ as
```
slosar: So, what happens if I ask CCL to use angpow to calculate other power spectra?

configure.ac
jeremyneveu: Hi Anze. I have just tried to make it optional, but then the autoreconf complains because of the CXX flags defined in Makefile.am for the check-cpp method. Any idea on how to go forward ?

examples/ccl_sample_angpow.cpp
```diff
@@ -0,0 +1,198 @@
+#include <stdlib.h>
```
damonge: Question: why is this a .cpp file? Angpow is C++ but I thought CCL is still a C library, so you could link it from any C code. Is this no longer true?

Also, I thought we were using .cc and not .cpp for C++ source files

tests/ccl_sample_angpow.c
```diff
@@ -1,181 +1 @@
...
-  ccl_cl_tracer_free(ct_gc_B);
-  ccl_cl_tracer_free(ct_gc_A);
-
-  //Always clean up!!
-  ccl_cosmology_free(cosmo);
-
-  return 0;
-}
-
+#include "ccl_sample_angpow.c"
```
damonge: What's the point of this file?!

examples/ccl_sample_angpow.cpp
jeremyneveu: You're right there is not point that this is a C++ file. I moved it to C only.
The files in examples use the extension .cpp, so I did same to preserve the harmony of CCL ;)

tests/ccl_sample_angpow.c
jeremyneveu: It is a mistake, I remove it.

jeremyneveu:
Thanks @damonge for this review. I ask for volunteers now.

damonge:
Looks good to me. Merge, merge, merge!

PR #284, opened 2017-11-18
Fix to neutrino example files

elisachisari:
The files: `ccl_example_neutrinos` and `ccl_test_neutrinos` in the `tests` file lag behind master.


c-d-leonard:
I have had a look at these files and found that "ccl_test_neutrinos.c", "ccl_sample_power_neutrinos.c" and "ccl_example_neutrinos.c" were all incorrectly added to the repo (by me) at some point. Sorry! I have removed them (and merged master into this branch). This is ready for review.

elisachisari:
Following feedback from @ellison-j on Slack, I think we should bring back one of the example files in this PR.

c-d-leonard:
Hi @elisachisari 

Okay, I modified one of the files to be a more up to date and hopefully more straightforward and helpful example file for neutrinos, including how to get a power spectrum and Omega_nu(z) with a single massive neutrino. This is now called CCL/tests/ccl_sample_neutrinos.c. I also removed a data file that one of this sample files was previously using to read in k values; this is no longer necessary with the updated sample file.

Also note: I will need to change this example file with the multiple non-degenerate neutrinos support when that is merged because the mass will go from a double to a double*. I will make sure to update this sample file in that pull request but I think this is good for now.

PR #285, opened 2017-11-18
Update swig files

elisachisari:
Swig files need to be updated in master. Current swig check doesn't pass.


philbull:
This should also be fixed in the `cosmicemu` branch, when that is merged.

elisachisari:
Given the delays in cosmicemu following conflicts with master, I would appreciate it we could merge this today to have swig checks up to date un master while we work on merging cosmicemu. @damonge, could you approve?

elisachisari:
Actually - these checks are passing in master now, apparently following a recent merge of another branch. So I'm closing this PR.

PR #287, opened 2017-11-20
Handle non-degenerate neutrino masses

c-d-leonard:
This is not ready to merge yet! I need to check / add support for python.

### DIFF
--- a/include/ccl_core.h
+++ b/include/ccl_core.h
@@ -38,7 +38,7 @@ typedef struct ccl_parameters {
   // At the moment, only support equal-mass massive neutrinos (cannot have two different masses)
   double N_nu_mass; // Number of different species of massive neutrinos
   double N_nu_rel;  // Neff massless
-  double mnu;  // total mass of massive neutrinos
+  double *mnu;  // total mass of massive neutrinos (This is a pointer so that it can hold multiple unequal masses.)
   double Omega_n_mass; // Omega_nu for MASSIVE neutrinos 
   double Omega_n_rel; // Omega_nu for MASSLESS neutrinos
  
@@ -139,7 +139,7 @@ ccl_cosmology * ccl_cosmology_create(ccl_parameters params, ccl_configuration co
 
 // Helper functions to create ccl_cosmology structs directly given a set of params
 ccl_cosmology * ccl_cosmology_create_with_params(
-        double Omega_c, double Omega_b, double Omega_k, double N_nu_rel, double N_nu_mass, double mnu, 
+        double Omega_c, double Omega_b, double Omega_k, double N_nu_rel, double N_nu_mass, double* mnu, 
         double w0, double wa, double h, double norm_pk, double n_s,
         int nz_mgrowth, double *zarr_mgrowth, double *dfarr_mgrowth, 
         ccl_configuration config, int *status);
@@ -170,7 +170,7 @@ ccl_cosmology * ccl_cosmology_create_with_lcdm_params(
  * For specific cases see documentation for ccl_error.c
  * @return void
  */
-ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k, double N_nu_rel, double N_nu_mass, double mnu, double w0, double wa, double h, double norm_pk, double n_s,int nz_mgrowth,double *zarr_mgrowth,double *dfarr_mgrowth, int *status);
+ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k, double N_nu_rel, double N_nu_mass, double* mnu, double w0, double wa, double h, double norm_pk, double n_s,int nz_mgrowth,double *zarr_mgrowth,double *dfarr_mgrowth, int *status);
 
 // Specific sub-models
 /**
@@ -231,10 +231,10 @@ ccl_parameters ccl_parameters_create_flat_wacdm(double Omega_c, double Omega_b, 
  */
 ccl_parameters ccl_parameters_create_lcdm(double Omega_c, double Omega_b, double Omega_k, double h, double norm_pk, double n_s, int *status);
 
-ccl_parameters ccl_parameters_create_flat_lcdm_nu(double Omega_c, double Omega_b, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double mnu, int *status);
-ccl_parameters ccl_parameters_create_flat_wcdm_nu(double Omega_c, double Omega_b, double w0, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double mnu, int *status);
-ccl_parameters ccl_parameters_create_flat_wacdm_nu(double Omega_c, double Omega_b, double w0,double wa, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double mnu, int *status);
-ccl_parameters ccl_parameters_create_lcdm_nu(double Omega_c, double Omega_b, double Omega_k, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double mnu, int *status);
+ccl_parameters ccl_parameters_create_flat_lcdm_nu(double Omega_c, double Omega_b, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double* mnu, int *status);
+ccl_parameters ccl_parameters_create_flat_wcdm_nu(double Omega_c, double Omega_b, double w0, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double* mnu, int *status);
+ccl_parameters ccl_parameters_create_flat_wacdm_nu(double Omega_c, double Omega_b, double w0,double wa, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double* mnu, int *status);
+ccl_parameters ccl_parameters_create_lcdm_nu(double Omega_c, double Omega_b, double Omega_k, double h, double norm_pk, double n_s, double N_nu_rel, double N_nu_mass, double* mnu, int *status);
 
 /**
  * Free a cosmology struct
--- a/include/ccl_neutrinos.h
+++ b/include/ccl_neutrinos.h
@@ -38,4 +38,4 @@ gsl_spline* calculate_nu_phasespace_spline(int *status);
  * For specific cases see documentation for ccl_error.c
  * @return OmNuh2 Fractional energy density of neutrions with mass mnu, multiplied by h squared. 
  */
-double Omeganuh2 (double a, double Neff, double mnu, double TCMB, gsl_interp_accel* accel, int * status);
+double Omeganuh2 (double a, double Neff, double* mnu, double TCMB, gsl_interp_accel* accel, int * status);
--- a/src/ccl_core.c
+++ b/src/ccl_core.c
@@ -163,7 +163,7 @@ norm_pk: amplitude of the primordial PS (either A_s or sigma_8)
 n_s: index of the primordial PS
 */
 ccl_cosmology * ccl_cosmology_create_with_params(double Omega_c, double Omega_b, double Omega_k,
-						 double N_nu_rel, double N_nu_mass, double M_nu,
+						 double N_nu_rel, double N_nu_mass, double* mnu,
 						 double w0, double wa, double h, double norm_pk, double n_s,
 						 int nz_mgrowth, double *zarr_mgrowth, 
 						 double *dfarr_mgrowth, ccl_configuration config,
@@ -171,7 +171,7 @@ ccl_cosmology * ccl_cosmology_create_with_params(double Omega_c, double Omega_b,
 {
   // Create ccl_parameters struct from input parameters
   ccl_parameters params;
-  params = ccl_parameters_create(Omega_c, Omega_b, Omega_k, N_nu_rel, N_nu_mass, M_nu, w0, wa,
+  params = ccl_parameters_create(Omega_c, Omega_b, Omega_k, N_nu_rel, N_nu_mass, mnu, w0, wa,
 				 h, norm_pk, n_s, nz_mgrowth, zarr_mgrowth, dfarr_mgrowth, status);
   // Check status
   ccl_check_status_nocosmo(status);
@@ -281,7 +281,7 @@ A_s: amplitude of the primordial PS
 n_s: index of the primordial PS
  */
 ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k,
-				     double N_nu_rel, double N_nu_mass, double mnu,
+				     double N_nu_rel, double N_nu_mass, double* mnu,
 				     double w0, double wa, double h, double norm_pk,
 				     double n_s,int nz_mgrowth,double *zarr_mgrowth,
 				     double *dfarr_mgrowth, int *status)
@@ -346,9 +346,12 @@ ccl_parameters ccl_parameters_create_flat_lcdm(double Omega_c, double Omega_b, d
   double Omega_k = 0.0;
   double N_nu_mass = 0.0;
   double N_nu_rel = 3.046;
-  double mnu = 0.0;
   double w0 = -1.0;
   double wa = 0.0;
+  double *mnu;
+  double mnuval = 0.;
+  mnu = &mnuval;
+  
   ccl_parameters params = ccl_parameters_create(Omega_c, Omega_b, Omega_k, N_nu_rel, N_nu_mass,
 						mnu, w0, wa, h, norm_pk, n_s, -1, NULL, NULL, status);
   return params;
@@ -361,7 +364,7 @@ INPUT: some cosmological parameters needed to create a flat LCDM model with neut
 TASK: call ccl_parameters_create to produce an LCDM model
 */
 ccl_parameters ccl_parameters_create_flat_lcdm_nu(double Omega_c, double Omega_b, double h, double norm_pk,
-						  double n_s, double N_nu_rel, double N_nu_mass, double mnu,
+						  double n_s, double N_nu_rel, double N_nu_mass, double *mnu,
 						  int *status)
 {
   double Omega_k = 0.0;
@@ -383,9 +386,12 @@ ccl_parameters ccl_parameters_create_lcdm(double Omega_c, double Omega_b, double
 {
   double N_nu_mass = 0.0;
   double N_nu_rel = 3.046;
-  double mnu = 0.0;
   double w0 = -1.0;
   double wa = 0.0;
+  double *mnu;
+  double mnuval = 0.;
+  mnu = &mnuval;
+  
   ccl_parameters params = ccl_parameters_create(Omega_c, Omega_b, Omega_k, N_nu_rel, N_nu_mass, mnu, w0, wa,
 						h, norm_pk, n_s,-1,NULL,NULL, status);
   return params;
@@ -398,7 +404,7 @@ TASK: call ccl_parameters_create for this specific model
 */
 ccl_parameters ccl_parameters_create_lcdm_nu(double Omega_c, double Omega_b, double Omega_k, double h,
 					     double norm_pk, double n_s, double N_nu_rel, double N_nu_mass,
-					     double mnu, int *status)
+					     double* mnu, int *status)
 {
   double w0 = -1.0;
   double wa = 0.0; 
@@ -421,8 +427,11 @@ ccl_parameters ccl_parameters_create_flat_wcdm(double Omega_c, double Omega_b, d
   double Omega_k = 0.0;
   double N_nu_mass = 0.0;
   double N_nu_rel = 3.046;
-  double mnu = 0.0;
   double wa = 0.0;
+  double *mnu;
+  double mnuval = 0.;
+  mnu = &mnuval;
+
   ccl_parameters params = ccl_parameters_create(Omega_c, Omega_b, Omega_k, N_nu_rel, N_nu_mass, mnu, w0, wa,
 						h, norm_pk, n_s,-1,NULL,NULL, status);
   return params;
@@ -435,7 +444,7 @@ TASK: call ccl_parameters_create for this specific model
 */
 ccl_parameters ccl_parameters_create_flat_wcdm_nu(double Omega_c, double Omega_b, double w0, double h,
 						  double norm_pk, double n_s, double N_nu_rel, 
-						  double N_nu_mass, double mnu, int *status)
+						  double N_nu_mass, double *mnu, int *status)
 {
 
   double Omega_k = 0.0;
@@ -456,7 +465,10 @@ ccl_parameters ccl_parameters_create_flat_wacdm(double Omega_c, double Omega_b, 
   double Omega_k = 0.0;
   double N_nu_mass = 0.0;
   double N_nu_rel = 3.046;
-  double mnu = 0.0;
+  double *mnu;
+  double mnuval = 0.;
+  mnu = &mnuval;
+  
   ccl_parameters params = ccl_parameters_create(Omega_c, Omega_b, Omega_k,N_nu_rel, N_nu_mass, mnu, w0, wa,
 						h, norm_pk, n_s,-1,NULL,NULL, status);
   return params;
@@ -469,7 +481,7 @@ TASK: call ccl_parameters_create for this specific model
 */
 ccl_parameters ccl_parameters_create_flat_wacdm_nu(double Omega_c, double Omega_b, double w0, double wa,
 						   double h, double norm_pk, double n_s, double N_nu_rel,
-						   double N_nu_mass, double mnu, int *status)
+						   double N_nu_mass, double* mnu, int *status)
 {
 
   double Omega_k = 0.0;
--- a/src/ccl_neutrinos.c
+++ b/src/ccl_neutrinos.c
@@ -93,38 +93,45 @@ double nu_phasespace_intg(gsl_interp_accel* accel, double mnuOT, int* status)
 INPUTS: a: scale factor, Neff: number of neutrino species, mnu: total mass in eV of neutrinos, TCMB: CMB temperature, accel: pointer to an accelerator which will evaluate the neutrino phasespace spline if defined, status: pointer to status integer.
 TASK: Compute Omeganu * h^2 as a function of time.
 */
-double Omeganuh2 (double a, double Neff, double mnu, double TCMB, gsl_interp_accel* accel, int* status)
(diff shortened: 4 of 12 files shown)


c-d-leonard:
Update: this branch now handles multiple non-degenerate neutrino masses and passes all current python and c tests * which are all written for the case of 0 masses neutrinos *. To do before review: introduce tests with massive neutrinos.

c-d-leonard:
Another update: 

Tests have now been added, in C and in python, for initiating a cosmology with non-zero massive neutrinos, and for calculating comoving distances and distance moduli with non-zero massive neutrinos. The benchmarks for this were produced using astropy.

One thing to note is: astropy uses a fitting function for the neutrino phasespace integral whereas we do the integral numerically and spline it. The fitting function they use is found in equation 26 of https://arxiv.org/pdf/1001.4538.pdf, which states that it is only expected to match the full integral to about 0.3%, higher than our nominal error tolerance. For this reason, I have set the tolerance on distance checks involving non-zero massive neutrinos to 1e-3 rather than 1e-4; with this the tests pass.

As long as people are okay with this I think this is basically ready for review.

c-d-leonard:
Hi again - I've now manually checked that the power spectrum from CLASS directly matches the power spectrum as called through CCL (still from CLASS) in massive neutrinos cosmologies for both the linear and nonlinear case. With this sanity check complete and the note updated, I think this branch may be ready for review.

As I mentioned before, the only formal tests I've added are for the distance quantities. However, this seems to me like it is sufficient, given that as detailed in the note we only claim to support massive neutrinos for distances, P(k), and lensing C_ells, as anything involving the growth rate would need support for scale-dependent growth rates which right now we don't have.

c-d-leonard:
Hi - I've now completed the neutrinos-related overhaul discussed in Issues #338 and #335 . You can see what I've done there, but to summarize:
- I've altered the API such that the user passes neutrino-related parameters of Neff (= effecitve number of neutrino species in the early universe) and mnu, where mnu can either be a single number, in which case it is assumed to be a sum of the masses of three neutrino species, or a list or array or length 3. In the former case, the default behaviour (from the python wrapper) is to split the sum of mass into the normal hierarchy, but a flag has been introduced to allow also inverted hierarchy or equal splitting. If working from the C library, the user must specify the splitting using a new argument to the parameters initializer. 
- To get the number of relativistic neutrinos given the number of massive neutrinos and Neff (internally - the user doesn't have to deal with this), we use the same math that the CLASS people used to get the numbers in explanatory.ini which confused us for a long time. I've added a bit in the note explaining this; if you're interested look there.
- In the case where the emulator is used for the power spectrum, I've made the default behaviour to require the user to pass a list or array of three equal masses, or to pass a sum but requesting the split into three masses be equal. By default, other configurations with the emulator switched on will cause an error message and CCL will quit. However, I've also included an option for the user to force CCL to equalize the neutrino masses before passing to the emulator. This will result in internal inconsistencies and the user will have to consciously switch it on; this is a use-at-your-own-risk feature.
- I also made changes to the function which takes Omeganuh2 today and converts to mnu, to allow the conversion to be along different assumptions (normal, inverted, equal, just pass out the sum). This then becomes more of a convenience function rather than just used for testing the emulator.
- I added checks to make sure the user is not asking CCL to computing a quantity in a cosmology with massive neutrinos for which we don't support that yet (e.g. growth rate, halo mass function).
- I added tests to the python interface with this new functionality and an automatic test against benchmarks for the power spectrum produced by externally calling CLASS vs calling in CCL.

From my perspective this PR is now ready for review. @philbull @slosar @christianel

include/ccl_neutrinos.h
```diff
@@ -20,6 +21,13 @@ extern "C" {
 // The combination of constants required in Omeganuh2
 #define NU_CONST (8. * pow(M_PI,5) *pow((KBOLTZ/ HPLANCK),3)* KBOLTZ/(15. *pow( CLIGHT,3))* (8. * M_PI * GNEWT) / (3. * 100.*100.*1000.*1000. /MPC_TO_METER /MPC_TO_METER  * CLIGHT * CLIGHT))
 
+typedef enum ccl_nu_masses_label {
+  ccl_nu_masses_normal_label=0,
```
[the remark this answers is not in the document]
philbull: Also, how does this relate to `ccl_mnu_type_label`?

CHANGELOG.md
```diff
@@ -0,0 +1,79 @@
+v 0.3 API changes:
+
+C library:
```
philbull: Other API changes have been made, e.g. you replaced an old function with `nu_masses()`.

pyccl/cls.py
```diff
@@ -85,7 +85,7 @@ def __init__(self, cosmo, tracer_type=None, has_rsd=False,
 
         # Construct new ccl_cl_tracer
         status = 0
-        self.cltracer, status = lib.cl_tracer_new_wrapper(
+        output = lib.cl_tracer_new_wrapper(
```
philbull: Could use a better name than `output`, e.g. `return_val` or something?

pyccl/core.py
```diff
@@ -39,6 +39,18 @@
...
 
+emulator_neutrinos_types = {
+	'strict': 	lib.strict,
+	'equalize': lib.equalize
+}
+
+mnu_types = {
+	'list': lib.mnu_list,
+	'sum': lib.mnu_sum,
+	'sum_inverted': lib.mnu_sum_inverted,
```
philbull: Can probably just call these 'inverted' and 'equal'.

pyccl/core.py
```diff
@@ -82,9 +94,12 @@ def __init__(self, Omega_c=None, Omega_b=None, h=None, A_s=None, n_s=None,
...
             n_s (float): Primordial scalar perturbation spectral index.
             Omega_k (float, optional): Curvature density fraction. Defaults to 0.
-            N_nu_rel (float, optional): Number of massless neutrinos present. Defaults to 3.046
-            N_nu_mass (float, optional): Number of massive neutrinos present. Defaults to 0.
-            m_nu (float, optional): total mass in eV of the massive neutrinos present (current must be equal mass). Defaults to 0.
+            Neff (float, optional): Effective number of neutrino species. Defaults to 3.046
+            m_nu (float or array-like, optional): If float: total mass in eV of 
+							the massive neutrinos present. If array-like, masses of 3 neutrino
+							species (must have length 3).
+			mnu_type (string): treatment for neutrinos. Available: 'sum', 'sum_inverted',
```
philbull: Weird indent here? We should always use spaces, not tabs.

pyccl/core.py
```diff
@@ -136,31 +151,47 @@ def __init__(self, Omega_c=None, Omega_b=None, h=None, A_s=None, n_s=None,
...
         
+        if isinstance(m_nu, float):
+            if mnu_type == None :
+                mnu_type = 'sum'
+            m_nu = [m_nu]
+        elif hasattr(m_nu, "__len__"):
+            if (len(m_nu)!=3):
+                raise ValueError("m_nu must be a float or array-like object with length 3.")
+            elif ((mnu_type=='sum') or (mnu_type=='sum_inverted') or (mnu_type=='sum_equal')):
+                raise ValueError("That mnu type cannot be passed with a list of neutrino masses, only with a sum.");
```
philbull: Replace `"That mnu type..."` with: `"mnu_type '%s' cannot be passed..." % mnu_type`, so it's clear to the user what setting has been detected.

pyccl/core.py
philbull: Also, mention what the default settings are.

pyccl/core.py
```diff
@@ -249,14 +280,17 @@ def __init__(self,
             effects to be implemented. Defaults to `nobaryons`.
             mass_function (:obj:`str`, optional): The mass function to use. 
             Defaults to `tinker` (2010).
+            emulator_neutrinos: `str`, optional): If using the emulator for
```
philbull: Would be good to mention the two options, and what each of them means.

pyccl/neutrinos.py
```diff
@@ -2,6 +2,13 @@
 from pyccl import ccllib as lib
 from pyccl.pyutils import check
 
+neutrino_mass_splits = {
```
philbull: I like this name for these settings -- could maybe use it in the C code as well?

pyccl/neutrinos.py
```diff
@@ -12,8 +19,8 @@ def Omeganuh2(a, Neff, mnu, TCMB=2.725):
 
     Args:
         a (float): Scale factor, normalized to 1 today.
-        Neff (float): Number of relativistic neutrino species (NB: to all practical purposes, Neff is simply N_nu_mass)
-        mnu (float): Neutrino mass (in eV)
+        Neff (float): Number of massive neutrino species (NB: to all practical purposes, Neff is simply N_nu_mass)
```
philbull: This comment confuses me -- I thought it was the effective number of relativistic neutrinos?

pyccl/neutrinos.py
```diff
@@ -25,41 +32,49 @@ def Omeganuh2(a, Neff, mnu, TCMB=2.725):
     # Convert to array if it's not already an array
     if not isinstance(a, np.ndarray):
         a = np.array([a,]).flatten()
+    if not isinstance(mnu, np.ndarray):
+        mnu = np.array([mnu,]).flatten()
+    
+    # FIXME: Implement length checking of mnu (should it be a certain length?)
```
philbull: FIXME here.

pyccl/neutrinos.py
```diff
@@ -25,41 +32,49 @@ def Omeganuh2(a, Neff, mnu, TCMB=2.725):
...
     
     # Check status and return
     check(status)
     if scalar: return OmNuh2[0]
     return OmNuh2
 
 
-def Omeganuh2_to_Mnu(a, Neff, OmNuh2, TCMB=2.725):
+def nu_masses(OmNuh2, label, ccl_TCMB=2.725):
     """Omeganu2h_to_Mnu
```
philbull: This function needs a description.

pyccl/neutrinos.py
```diff
@@ -25,41 +32,49 @@ def Omeganuh2(a, Neff, mnu, TCMB=2.725):
...
 
     Args:
-        a (float or array_like): Scale factor(s), normalized to 1 today.
-        Neff (float): Number of relativistic neutrino species (NB: to all practical purposes, Neff is simply N_nu_mass)
-        OmNuh2 (float): Neutrino energy density times h^2
+        OmNuh2 (float): Neutrino energy density at z=0 times h^2
+        label: indicates how the masses should be split up
```
philbull: Just flagging up this use of the word `label`, in case the terminology changes (see previous comments).

src/ccl_core.c
```diff
@@ -281,31 +288,113 @@ Omega_b: baryons
...
-
-  // Neutrinos
+  params.Neff = Neff;
+  
+  // If mnu is a sum of neutrino masses, set these by default to 
+  // obey the normal hierarchy.
+  if (mnu_type==ccl_mnu_sum){
+	  params.sum_nu_masses = *mnu;
+	  mnu=malloc(3*sizeof(double));
+	  if (params.sum_nu_masses>0.059){
```
philbull: Please add a comment saying where these numbers come from.

src/ccl_error.c
```diff
@@ -69,11 +69,16 @@ void ccl_check_status_nocosmo(int * status)
     fprintf(stderr,"%s", "CCL_ERROR_HMF_INTERP");
     exit(1);
```
philbull: I'm worried that these `exit()` statements could crash the Python wrapper. I will look into this (it's not your fault though).

src/ccl_neutrinos.c
```diff
@@ -95,114 +95,123 @@ INPUTS: a: scale factor, Neff: number of neutrino species, mnu: total mass in eV
...
+  
+  // Now split the sum up into three masses depending on the label given:
+  
+  if(label==ccl_nu_masses_normal_label){
+	  
+     double *mnu;
+	 mnu = malloc(3*sizeof(double));
+			
+	 if (sumnu>0.059){
+		 mnu[0] = (sumnu - 0.059) / 3.;
```
philbull: Mention where these numbers are from. If they are measured physical parameters, they should probably be stored with the other physical constants, like G.

include/ccl_core.h
```diff
@@ -141,19 +143,37 @@ typedef struct ccl_cosmology
...
+typedef enum ccl_mnu_type_label {
+  ccl_mnu_list=0,   // you pass a list of three neutrino masses
+  ccl_mnu_sum =1,  // sum, defaults to splitting with normal hierarchy
+  ccl_mnu_sum_inverted = 2, //sum, split with inverted hierarchy
+  ccl_mnu_sum_equal = 3, //sum, split into equal masses
+  // More options could be added here
+} ccl_mnu_type_label;
+
+
+// Label for which method you want to use to distribute the neutrino masses if you pass a sum.
```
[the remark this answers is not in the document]
c-d-leonard: no, removing.

pyccl/core.py
c-d-leonard: I did it this way because I feel like it might be confusion to say just 'inverted' and 'equal' - the user might think they are specifying that they need to pass a list of masses in agreement with the inverted hierarchy. This way they know they are passing a sum. Do you feel strongly about changing this?

CHANGELOG.md
```diff
@@ -0,0 +1,79 @@
...
+Old call signature:
+ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k,
+				     double N_nu_rel, double N_nu_mass, double mnu,
+				     double w0, double wa, double h, double norm_pk,
+				     double n_s, double bcm_log10Mc, double bcm_etab, double bcm_ks,
+				     int nz_mgrowth,double *zarr_mgrowth,
+					 double *dfarr_mgrowth, int *status)
+					 
+New call signature:
+ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k,
```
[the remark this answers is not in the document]
c-d-leonard: Sounds good, I wasn't sure what to do since this is the first time we've changed the API since the new API-change-documentation guidance so I was just winging it. I've made it shorter.

CHANGELOG.md
```diff
@@ -0,0 +1,79 @@
...
+C library:
+
+Changes in ccl_core.c:
+
+Old call signature:
+ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k,
+				     double N_nu_rel, double N_nu_mass, double mnu,
+				     double w0, double wa, double h, double norm_pk,
+				     double n_s, double bcm_log10Mc, double bcm_etab, double bcm_ks,
+				     int nz_mgrowth,double *zarr_mgrowth,
```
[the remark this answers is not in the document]
c-d-leonard: Fixed.

include/ccl_config.h
```diff
@@ -79,6 +79,18 @@ typedef enum mass_function_t
...
 
+/** 
+ * Emulator neutrinos typedef
+ * Specified whether, when the cosmic emulator is switched on,
+ * CCL should exit if non-equal neutrino masses are passed (strict)
+ * or equalize the masses, resulting in potential slight inconsistencies.
+*/
+typedef enum emulator_neutrinos_t
+{
+  ccl_strict   = 1,
```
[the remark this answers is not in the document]
c-d-leonard: Changed to ccl_emu_strict and ccl_emu_equalize

include/ccl_core.h
```diff
@@ -166,9 +186,9 @@ ccl_cosmology * ccl_cosmology_create_with_lcdm_params(
  * @param Omega_c Omega_c 
  * @param Omega_b Omega_b 
  * @param Omega_k Omega_k 
- * @param N_nu_rel Number of relativisitic species
- * @param N_nu_mass Number of massive neutrinos
- * @param mnu neutrino mass
+ * @param Neff Number of relativistic neutrino species in the early universe
+ * @param mnu neutrino mass, either sum or list of length 3
+ * @param mnu_is_sum determines nature of mnu, 1 for sum and 0 for list
```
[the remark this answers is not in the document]
c-d-leonard: Yes, fixed.

include/ccl_core.h
```diff
@@ -184,7 +204,14 @@ ccl_cosmology * ccl_cosmology_create_with_lcdm_params(
  * For specific cases see documentation for ccl_error.c
  * @return void
  */
-ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k, double N_nu_rel, double N_nu_mass, double mnu, double w0, double wa, double h, double norm_pk, double n_s, double bcm_log10Mc, double bcm_etab, double bcm_ks, int nz_mgrowth,double *zarr_mgrowth,double *dfarr_mgrowth, int *status);
+
+ccl_parameters ccl_parameters_create(double Omega_c, double Omega_b, double Omega_k,
+				     double Neff, double* mnu, ccl_mnu_type_label mnu_type,
```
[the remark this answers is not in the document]
c-d-leonard: Yes, changed to ccl_mnu_convention.

include/ccl_neutrinos.h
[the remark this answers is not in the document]
c-d-leonard: I made these labels more user-friendly, they are now 'ccl_nu_normal' etc. This is very similar to ccl_mnu_type_label (which is now renamed ccl_mnu_convention), but the difference is that this is used for telling the nu_masses function how to split up the total mass equivalent to Omeganu0 if at all, whereas ccl_mnu_convention tells the parameter creation function whether the neutrino mass parameter is a list or a sum and if so what to do with the sum. I could merge them, but there would be irrelevant labels for each case in the superset of labels. I tried to rename this one to be more clear instead as you suggested below.

CHANGELOG.md
c-d-leonard: Thanks, I forgot to document this. Added now.

pyccl/cls.py
c-d-leonard: Sure, changed.

pyccl/core.py
c-d-leonard: Good point, fixed.

pyccl/core.py
c-d-leonard: Fixed.

pyccl/core.py
c-d-leonard: I've added this now.

pyccl/neutrinos.py
c-d-leonard: Good idea, change made.

pyccl/neutrinos.py
c-d-leonard: This is a very good catch. We shouldn't be passing this in at all, it should be the lengths of mnu. I've fixed this. This is another API change, added to the changelog.

pyccl/neutrinos.py
c-d-leonard: See above. I am pretty sure (?) I didn't write this FIXME, but maybe I did. In any case, it's fixed now.

pyccl/neutrinos.py
c-d-leonard: Added.

pyccl/neutrinos.py
c-d-leonard: I changed this to "mass_split" to be more description.

src/ccl_neutrinos.c
c-d-leonard: I added them to the constants.h file. They're the experimentally measured mass splittings. I changed this a bit to make it clearer I think.

c-d-leonard:
@philbull Thanks for the review! I've made changes to address your request, see details in my comments on your review.

src/ccl_error.c
philbull: I fixed this in a different PR that's now been merged with `master`.

philbull:
It would be good to get one more review before committing this, because it's such an extensive change. I'm happy with it though.

doc/0000-ccl_note/main.tex
```diff
@@ -233,22 +233,41 @@ \subsection{Creating a cosmology}
...
+
+In the case where {\tt mnu} corresponds to a sum, we first allocate the three neutrino masses according to the requested convention (by default the normal hierarchy). Having then a set of three masses regardless of {\tt mnu$\_$type}, we check for which of the three masses the coresponding neutrino species is non-reativistic today ($m_\nu>0.00017$, \citealt{Lesgourgues2012}), and obtain a number of massive neutrinos {\tt N$\_$nu$\_$mass}. We use this along with the {\tt N$\_$eff} value to set the number of relativistc neutrinos species {\tt N$\_$nu$\_$rel}. We must be slightly careful in doing so, as for massive neutrinos only we follow {\tt CLASS} in modifying the relationship between the temperature of the CMB and the neutrino temperature:
```
christianel: In line 255, there are a couple of typos: It should be '...for which of the three masses the corresponding neutrino species is non-relativistic...' and 'the number of relativistic neutrino species'.

doc/0000-ccl_note/main.tex
```diff
@@ -233,22 +233,41 @@ \subsection{Creating a cosmology}
...
+\begin{equation}
+T_{\nu}^{\rm eff} = T_{\rm CMB} T_{\rm NCDM}.
```
christianel: Can you define here what you mean with T_{\rm NCDM}?

doc/0000-ccl_note/main.tex
```diff
@@ -300,19 +319,14 @@ \subsection{Density parameter functions}
...
+For massive neutrinos, $\Omega_{\nu, {\rm m}}(a)$ is calculated by calling a set of functions contained in {\tt ccl\_neutrinos.c}. For each species of massive neutrino with mass $m_\nu^i$, we define
 \begin{equation}
-\mu = \frac{m_{\nu}^{\rm tot}a}{N_{\rm m}^\nu T_{\nu}^{\rm eff}}
+\mu^i = \frac{m_{\nu}^{i}a}{T_{\nu}^{\rm eff}}
 \label{mnuOT}
 \end{equation}
-in units such that $\mu$ is dimensionless. We then conduct the phase-space integral required to get the neutrino density, and multiply by appropriate factors to obtain $\Omega_{\nu, {\rm m}}(a)$:
+in units such that $\mu$ is dimensionless. We conduct the phase-space integral required to get the neutrino density once, and store the result as a {\tt gsl} spline for subsequent access. Given then the value of this integral, we multiply by appropriate factors to obtain $\Omega_{\nu, {\rm m}}(a)$:
```
christianel: Should we add a reference here for the exact form of the phase-space integral?

doc/0000-ccl_note/main.tex
c-d-leonard: Fixed, thanks.

doc/0000-ccl_note/main.tex
c-d-leonard: Added a sentence defining this.

doc/0000-ccl_note/main.tex
c-d-leonard: It's given explicitly in the equation on the next line, I added a phrase clarifying this.

c-d-leonard:
Thanks @christianel and @philbull . I have added some changes to the note to address the issues Christiane noted.

include/ccl_core.h
```diff
@@ -141,19 +143,28 @@ typedef struct ccl_cosmology
   // other flags?
 } ccl_cosmology;
 
+// Label for whether you are passing a pointer to a sum of neutrino masses or a pointer to a list of 3 masses.
+typedef enum ccl_mnu_convention {
+  ccl_mnu_list=0,   // you pass a list of three neutrino masses
+  ccl_mnu_sum =1,  // sum, defaults to splitting with normal hierarchy
```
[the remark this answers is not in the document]
c-d-leonard: I consciously close to do this because I think we want a clear default option in C which corresponds to the default option in python. In python, if the user does not indicate how mnu should be split up but passes a single value (rather than a list or array), the sum is split into masses given the normal hierarchy (modulo the issue you pointed out above). But in C we don't have the capability to check the length of mnu, so the user must pass something for the convention. Therefore I wanted to use names that indicate which convention they should pick to get the same as the default behaviour in python. What do you think about this logic?

src/ccl_core.c
```diff
@@ -281,31 +288,108 @@ Omega_b: baryons
...
   params.Omega_k = Omega_k;
-
-  // Neutrinos
+  params.Neff = Neff;
+  
+  
+  if (mnu_type==ccl_mnu_sum){
+	  params.sum_nu_masses = *mnu;
+	  mnu=malloc(3*sizeof(double));
+	  mnu[0] = (params.sum_nu_masses - DELTAM12 - DELTAM13) / 3.;
```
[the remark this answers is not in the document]
c-d-leonard: Yes, I was aware that the mass differences are differences in mass squares. I implemented the splitting in this way following e.g. Lesgourgues & Pastor 2012 (1212.6154, see page 15 above equation 25) in approximating (Delta m_{31}) ~ sqrt(Delta m_{31}^2). Obviously, this is not exact, you're right. However, I suspect that what I am doing has less than our tolerance effect on the observables. I can check this.

c-d-leonard:
Thanks for the review @slosar . To respond to your overall comments:
a) Good point re: where the 93.14 eV comes from. I will change the notes to account for this.
b) (see my comment on your inline comment re potential bug in mass splitting)
c)  I agree this would be good and probably something we should implement in the future. Could you make an issue for this?
d) Yes, that's approximately correct. You won't get * exactly * 2 relativistic neutrinos, because of the thing with TNCDM, but pretty close to that.

I'll fix the text of the latex note and check into the mass splitting question as you pointed out.

src/ccl_core.c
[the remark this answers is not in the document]
slosar: Danielle, I still don't this this is right. I'll post on slack, since I cannot put images here.

c-d-leonard:
@slosar , I've now implemented the corrected equations for neutrino mass splittings that you requested. If you could take a look that would be great.

c-d-leonard:
I just added a test suggested by @slosar which checks that given a sum of neutrino masses, when requesting the normal or inverted hierarchy, the masses that are output recover the correct mass squared differences.

philbull:
This is good to merge! Danielle, please do the honours :)

philbull:
Danielle has been chasing down some failing tests, which led to the discovery of a few memory leaks (one quite major one) and some incorrect error handling behaviour. All of these issues should now be fixed by the latest commits: we added some extra checks to prevent uninitialised power spectra in `ccl_cosmology` causing problems, and I've run valgrind on `ccl_sample_run` to check that all of the memory leaks have been cleared up. I also did a bit of cleanup in `ccl_core` and `ccl_power` (the EH-related functions).

PR #288, opened 2017-11-21
Changed the Stefan-Boltzmann constant to be a derived parameter.

matthewkirby:
As I was writing I noticed that the Stefan-Boltzmann constant was hard-coded into ccl_constants.h.

I turned it into a derived parameter. The relative difference is 3e-12 of its previous value.

I was able to get SWIG running (unlike last time) by adding in `-I/path/to/GSL` and it runs with no errors. However, even after doing this, the C test that tests for updated swig files fails. All other tests pass. I had this issue in the past and @damonge fixed it somehow.

### DIFF
--- a/include/ccl_constants.h
+++ b/include/ccl_constants.h
@@ -69,7 +69,8 @@ extern "C" {
 /**
  * Stefan-Boltzmann constant in units of kg/s^3 / K^4
  */
-#define STBOLTZ GSL_CONST_MKSA_STEFAN_BOLTZMANN_CONSTANT
+//#define STBOLTZ GSL_CONST_MKSA_STEFAN_BOLTZMANN_CONSTANT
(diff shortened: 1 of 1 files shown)


damonge:
@matthewkirby : why is this necessary? In principle both numbers should agree modulo floating point errors. I'm just worried that we change a quantity from its arguably most exact value, provided by GSL.

matthewkirby:
Not entirely necessary. It comes down to deciding whether to be exactly consistent with GSL or exactly consistent with the rest of the constants. I figured I would write it in and then the decision could be made either way. 

In the end, you are right, the difference is extremely small so the change is negligible.

elisachisari:
Should we continue with the GSL value then? Can this PR be closed in that case?

Issue #293, opened 2017-12-03
HMF and halo bias python notebook needed


elisachisari:
We currently don't have any python notebooks showing examples of how to obtain the halo mass function and halo bias.

PR #294, opened 2017-12-03
Jupyter notebook for HMF and halo bias

elisachisari:
There is a new notebook under tests that gives examples of HMF and halo bias. This addresses issue #293


villarrealas:
Everything checks out here as far as I can tell!

elisachisari:
Great! Merging now.

PR #295, opened 2017-12-04
Automatic class installation

damonge:
first version, working in C, not in python yet


damonge:
Alright, this should be working now. There's a number of things to check:
- [ ] That travis passes
- [ ] That you can follow the new instructions in README.md regarding the CLASS installation, and that those instructions are sufficiently clear.
- [ ] That the class_install.py script runs without any problems on both linux and mac
- [ ] That the C library can be installed in both linux and mac
- [ ] That the python module can be installed in both linux and mac
- [ ] That the C and python unit tests pass.

Obviously I've checked all of these (at least in linux). But it'd be good if the reviewer could verify them.

rainwoodman:
I had a glance at `.travis.yaml` What about setting it up to use miniconda which comes with gsl and is expected to work on osx? - I suspect it's going to be the way most people get Python on osx anyways.

fjaviersanchez:
@rainwoodman Does miniconda include GSL2x or just GSL1x?

fjaviersanchez:
@damonge the tests keep failing because it is not finding the CCL library nor the CLASS library. I'll take a look at the install script to see what's going on.

rainwoodman:
https://anaconda.org/anaconda/gsl

It looks like 2.0.

fjaviersanchez:
@damonge `wget` is not default in MAC so maybe we should switch to `curl` or try `git clone` from the class repository?

fjaviersanchez:
@damonge I solved the `wget` part but also MacOS X is not compatible with openmp by default. I am trying to change that as well

damonge:
@fjaviersanchez @elisachisari @philbull : OK, travis checks are passing now. Can you guys check that you can install the library following the new procedure described in the README? It'd be good to check that things work at least in linux and mac

fjaviersanchez:
@damonge It installs beautifully in my laptop!

elisachisari:
@damonge. I tried installing in the usual way in my Mac, and got the following problem:
screen shot 2017-12-20 at 10 09 11 am

README.md
```diff
@@ -23,9 +23,15 @@ In order to compile `CCL` you need a few libraries:
...
 * [FFTW3](http://www.fftw.org/) is required for computation of correlation functions.
 * FFTlog([here](http://casa.colorado.edu/~ajsh/FFTLog/) and [here](https://github.com/slosar/FFTLog))is provided within `CCL`, with minor modifications.
+* The C library associated to the CLASS code. The installation of this library is described below.
+
+## Installing CLASS
+CCL uses CLASS as one of the possible ways of computing the matter power spectrum. In order to communicate with CLASS, CCL must be linked to its library. Before installing CCL proper you must therefore install this library first. Since this process is not necessarily straightforward, we provide a python script `class_install.py` that automatically downloads and install the latest tagged stable version of CLASS. You should run this script (`python class_install.py`) before carrying out the next steps. By default, the script assumes that your main C compiler is `gcc`. If that's not the case, pass the name of your C compiler to the script via the command-line argument `--c_comp` (i.e. `python class_install.py --c_comp=[name of compiler]`). Type `python class_install.py -h` for further details.
```
elisachisari: I think this script downloads a specific version of CLASS - v2.6.3. Do we want to give instructions to the user to potentially replace that with a newer version when available?

damonge: I'd say we should be in charge of keeping that up to date. If they want their own version of CLASS they should follow the steps for the custom CLASS install.

elisachisari: Ok

elisachisari:
If I start from scratch in a new CCL folder which does not have a previous installation, everything works fine. For any new user downloading CCL after this merge, there won't be problems. I suggest adding a comment in the README that if you have a previous installation of CCL which didn't have the CLASS patch, you might want to start from scratch by cloning a new repo. Or you could run the CLASS installation twice as I did.

damonge:
Sounds good, I'll do that

elisachisari:
@damonge - could you also edit the CCL note to match these changes before you merge?

damonge:
Note modified to describe the new CLASS installation

README.md
jchiang87: Should the section "Compiling against an external version of CLASS" be omitted or revised?

damonge: Not really. These modifications automatically download and install a tagged version of CLASS that we have chosen. If users want to use any other CLASS (e.g. hi-CLASS), they have to follow the instructions in that section

jchiang87: It still says "CCL has a built-in version of CLASS" (the first sentence of that section).  I guess I'm missing something, but I thought the built-in version has been removed.

damonge: Oh, I see. I'll fix that quickly. Can you review a quick PR?

PR #296, opened 2017-12-06
New CosmicEmu branch 

elisachisari:
This is almost the same as the old cosmic emu branch. The only difference is in the latest merge with master, which incorporates CMBlensing and was causing trouble in the previous branch. @philbull opened a new branch where merge conflicts with master are now (almost) resolved.


philbull:
I've now fixed up the remaining issues (to do with how errors causing integration errors are passed through to Python). This is now ready for a final review.

tests/ccl_sample_pkemu.c
```diff
@@ -0,0 +1,49 @@
...
+  double Omega_k = 0.;
+  double Neff = 3.04;
+  double Nmass = 1;
+  double mnu = 0.06;
+  
+  ccl_configuration config = default_config;
+  config.transfer_function_method = ccl_emulator;
+  config.matter_power_spectrum_method = ccl_emu;
+
+  //ccl_parameters params = ccl_parameters_create_flat_lcdm(Omega_c, Omega_b, h, normp, n_s, &status);
```
c-d-leonard: why is this line commented out? If it's not needed should we delete it?

tests/ccl_sample_pkemu.c
```diff
@@ -0,0 +1,49 @@
...
+    pnl = ccl_nonlin_matter_power(cosmo, k,1.0,&status);
+    p1nl = ccl_nonlin_matter_power(cosmo,k, a_at_z1,&status);
+    p2nl = ccl_nonlin_matter_power(cosmo,k, a_at_z2,&status);
+    p = ccl_linear_matter_power(cosmo, k,1.0, &status);
+    p1 = ccl_linear_matter_power(cosmo,k, a_at_z1,&status);
+    p2 = ccl_linear_matter_power(cosmo,k, a_at_z2,&status);
+    printf("%le %le %le %le %le %le %le\n", k, p,p1,p2,pnl,p1nl,p2nl);
+  }
+
+  /*printf("sigma_8 = %.6lE\n", ccl_sigmaR(cosmo,8./h,&status));
```
c-d-leonard: Do we want to leave these commented-out lines in?

tests/ccl_sample_pkemu.c
```diff
@@ -0,0 +1,44 @@
...
+{
+  int status=0;
+  double Omega_c = 0.25;
+  double h = 0.7;
+  double Omega_b = 0.022/h/h;
+  double normp = 0.8; //2.1e-9
+  double n_s = 0.96;
+  double Omega_k = 0.;
+  double Neff = 3.04;
+  double Nmass = 1;
```
c-d-leonard: I think that in speaking to Katrin we established that the emulator requires the total number of neutrinos to be fixed to 3.04. This sets the total number to 4.04 so I think it will give incorrect results.

tests/ccl_test_emu_nu.c
```diff
@@ -0,0 +1,196 @@
...
+  double Omega_b[4];
+  double n_s[4];
+  double h[4];
+  double w_0[4];
+  double w_a[4];
+};
+
+CTEST_SETUP(emu_nu) {
+  
+  data->N_nu_rel=3.04;
```
c-d-leonard: Again, I think this will set the total number of neutrinos to 4.04.

tests/ccl_test_pyccl_interface.py
```diff
@@ -29,17 +29,25 @@ def reference_models():
     p5 = ccl.Parameters(Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96)
     cosmo5 = ccl.Cosmology(p5,transfer_function='eisenstein_hu')
 
+    # Emulator Pk
+    p6=ccl.Parameters(Omega_c=0.27, Omega_b=0.022/0.67**2, h=0.67, sigma8=0.8, n_s=0.96,N_nu_rel=3.04,N_nu_mass=0.,m_nu=0.)
+    cosmo6 = ccl.Cosmology(p6,transfer_function='emulator',matter_power_spectrum='emu')
+    
+    # Emulator Pk w/neutrinos
+    p7=ccl.Parameters(Omega_c=0.27, Omega_b=0.022/0.67**2, h=0.67, sigma8=0.8, n_s=0.96,N_nu_rel=3.04,N_nu_mass=1,m_nu=0.06)
```
c-d-leonard: Again, concerned about setting total number of neutrinos to 4.04.

katrinheitmann:
Just one last comment on that: you can set Omega_nu=0 in the emulator, 
in which case you would have massless neutrinos.

elisachisari:
@katrinheitmann @elikrause @christianel: 

@c-d-leonard and I have been discussing this issue and here’s the summary of the situation:
- CCL implements the cosmic emulator of Lawrence et al. for the power spectrum with massive neutrinos.
- We need to make sure that our set-up of the cosmology that we pass the emulator via CCL is consistent with what was used for the original emulator. This is important because we want distances obtained by CCL to use the same cosmology as the emulator team used.
- In the case of neutrinos, we believe that the original configuration was 3.04 massive neutrino species and 0 massless neutrinos, in a manner consistent with the CAMB implementation (which we are not very familiar with).
- But we can’t really pass non-integer massive neutrinos to CCL.

So, given CCL options:

What values of N_nu_rel (number of massless neutrinos) & N_nu_mass (number of massive neutrinos) should we pass to the ccl_parameters_create routine to get cosmologies consistent with emulator set-up? 

And what value of Neff should we pass to the Omeganuh2 routines in ccl_neutrinos.c? Normally, this should be consistent with N_nu_mass, but the emulator configuration would suggest using 3.04.

elisachisari:
@c-d-leonard @katrinheitmann: We have been discussing how to call CCL in a manner that is compatible with the emulator set up. The emulator specified N_eff=3.04 and a value for Omega_nu. CCL on the other hand requires {N_nu_mass,N_nu_rel,m_nu}, where N_nu_mass should be an integer. For best compatibility with the emulator set-up, I've enforced that whenever CCL is called with the emulator option 'on', only two possibilities are allowed:

- {N_nu_mass,N_nu_rel,m_nu}={3,0.00641,m_nu} for massive neutrinos. Only 3 are allowed, not 2 nor 1. And m_nu is allowed to take any value compatible with the emulator Omega_nu allowed range.
- {N_nu_mass,N_nu_rel,m_nu}={0,3.04,0} for massless neutrinos.

We worried that setting N_nu_mass to be 3.04 or 3.046 would make a difference. I checked growth and distances and this doesn't make a difference above 1e-4 for a>0.01. But it can be greater otherwise. All of this is documented in the note.

Please take a look and let me know if you are happy with these changes. If you approve, I can then merge. Master has been recently merged onto this branch as well.

pyccl/neutrinos.py
```diff
@@ -1,33 +1,61 @@
...
     """Omeganuh2
     
     Args:
         a (float): Scale factor, normalized to 1 today.
         Neff (float): Number of relativistic neutrino species
         mnu (float): Neutrino mass (in eV)
-        TCMB (float): Temperature of the CMB (K)
+        TCMB (float, optional): Temperature of the CMB (K). Default: 2.725.
     Returns:
         Omeganuh2 (float or array_like)
```
c-d-leonard: I think it would be ideal if we could include a comment in the documentation here which makes clear the difference between Omeganuh2 and ccl_omega_x called for neutrinos (they are not the same as z!=0 and I don't think this is obvious from the name).

pyccl/neutrinos.py
```diff
@@ -1,33 +1,61 @@
...
 
 
-def Omeganuh2_to_Mnu(a,Neff,OmNuh2,TCMB):
+def Omeganuh2_to_Mnu(a, Neff, OmNuh2, TCMB=2.725):
     """Omeganu2h_to_Mnu
 
     Args:
         a (float or array_like): Scale factor(s), normalized to 1 today.
         Neff (float): Number of relativistic neutrino species
         OmNuh2 (float): Neutrino energy density times h^2
```
c-d-leonard: Again I think it would be good to make it very clear how this is different from ccl_omega_x in the doc here.

katrinheitmann:
So case for massless neutrinos is perfect now, thanks! I am still wondering though if it is more accurate to say for the massive case (3.0, 0, m_nu). I checked with CAMB, it basically doesn't make any difference for the transfer function and that's the only place where the N-body code gets to know about N (otherwise, it's Omega_nu that's entering). What do you you think?

elisachisari:
The difference would be in distances and growth, particularly at high redshift. If we set (3.0, 0, m_nu), we don't guarantee 3.04 (actually, 3.046) in the early universe. Do you think this is more consistent with the emulator neutrino cosmologies? (The paper, Lawrence+, says Neff=3.04, that's why we enforced this it.)

PR #297, opened 2017-12-06
Implemented hybrid linear-logarithmic sampling on a

christianel:
I have implemented a function "ccl_linlog_spacing" in src/ccl_utils.c that creates a number of logarithmically spaced (Nlog) bins and then linearly spaced bins (Nlin) in a. This function allows to use smaller scale factors (down to a=0.0001) compared to before, where a_min was 0.1. Two new splines (A_SPLINE_MINLOG and A_SPLINE_NLOG) are defined in include/ccl_params.ini and include/ccl_params.h. The function is then called in src/ccl_background.c in ccl_compute_distances and ccl_cosmology_compute_growth. 

In addition, I have tested different values of A_SPLINE_NA and A_SPLINE_NLOG by comparing the derived values for the growth factor, the growth rate, the comoving radial distance and h0 over h with those that I obtained when using A_SPLINE_NA and A_SPLINE_NLOG=10000. I found that A_SPLINE_NLOG=250 and A_SPLINE_NLIN=250 are already enough to guarantee that the relative error coming from using these smaller values is below 0.0001 everywhere for these four functions. 

In addition, I removed "A_SPLINE_DELTA" from include/ccl_params.ini, include/ccl_params.h and src/ccl_core.c because it was not used anywhere.


elisachisari:
@christianel @damonge: is this ready for review?

christianel:
@elisachisari @damonge : I have just updated the documentation (main.tex and authors.csv), so this is ready for review.

damonge:
still missing a unit test, but I'll make that today

damonge:
OK, all done. @elisachisari @tmcclintock : I would suggest merging this one  before #295 , which is mostly ready too, so let us know if you can review.

tmcclintock:
Looks good so far, but I think that there should always be a unit test in C and in Python. You've done the former, but the latter guarantees that the interface is working correctly. This is important because for people in my group, for instance, they will only end up using pyccl, so it has to be as rigorously tested as the C back end.

Btw I can't officially review it until assigned.

damonge:
@tmcclintock : I think this is gonna set an unrealistic standard for developers, given the workforce we have. This is certainly a super simple case for which we can easily build unit tests both in C and python, but this isn't always the case. I'm worried that if we always ask for this (which we haven't so far), it will end up putting people off and preventing new contributions. I'm sure there are instances where asking for this is very necessary, but since this PR hasn't touched the python modules and python is basically calling C, we're not really testing anything by adding a python unit test.

In any case, after this rant, I've generated a python unit test, so there you go ;-)

tmcclintock:
Perhaps. On the other hand python unit tests don't take up more than, like, a line or two. I guess doing it on a case-by-case basis makes sense, at least for now.

tests/ccl_test_growth_hiz.c
```diff
@@ -0,0 +1,119 @@
...
+  ccl_cosmology * cosmo = ccl_cosmology_create(params, default_config);
+  ASSERT_NOT_NULL(cosmo);
+  
+  // Compare to benchmark data
+  for (int j=0; j<7; j++) {
+    double a = 1/(1.+data->z[j]);
+    double gf_ij=ccl_growth_factor_unnorm(cosmo,a, &status);
+    if (status) printf("%s\n",cosmo->status_message);
+    double absolute_tolerance = GROWTH_HIZ_TOLERANCE*data->gf[model][j];
+    printf("%lE %lE %lE\n",data->z[j],fabs(data->gf[model][j]-gf_ij), absolute_tolerance);
```
elisachisari: Could we remove these print statements and all the ones that print to screen for tests 21-23?

include/ccl_utils.h
```diff
@@ -15,6 +15,10 @@ extern "C" {
  */
 double * ccl_linear_spacing(double xmin, double xmax, int N);
 
+/**
+ * Compute bin edges of N-1 logarithmically and then linearly spaced bins on the interval [xmin,xmax]
```
elisachisari: The header of the function in ccl_utils.h needs to follow the format of the other headers, so that the docs are processed automatically.

src/ccl_utils.c
```diff
@@ -30,6 +30,56 @@ double * ccl_linear_spacing(double xmin, double xmax, int N)
...
 
+/* ------- ROUTINE: ccl_linlog spacing ------
+ * INPUTS: [xminlog,xmax] of the interval to be divided in bins
+ *         xmin when linear spacing starts
+ *         Nlog number of logarithmically spaced bins 
+ *         Nlin number of linearly spaced bins 
+ * OUTPUT: bin edges in range [xminlog,xmax]
+ * */
+
+double * ccl_linlog_spacing(double xminlog, double xmin, double xmax, int Nlog, int Nlin)
```
elisachisari: The linlog spacing routine exits for some values of the parameters. Have you checked that CCL exits correctly and prints the right message for these cases both in C and python?

include/ccl_params.ini
```diff
@@ -13,11 +13,12 @@
 ;of filling these splines
 
 ;Scale factor spline
-A_SPLINE_DELTA=0.001
-A_SPLINE_NA=1000
-A_SPLINE_MIN=0.0003
+A_SPLINE_NA=250
```
elisachisari: I was confused at first by there being two "NA"-related parameters. We have  A_SPLINE_NA and N_A. The latter is in also in the ccl_splines struct and it is used for the power spectrum spline. Perhaps we could rename N_A to A_SPLINE_NA_PK to make the difference more clear?

elisachisari:
@christianel Thanks for implementing this. I gave you a review suggesting some minor additions/changes.

include/ccl_params.ini
christianel: Ok, I have renamed N_A to A_SPLINE_NA_PK (in include/ccl_params.h, include/ccl_params.ini, src/ccl_core.c, src/ccl_power.c and tests/ccl_test_cmblens.c).

include/ccl_utils.h
christianel: Ok, I have changed the header of the function in include/ccl_utils.h.

tests/ccl_test_growth_hiz.c
christianel: @damonge : Can I just remove these print statements or are they used for anything at the moment?

src/ccl_utils.c
christianel: There was indeed one case where CCL did not exit correctly. I have corrected the error in the linlog function in src/ccl_utils.c and checked that CCL exists correctly and prints the right message for all cases in C. @ damonge: What would you suggest to test whether CCL also exists correctly for all cases in python?

christianel: @damonge : What would you suggest to test whether CCL also exists correctly for all cases in python?

damonge: @elisachisari : this function is never called from python (the same way ccl_linear_spacing wasn't before), so the errors have to be caught at the C level. We have verified that we always check the output within C, and that if the function has failed, the error gets delivered to python.

PR #303, opened 2017-12-13
disabled full-sky shear correlation function calculation

elikrause:
This PR disables the full-sky calculation of xi+-, and prints an error message pointing the user to Kilbinger et al. 2017.
The corresponding ccl_paper subsection has been updated accordingly.


elikrause:
@sukhdeep1989 Sukhdeep, are you OK with this being merged, or are you working on a proper implementation at the moment?

NiallMac:
Just thought I'd flag in case useful - https://arxiv.org/abs/1704.01054 claims that the expression in Kilbinger 2017 is incorrect (and also see e.g. Stebbins 1996 https://arxiv.org/abs/astro-ph/9609149 which they probably should have cited!), so not sure how useful it is as a reference for this particular point.

elikrause:
Thanks @NiallMac , yes we also noticed that - it is corrected in the journal version

sukhdeep1989:
@elikrause Sorry, I missed your message. I am ok with merging for now. I have a rough implementation of wigner symbols but it requires cleanup and testing. We can put it in later once finalized. 

I had some trouble in compiling CCL in december, so kind of dropped the ball on this. I was able to compile on this branch, so will look into it. If I can't get into it sooner, this will be my hack day project :).

PR #306, opened 2018-01-05
3dcorrelation

zdu863:
Added new function ccl_correlation_3d to calculate 3d correlation function from power spectrum. New function added to `ccl_correlation.c` and new test file `ccl_3dcorr_sample.c` created.


dkirkby:
It looks like this new function assumes an isotropic P(kperp, kpar).  Are there plans to extend this to higher multipoles in order to handle redshift-space distortion?

zdu863:
Yes, we can look into this as a next step if this functionality is useful to have in CCL.

zdu863:
@slosar Thanks for the comments. We will work on revising the benchmark.

ellison-j:
Hi Anze @slosar . We have changed the implementation of the unit test based on your suggestion as follows:

1. Use absolute difference in r^2 xi(r)
3. Use CLASS/Halofit instead of linear BBKS, since we want to check the BAO peak 
2. For Delta log k = 1e-3 we find a maximum absolute difference in r^2 xi(r) of 0.17 for the range r = 0.1 - 100 Mpc for redshift zero. The relative error [Delta xi(r) / xi(r)] is 2.7e-3 at r = 50 Mpc, 7.4e-3 at 100 Mpc, and smaller for r < 50 Mpc.
3. In the range r = 50 - 250 Mpc, with Delta log k = 1e-3 and at redshift zero, we find a maximum absolute difference in r^2 xi(r) of 0.7, corresponding to 2% of the BAO peak value of r^2 xi(r). The absolute difference decreases with redshift.
4. Zilong has rewritten the CTEST to check the range r = 0.1 - 100 Mpc with 40 points and a tolerance in r^2 xi(r) of < 0.2 and the range r = 50 - 250 Mpc with 100 points and a tolerance in r^2 xi(r) of < 0.7. The tolerance for higher redshifts is lower.
5. If more accuracy is desired, the user can increase the number of points in k used for P(k) by changing N_K_3DCOR from the default value of 1000 to a bigger number. Very roughly, the precision increases by an order of magnitude if the number of points per decade is increased by a factor of 10.

Can you take another look please? If it is OK, we will go ahead and change the description in the note.

slosar:
Hi, 
Thanks for this extra work. So, what do you use as truth now? I think we should ramp the default N_K to something larger and perhaps a power of two (there is the FFT underneath so it might work better with 2^N, no)? I suggest N_k=131072 or 2^17, it is is not like one is going to waste much time on this and I'd rather have excessive default precision. 2% on BAO peak height is OK, but not great. Typically the ratio of error on position and error on height is 7 in spectro surveys, so for 1% position determination, we get 7% height determination. 2% is smaller than 7%, but not much smaller.

dkirkby:
You could use a P(k) -> xi(r) -> P(k) round trip to test accuracy.  I would also check against a power law with an analytic transform.

ellison-j:
@slosar and @dkirkby thanks for your review and comments. Anze: we are using a benchmark xi(r) calculated from CCL Halofit P(k) and a precise numerical transform form Tom Mclintock. We will change the default N_K to 131072. David: good idea, yes we will try this.

slosar:
I think checking against known analytic solutions is indeed a very good idea. I'm not so sure about round-trip, because I think it is going to work deceptively well due to perfect quadrature. However, going to xi, resampling at half-shifted points and going back might be an appropriate test, yes. Or even comparing two forward transforms where sampling has been shifted by half spacing in log.

ellison-j:
OK, good point. We will look into it and let you know.

ellison-j:
Dear @slosar and @dkirkby:

Here are the results of using a higher number of bins/per decade and using an analytic function to test the FFT. 

Use CLASS/Halofit, since we want to check the BAO peak. The benchmark is the CCL P(k) transformed using a precise numerical transform from Tom Mclintock. 

K range: default, K = 5e-5 to 1e3 Mpc^-1.  

Use N_K_3DCOR = 100,000 bins per decade sampling as new default.
We see no obvious difference using 2^17 and in the FFTlog paper they don't use 2^n. 

** Results:
For Delta log k = 1e-5 we find a maximum absolute difference in r^2 xi(r) of 2.7e-2 for the range r = 0.1 - 100 Mpc for redshift zero. The relative error [Delta xi(r) / xi(r)] is 1e-4 at r = 1-50 Mpc and 3e-4 at 100 Mpc.

In the range r = 50 - 250 Mpc, with Delta log k = 1e-3 and at redshift zero, we find a maximum absolute difference in r^2 xi(r) of 0.025, corresponding to 0.7% of the BAO peak value of r^2 xi(r). The difference at the peak (r = 150 Mpc) is <0.015, corresponding to 0.04% of the peak value. The absolute difference decreases with redshift.

** Accuracy tests using analytic function:

Used power law approximation of correlation function: xi(r) = (r/r_0)^a, where a = -1.67 and r_0 = 5 h^-1 Mpc with h=0.7.
The transform P(k) is known analytically.

Compared CCL calculation of xi(r) to known analytic transform result. The results are:

Relative difference Delta-xi/xi:
r = 0.1-0.8 Mpc: 10%
r = 1: 1%
r = 2 - 100: 0.3%
r = 150: 3e-3 = 0.3% 
Rising to 5% at r = 1000

The accuracy at low r (r=0.1-0.8 Mpc) can be improved by increasing the k range over which P(k) is sampled. Using K_MAX = 1e5 Mpc^-1 the relative difference at r = 0.1-2.0 is ~0.3%.

However, we are now having some issue with the 3dcorrelation branch. Something may have gotten corrupted since now we get inconsistent results running the CTEST using 'make check' compared to running a standard c program containing the same code.   Zilong is trying to figure it out. @elisachisari  any idea how to tell if there was some problem when it was merged with master, or at some other stage?

dkirkby:
I guess you are already making some useful diagnostic plots for these tests. Could you post them here?

ellison-j:
Here are the plots showing the benchmark tests:
image
image

And this plot shows the result of the analytical function test:
image

slosar:
This all looks reasonable to me. It is clear that power-law will be bad at the edges of transform. I think this accuracy is sufficient for default as long as it is documented.

dkirkby:
Thanks for the plots @ellison-j.  I agree that this looks reasonable for FFTlog.  Can these plots be added to the CCL note or sphinx docs?  Are both of these comparisons included in the travis tests now, with some threshold on the required accuracy?

tmcclintock:
@dkirkby I'll include them in the RTD documentation once they are merged into master.

ellison-j:
Thanks @dkirkby and @tmcclintock. We are having an issue with the unit test, which will hopefully be resolved in the next day or so. At the moment the ctest code uses the first benchmark with three cosmology models and redshift 0-5 with thresholds dependent on redshift. Yes, I guess we could add these plots to the note; I'll check with @elisachisari to see if she is ok with it.

ellison-j:
Hi @slosar and @dkirkby: just to let you know the status, we have resolved the issue with the C unit test and have committed it to github. The problem was not in our code but tracked down to another C unit test (cmblens) which was overwriting the power spectrum splines. The unit test has now been fixed (thanks to Elisa and David) in both the master and 3dcorrelation branches, so now everything works ok. I am now updating the note and adding the plots as per your suggestion which Elisa is happy with.

slosar:
Ok, this looks good enough for me.

ellison-j:
Thanks for approving @slosar. I've updated the note now. It should be ready to merge is you approve @dkirkby.

ellison-j:
Thanks @dkirkby and @slosar. I've merged into master.

PR #313, opened 2018-02-02
Fix issue with `pip` always trying to reinstall `mkauthlist` dependencies

philbull:
Myself and other people have been having problems with `mkauthlist` always trying to change the `numpy` version through pip and other annoyances.

This PR:
- Adds test in Makefile to see if `mkauthlist` command exists, instead of just trying to `pip upgrade` it every time we run `make`.
- Allows the user to manually specify the path to the `mkauthlist` command themselves, by setting the `MKAUTHLIST` environment variable.

Example usage:
```sh
$ make
Makefile:107: *** "mkauthlist not found. Install using 'pip install mkauthlist'". Stop.
```

```sh
$ MKAUTHLIST=/home/mariecurie/.local/bin/mkauthlist make
<makefile runs>
```

### DIFF
--- a/doc/0000-ccl_note/Makefile
+++ b/doc/0000-ccl_note/Makefile
@@ -78,10 +78,15 @@ tar : main
 	cd ${tardir} && tar -czf ../${outname}.tar.gz . && cd ..
 	rm -rf ${tardir}
 
-authlist :
-	pip install --upgrade mkauthlist
+authlist : mkauthlist-exists
+	#pip install --upgrade mkauthlist
+ifdef MKAUTHLIST
+	$(MKAUTHLIST) -j ${style} -f -c "LSST Dark Energy Science Collaboration" \
+		--cntrb contributions.tex authors.csv authors.tex
+else
 	mkauthlist -j ${style} -f -c "LSST Dark Energy Science Collaboration" \
 		--cntrb contributions.tex authors.csv authors.tex
+endif
 
 # http://stackoverflow.com/q/8028314/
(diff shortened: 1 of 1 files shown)


elisachisari:
Pinging @kadrlica @drphilmarshall regarding this PR, as these might be useful modifications for mkauthlist beyond CCL.

kadrlica:
@elisachisari I'm pretty sure that @abmantz has already made similar changes in `start_paper` (see [here](https://github.com/LSSTDESC/start_paper/blob/master/Makefile#L38-L49)). 

Also, just for clarity, these are *not* issues/changes to `mkauthlist`, but rather to the `start_paper` Makefile. I would suggest changing the title of this PR to reflect that.

philbull:
@kadrlica Thanks, I've changed the title to be clearer. Part of the issue seems to be that `pip` doesn't install the `mkauthlist` command into the default path. Is this a known/intentional behaviour?

kadrlica:
Thanks @philbull. There are some issues surrounding users who do not have root access to the python default systems path.  Another issue is the `--upgrade` flag, which was originally requested by @drphilmarshall to make sure that the version of `mkauthlist` was kept current; however, my personal opinion is that we shouldn't be upgrading `mkauthlist` without the user specifically asking for it.

kadrlica:
You can see how Adam solved the mkauthlist path issue in updates to start_paper [here](https://github.com/LSSTDESC/start_paper/blob/master/Makefile#L38-L49).

philbull:
Thanks folks, merging this now.

PR #211, opened 2017-06-30
Tidy up and improve tests

philbull:
This pull request makes some minor changes to the way tests are performed.
- Moves example code out of the `tests/` directory and into a new `examples/` directory
- Generalises some tests to use CLASS P(k) as well as BBKS P(k)
- Adds validation code for CLASS P(k) accuracy, using high-precision CLASS run as a reference

PR #331, opened 2018-02-14
Move example code/scripts to a separate directory

philbull:
Moves example code to its own directory, away from the tests, and updates the README and Makefile as appropriate.

Also removes some old entries in the Makefile, and some legacy examples that were still hanging around.


philbull:
(This replaces some parts of PR #211, which is outdated.)

elisachisari:
Looks good. Thanks for sandbox additions too. Could you edit the README and CCL note to match your changes?

philbull:
The README should be updated, but I forgot about the note. I'll fix that.

philbull:
OK, done.

philbull:
I checked, and I don't see any mention of the sandbox files anywhere else. They were very old, and didn't look like functioning code anyway. So I think we're good to drop them. I'll merge into `master` now. Thanks!

Issue #324, opened 2018-02-09
pyccl calls for sigmaR (and possibly others) fail on integer passes


villarrealas:
The function sigmaR takes parameters cosmo + a sphere size R. However, if it is passed an integer for R (something like 8), it simply crashes because it cannot take an integer value. Since it is feasible people will want to pass it something like linspace with integers, this might be a concern?

I suspect this may be elsewhere in pyccl wrapper work.

tmcclintock:
@philbull wrote the vectorization wrappers. I think the cleanest thing to do would be to do type casting there. What do you think, Phil?

philbull:
This should now be fixed in PR #334 -- please review!

philbull:
Fixed and merged.

PR #334, opened 2018-02-19
Do type check for integers in Python wrapper, to prevent crashes

philbull:
Fixes issue #324, crashes caused by improper handling of integer arguments.


philbull:
OK, merged. Thanks!

Issue #337, opened 2018-02-20
Different transfer_functions and matter_power_spectrum


alexander-mead:
When I generate a cosmology object I have the option to specify a ```transfer_function``` and a ```matter_power_spectrum```. I could not find an explanation of the difference between these two things in either the README, Doxygen docs or the pdf note that can be generated (000-ccl_note.pdf).

When specifying a cosmological model I can understand that one needs to specify a means of generating the linear power spectrum, and possibly a non-linear power spectrum, but I'm not sure how the ```transfer_function``` or ```matter_power_spectrum``` attributes of the cosmology structure map onto this.

It was also not clear to me what to make of the list of ```transfer_function``` keys, which includes 'emulator' and 'fitting_function'... ?

Apologies if this question is moronic or if I've missed something critical from a section of the documentation.

philbull:
Thanks Elisa, this is super useful! When you say `matter_power_spectrum=any` in the last two items, there are situations where this would fail I thought? e.g. `transfer_function=ccl_eisenstein_hu` and `matter_power_spectrum=ccl_emu` wouldn't work?

I think having the flexibility you mentioned is worthwhile, but I feel that it's too confusing to have separate lists for the `transfer_function` and `matter_power_spectrum` settings when they are often (but not always) tightly coupled to one another. As an alternative, could we replace these two settings with a single `power_spectrum` setting (the name `matter_power_spectrum` is too verbose!), and just add each possible power spectrum + transfer function pair as a separate option in the list?

For example, the combinations you listed above would turn into:
```
ccl_class_linear, ccl_class_halofit
ccl_emulator
ccl_bbks_linear, ccl_bbks_halofit, ...
ccl_eh_linear, ccl_eh_halofit...
````
I got rid of some duplicate definitions, like `fitting_function`, to keep the list short. (We could potentially even get rid of the `linear` settings, as I'm not sure what the use-case for those is given that we have separate `linear` and `nonlin` matter power spectrum functions.)

That's only a handful of combinations, and there's no risk of choosing an incompatible transfer/matter power spectrum pair, as only allowed combinations are in the list. It's also easy to add new methods, e.g. `ccl_emulator_ostrich` or whatever. Would this alternative work for you? @elisachisari @alexander-mead ?

alexander-mead:
Okay, I think I understand this a bit better now. I think my suggestion would be the following:

[...]

PR #339, opened 2018-02-21
Halo model for power spectrum calculation

villarrealas:
This branch works to add a halo model calculation and refactor some of the code in ccl_massfunc as a result of this. This will allow a new method of power spectrum calculation.

I think one key thing prior to pushing to master will be to develop benchmarks of this for validation purposes.


src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+#include "ccl_params.h"
+#include "ccl_emu17.h"
+#include "ccl_emu17_params.h"
+
+double delta_c(){
+  // Linear collapse threshold for haloes
+  return 1.686;
+}
+
+// TODO: referred to as odelta in ccl_massfunc, as a free parameter.
```
elikrause: should these TODO items be addressed before the pull request can be approved?

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+double Delta_v() {
+  // Halo mean density
+  return 200.;
+}
+
+// TODO: Check growth factor is normalised g(a=1)=1
```
elikrause: yes, you're calling the growth factor normalized as g(a=1) =1
ccl_growth_factor_unnorm would be the the version with g(a) =a during matter domination

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+double f_damp(ccl_cosmology *cosmo, double a, int * status){
+  // HMcode parameter: f  
+  double sigma8z = ccl_sigma8(cosmo, status)*ccl_growth_factor(cosmo, a, status);
+  return 0.188*pow(sigma8z, 4.29);
+}
+
+//TODO: Actually code this up
```
elikrause: This TODO should probably be included for the PR, no?
Or is this only for HMCode, not the generic halo model?

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+  return pow(halomass*3.0/(4.0*M_PI*rho_matter*Delta_v()),1.0/3.0);
+}
+
+double r_Lagrangian(ccl_cosmology *cosmo, double halomass, double a, int * status){
+  // Calculates the halo Lagrangian radius as a function of halo mass
+  double rho_matter = comoving_matter_density(cosmo);  
+  return pow(halomass*3.0/(4.0*M_PI*rho_matter),1.0/3.0);
+}
+
+// TODO: Why is 'sinl' used in the below routine as well as 'sin'?
```
elikrause: in case this question is based on my code snippets: you're free to use either, I had just found sinl to be more stable for small arguments.

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+  // Various bits for summing together to get final result
+  f1 = sin(ks)*(gsl_sf_Si(ks*(1.+c))-gsl_sf_Si(ks));
+  f2 = cos(ks)*(gsl_sf_Ci(ks*(1.+c))-gsl_sf_Ci(ks));
+  f3 = sinl(c*ks)/(ks*(1.+c));
+  fc = log(1.+c)-c/(1.+c);
+  
+  return (f1+f2-f3)/fc;
+}
+
+// TODO: move this into ccl_massfunc as standard function conversion.
```
elikrause: yes, it would be good to move this before the PR is approved

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+  double neff = collapse_index(cosmo, a, status);
+  return 2.93*pow(1.77,neff);
+}
+
+double comoving_matter_density(ccl_cosmology *cosmo){
+  // The comoving density of matter. This is a constant. Units are Msun/Mpc^3 (no factors of h)
+  return RHO_CRITICAL*cosmo->params.Omega_m*cosmo->params.h*cosmo->params.h;
+}
+
+// TODO: possible that delta should be passed around for consistency checks
```
elikrause: agreed!

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+  return delta_c()/ccl_sigmaM(cosmo, halomass, a, status);
+}
+
+// TODO: Actually write this
+double z_formation_Bullock(ccl_cosmology *cosmo, double halomass, double a, int * status){
+  // Halo "formation redshift" calculated according to the Bullock et al. (2001) prescription
+  return 0.;
+}
+
+// TODO: make consistency check so that ccl_halo_concentration only runs if called with appropriate definition
```
elikrause: yes, this should be implemented before merging into master!

src/ccl_halomod.c
```diff
@@ -0,0 +1,300 @@
...
+  double wk_squared = pow(u_nfw_c(p->cosmo,c,halomass,p->k,p->a,p->status),2); 
+
+  // The mean background matter density in Msun/Mpc^3
+  //double rho_matter = RHO_CRITICAL*p->cosmo->params.Omega_m*pow(p->cosmo->params.h,2);
+  double rho_matter = comoving_matter_density(p->cosmo);
+
+  // Fairly sure that there should be no ln(10) factor should be here since the integration is being specified in log10 range
+  double dn_dlogM = ccl_massfunc(p->cosmo,halomass,p->a,Delta_v(),p->status);
+    
+  return pow(halomass,2)*dn_dlogM*wk_squared/pow(rho_matter,2);
```
elikrause: isn't there a factor of halo mass, and possibly some numeric prefactor missing to convert d M -> d lg_10(M)?

alexander-mead: I think not in this case, because the integral is done over a range that is set in log10 I think this cancels out in the end. Probably there is a ln(10) that is missing in both a numerator and denominator. I'll think about whether it is better to include this explicitly to make what is going on in the code more transparent...

alexander-mead: Certainly there is no factor of mass missing. I think the difference is in our definition of the halo profile, where mine carries a mass scaling. So I have W(M,k->0)=M/\bar\rho whereas you have U(M,k->0)=1.

.gitignore
```diff
@@ -37,3 +44,15 @@ tests/.ipynb_checkpoints/
 
 # Ignore VCS files
 .idea/*
+
+# Ignore class directory
+class/
+Mead/
```
damonge: I imagine there's a bunch of these things that won't be there in the end

compile_halomod_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+gcc -Wall -Wpedantic -g -O0 -I/usr/local/include -std=gnu99 -fPIC tests/ccl_test_halomod.c -o tests/ccl_test_halomod -L/usr/local/lib -lgsl -lgslcblas -lm -Lclass -lclass -lccl
```
damonge: This whole file shouldn't be in the repo in the end

include/ccl_halomod.h
```diff
@@ -0,0 +1,24 @@
...
+extern "C" {
+#endif
+
+#pragma once
+
+#include "ccl_core.h"
+
+  /*
+double u_nfw_c(ccl_cosmology *cosmo, double c, double halomass, double k, double a, int * status);
+  */
```
damonge: Remove comment

include/ccl_power.h
```diff
@@ -6,6 +6,8 @@ extern "C" {
 
 #include "ccl_core.h"
 
+double ccl_comoving_matter_density(ccl_cosmology *cosmo);
```
damonge: Do you need these after Shahab's modifications?

include/ccl_power.h
```diff
@@ -64,6 +66,19 @@ double ccl_nonlin_matter_power(ccl_cosmology * cosmo, double k, double a,int * s
  * @return sigma(R).
  */
 double ccl_sigmaR(ccl_cosmology *cosmo, double R, int * status);
+double ccl_sigmaRz(ccl_cosmology *cosmo, double R, double a, int * status);
```
damonge: I'd suggest to actually replace sigmaR with this function
It also seems like you're only making sigmaR (and sigmaV) transparent to python, but I would actually do it for the more general ones. That's why I would just replace sigmaR with a new sigmaR that also takes in an 'a' parameter. This could be made to default to a=1 at the python level. Does this make sense?

pyccl/power.py
```diff
@@ -46,6 +46,19 @@ def sigmaR(cosmo, R):
     return _vectorize_fn(lib.sigmaR, 
                          lib.sigmaR_vec, cosmo, R)
 
+def sigmaV(cosmo, R):
```
damonge: This is what I meant above, you could just add an optional argument a=1 here and pass it to C.

src/ccl_background.c
```diff
@@ -14,6 +14,10 @@
 #include "gsl/gsl_roots.h"
 #include "ccl_params.h"
 
+// The comoving density of matter. This is a constant in time. Units are Msun/Mpc^3 (no factors of h)
+double ccl_comoving_matter_density(ccl_cosmology *cosmo){
```
damonge: Just repeating my comment about shahab's implementation

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
+#include "ccl_utils.h"
+#include "math.h"
+#include "stdio.h"
+#include "stdlib.h"
+#include "gsl/gsl_integration.h"
+#include "gsl/gsl_interp.h"
+#include "gsl/gsl_spline.h"
+#include "gsl/gsl_sf_expint.h"
+#include "ccl_placeholder.h"
```
damonge: Why do you need this?

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
...
+#include "ccl_background.h"
+#include "ccl_power.h"
+#include "ccl_massfunc.h"
+#include "ccl_error.h"
+#include "class.h"
+#include "ccl_params.h"
+#include "ccl_emu17.h"
+#include "ccl_emu17_params.h"
+
```
damonge: Do you really need all of those includes?

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
...
+// Set two-halo term
+// 1 - Standard two-halo term
+// 2 - Linear theory
+int i2h=1;
```
damonge: No global variables!!!!

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
...
+double mmin=1e7;
+double mmax=1e17;
+
+// Virial density of haloes
+double Delta_v=200.;
+
+// HMcode parameter: one-halo damping wavenumber: k*  
+double k_star(ccl_cosmology *cosmo, double a, int * status){  
+  return 0.584/ccl_sigmaVz(cosmo, 0., a, status);
+}
```
damonge: If a function is not gonna be used by users, it should be made static.
Also, it seems like you're not even using it, so get rid of anything you don't use.

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
...
+double collapse_index(ccl_cosmology *cosmo, double a, int * status){
+  return -2.;
+}
+
+// HMcode parameter: smoothing for transition between one- and two-halo regimes: alpha
+double alpha(ccl_cosmology *cosmo, double a, int * status){  
+  double neff = collapse_index(cosmo, a, status);
+  //return 2.93*pow(1.77,neff); //This is the relation from Mead et al. (2015)
+  return 3.24*pow(1.85,neff); //This is the relation from Mead et al. (2016)
+}
```
damonge: Same for all of these functions

src/ccl_halomod.c
```diff
@@ -0,0 +1,428 @@
...
+
+// analytic FT of NFW profile, from Cooray & Sheth (2002; Section 3 of https://arxiv.org/abs/astro-ph/0206508)
+// Normalised such that U(k=0)=1
+double u_nfw_c(ccl_cosmology *cosmo, double c, double halomass, double k, double a, int * status){
```
damonge: Again, make this static.

.gitignore
alexander-mead: Yes, what is the general way of dealing with this type of thing? Basically I want a .gitignore that applies to the master CCL and that I don't mess around with, but then I also want another .gitignore that avoids any extra files or folders I may be messing around with in the main CCL directory. Or should I just not be adding files or folders here at all?

src/ccl_halomod.c
alexander-mead: Okay, these functions I added in because they pertain to the HMcode that I plan to add at a later date. So they will eventually be used. What is the protocol here? Should I make a new branch for HMcode and just leave all this half-finished stuff there? Or is it okay to leave the half-finished functions here for the time being?

include/ccl_power.h
alexander-mead: Yes, let me try and do this.

The only concern I have relates to the approximation of scale-independent growth here. This somehow relates to my issue #344 . Clearly this is wrong in massive-neutrino cosmologies, but it is also slightly wrong all the time anyway because the baryons and CDM have different power spectra really. I guess that by using the approximation of scale-independent growth here you run the risk of someone calling ```sigmaR(z=1000)``` and getting some answer the meaning of which is debatable.

include/ccl_halomod.h
alexander-mead: Done

compile_halomod_test.sh
alexander-mead: Done

.gitignore
alexander-mead: Done, although I kept the .gitignore for the class directory. Is that okay?

include/ccl_power.h
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Removed

src/ccl_halomod.c
alexander-mead: Removed

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Removed things I don't use (into a new branch) and made other functions explicitly static.

alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

include/ccl_power.h
alexander-mead: No, not really. Is the correct thing to do to 'rebase' my branch on the now updated master branch so that I can have access to Shahab's new density functions? I tried to do this but ran into conflicts that were to do with files created by SWIG. Is there some general preferred method for rebasing, or any idea how to deal with the conflicts?

include/ccl_power.h
alexander-mead: How do I make sure that a=1 is the default with python? What files do I have to change to make this happen? Or is it going to be easier to change the sigmaR function so that the user needs to specify a?

pyccl/power.py
alexander-mead: Okay, how exactly do I do this? As in, how do I make the argument optional in python? Or is it better just to change the sigmaR function that is exposed and make it need the use to specify an 'a'?

src/ccl_halomod.c
alexander-mead: Done

alexander-mead: Although, what is the best thing to do here? There are some variables that I would like to be accessed by multiple functions. It seems laborious to pass them all explicitly between all the routines that might use them, and repeating their definitions in each routine seems like it could lead to errors.

include/ccl_power.h
alexander-mead: Done

pyccl/power.py
alexander-mead: DOne

include/ccl_power.h
alexander-mead: Done now

src/ccl_background.c
alexander-mead: Done

EiffL:
I've tried to resolve all the conflicts, all the python tests pass, and only the `tests/ccl_test_power_nu.c` fails in check_ccl, with the following error:
```
TEST 60/65 power_nu_nl:model_3 [OK]
TEST 61/65 power_nu_nl:model_2 ERROR 1032: ccl_power.c: ccl_cosmology_compute_power_class(): Error computing CLASS power spectrum
```

pyccl/power.py
```diff
@@ -31,19 +31,35 @@ def nonlin_matter_power(cosmo, k, a):
...
+        float or array_like: RMS variance in the density field in top-hat sphere; Mpc.
 
     """
-    return _vectorize_fn(lib.sigmaR, 
-                         lib.sigmaR_vec, cosmo, R)
+    return _vectorize_fn2(lib.sigmaR, 
+                         lib.sigmaR_vec, cosmo, R, a)
+
+def sigmaV(cosmo, R, a=1.):
+    """RMS variance in the displacement field in a top-hat sphere of radius R.
```
philbull: Would be good to define displacement field.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+double ccl_halo_concentration(ccl_cosmology *cosmo, double halomass, double a, int * status)
+{
+
+  // Set concentration-mass relation
+  // 1 - Bhattaharya et al. (2011)
+  // 2 - Full Bullock et al. (2001)
+  // 3 - Virial Duffy et al. (2008)
+  // 4 - Constant concentration (useful for testing)
+  // 5 - Simple Bullock et al. (2001)
+  int iconc=3;
```
philbull: This seems to be hard-coded. It would be good to add this as a new option, with an `enum` somewhere to label each option with a keyword (c.f. the `matter_power_spectrum_method` in `ccl_power.h`).

philbull: Although I wouldn't add this as an option into `ccl_config`, as there don't seem to be any splines/precomputations going on here. It's fine to just add a new argument to the function.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  // Simple Bullock et al. (2001) relation
+  else if(iconc==5){
+    return 9.*pow(halomass/Mstar(),-0.13);
+  }
+
+  // Something went wrong
+  else{    
+    exit(0);
```
philbull: This should raise an error. See `ccl_error.c` for helper functions to help you do that. Here are the essentials:

1. Define a new error variable in `ccl_error.h`, e.g. `#define CCL_ERROR_HALOCONC xxx`
1. Set the status variable `*status = CCL_ERROR_HALOCONC;`
1. Raise an error, using `ccl_raise_exception(int err, char* msg)`.
1. Return `NAN`.

Calling `exit()` is bad because it crashes the Python interpreter :)

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+
+// Parameters structure for the one-halo integrand
+typedef struct{  
+  ccl_cosmology *cosmo;
+  double k, a;
+  int * status;
+} Int_one_halo_Par;
+
+// Integrand for the one-halo integral
+static double one_halo_integrand(double log10mass, void *params){
```
philbull: Lots of static functions here, but not sure why.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  double b = ccl_halo_bias(p->cosmo,halomass,p->a,Delta_v,p->status);
+    
+  return b*dn_dlogM*wk;
+}
+
+// The two-halo term integral using gsl
+static double two_halo_integral(ccl_cosmology *cosmo, double k, double a, int *status){
+
+  double mmin=1e7; // Minimum mass for the halo-model integration
+  double mmax=1e17; // Maximum mass for the halo-model integration
```
philbull: Hard-coded variables again.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  return result;
+}
+
+// Computes the two-halo term
+double ccl_twohalo_matter_power(ccl_cosmology *cosmo, double k, double a, int *status){
+
+  // Set two-halo term
+  // 1 - Standard two-halo term
+  // 2 - Linear theory
+  int i2h=1;
```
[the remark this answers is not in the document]
philbull: Also, better to use an `enum` so that you can name each of these options, instead of using the numbers explicitly.

src/ccl_massfunc.c
```diff
@@ -183,18 +183,36 @@ TASK: Outputs fitting function for use in halo mass function calculation;
...
   double sigma=ccl_sigmaM(cosmo, halomass, a, status);
   int gslstatus;
   
   switch(cosmo->config.mass_function_method) {
+
+  //Equation (10) in arxiv: 9901122
+  // Note that Sheth & Tormen use nu=(dc/sigma)^2 whereas we use nu=dc/sigma
+  case ccl_shethtormen:
+
+    //TODO: Mead: Somehow enforce that this should only work with a virial-density Delta_v
```
philbull: Does this comment need action?

src/ccl_massfunc.c
```diff
@@ -183,18 +183,36 @@ TASK: Outputs fitting function for use in halo mass function calculation;
...
+
+    //ST mass function fitting parameters
+    fit_A = 0.21616;
```
philbull: I also wonder if these should be defined in the header instead of hard-coded.

src/ccl_power.c
```diff
@@ -1638,30 +1638,55 @@ typedef struct {
...
 {
-  SigmaR_pars *par=(SigmaR_pars *)params;
-  
-  double k=pow(10.,lk);
-  double pk=ccl_linear_matter_power(par->cosmo,k, 1.,par->status);
-  double kR=k*par->R;
   double w;
+
   if(kR<0.1) {
     w =1.-0.1*kR*kR+0.003571429*kR*kR*kR*kR
```
philbull: Would be good to document where these magic numbers come from.

examples/Halo model.ipynb
```diff
@@ -0,0 +1,281 @@
+{
```
philbull: Comments about the notebook:
1. Should use only numpy math functions instead of doing `import math`.
1. Should add units too the plots.
Otherwise, looks good!

alexander-mead: Done

src/ccl_power.c
alexander-mead: Done

src/ccl_power.c
```diff
@@ -1681,10 +1706,39 @@ double ccl_sigmaR(ccl_cosmology *cosmo,double R, int *status)
...
+																				0.0, ccl_gsl->INTEGRATION_SIGMAR_EPSREL,
+																				workspace,&sigma_V,NULL,NULL);
+
+  if(gslstatus != GSL_SUCCESS) {
+    ccl_raise_gsl_warning(gslstatus, "ccl_power.c: ccl_sigmaV():");
+    *status |= gslstatus;
+  }
+
+  //TODO: log10 could be taken already in the macros.
+  //TODO: 1E-5 should be a macro
```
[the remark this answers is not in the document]
alexander-mead: Hmm... I didn't code this up. @villarrealas, do you have any input?

doc/0000-ccl_note/main.tex
```diff
@@ -883,11 +883,84 @@ \subsection{Halo mass \& halo bias functions}
...
+where $M$ is the halo mass, $\mathrm{d}n/\mathrm{d}M$ is the halo mass function defined as the first of equations~(\ref{eq:halo_bias_and_mass_function}) and $b(M)$ is the linear halo bias with respect to the linear matter density field, defined as the large-scale limit of the second of equations~(\ref{eq:halo_bias_and_mass_function}).
+
+Equations~(\ref{eq:two_halo}) and (\ref{eq:one_halo}) contain the (spherical) Fourier transform of the halo profile, or halo `window function':
+\begin{equation}
+W_i(M,k)=\int_0^\infty4\pi r^2\frac{\sin(kr)}{kr}\rho_{\mathrm{H},i}(M,r)\;\mathrm{d}r\ ,
+\label{eq:window_function}
+\end{equation}
+where $\rho_{\mathrm{H},i}(M,r)$ is the radial profile for the field $i$ in a host halo of mass $M$. For example, if one is interested in matter fields then this would be the halo density profile, if one were interested in galaxies then this would be the number density and distribution of galaxies around a halo of mass $M$.
+
+%The dimensionless mass function, $f(\nu)$, is related to $n(M)$ via
```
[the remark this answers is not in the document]
alexander-mead: Done

doc/0000-ccl_note/main.tex
```diff
@@ -883,11 +883,84 @@ \subsection{Halo mass \& halo bias functions}
...
+\begin{equation}
+\frac{1}{\bar\rho_\mathrm{m}}\int_0^\infty M\frac{\mathrm{d}n}{\mathrm{d}M}\;\mathrm{d}M=1\ ,
+\label{eq:mf_normalisation}
+\end{equation}
+\begin{equation}
+\frac{1}{\bar\rho_\mathrm{m}}\int_0^\infty Mb(M)\frac{\mathrm{d}n}{\mathrm{d}M}\;\mathrm{d}M=1\ .
+\label{eq:bias_normalisation}
+\end{equation}
+If one uses a mass function and bias pair that are related via the peak-background split formalism \citep{Mo1996,Sheth2001} the these conditions are automatically satisfied. In words these equations enforce that all matter is associated to a halo and that matter is on average unbiased with respect to itself. In the convention used in CCL the units of $P(k)$ will be exactly the units of $\rho_i\rho_j / \mathrm{Mpc}^3$. The units of the $W_i$ are those of $\rho_i$ multiplied by volume.
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/halomodel.py
```diff
@@ -0,0 +1,55 @@
+from pyccl import ccllib as lib
+from pyccl.pyutils import _vectorize_fn, _vectorize_fn2, _vectorize_fn3, _vectorize_fn4
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_background.c
```diff
@@ -1023,7 +1022,10 @@ void ccl_scale_factor_of_chis(ccl_cosmology * cosmo, int nchi, double chi[], dou
 
 double ccl_growth_factor(ccl_cosmology * cosmo, double a, int * status)
 {
-  if(a>1.) {
+  if(a==1.){
```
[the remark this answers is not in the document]
alexander-mead: Okay, this is a little bit of a hack. Let me explain: as part of my coding it was necessary to 'upgrade' the function sigma(R) so that it always took the scale factor as an additional argument (which defaults to a=1 in python). This meant that I changed sigma(R) -> sigma(R,a)=g(a)*sigma(R,a=1). This means that every time the new sigma(R,a) is now accessed the growth function is also accessed. This meant that some of the tests that were previously passing for massive-neutrino cosmologies ended up failing because the neutrino cosmologies do not allow the growth function to be accessed because it is strongly scale dependent. As far as I remember these tests were of the mass function for massive-neutrino models. The only failed because they access the growth function. sigma(R,a=1) is well defined for the massive neutrino cosmologies because you have the power spectrum at a=1, so I figured the quickest way to prevent these fails was to enforce g(a=1)=1 always.

I figured it wouldn't be a problem because I couldn't imagine a scenario where you wouldn't want g(a=1)=1... but it is a bit of a hack. Happy to hear of any alternatives or better ideas to circumvent this problem.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+#include "gsl/gsl_integration.h"
+#include "gsl/gsl_sf_expint.h"
+#include "ccl_background.h"
+#include "ccl_power.h"
+#include "ccl_massfunc.h"
+
+// Cosmology dependence of the virial collapse density according to the spherical-collapse model
+// Fitting function from Bryan & Norman (1998; arXiv:astro-ph/9710107)
+// Here, this is defined relative to the background matter density, not the critical density
+static double Dv_BryanNorman(ccl_cosmology *cosmo, double a, int *status){
```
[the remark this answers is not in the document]
alexander-mead: I thought it was correct to make all functions static that would not be accessed by other *.c files. In this case Dv_BryanNorman is only used within ccl_halomod.c.

include/ccl_power.h
```diff
@@ -59,11 +59,26 @@ double ccl_nonlin_matter_power(ccl_cosmology * cosmo, double k, double a,int * s
  * Returns sigma(R) for specified cosmology at a = 1.
  * @param cosmo Cosmology parameters and configurations
  * @param R Smoothing scale, in [Mpc] units
+ * @param a scale factor
  * @param status Status flag. 0 if there are no errors, nonzero otherwise.
  * For specific cases see documentation for ccl_error.c
  * @return sigma(R).
  */
-double ccl_sigmaR(ccl_cosmology *cosmo, double R, int * status);
+/*double ccl_sigmaR(ccl_cosmology *cosmo, double R, int * status);*/
```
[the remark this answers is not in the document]
alexander-mead: Assuming that no one raises any objections to the API change on Slack, do I manually update the version number (to 0.4) in the CHANGELOG and then just add some details of what I did?

include/ccl_power.h
```diff
@@ -59,11 +59,26 @@ double ccl_nonlin_matter_power(ccl_cosmology * cosmo, double k, double a,int * s
...
+ * Variance of the displacement field with (top-hat) smoothing scale R [Mpc]
+ * Returns sigma(V(R)) for specified cosmology at a = 1.
+ * @param cosmo Cosmology parameters and configurations
+ * @param R smoothing scale, in [Mpc] units
+ * @param a scale factor
+ * @param status Status flag. 0 if there are no errors, nonzero otherwise.
+ * For specific cases see documentation for ccl_error.c
+ * @return sigma(R).
+ */
+/*double ccl_sigmaV(ccl_cosmology *cosmo, double R, int * status);*/
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_massfunc.c
alexander-mead: I feel that maybe these ones shouldn't be in the header since they are taken from papers and the user shouldn't have any choice about what they are. They are on the same level as the fit_A, fit_B, ... that are in the Tinker mass function already.

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+    double A = 1.-two_halo_integral(cosmo, 0., a, status);
+    //printf("A: %10.5f\n", A);
+
+    // ...multiplied by the ratio of window functions
+    A=A*window_function(cosmo, mmin, k, a, status)/window_function(cosmo, mmin, 0., a, status);    
+
+    // Add the additive correction to the calculated integral
+    I2h=I2h+A;
+      
+    return ccl_linear_matter_power(cosmo, k, a, status)*pow(I2h,2);
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+    // The halo concentration for this mass and scale factor  
+    double c = ccl_halo_concentration(cosmo,m,a,status);
+
+    // The function U is normalised so multiplying by M/rho turns units to overdensity
+    return m*u_nfw_c(cosmo,c,m,k,a,status)/rho_matter;
+  }
+
+  // Something went wrong
+  else{
+    exit(0);
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+    f2 = cos(ks)*(gsl_sf_Ci(ks*(1.+c))-gsl_sf_Ci(ks));
+    f3 = sin(c*ks)/(ks*(1.+c));
+    fc = log(1.+c)-c/(1.+c);
+  
+    return (f1+f2-f3)/fc;
+  }
+}
+
+// Halo "formation redshift" calculated according to the Bullock et al. (2001) prescription
+// TODO: Actually code this up
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+
+  // Something went wrong
+  else{    
+    exit(0);
+  }
+	  
+}
+
+// Fourier Transforms of halo profiles
+static double window_function(ccl_cosmology *cosmo, double m, double k, double a, int *status){
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  // Fairly sure that there should be no ln(10) factor should be here since the integration is being specified in log10 range
+  double dn_dlogM = ccl_massfunc(p->cosmo,halomass,p->a,Delta_v,p->status);
+    
+  return dn_dlogM*pow(wk,2);
+}
+
+// The one-halo term integral using gsl
+static double one_halo_integral(ccl_cosmology *cosmo, double k, double a, int *status){
+
+  double mmin=1e7; // Minimum mass for the halo-model integration
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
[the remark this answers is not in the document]
alexander-mead: Removed this entirely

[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+
+// Computes the two-halo term
+double ccl_twohalo_matter_power(ccl_cosmology *cosmo, double k, double a, int *status){
+
+  // Set two-halo term
+  // 1 - Standard two-halo term
+  // 2 - Linear theory
+  int i2h=1;
+
+  double mmin=1e7; //Minimum mass for the integrals
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_massfunc.c
```diff
@@ -300,13 +318,40 @@ static double massfunc_f(ccl_cosmology *cosmo, double halomass, double a, double
...
 
+  // Equation (12) in  arXiv: 9901122
+  // Derived using the peak-background split applied to the mass function in the same paper
+  // Note that Sheth & Tormen use nu=(dc/sigma)^2 whereas we use nu=dc/sigma
+  case ccl_shethtormen:
+
+    // TODO:Mead: Somehow enforce that this should only work with a virial-density Delta_v
+
+    // ST bias fitting parameters (which are the same as for the mass function)
+    fit_p = 0.3;
```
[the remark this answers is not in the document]
alexander-mead: I think this is okay for the same reason as above.

pyccl/__init__.py
```diff
@@ -23,10 +23,11 @@
 
 # Power spectrum calculations and sigma8
 from pyccl.power import linear_matter_power, nonlin_matter_power, sigmaR, \
-    sigma8
+    sigmaV, sigma8
 
 # Halo mass function
-from pyccl.massfunction import massfunc, massfunc_m2r, sigmaM, halo_bias
+from pyccl.massfunction import massfunc, massfunc_m2r, sigmaM, halo_bias, \
+    r_delta, r_Lagrangian, nu
```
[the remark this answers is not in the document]
alexander-mead: Yes, I think these are esoteric enough that they may not be used generally. I have removed them for the time being.

Done

pyccl/halomodel.py
```diff
@@ -0,0 +1,55 @@
...
+
+def onehalo_matter_power(cosmo, k, a):
+    """one-halo term for matter power spectrum
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/massfunction.py
```diff
@@ -63,3 +63,52 @@ def halo_bias(cosmo, halo_mass, a, odelta=200):
     """
     return _vectorize_fn4(lib.halo_bias, 
                           lib.halo_bias_vec, cosmo, halo_mass, a, odelta)
+
+def r_delta(cosmo, halo_mass, a, odelta):
+    """ something something rdelta
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/massfunction.py
```diff
@@ -63,3 +63,52 @@ def halo_bias(cosmo, halo_mass, a, odelta=200):
...
+
+    Args:
+        cosmo (:obj:`ccl.cosmology`): Cosmological parameters.
+        halo_mass (float or array_like): Halo masses; Msun.
+        a (float): Scale factor.
+        odelta (float): overdensity parameter
+
+    Returns:
+        r_delta (float or array_like): something something
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/massfunction.py
```diff
@@ -63,3 +63,52 @@ def halo_bias(cosmo, halo_mass, a, odelta=200):
...
+
+    """
+    return _vectorize_fn4(lib.r_delta,
+			  lib.r_delta_vec, cosmo, halo_mass, a, odelta)
+
+def r_Lagrangian(cosmo, halo_mass, a, odelta):
+    """ something something r Lagrangian
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/power.py
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  // Structure required for the gsl integration
+  ipar.cosmo = cosmo;
+  ipar.k = k;
+  ipar.a = a;
+  ipar.status = &one_halo_integral_status;
+  F.function = &one_halo_integrand;
+  F.params = &ipar;
+
+  // Actually does the integration
+  qagstatus = gsl_integration_qag(&F, log10massmin, log10massmax, 0, 1E-4, 1000, GSL_INTEG_GAUSS41, w, &result, &eresult);
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
```diff
@@ -0,0 +1,310 @@
...
+  // Structure required for the gsl integration
+  ipar.cosmo = cosmo;
+  ipar.k = k;
+  ipar.a = a;
+  ipar.status = &two_halo_integral_status;
+  F.function = &two_halo_integrand;
+  F.params = &ipar;
+
+  // Actually does the integration
+  qagstatus = gsl_integration_qag(&F, log10massmin, log10massmax, 0, 1E-4, 1000, GSL_INTEG_GAUSS41, w, &result, &eresult);
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_massfunc.c
alexander-mead: Done

src/ccl_massfunc.c
```diff
@@ -528,3 +579,25 @@ double ccl_sigmaM(ccl_cosmology * cosmo, double halomass, double a, int * status
...
+// Converts halo mass to rdelta.
+// TODO: possible that delta should be passed around for consistency checks
+double ccl_r_delta(ccl_cosmology *cosmo, double halomass, double a, double odelta, int * status){    
+  //double rho_matter = ccl_comoving_matter_density(cosmo);
+  double rho_matter = ccl_rho_x(cosmo, 1., 1, 1, status);
+  return pow(halomass*3.0/(4.0*M_PI*rho_matter*odelta),1.0/3.0);
+}
+
+// Peak threshold: nu(M,z) = delta_c(z) / sigma(M,z)
+// TODO: Alternatively - just use this in ccl_massfunc as necessary
```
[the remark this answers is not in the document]
alexander-mead: Done

philbull:
OK, something broke with commit 6279f4d -- it's causing a segfault on my system, and I'm getting mysterious Travis errors. What happened?

EiffL:
@alexander-mead  I have just pushed a tiny fix that should solves the problem you were having. Here is what was happening:

In `ccl_massfunc.c`, the  `double nu_mass(ccl_cosmology *cosmo, double halomass, double a, int *status)` function has been commented out, but it was still declared in `ccl_massfunc.h`. It is not a problem for the C executable and library, but SWIG creates a symbol, without and actual definition for it. So when `_ccllib.so` gets loaded by the python interpreter, it looks for this symbol and can't find it's definition, hence the following error message:
```
In [1]: import pyccl
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
~/repo/CCL/pyccl/ccllib.py in swig_import_helper()
     13         try:
---> 14             return importlib.import_module(mname)
     15         except ImportError:

/usr/lib/python3.6/importlib/__init__.py in import_module(name, package)
    125             level += 1
--> 126     return _bootstrap._gcd_import(name[level:], package, level)
    127 

/usr/lib/python3.6/importlib/_bootstrap.py in _gcd_import(name, package, level)

/usr/lib/python3.6/importlib/_bootstrap.py in _find_and_load(name, import_)

/usr/lib/python3.6/importlib/_bootstrap.py in _find_and_load_unlocked(name, import_)

/usr/lib/python3.6/importlib/_bootstrap.py in _load_unlocked(spec)

/usr/lib/python3.6/importlib/_bootstrap.py in module_from_spec(spec)

/usr/lib/python3.6/importlib/_bootstrap_external.py in create_module(self, spec)

/usr/lib/python3.6/importlib/_bootstrap.py in _call_with_frames_removed(f, *args, **kwds)

ImportError: /home/francois/repo/CCL/pyccl/_ccllib.so: undefined symbol: nu_mass

During handling of the above exception, another exception occurred:

ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-1-7837a010f21e> in <module>()
----> 1 import pyccl

~/repo/CCL/pyccl/__init__.py in <module>()
     15     environ["CLASS_PARAM_DIR"] = path.dirname(path.abspath(__file__))
     16 
---> 17 from pyccl import ccllib as lib
     18 
     19 from pyccl import core, constants, background, power, massfunction, halomodel

~/repo/CCL/pyccl/ccllib.py in <module>()
     15         except ImportError:
     16             return importlib.import_module('_ccllib')
---> 17     _ccllib = swig_import_helper()
     18     del swig_import_helper
     19 elif _swig_python_version_info >= (2, 6, 0):

~/repo/CCL/pyccl/ccllib.py in swig_import_helper()
     14             return importlib.import_module(mname)
     15         except ImportError:
---> 16             return importlib.import_module('_ccllib')
     17     _ccllib = swig_import_helper()
     18     del swig_import_helper

/usr/lib/python3.6/importlib/__init__.py in import_module(name, package)
    124                 break
    125             level += 1
--> 126     return _bootstrap._gcd_import(name[level:], package, level)
    127 
    128 

ModuleNotFoundError: No module named '_ccllib'
```
It looks like it's not finding _ccllib, but the `.so` file is there, it's just because the import is crashing in the middle when it encounters the `nu_mass` symbol.

Long story short, I've commented out `nu_mass()` in `ccl_massfunc.h` and everything seems to be working for me. If that function is no longer needed, it should be completely removed from `ccl_massfunc.h` and `ccl_massfunc.c`

.travis.yml
```diff
@@ -35,6 +35,7 @@ install:
 - python --version
 script:
 - python setup.py build
+- python setup.py install
```
EiffL: This doesn't hurt, but it's not really necessary.

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+    a_form = gsl_root_fsolver_root (s);
+    a_min = gsl_root_fsolver_x_lower (s);
+    a_max = gsl_root_fsolver_x_upper (s);
+    gslstatus = gsl_root_test_interval (a_min, a_max, 0., 0.001);
+
+    // testing print statements until it actually works
+    printf("%5d [%.7f, %.7f] %.7f", iter, a_min, a_max, a_form);
+    }
+  while (gslstatus == GSL_CONTINUE && iter < max_iter);
+
```
philbull: Should this check `gslstatus` for errors too?

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+// The concentration-mass relation for haloes
+double ccl_halo_concentration(ccl_cosmology *cosmo, double halomass, double a, double odelta, ccl_conc_label label, int *status){
+
+  double gz, g0, nu, delta_c, a_form;
+  double Mpiv, A, B, C;
+
+  switch(label){
+
+    // Bullock et al. (2001; xxxx.xxxx; Delta = Virial)
```
philbull: Missing arXiv number.

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+    printf("%5d [%.7f, %.7f] %.7f", iter, a_min, a_max, a_form);
+    }
+  while (gslstatus == GSL_CONTINUE && iter < max_iter);
+
+  gsl_root_fsolver_free (s);
+
+  return a_form;    
+}
+
+// The concentration-mass relation for haloes
```
philbull: Since this is a `ccl_` function (i.e. user-facing), it should have a proper docstring comment.

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+  
+  switch(label){
+
+  case NFW:
+
+    // The mean background matter density in Msun/Mpc^3
+    rho_matter = ccl_rho_x(cosmo, 1., ccl_species_m_label, 1, status);
+
+    // The halo concentration for this mass and scale factor  
+    c = ccl_halo_concentration(cosmo, m, a, odelta, Duffy2008_virial, status);
```
philbull: Should this really be fixed to Duffy?

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+
+// Integrand for the one-halo integral
+static double one_halo_integrand(double log10mass, void *params){  
+  
+  Int_one_halo_Par *p = (Int_one_halo_Par *)params;;
+  double halomass = pow(10,log10mass);
+  double Delta_v = Dv_BryanNorman(p->cosmo, p->a, p->status); // Virial density of haloes
+
+  // The squared normalised Fourier Transform of a halo profile (W(k->0 = 1)
+  double wk = window_function(p->cosmo,halomass, p->k, p->a, Delta_v, NFW, p->status);
```
philbull: Fixed to NFW?

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+
+  // Actually does the integration
+  qagstatus = gsl_integration_qag(&F, log10mmin, log10mmax, HM_EPSABS, HM_EPSREL, HM_LIMIT, HM_INT_METHOD, w, &result, &eresult);
+
+  // Clean up
+  gsl_integration_workspace_free(w);
+
+  // Check for errors
+  if (qagstatus != GSL_SUCCESS) {
+    ccl_raise_gsl_warning(qagstatus, "ccl_halomod.c: one_halo_integral:");
```
philbull: Is this warning string complete? It ends with a colon.

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+
+// Integrand for the two-halo integral
+static double two_halo_integrand(double log10mass, void *params){  
+  
+  Int_two_halo_Par *p=(Int_two_halo_Par *)params;
+  double halomass = pow(10,log10mass);
+  double Delta_v = Dv_BryanNorman(p->cosmo, p->a, p->status);
+
+  // The window function appropriate for the matter power spectrum
+  double wk = window_function(p->cosmo,halomass, p->k, p->a, Delta_v, NFW, p->status);
```
philbull: Also fixed to NFW.

src/ccl_halomod.c
```diff
@@ -0,0 +1,355 @@
...
+  double I2h = two_halo_integral(cosmo, k, a, status);
+    
+  // The addative correction is the missing part of the integral below the lower-mass limit
+  double A = 1.-two_halo_integral(cosmo, 0., a, status);
+
+  // Virial overdensity
+  double Delta_v = Dv_BryanNorman(cosmo, a, status);
+
+  // ...multiplied by the ratio of window functions
+  double W1 = window_function(cosmo, HM_MMIN, k,  a, Delta_v, NFW, status);
```
philbull: Fixed to NFW.

src/ccl_massfunc.c
```diff
@@ -10,11 +10,66 @@
...
+  double Om_mz = ccl_omega_x(cosmo, a, ccl_species_m_label, status);
+  double x = Om_mz-1.;
+  double Dv0 = 18.*pow(M_PI,2);
+  double Dv = (Dv0+82.*x-39.*pow(x,2))/Om_mz;
+
+  return Dv;
+
+}
+
+/*----- ROUTINE: nu_mass -----
```
philbull: Remove commented code?

src/ccl_massfunc.c
```diff
@@ -300,13 +378,38 @@ static double massfunc_f(ccl_cosmology *cosmo, double halomass, double a, double
...
-  double fit_A, fit_B, fit_C, fit_a, fit_b, fit_c, overdensity_delta, y;
+  double fit_A, fit_B, fit_C, fit_a, fit_b, fit_c, fit_p, overdensity_delta, y;
   double delta_c_Tinker, nu;
   double sigma=ccl_sigmaM(cosmo,halomass,a, status);
   switch(cosmo->config.mass_function_method) {
 
+  // Equation (12) in  arXiv: 9901122
+  // Derived using the peak-background split applied to the mass function in the same paper
+  // Note that Sheth & Tormen (1999) use nu=(dc/sigma)^2 whereas we use nu=dc/sigma
+  // TODO: Mead: Somehow enforce that this should only work with a virial-density Delta_v
```
philbull: TODO: Does this need to be addressed?

src/ccl_massfunc.c
```diff
@@ -415,87 +518,101 @@ void ccl_cosmology_compute_sigma(ccl_cosmology * cosmo, int *status)
...
   }
   ccl_check_status(cosmo, status);
-  return f*rho_m*deriv/halomass;
+
+  return val;
+}
+
+/*----- ROUTINE: ccl_massfunc -----
+INPUT: ccl_cosmology * cosmo, double halo mass in units of Msun, double scale factor
+TASK: returns halo mass function as dn / dlog10 m
```
philbull: Need to mention units, e.g. is it in comoving or physical Mpc^-3, is it in h^-1 units etc.

src/ccl_power.c
```diff
@@ -1676,15 +1714,55 @@ double ccl_sigmaR(ccl_cosmology *cosmo,double R, int *status)
...
+																				0.0, ccl_gsl->INTEGRATION_SIGMAR_EPSREL,
+																				workspace,&sigma_V,NULL,NULL);
+
+  if(gslstatus != GSL_SUCCESS) {
+    ccl_raise_gsl_warning(gslstatus, "ccl_power.c: ccl_sigmaV():");
+    *status |= gslstatus;
+  }
+
+  // TODO: log10 could be taken already in the macros.
+  // TODO: 1E-5 should be a macro
```
philbull: Address these TODO items.

tests/ccl_test_pyccl_interface.py
```diff
@@ -338,6 +343,53 @@ def check_massfunc_nu(cosmo):
...
+
+    # Wave-vector variables
+    k_scalar = 1.
+    k_list = [1e-3, 1e-2, 1e-1, 1e0, 1e1]
+    k_array = np.array([1e-3, 1e-2, 1e-1, 1e0, 1e1])
+
+    # halo concentration
+    #assert_( all_finite(ccl.halo_concentration(cosmo, mass_scalar, a)) )
+    #assert_( all_finite(ccl.halo_concentration(cosmo, mass_list,   a)) )
+    #assert_( all_finite(ccl.halo_concentration(cosmo, mass_array,  a)) )
```
philbull: Remove commented code.

tests/ccl_test_pyccl_interface.py
```diff
@@ -732,8 +784,8 @@ def test_valid_transfer_combos():
     
     assert_raises(ValueError, ccl.Cosmology, transfer_function='emulator', 
                               matter_power_spectrum='linear', **params)
-    assert_raises(ValueError, ccl.Cosmology, transfer_function='boltzmann', 
-                              matter_power_spectrum='halomodel', **params)
+    #assert_raises(ValueError, ccl.Cosmology, transfer_function='boltzmann',
```
philbull: Why commented out?

philbull:
Also, I see that we have a test failing due to a precision issue:
```
TEST 58/65 halomod:model_2 [FAIL]
  ERR: /home/travis/build/LSSTDESC/CCL/tests/ccl_test_halomod.c:156  expected 1.388688e-01, got 1.390115e-01 (diff -1.427826e-04, tol 1.388688e-04)
```

philbull:
@elisachisari suggests that this might be caused by the recent changes to the default set of physical constants.

elisachisari:
Here is a concrete set of tasks that need to be addressed concerning this PR.

- [ ] Address review by @philbull 
- [ ] Fix failure in CTEST, most likely due to the new physical constants adopted by CCL @damonge 
- [ ] Upload benchmark code that produces the benchmark to the relevant branch @damonge 
- [ ] Make a plot for the paper comparing to the benchmarks - add this in the paper and in the jupypter notebook #369.
- [ ] Address comments in the paper itself: making sure notation is homogeneous with respect to HMF section, add implementation and validation details.

src/ccl_massfunc.c
alexander-mead: It had already been addressed. Only the comment remained. Done

src/ccl_massfunc.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

.gitignore
```diff
@@ -61,6 +70,16 @@ pyccl.egg*
 # Ignore VCS files
 .idea/*
 
+
+
+# Ignore document pdf and bibliography
```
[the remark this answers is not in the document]
alexander-mead: Okay, I removed it. Done.

include/ccl_halomod.h
```diff
@@ -0,0 +1,74 @@
...
+#define HM_EPSABS 0 // Absolute error for the halo-model integration
+#define HM_EPSREL 1E-4 // Relative error for the halo-model integration
+#define HM_LIMIT 1000 // Maximum sub intervals for the halo-model integration
+#define HM_INT_METHOD GSL_INTEG_GAUSS41 // Integration scheme for halo-model integration
+  
+#include "ccl_core.h"  
+
+  // concentration-mass relation
+  typedef enum ccl_conc_label {
+    Bhattacharya2011 = 1,
```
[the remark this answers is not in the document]
alexander-mead: Done

include/ccl_halomod.h
```diff
@@ -0,0 +1,74 @@
...
+    Duffy2008_virial = 2,
+    constant = 3,
+    Bullock = 4,
+  } ccl_conc_label;
+
+  // halo window profiles
+  typedef enum ccl_win_label {
+    NFW = 1,
```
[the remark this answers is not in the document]
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Okay, I changed this and made it similar to other gsl warnings that are raised in e.g., `ccl_power.c`. I followed the convention:

src/ccl_massfunc.c
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: Ah, @villarrealas wrote this parts. Any comments Antonio? At the moment the Bullock relation is not used at all, but it should be when I add HMcode in the future... maybe this should be stashed away until then.

pyccl/ccl_massfunc.i
```diff
@@ -64,4 +64,47 @@ void halo_bias_vec(ccl_cosmology * cosmo,
         output[i] = ccl_halo_bias(cosmo, halo_mass[i], a, odelta, status);
     }
 }
+
+/*
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/__init__.py
```diff
@@ -29,10 +29,11 @@
 
 # Power spectrum calculations and sigma8
 from pyccl.power import linear_matter_power, nonlin_matter_power, sigmaR, \
-    sigma8
+    sigmaV, sigma8
 
 # Halo mass function
-from pyccl.massfunction import massfunc, massfunc_m2r, sigmaM, halo_bias
+from pyccl.massfunction import massfunc, massfunc_m2r, sigmaM, halo_bias#, \
```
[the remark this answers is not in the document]
alexander-mead: Done

pyccl/halomodel.py
```diff
@@ -0,0 +1,45 @@
+from pyccl import ccllib as lib
+from pyccl.pyutils import _vectorize_fn2
+
+def onehalo_matter_power(cosmo, k, a):
+    """One-halo term for matter power spectrum
```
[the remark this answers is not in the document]
alexander-mead: Done

tests/ccl_test_pyccl_interface.py
alexander-mead: Okay,  a bit of a long story here:

First; previously there were `halo_model` and `halomodel` cases for `ccl_nonlin_matter_power`, neither of which were wired up to anything. If you called `ccl_nonlin_matter_power` with each method I think you just got the linear power back. I removed one of them so we are now left with `halo_model` only. I couldn't think of any good reason to have both.

Second; I looked at wiring the `halo_model` option up to the new calculation from this new halomod branch, but this is impossible I think because `ccl_halomod.c` uses `ccl_power.h`, so I can't use the correct routines without entering some sort of infinite calling loop. So at the moment the `halo_model` option remains useless.

This sort of relates to one of the issues I raised previously (https://github.com/LSSTDESC/CCL/issues/337) and also sort of to https://github.com/LSSTDESC/CCL/issues/438

Generally, I think it would be better to have a source file dedicated to getting the linear power spectrum (`ccl_linear_power.c` ?) and then a separate file(s) dedicated to whatever non-linear method you wanted to use... and then one final source file that had a way of choosing between whatever power spectrum calculation the user needed, be that linear, emu, halomodel, halofit... whatever.

tests/ccl_test_pyccl_interface.py
alexander-mead: Done

src/ccl_halomod.c
alexander-mead: See below. Done

src/ccl_halomod.c
alexander-mead: See comment below. Done.

src/ccl_power.c
villarrealas: I am not entirely sure what these TODO items are for in sigmaR (I forget who wrote ccl_power off the top of my head), but I believe both of these look like they are already resolved.

src/ccl_halomod.c
alexander-mead: Okay, I actually removed the Bullock concentration relation entirely, thus side-stepping this problem. It's still in my separate HMcode branch though, so this is a battle for a later date...

Done

src/ccl_power.c
alexander-mead: Apparently this is sorted, so I've now done

tilmantroester:
Benchmark plot is now in the notebook and in the paper.

src/ccl_halomod.c
```diff
@@ -181,7 +184,7 @@ static double window_function(ccl_cosmology *cosmo, double m, double k, double a
     rho_matter = ccl_rho_x(cosmo, 1., ccl_species_m_label, 1, status);
 
     // The halo concentration for this mass and scale factor  
-    c = halo_concentration(cosmo, m, a, odelta, Duffy2008_virial, status);    
+    c = ccl_halo_concentration(cosmo, m, a, odelta, Duffy2008_virial, status);
```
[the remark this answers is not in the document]
alexander-mead: Okay, I did a fairly substantial re-write of things to change this. The `halo_concentration` is now elevated to the config for the cosmology object, putting on the same status as `mass_function`. The user can now choose combinations of mass function and halo concentration that they would like to use for the halo-model power-spectrum calculation. The only snag is that the halo-model calculation is fixed to assume that haloes are defined by the virial definition, rather than say M200 or M500. The calculation enforces consistency though, so you *cannot* use a mass function for haloes with M500 in tandem with a concentration relation for haloes defined by M200. For the power spectrum calculation you have to use relations that are defined for virial haloes. I think this is now done!

PR #340, opened 2018-02-22
Add simple unit tests for all power spectrum methods

philbull:
Adds a set of unit tests that calculate the linear and nonlinear power spectrum over a range of scales and redshifts for a couple of cosmologies. A number of choices of transfer function and matter power spectrum types are tested.

(Also fixes an initialisation issue with the `status_message` string inside the `ccl_cosmology` object.)


philbull:
I enabled coveralls too, so now we can track unit test coverage.

tests/ccl_test_power.py
```diff
@@ -0,0 +1,104 @@
...
+Omega_v_vals = np.array([0.7, 0.7, 0.7, 0.65, 0.75])
+w0_vals = np.array([-1.0, -0.9, -0.9, -0.9, -0.9])
+wa_vals = np.array([0.0, 0.0, 0.1, 0.1, 0.1])
+
+# List of supported transfer functions
+transfer_fns = ['boltzmann', 'bbks', 'boltzmann_camb', 'boltzmann_class', 
+                'camb', 'class', 'eisenstein_hu', 'emulator', 
+                'fitting_function', 'none']
+
+transfer_fns = ['boltzmann_class',
```
elisachisari: Why is this redefined here?

tests/ccl_test_power.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    return np.alltrue( np.isfinite(vals) )
+
+
+def calc_power_spectrum(Omega_v, w0, wa, transfer_fn, matter_power, linear):
+    """
+    Calculate linear and nonlinear power spectrum for a given set of parameters 
+    and choices of transfer function and matter power spectrum.
+    """
+    k = np.logspace(-4., 0., 300)
```
elisachisari: Perhaps extend k range to below/above extrapolation ranges, to make sure extrapolation doesn't result in errors.

tests/ccl_test_power.py
```diff
@@ -0,0 +1,104 @@
...
+    print(">>> (%s; %s)" % (transfer_fn, matter_power))
+    
+    # Loop over parameters
+    for i in [0,2]: #w0_vals.size):
+        calc_power_spectrum(Omega_v_vals[i], w0_vals[i], wa_vals[i], 
+                            transfer_fn=transfer_fn, matter_power=matter_power,
+                            linear=lin)
+
+@decorators.slow
+def test_power_spectrum_linear():
```
elisachisari: I'm not sure all these combinations should be allowed. For example, I think we only support 'emu' transfer with 'emu' matter_power.  There are details of the allowed combinations here:
https://github.com/LSSTDESC/CCL/issues/337
Should we take advantage of this PR to implement graceful exits for the combinations that aren't allowed and check here that indeed CCL exits?

tests/ccl_test_power.py
```diff
@@ -0,0 +1,104 @@
...
+
+@decorators.slow
+def test_nonlin_power_spectrum_halofit():
+    for tfn in transfer_fns: loop_over_params(tfn, 'halofit', lin=False)
+
+@decorators.slow
+def test_nonlin_power_spectrum_emu():
+    transfer_fns = ['emulator',]
+    for tfn in transfer_fns: loop_over_params(tfn, 'emu', lin=False)
+
```
elisachisari: I could recommend adding a unit test with baryons too. (Do we already have one elsewhere?)

philbull:
Sorry for the delay. I've now implemented the requested changes, including adding a check (in the Python wrapper only) to make sure that `transfer_function` and `matter_power_spectrum` are compatible. (I didn't add any tests for baryons though.)

tests/ccl_test_power.py
elisachisari: Actually, I checked and cosmo9 among the test cosmologies includes baryons, so no need to add anything else.

PR #343, opened 2018-02-28
Guard against errors that would kill the Python interpreter

philbull:
Protect against killing the Python interpreter by using `ccl_raise_exception()` instead of `exit()` statements in `ccl_check_status_nocosmo()`.


philbull:
Sort of. We could implement error handling triggers using SWIG itself, but we're not doing that at the moment. Instead, the Python wrapper does the status checks when SWIG hands back control to Python, by calling the `check()` function. The behaviour is pretty much the same, except this gives us slightly more control over how to handle errors.

philbull:
Great, thanks!

Issue #347, opened 2018-03-08
Artefacts in lensing Cls


tilmantroester:
I noticed that the lensing Cls have small artefacts (peaks at certain ell values). For the GG terms they are small (in this example < 0.1%) but become large (>1%) for the IA terms. The following plot shows the fractional difference between neighbouring ell-values:

![cl_artefacts](https://user-images.githubusercontent.com/26582252/37162250-aeab9f3a-22ec-11e8-84e1-78c1b81c012a.png)

I extract the GI term by computing G(G+I) and then subtracting the GG term. Since the GI term is much smaller than GG, this amplifies the effect of these artefacts.

I tried changing the A_SPLINE_NA_PK parameter but this doesn't seem to have an effect, nor changing the number of samples in the provided n(z).

Are there other parameters that can be used to tune the precision of the Limber integrals?

[Example code](https://github.com/LSSTDESC/CCL/files/1793917/test_ccl_stuff.py.txt)

damonge:
There are GSL integration parameters that could be tweaked for this, but they're not accessible to the user right now. Some of this may be fixed when the Angpow PR gets merged, so I'd suggest waiting until then and checking again (I'll rerun the code).

tilmantroester:
Since I was playing around with the GSL parameters for the other issue anyways, I figured I could check how ccl_angular_cl depends on the choice of GSL parameters. Plotted below are the fractional differences for different choices of epsrel and the number of Gauss-Kronrod points compared to the case where epsrel=1e-6 and 61 Gauss-Kronrod points. The grey band denotes a fractional difference of 1e-4.

![cl_artefacts_2](https://user-images.githubusercontent.com/26582252/38153576-64901b6c-3465-11e8-9694-d72caa104179.png)

The upshot is that the current settings aren't sufficient if a 1e-4 accuracy is required everywhere.

elisachisari:
@tilmantroester thanks for doing this - none of the curves lie within the gray band then? What do you suggest to improve this? Is it as simple as decreasing epsrel (at the cost of extra computational time)?

tilmantroester:
@elisachisari In case of the n(z) I used for the comparison, either using epsrel=1.0e-4 and GSL_INTEG_GAUSS61 or epsrel=1.0e-5 and GSL_INTEG_GAUSS51 gets us within the 10^-4 band everywhere. I haven't checked the impact on the computational time yet. 
[...]

PR #348, opened 2018-03-12
corrected spin-2 prefactors

damonge:
This addresses part of issue #347 , related to the factors of sqrt((l+2)!/(l-2)!) associated to the transfer functions of spin-2 quantities.


elisachisari:
@damonge Are the note and paper consistent with these changes?

damonge:
Yes, I checked

damonge:
In case it's useful for future reference, useful expressions can be found here:
https://github.com/damonge/LimberJack/blob/master/doc/limber.pdf

PR #349, opened 2018-03-12
added functionality to return internal tracer functions (e.g. dndz)

damonge:
For debugging purposes (among other things) we'll probably want to know how a certain quantity, such as the N(z) or the lensing window function, is being used internally to compute power spectra. This PR adds a method to the ClTracer classes to evaluate any internal function of redshift. Currently it allows for the N(z), lensing and magnification window functions and different types of biases.


elisachisari:
Thanks for implementing this. Some comments on this PR: 
Why 209 file changes? It seems most of the changes are in the doxygen files and some changes relate to specifying an absolute path which refers to your /home/. Is this expected? 
Could you add an example for retrieval of at least one this functions in ccl_sample_run and the python notebooks? 
Could you also add something in the CCL note?

damonge:
@elisachisari thanks for the review.

I had to update the doxygen files because I implemented new functionality in C and documented it. Good catch with the absolute paths in the doxygen files. Apparently by default Doxygen uses the absolute path of source files to reference them in the html files, and actually the current master branch contains some references to my absolute paths, since I was the last one to update the doxygen documentation. Apparently that doesn't affect users being able to see the doxygen pages (I checked through https://htmlpreview.github.io/), but I agree that it doesn't look good. I've fixed this, and now only relative paths appear. I've also checked that the html pages are still visible.

I've added examples of this in ccl_sample_run.c and in the lensing angular power spectrum notebook, and mentioned this functionality in the note.

elisachisari:
Haven't looked at all the changes yet, will do soon, but travis checks are failing:
https://travis-ci.org/LSSTDESC/CCL/jobs/355236600
It looks as if the most recent changes to ccl_sample_run are not compatible with cpp.

damonge:
Fixed!

examples/ccl_sample_run.c
```diff
@@ -109,6 +115,10 @@ int main(int argc,char **argv)
   
   //Cosmic shear tracer
   CCL_ClTracer *ct_wl=ccl_cl_tracer_lensing_simple_new(cosmo,NZ,z_arr_sh,nz_arr_sh, &status);
+
+  //This function allows you to retrieve some of the tracer's internal functions of redshift
+  ccl_get_tracer_fas(cosmo,ct_gc,2*NZ,a_arr_resample,nz_resampled,CCL_CLT_NZ,&status);
```
elisachisari: Can na be any number? I.e., did it need to be 2NZ here?

Issue #341, opened 2018-02-28
convenience function for critical density in a given cosmology in Python


c-d-leonard:
It would be great to have a function for getting rho_critical in a given cosmology from the python wrapper (I know I'm not the only one who has to rederive this every time ...)

alexander-mead:
I guess it would also be useful to have a function for rho_matter... and also for physical and comoving versions of rho_matter and rho_critical.

sjoudaki:
I can do it.

sjoudaki:
Hi Danielle, are you happy for the units to be M_sun/h / (Mpc/h)^3?

alexander-mead:
I think that, in order to be consistent with everything else, the units should not have h factors in them.

alexander-mead:
I wrote this function for the comoving matter density already, which I'm fairly sure is correct:

double comoving_matter_density(ccl_cosmology *cosmo){
  // The comoving density of matter. This is a constant. Units are Msun/Mpc^3 (no factors of h)
  return RHO_CRITICAL*cosmo->params.Omega_m*cosmo->params.h*cosmo->params.h;
}

sjoudaki:
Sounds good, I can certainly make sure no factors of h. What you've written is correct for the present. I've created the redshift-dependent "ccl_rho_x", will try to push soon.

alexander-mead:
I think ```CCL``` already has the constant z=0 ```RHO_CRITICAL``` defined, which has units of Msun/h / [Mpc/h]^3 (note this is with h factors) and has the value ~ 2.77466e+11 

Copying out from my Fortran halo-model code, I have four other types of densities defined. I think these are sensible things to define, but maybe you should check... note that I use units *with* h factors and my constant ```critical_density``` is the same as the ```RHO_CRITICAL``` from ```CCL```.

```
 FUNCTION comoving_critical_density(z,cosm)

    !Comoving critical density in (Msun/h) / (Mpc/h)^3
    IMPLICIT NONE
    REAL :: comoving_critical_density
    REAL, INTENT(IN) :: z
    TYPE(cosmology), INTENT(IN) :: cosm

    comoving_critical_density=physical_critical_density(z,cosm)/(1.+z)**3

  END FUNCTION comoving_critical_density

  FUNCTION physical_critical_density(z,cosm)

    !Physical critical density in (Msun/h) / (Mpc/h)^3
    IMPLICIT NONE
    REAL :: physical_critical_density
[...]

PR #351, opened 2018-03-15
Physical density function

sjoudaki:
Addressing issue #341, new function (ccl_rho_x) created for physical density of a given species with redshift.


sjoudaki:
Thanks Alex! I wanted the function to be self-contained, in the same way as \Omega_x, but could certainly express the code in terms of the existing function for H(a). Would be simple. As for including components beyond critical and matter, I thought better to be completely general. But I'm flexible. Happy for others like David (@damonge) to chime in.

elisachisari:
Python 3 checks aren't passing in Travis, though python 2 seems ok. We'll have to find out what is causing this.

elisachisari:
Perhaps @c-d-leonard can comment on this PR in the context of issue #341

c-d-leonard:
To respond (late) to @elisachisari 's request for comment, this certainly fulfills and more what I mentioned in Issue #341 . Like Alex I was mostly worried about avoiding silly unit mistakes for the comoving quantities. It's great that you've included physical densities too though. 

I haven't checked the equations in detail but they look as I would think they should at first glance. One comment would be that it would be great to see a test which checks the value itself of at least one of these rho quantities against something calculated externally, in addition to the interface test that has been added.

sjoudaki:
@c-d-leonard In case it's helpful, updated ccl_sample_run which now also outputs the critical and matter densities. @alexander-mead Both ccl_rho_x and ccl_omega_x now expressed in terms of existing h_over_h0 function. In discussing with @elisachisari, may as well keep rho_x for all components.

include/ccl_background.h
```diff
@@ -6,6 +6,24 @@ extern "C" {
 #pragma once
 #include "ccl_core.h"
 
+//rho_x labels
+typedef enum ccl_rho_x_label {
```
damonge: I believe this adds a lot of extra nuisance constants for very little payoff. Couldn't we recycle ccl_omega_x_label for this?

include/ccl_background.h
```diff
@@ -243,6 +261,17 @@ double ccl_scale_factor_of_chi(ccl_cosmology * cosmo, double chi, int * status);
...
+/**
+ * Physical density (rho) as a function of scale factor.  Critical density is defined as rho_critical = 3 H^2(a)/ (8 pi G). Density of a given species is then rho_x = Omega_x(a) rho_critical(a). For example, rho_matter(a) = Omega_m  a^{-3} / (H^2/H0^2)  3H^2 / (8 pi G) =  Omega_m  a^{-3}  3H0^2 / (8 pi G) =  Omega_m a^{-3}  rho_critical_present. Units of M_sun/(Mpc)^3.
+ * @param cosmo Cosmological parameters
+ * @param a scale factor, normalized to 1 for today
+ * @param label species type. Available: 'critical'(0,1), 'matter'(2,3), 'dark_energy'(4,5), 'radiation'(6,7), 'curvature'(8,9), 'massless neutrinos'(10,11), 'massive neutrinos'(12,13). Even label numbers refer to the physical densities. Odd label numbers refer to 'comoving' densities, which multiply the physical densities by a^3.
+ * @param status Status flag. 0 if there are no errors, nonzero otherwise. 
+ * For specific cases see documentation for ccl_error.
+ * @return rho_x, physical density at scale factor a. 
+ */
+double ccl_rho_x(ccl_cosmology * cosmo, double a, ccl_rho_x_label label, int* status);
```
damonge: here you could use ccl_omega_x_label instead of ccl_rho_x_label. The former does not have "comoving" options, but you could just add one more parameter "is_comoving" that can be 0 or 1 depending on whether the user wants the comoving density. I'm pretty sure this would actually simplify the code a lot too (by a factor of 2 in number of lines ;-))

include/ccl_constants.h
```diff
@@ -45,6 +45,7 @@ extern "C" {
  * Solar mass in units of kg (from GSL)
  */
 #define SOLAR_MASS GSL_CONST_MKSA_SOLAR_MASS
+//#define SOLAR_MASS 1.9885e30 //(from PDG 2015) in Kg
```
damonge: This seems unrelated to the point of this PR, but it's fine to have it here in any case.

pyccl/__init__.py
```diff
@@ -10,7 +10,7 @@
 from pyccl.background import growth_factor, growth_factor_unnorm, \
     growth_rate, comoving_radial_distance, comoving_angular_distance, \
     h_over_h0, luminosity_distance, distance_modulus, scale_factor_of_chi, \
-    omega_x
+    omega_x, rho_x
```
damonge: As I said above, I'd remove the need for the extra rho_x here

pyccl/background.py
```diff
@@ -11,6 +11,23 @@
     'neutrinos_massive': lib.omega_nu_label,
 }
 
+species_types_rho = {
```
damonge: Again, let's try to recycle 'species_types' here. I think you'll introduce a lot of extra complexity otherwise

src/ccl_background.c
```diff
@@ -34,6 +34,78 @@ static double h_over_h0(double a, ccl_cosmology * cosmo, int *status)
...
+  else {
+    OmNuh2 = 0.;
+  }
+
+  double rhocrit_present = RHO_CRITICAL * (cosmo->params.h) * (cosmo->params.h); //units of M_sun/(Mpc)^3
+  
+  switch(label) {
+  case ccl_rho_crit_label :
+      return rhocrit_present * h_over_h0(a, cosmo, status) * h_over_h0(a, cosmo, status);
+  case ccl_rho_crit_comoving_label :
```
damonge: This is what I mean. If you add an 'is_comoving' argument to this function you can remove all the _comoving_ cases here (half of them) and just multipy by a^3 at the end if is_comoving. This has the added advantage that you could use the ccl_omega_x labels

src/ccl_background.c
```diff
@@ -61,17 +133,17 @@ double ccl_omega_x(ccl_cosmology * cosmo, double a, ccl_omega_x_label label, int
 	
   switch(label) {
   case ccl_omega_m_label :
-    return cosmo->params.Omega_m/(cosmo->params.Omega_m+cosmo->params.Omega_l*pow(a,-3*(cosmo->params.w0+cosmo->params.wa))*exp(3*cosmo->params.wa*(a-1))+cosmo->params.Omega_k*a+ (cosmo->params.Omega_g + cosmo->params.Omega_n_rel)/a + OmNuh2*a*a*a / (cosmo->params.h) / (cosmo->params.h));
+    return cosmo->params.Omega_m/(a*a*a) / h_over_h0(a, cosmo, status) / h_over_h0(a, cosmo, status);
```
damonge: What's happening here? Having to evaluate h_over_h0 twice seems very inefficient (and I'm not sure the complier will be smart enough to optimize it automatically). It'd be better to do
```
#At the beginning of this function
hnorm=h_over_h0(a,cosmo,status)
...
#Then, later on
return cosmo->params.Omega_m/(a*a*a) /(hnorm*hnorm);
```

sjoudaki:
New files have been uploaded that addresses David's comments, mainly in regards to common labels and optional argument for the comoving cases. Happy for everyone to have another look. Thanks!

Issue #302, opened 2017-12-13
Integration error depending on IA amplitude


joezuntz:
I'm seeing an integration error when calculating C_ell for the DES SV n(z) that depends on the value of the IA amplitude, which seems a bit odd.

I wanted to double check if I'm doing something wrong with ranges or input values or something else like that.

I've attached an example which uses the current CCL master branch.  I had to call it issue.py.txt so that github would let me attach it.  It loads the attached number densities.

If you use A=1.0 then it works, but if you use A=3.0 then it fails with  
```
RuntimeError: Error CCL_ERROR_INTEG: ccl_cls.c: ccl_angular_cl(): error integrating over k
```
Can anyone advise?

[bin_1.txt](https://github.com/LSSTDESC/CCL/files/1556534/bin_1.txt)
[bin_3.txt](https://github.com/LSSTDESC/CCL/files/1556535/bin_3.txt)
[z.txt](https://github.com/LSSTDESC/CCL/files/1556536/z.txt)
[issue.py.txt](https://github.com/LSSTDESC/CCL/files/1556538/issue.py.txt)

joezuntz:
Also i was giving it a very wide range in ell as I thought that this would be needed for the conversion to xi(theta).  If that's not the case then let me know.

elisachisari:
@joezuntz The "intrinsic alignment bias" as defined in CCL is not normalized following usual conventions in IA literature. See issue #280. I thought I'd point this out. I think you are passing a huge IA amplitude if you set IAbias = 3 and this is somehow causing integration problems (which we should investigate).

joezuntz:
@elisachisari thanks!  Can you advise how I should change the code attached above to make it sensibly normalized?  I included the C_1 factor already.

jablazek:
@elisachisari : assuming the IAbias is linear in A, then it is strange if A=3 causes problems when A=1 does not. But either way, yes we should address #280.

jablazek:
From issue.py above:
> bias_ia = -A * C1_rhocrit * Omega_m / D

This looks "standard." @elisachisari : is something happening downstream that rescales this bias, or is it bias_ia or bias_ia^2 that multiplies P_delta in the NLA model?

damonge:
I have to look into this for a different project, so let me get back to you today

damonge:
[...]

PR #353, opened 2018-03-21
Adds backup integration function for computing Cls

EiffL:
This pull request aims at addressing #302 where it was realized that the computation of Cls could fail unexpectedly when changing the IA amplitude. At the root of the problem was that for some `ell` the `gsl_integration_qag` function in `ccl_angular_cl` failed due to round-off errors before reaching the desired relative accuracy. To try to recover from such failures, I have added a backup integration strategy based on another gsl function which will try to compute the integral if the first method fails. This alternative method, based on  `gsl_integration_cquad` is in principle slower than the default method but should be more robust.


EiffL:
One detail though, is there any mechanism in CCL to display warnings ? For instance to notify the user that the default integration failed?

elisachisari:
I don't think we have a mechanism for warnings. Take a look at ccl_error.c in src/, there we have ccl_raise_exception. We could similarly have a ccl_raise_warning for this application.

elisachisari:
I checked this out and used a jupyter notebook for computing Cl's for cosmic shear and gxy-gxy lensing including IA with amplitudes (bias_ia) ranging from 0 to 100 and for ells in the range [2,3000]. Everything worked well for me and checks pass in C and python. However, the same notebook worked fine for me in master as well, so I'd suggest for @joezuntz to check this PR solves his issue #302 specifically.

EiffL:
I've  opened a  new  PR for this, starting from a clean version of master, we should close this PR and continue with the other one: #406

EiffL:
This is now outdated and solved by another PR, closing this one.

Issue #270, opened 2017-10-31
Distutils screws up rpath on some systems


philbull:
The `setup.py` script passes a `runtime_library_dirs` kwarg to the distutils `setup()` function. This is supposed to setup the CCL library so that it knows where other libraries that it depends on have been installed, without those libraries necessarily being in the user's library path (a feature called 'rpath').

I just tried to install CCL on a system where distutils is doing the wrong thing with this argument. The value of the kwarg is generally passed to the compiler using the `-R` switch (which is then passed to the linker). This switch doesn't seem to be present on some systems though. I don't know why.

A possible remedy is to explicitly specify the rpath argument. To do this, we would comment out the following line:
    `runtime_library_dirs=[libdir],`
and modify the line below it to read:
    `extra_compile_args=['-O4', '-std=c99', "-Wl,-rpath=%s"%libdir],`

GCC version: gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-16)
python: Python 2.7.13 :: Anaconda, Inc.
uname -a
Linux 3.10.0-693.2.2.el7.x86_64 #1 SMP x86_64 x86_64 x86_64 GNU/Linux

fjaviersanchez:
I am adding setuptools as part of the solution of #269. Maybe it helps in this case too?

fjaviersanchez:
@philbull did you find the error while installing or at runtime?

elisachisari:
@EiffL can we close this following the new installation?

EiffL:
I think we should be ok now, cmake should handle all of this, but I have to admit I'm not 100% sure why the problem happened in the first place...

beckermr:
I am closing this issue since it appears we are not too worried anymore. Feel free to reopen if desired!

Issue #269, opened 2017-10-27
make pip install work


fjaviersanchez:
I have been working on trying to make CCL installable using `pip`. The problem is that it installs correctly but one of the CLASS files is not linked properly so it causes a segfault when using the methods that internally call CLASS. Any help is welcome!

Issue #160, opened 2017-04-05
generated configure script is included in github.


rainwoodman:
I noticed the generated configure script is included in github.

The usual practice is not to include generated configure scripts on github.
autotools will automatically include these scripts in a release tarball.

I think people do provide an autogen.sh to preproduce the scripts after a checkout.

e.g. 

https://git.gnome.org/browse/gtk+/tree/autogen.sh

autotools also have a 'maintainer-mode' which may modify these generated files at times

The idea is that the 'developers' who obtain the source code on github shall have a development environment with autoconf automake etc properly installed. The distributed release tar files are `platform independent` (in the autotools sense).

I don't see a discussion about this on the issue tracker -- so I am starting one here.

beckermr:
Closing this since we use cmake now.

Issue #70, opened 2017-01-31
Register CCL Python package with PyPI


philbull:
Once the API has stabilised and we have a public release of CCL, the package should be registered with PyPI. This will allow people to install it using the `pip` command.

Instructions for doing this are here:
https://packaging.python.org/distributing/

elisachisari:
We now have a [registered package](https://pypi.org/project/pyccl/), so I'm closing this issue.

PR #357, opened 2018-03-30
Complete overhaul of CCL compilation framework

EiffL:
Ok, so I know how a pull request affecting 41 files with 70,161 deletions is going to look, but just hear me out first ;-)
Long story short, here is what this PR will allow you to do (once approved and uploaded to PyPi):
```sh
$ pip install pyccl
```
No questions asked, whether running on cori, a random linux or even on Mac OS X.
Here are a few more details:
  - pip installable CCL package from PyPi 
  - compiles seamlessly on Mac OS X and Linux with support of clang and gcc
  - automatically finds all dependencies if installed on the system and downloads missing requirements (including swig, class, fftw, gsl) with no intervention from the user
  - automated (re-)generation of swig interface (as needed, no intervention from the user)
  - automatically finds the correct python and numpy libraries to link to 
  - proper install of C library with pkg-config file
  - Adds Travis CI tests for Mac OS X [Build Status](https://travis-ci.org/EiffL/CCL)

The main motivation for these changes is that I got annoyed that I couldn't pip install ccl on my local cluster... but they also address a number of open issues #269 #270 #160 #70  

The price to pay for all of this automation is to get rid of the current autotools-based build system and switch to `CMake` goodness instead. `CMake` takes care of everything for you (including downloading and building missing dependencies) and is completely cross-platform. The build process for the C library now becomes:
```sh
$ mkdir build && cd build    # creates a separate build directory to store all the generated files 
$ cmake ..                   # CMake configuration: looks for C compiler, python interpreter, fftw, gsl...
$ make                       # Builds  C library along with missing dependencies
```
and for the python module it's even easier:
```sh
$ python setup.py install
```
no need to worry about class, swig, gsl, or fftw.

Now, in terms of modifications, here is what I changed:
  - removed all the makefiles, configure, `install_gsl.sh`, and `class_install.py ` files.
  All of this is now replaced by CMake. The main script is in `CMakeLists.txt` with a few extra modules in `cmake/Module` to take care of finding/compiling external libraries (downloading, compiling and adding CLASS to the project is just one CMake command).
  - simplified `setup.py` and added extra fields for PyPi
  Because all of the  heavy lifting is done by CMake, the `setup.py` only needs to call cmake and collect the output library.
  - python module: automated swig generation, standalone install, includes `ccl_params.ini`
  To make sure the python code is portable, I have removed the pre-computed swig interface, it is now automatically regenerated  (not a problem because swig is found or downloaded by cmake so will always be available). In order to make the python module standalone, it is linked against static libraries so that you don't have to install the  C library and can erase the CCL folder afterwards without breaking the python module. Also, `ccl_params.ini` gets added to the python module when it is built and at runtime in `__init__.py` will check if the environment variable `CCL_PARAM_FILE` already exists, otherwise it defaults to the `ccl_params.ini` included in the package. Finally, I have switched to `nose` tests which automatically discovers all the python tests without needing a separate script.
  - created pkg-config file:
  A pkg-config file for CCL is now automatically created and installed on the system, so that other projects can easily link to CCL. In particular, compiling a c file and linking it to ccl becomes:
  ```sh
  $ gcc ccl_sample_run.c `pkg-config --libs --cflags ccl` -o ccl_sample_run
  ``` 
  - Updated documentation:
  I have updated the documentation on the install procedure to reflect the new cmake process

Concerning license issues, I think my modifications are OK, I am not packaging CLASS along with the code, it is downloaded and compiled by the user everytime. As long as we only upload a src package to PyPi it should be ok. 

I have tested all of this on my various linux PCs, my local cluster, cori and Mac OS X using travis CI, but that doesn't cover all the variety of setups out there, if people find any difficulties in building CCL I will be  happy to provide all the necessary support.


dkirkby:
I just tried the following install:
```
conda create -n CCLTEST python=3.6 pip
source activate CCLTEST
pip install git+git://github.com/EiffL/CCL.git
```
but it failed to download CLASS:
```
    ...
    [  2%] Creating directories for 'CLASS'
    [  5%] Performing download step (download, verify and extract) for 'CLASS'
    -- downloading...
         src='https://github.com/lesgourg/class_public/archive/v2.6.3.tar.gz'
         dst='/private/var/folders/7j/j43qxbs15w73xp3ldj323sh40000gn/T/pip-7rtv5m_3-build/build/CLASS/src/v2.6.3.tar.gz'
         timeout='none'
    CMake Error at CLASS-stamp/download-CLASS.cmake:27 (message):
      error: downloading
      'https://github.com/lesgourg/class_public/archive/v2.6.3.tar.gz' failed

        status_code: 35
        status_string: "SSL connect error"
        log:   Trying 192.30.255.113...

      Connected to github.com (192.30.255.113) port 443 (#0)

      Unknown SSL protocol error in connection to github.com:-9836

      Closing connection 0
```

dkirkby:
The direct link https://github.com/lesgourg/class_public/archive/v2.6.3.tar.gz works in my browser, so this looks like a problem with the SSL credentials used by the python downloader.

EiffL:
Ok thanks David. No worries, if SSL is an issue we can download class using
git. It may be even a better idea, people could specify as an option which
class repo/branch/tag to use when building CCL.

EiffL:
There, that should fix it.

damonge:
@EiffL this looks great. I have been earnestly trying to break it on my linux machine but it seems to work well. I have a few questions/queries/commets:
 - Would it make sense to add a 'make install' target to uninstall the C library? I don't think there is one (and it's nicer than the xargs rm solution, I think. Also, the xargs method does not remove all the library files, it seems). 
 - Currently cmake doesn't seem to be satisfied with my GSL2.1 installation and it insists on downloading 2.4 (which takes a while to install). Does it always need to have 2.4 (although there's no functionality in CCL that requires it)? The documentation actually says 2.1
 - In the README, I think it's not clear whether you should run 'pip install pyccl' after having installed the C library (I assume not, but a user might thing that they have to go through the C installation right above that instruction in the README).
 - One more thing (and I think this is the main one): I don't think the CONTRIBUTING.md documents what to do when adding a new .c .i or .py file.

EiffL:
Ok, I hope this makes the documentation clearer. `contributing.md` doesn't change much. I may not have thought of every use case, let me know if there is anything confusing.

I've downgraded the GSL requirement to 2.1, I don't know why I thought it required 2.4.

I've also added a `make uninstall` target despite the warnings on the internet that uninstalling files automatically like this can be dangerous ;-) A situation where it can misbehave is if let's say you have 2 software that use `fftlog.h` which is by default installed by both softs as `/usr/local/include/fftlog.h` if you uninstall `CCL` you remove this file and break the other software `¯\_(ツ)_/¯` 

I can also add `make test` targets if needed (CMake has like ridiculously overkill testing abilities). But I thought the `check_ccl` binary was cool because it lets you test the install environment on top of just the code, it will tell you right-away (by crashing ;-) ) if you are missing an LD_LIBRARY_PATH.

Dockerfile
```diff
@@ -1,45 +1,20 @@
...
+RUN apt-get install -y git cmake make g++ gcc wget swig python-pip pkg-config
+RUN apt-get install -y libgsl-dev libfftw3-dev
+RUN pip install numpy scipy matplotlib jupyter
 
-ENV GSL_TAR="gsl-2.3.tar.gz"
-ENV GSL_DL="http://ftp.wayne.edu/gnu/gsl/$GSL_TAR"
-ENV FFTW_TAR="fftw-3.3.6-pl2.tar.gz"
-ENV FFTW_DL="http://www.fftw.org/$FFTW_TAR"
+# Installing CCL C library
+RUN git clone https://github.com/EiffL/CCL && cd CCL && \
```
fjaviersanchez: Once merged you'll probably want to change this and point to the LSSTDESC/CCL repo, right?

cmake/Modules/BuildCLASS.cmake
```diff
@@ -0,0 +1,31 @@
...
+  set(CLASS_OMPFLAG "#OMPFLAG   = -fopenmp")
+else()
+  set(CLASS_OMPFLAG "OMPFLAG   = -fopenmp")
+endif()
+
+
+# Downloads and compiles CLASS
+ExternalProject_Add(CLASS
+        PREFIX CLASS
+        GIT_REPOSITORY https://github.com/lesgourg/class_public.git
```
fjaviersanchez: Nice!

Dockerfile
EiffL: Yes :-) It's there just so that I could build and test the Docker container before the merge

EiffL:
Thanks @fjaviersanchez

.gitignore
```diff
@@ -25,15 +25,33 @@ tests/ccl_test_power
...
 build/*
+dist/*
 tests/.ipynb_checkpoints/
 *ipynb*
 *log
 *dat
 *bak
 *dylib
+pyccl/ccllib.py
+pyccl/ccl_params.ini
```
philbull: Not sure why this file is in .gitignore -- it's a config file.

.travis.yml
```diff
@@ -1,20 +1,34 @@
...
-  - sudo apt-get install fftw3 fftw3-dev pkg-config
-#  - sudo apt-get install latexmk
-  - pip install numpy
-  - ./install-gsl.sh
-  - python class_install.py
+# Caching all the depencencies so that they don't have to be redownloaded and
+# compiled everytime
+cache:
+  directories:
+    - $HOME/build/EiffL/CCL/build/GSL/src
```
philbull: Travis has a `$TRAVIS_BUILD_DIR` environment variable, which points to the default Travis working directory -- it might be a bit cleaner to put the cached libraries in there instead.

.travis.yml
```diff
@@ -23,10 +37,22 @@ before_install:
...
+    - export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+    - export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig
+    - cmake --version
+    - python --version
+
 script:
-    - python tests/run_tests.py --debug --detailed-errors --verbose --process-restartworker
-    - make check-cpp
+    - python setup.py build
+    - python setup.py test
```
philbull: Some of the tests are decorated with a `slow` tag, so that they can be ignored by default when users run tests on their local machine. We want them to be run by Travis though. Is this what the `test` target does, or do we need to add a `--all` argument or similar?

cmake/Modules/BuildSWIG.cmake
```diff
@@ -0,0 +1,22 @@
+include(ExternalProject)
```
philbull: We originally had swig set up as a developer-only dependency, as it's really only needed to autogenerate some source files -- no need to add another dependency for users. (This came at the cost of keeping autogenerated files in version control, but that wasn't a big problem.) Perhaps CMake makes it easy enough to install that we don't need to worry about this any more? Let's see how testing goes, I guess.

pyccl/__init__.py
```diff
@@ -1,3 +1,7 @@
+# Sets the environment variable for default config path if it does not exist yet
+from os import environ, path
+if environ.get("CCL_PARAM_FILE") is None:
+    environ["CCL_PARAM_FILE"] = path.dirname(path.abspath(__file__)) + '/ccl_params.ini'
```
philbull: Just checking -- does the C code also have an environment variable check like this?

.gitignore
EiffL: Because these are regenerated automatically on build, on the local machine, they should not be stored on git

EiffL: (for the .ini file, it is copied over by from the include directory on build)

pyccl/__init__.py
EiffL: Well, the C code will first try the environment, then look for the hard coded install_dir/ccl_params.ini, then crash if not found. If the environment variable is defined by the user then both C and python codes will use the same config file

cmake/Modules/BuildSWIG.cmake
EiffL: Yep, sure, but in all my tests, swig compiled fine when missing from a system. The benefit is that it becomes completely transparent, the interface is automatically regenerated and user/developers don't even have to think about it.

.travis.yml
EiffL: good point, looks like doing python setup.py test just runs everything (even the slow tests). To have more control of the tests, one could use the `python setup.py nosetests` command (from what I understand `numpy.testing.run_module_suite` is just a wrapper for nosetests)

.travis.yml
EiffL: sure, thanks for the pointer, will update the Dockerfile

EiffL:
@elisachisari Ok done. Everything works now, including the Angpow integration, I think it's ready to merge. There is one last reference to my fork of Angpow4CCL but hopefully Jeremy can get that merged tomorrow.

I have uploaded a release to the test pypi:
https://test.pypi.org/project/pyccl
Looks very nice and can be installed with:
```
$ pip install --user --index-url https://test.pypi.org/simple/ pyccl
```

@fjaviersanchez you had created a pyccl package on pypi I think. I don't have the password for the desc ccl author, but I have created a section (untested) in the .travis.yml which should upload automatically to pypi whenever a new version of ccl is tagged.
Could you encrypt the password using the procedure described here: https://docs.travis-ci.com/user/deployment/pypi/

EiffL:
Thanks @fjaviersanchez, I tried to test it on my fork, but of course it didnt work because the sure 
 credentials  only work from this repo. I'm pretty it'll work fine though, we can always  fix it if it doesn't work once it's  merged in master

PR #367, opened 2018-04-07
Improve tests in cls.py

philbull:
Adds extra test coverage for several Python files, which gets most of them over 90% coverage. Also cleans up some docstrings and adds a couple of new error checks in `cls.py`.

Also in this PR: Travis now caches the version of GSL that it compiles, which speeds up the Travis tests by a few minutes.


tests/ccl_test_core.py
```diff
@@ -183,9 +183,20 @@ def test_cosmology_init():
     
     # Make sure error raised if incorrect type of Parameters object passed
     assert_raises(TypeError, ccl.Cosmology, params=params.parameters)
+    assert_raises(TypeError, ccl.Cosmology, params="xx")
```
fjaviersanchez: Just to be consistent with the rest why not use `params='x'`?

tests/ccl_test_pyccl_interface.py
```diff
@@ -364,13 +406,30 @@ def check_corr(cosmo):
...
-
-    t=np.logspace(-2,np.log10(5.),20) #degrees
-    corrfunc=ccl.correlation(cosmo,ells,cls,t,corr_type='L+',method='FFTLog')
-    assert_( all_finite(corrfunc))
-
+    ells = np.arange(3000)
+    cls = ccl.angular_cl(cosmo, lens1, lens2, ells)
+
+    t_arr = np.logspace(-2., np.log10(5.), 20) # degrees
+    t_lst = [t for t in t_arr]
```
fjaviersanchez: what about using `t_lst = t_arr.tolist()`?

philbull: That's a good tip, but I don't think it matters too much. Let's leave this unchanged.

tests/ccl_test_core.py
philbull: I'll change this now. Thanks!

philbull:
Awesome, thanks for the review @fjaviersanchez -- I'll merge this in once the massive neutrinos PR has gone in (as it might require some merge magic).

tests/ccl_test_pyccl_interface.py
fjaviersanchez: Yeah, here it doesn't make any difference.

Issue #368, opened 2018-04-11
Round-off errors in Omega_l affecting benchmarks


elisachisari:
If I create the following cosmology in python:
`cosmo_ccl1 = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7, A_s = 2.1e-9, n_s=0.96)`
and then print it, I see that Omega_l is not 0.7 but 0.6999... This is sufficient to make CCL python outputs deviate from the benchmark accuracy, for example, when computing comoving distances. Is there a way we can correct this behaviour even if it requires passing Omega_l directly?

tilmantroester:
According to https://github.com/LSSTDESC/CCL/blob/841a085ef7da5eaf2c90cb28c15caf32a1c05d27/src/ccl_core.c#L260,
Omega_l has contributions from Omega_g and Omega_n_rel, both of which are ~1e-5. Together they make up the difference between the reported Omega_l and 0.7.

elisachisari:
Thanks @tilmantroester. I managed to match the cosmology from the benchmarks better now. I force omega_g to 0 and restore omega_l to the benchmark value. But it still doesn’t fully solve the problem.

elisachisari:
Issue found and solved! The trick is to set up the parameters struct, then correct the cosmological parameters to the ones I want, and only then set up the cosmology. This does the trick and I'm now starting to reproduce benchmark checks in python. Thanks to @reneehlozek who helped with identifying this. See the notebook in the benchmark_notebook branch for more details.

PR #369, opened 2018-04-11
Jupyter notebook with benchmark plots for paper

elisachisari:
This PR incorporates a Jupyter notebook where we will develop all the plots needed to illustrate table 3 in the paper. The first commit includes an example for distances. Further examples need to be added. Also, results are currently limited by issue #368.


elisachisari:
Jupiter notebook now working! The next task is to add all the remaining plots

tilmantroester:
General:
iii) I think in the case of 10^-B, A=1.

v) Python 2 will get officially retired on January 1, 2020 (https://pythonclock.org/). New features in NumPy will only support Python 2 until **January 1, 2019** before deprecating support for Python 2 at the end of 2019 as well(https://docs.scipy.org/doc/numpy-1.14.0/neps/dropping-python2.7-proposal.html). Do we want to put effort into supporting a version of Python that will be deprecated before the time LSST goes live?

Background:
i) The old background benchmark plots that have less than 1e-4 agreement are the massive neutrino benchmarks that use astropy. As described in the note and paper, astropy uses an approximation that is only accurate at the 1e-3 level. Even though the CLASS benchmarks yield much better agreements, having benchmarks from an other, independent code is a good thing to have.

ii) I was under the impression that the accuracies quoted in table 2 are accuracy _targets_ and the actual achieved accuracy is some number smaller than this. I think that quoting the achieved accuracy too precisely is problematic, since the achieved accuracy depends on the integrators/constants/cosmological parameters/benchmark code/etc, which can change over time.

Consolidated background benchmark plots:
ii) CLASS uses an approximation when solving the growth factor ODE, which isn't sufficiently accurate. Note that CLASS doesn't use the growth factor internally, so this doesn't affect the accuracy of other CLASS results.

Angular power spectra:
i) The size of the plots is given by the size of the plots in the paper. This can be changed in cell 2. I guess we should discuss what the point of the benchmark notebook is. Is it to showcase CCL or to (re)create the plots in the paper?

elisachisari:
Answers in italics.

i) *Re: the overall comment on what sets the numerical accuracy: Table 2 does not show target accuracies but achieved ones. It is a summary of the plots presented. I would prefer to call this accuracy vs precision (in some cases of the comparisons we see systematics trends in the differences). The only quantity for which we officially have a target accuracy is the correlation function (which is the ultimate result to be used in the fitting of the data). And in this case, we set the target accuracy as 0.5 x expected error bar.*

ii) Can we see how all of this translates into the final observables, i.e. to show final observables when computed with high-speed settings vs high-precision settings? -> *We made an attempt at starting this work and a PR is open that has a timing notebook, but we think we need to make further improvements on speeding up CCL, so I don't think we can make any further statements on this at this point.*

iii) We should perhaps try to be consistent in how we quote the precision of the tests in Table 2, in some cases we quote A x 10^{-B}, where A and B are integers, while in other cases we simply quote 10^{-B}. -> *I think this would require some rounding-up, for example we could have 2x1E-4 -> 1E-3. I am more keen on quoting the actual accuracy we achieved and not downgrading it.*

iv) If we compare CCL against two separate codes, where one comparison is at the 10^{-A} level and the other comparison is at the 10^{-B} level, where A and B differ, do we quote 10^{-A} or 10^{-B} in Table 2? At some level, I guess we need to make a judgement call on which code we trust more? An example of this is our comoving distance comparison with astropy versus CLASS, obtaining a higher degree of agreement with the latter. -> *We rarely have the luxury of having more than one code. But in general, if we do, we should take the worst number as the resulting accuracy. For the specific astropy case, we did expect less accuracy overall. *

Background functions:

i) In Table 2 we list comoving distances as accurate at the 10^{-4} level, but in cell 11 we find a plot that plateaus at 6.5 x 10^{-4} for 1 < z < 5. Similar in cell 14. Perhaps update Table 2 to account for this? -> *Yes, I think this is for the cosmologies with neutrinos. I will make two separate entries in Table 2 for cosmologies with and without neutrinos.*

ii) Growth function plots seem to show agreement at ~5 x 10{-6} level in cells 14 and 15, but we list this as 10^{-4} in Table 2. Perhaps update the table? -> *Yes, will update.*

Consolidated background benchmark plots

i) The precision of the comoving distances are here at the 5 x 10^{-7} level (compared with CLASS), significantly better than the 6.5 x 10^{-4} found with astropy. Unless we have a reason to doubt CLASS, perhaps quote the stronger agreement in Table 2? If we were to do so, it seems the astropy comparison is redundant and can be removed? -> *Given that the automated checks use the astropy benchmark, I would leave this as is. Though perhaps we could say in the paper that a direct comparison to CLASS yields even better agreement and explain why.*

Angular power spectra:

i) Although the three plots are also found in the paper, along with descriptions, can we perhaps add some text in the notebook that clarifies to the user what is being plotted? For visual clarity, perhaps increase the size of the plots as they appear in the notebook? (Clarifying text and/or increased plot sizes applies to various other sections, e.g. Angpow, projected correlation function, 3d correlation, mass function, etc.) Arguably go beyond ell of 3 x 10^{4} as lensing integral will require larger ell on small angular scales. -> *Sizes of all figures have been increased. I made a general pass though the notebook ad removed all comments. I clarified what is being plotted in each section as you suggested.*

Angpow:

i) Cell 23, change “native” to “non-Limber” to be consistent in labeling with cell 22? -> *Changed*

ii) The target accuracy in cell 23 is set to 0.01. It seems it would be more useful to set this to some fraction of the cosmic variance instead? -> *I don't think the target accuracy means anything in this plot, I opted for removing it.*

iii) In Table 2, the range for Angpow is listed as 1 < ell < 1000. Perhaps more useful to simply list this as ell > 1, since the relative error decreases towards larger ell (as we can see in e.g. cell 24)? Moreover, why is the range for non-Limber in Table 2 listed as 500 < ell < 1000? It can be confusing to a reader that the relative error for Angpow is lower than for non-Limber, perhaps clarify in the caption that Angpow’s accuracy is relative to non-Limber, while non-Limber’s accuracy is relative to a different implementation in CLASS. -> *Ok, this is a comment on the paper, which I'll address separately*

Projected correlation function:

i) Cell 27 seems redundant (considering the plots in cell 29; moreover the y-label seems incorrect for cell 27 since divided by error and not 2*error). -> *Removed*

ii) For the projected correlation function benchmarks, can we please clarify if this assumes the same Cls for the two codes as the starting point? (i.e. are we only testing the angular power spectrum to correlation function part of the calculation, or the full calculation?) -> *It does not assume the same Cls, good point. @sukhdeep1989 @damonge can you confirm you agree with me?*

iii) In the paper we write that we calculated the expected statistical uncertainty of the clustering and lensing correlation functions of the LSST gold sample, and that we split the galaxy samples into 10 tomographic bins. But for Fig 14 we’re using the 2-bin n(z) in Fig 11, so a bit unclear how these are consistent approaches? -> *Pinging @elikrause*

Matter power spectrum

i) Could be useful to clarify in the notebook that “baryonic correction model” refers to Schneider & Teyssier (2015). Also please clarify/give reference for Cosmic Emulator (Lawrence et al 2017). -> *Done*

ii) Would it be more useful/conventional to have x-axis in units of h/Mpc for these plots? -> *CCL makes predictions in 1/Mpc, so I think it's preferable to have them this way.* 

Mass function

i) The plot in cell 37 seems to go from log M = 10 to 16, but we give the range 6 to 16 in Table 2. In the plot, the units on the x-axis are in terms of M_{\odot}, while we quote M_{\odot}/h in Table 2. We moreover list an accuracy of 10^{-4} while it seems to be significantly stronger, at the level of 2 x 10^{-5}. Perhaps show test for multiple redshifts? (and quote z in Table 2) Lastly, in the paper we list the test is for CCL1-3, while only a single cosmology is shown in the plot (perhaps add a label in the plot that CCL1 is shown?). -> *Pinging @villarrealas for clarification*

ii) For the plot in cell 38, same applies as in the previous plot in regards to x-range, x-axis label, single cosmology, and possibility of multiple redshifts. Here, the precision seems to be at the 10^{-3} level, not 10^{-4} quoted in the paper. Perhaps clarify the significant change in target accuracy compared to previous cell? -> *I think it suffices to clarify this is the accuracy we achieve and say it still gives us good accuracy for the full HMF calculation. Again pining @villarrealas *

iii) For the plot in cell 38, same applies in regards to single cosmology and possibility of multiple redshifts. Here, the precision seems to be about an order of magnitude stronger than quoted in Table 2 (i.e. 5 x 10^{-4} instead of 5 x 10^{-3}). Perhaps change y-axis label to be consistent with Table 2 (mathcal{H} seems unusual)? -> *I'll modify Table 2. The reviewers requested an axis change, that's why we adopted \mathcal{H}. I can change that in Table 2 instead to be consistent.*

3d correlation

i) For cells 41, 42, could be useful to clarify the difference between “analytic_xi.txt” and “analytic_xi_1.txt”. Seems cell 41 may be redundant? *Done and removed one of the cells* Given the plot in cell 42, perhaps downgrade the precision of 3d correlation from 4 x 10^{-3} quoted in Table 2? -> *Will do*

Power spectrum spline accuracy check

i) For the linear plot in cell 43, do we understand why the precision degrades as we move to higher redshift (as would have naively expected it the other way round)? Do we understand why the z = 20 case has a higher precision in the nonlinear case around k = 1 h/Mpc than in the linear case? Could be useful to better describe in the notebook how exactly the “high-accuracy settings” version is defined relative to default spline. For completeness, I would suggest increasing the y-axis and extending to larger k. -> *Ok, I think we should discuss this and add it to the paper.*

VARRIC plots -> *Pinging @philbull *

i) The plots have the label \Omega_{cdm} while the caption refers to \Omega_c, perhaps make sure the notation is consistent? *Done* We note that CCL always uses standard precision settings, which is why the agreement degrades when switching to high-precision for CLASS; would it be useful to allow a CCL user to access the matter power spectrum with both options for precision settings (if not already)? *Yes, and we have discussed this. We have made some improvements to pass these type of parameters more easily, but I think we still need to make more progress on that.*

ii) For the z = 0 plot, we find that the “Halofit, standard precision” agrees with the “Linear, standard precision” on large scales (low k) as expected. Why is this also not true for the z = 2 plots? (i.e. for sufficiently low k, Halofit should not be modifying the matter power spectrum)

iii) While the contribution is sub-dominant, perhaps expand plots beyond k_max = 20/Mpc as lensing integral will ask for larger k on small angular scales?

Halo model plots

i) This benchmark seems fine. Perhaps clarify in caption of Fig 10 that the three different cosmologies correspond to CCL1, WMAP, Planck (specifying the specific incarnations of WMAP/Planck)? *Pinging @alexander-mead*

ii) Perhaps add Fig 9 in the paper to the notebook? (only one currently not there) -> *Pinging @alexander-mead*

villarrealas:
Apologies for the delay; soooo many telecons.

The mass function benchmarks only cover a single cosmology, which is why only a single line is shown. I believe that line is for CCL1.

I also agree with you regarding the accuracy of Cell 38. I think this is associated with a strange spike in the splining (which I want to track down someday...), but we decided not to pursue since the variation between mass function models is significantly higher than this error.

damonge:
@elisachisari  @sjoudaki : yes, the correlation function benchmarks use the same cosmological parameters and P(k) to begin with, but all intermediate calculations (including C_ells) are different.

damonge:
@sjoudaki @tilmantroester : regarding python2.7, I agree that at some point we should stop wasting effort supporting obsolete python versions, but I think there are still some users in DESC that would appreciate python2.7 for some time (Shahab is a good example - although c'mon man, move on!).

In any case I don't think this one script is a sufficiently relevant reason to make the decision to stop supporting 2.7 (we clearly do, since travis tests for it). Would it be very hard to change the aspects of it that are python3-specific?

I still think Tilman's point stands, and we should (very) soon start thinking about moving on from 2.

tilmantroester:
@damonge The python 3 stuff is only some dict combinations if I remember correctly. This should be easy to fix. I just need to figure out how to combine two dicts in python 2.

sjoudaki:
Hi all, thanks a lot for your replies and upgrades to the notebook and paper! Just a few minor items below:

i) Can we perhaps clarify the choice of 0.5-sigma for the target accuracy of the correlation functions? Ideally, it seems we would want the theory to be precise at the <0.1-sigma level?

ii) In the notebook we have "accuracy = X" (where X is a real number) in multiple instances, which now refers to achieved precision. Can we perhaps go over the plots to make sure it is fair reflection of the results, e.g. \Delta Cl / Cl in cell 24 has significantly stronger precision than 0.01, while \Delta \chi / \chi in cell 28 has weaker precision than 1? In the former case, I understand that we want to round up, but perhaps make sure this is consistently applied?

iii) David writes: “yes, the correlation function benchmarks use the same cosmological parameters and P(k) to begin with, but all intermediate calculations (including C_ells) are different.” Can we please add this clarification in the opening text of that section (projected correlation function)?

iv) The precision of \script(H) seems to be 5 x 10^{-5}, not 5 x 10^{-3}. The precision in log \sigma^{-1} seems to ~8 x 10^{-4}, not 5 x 10^{-3}. And, the precision of \Delta \sigma seems to be 2 x 10^{-5}, not 10^{-4}. Since the mass function benchmarks only cover a single cosmology, please update Table 2.

v) For the "generalized validation..." section, it seems this wasn't addressed: “For the z = 0 plot, we find that the ‘Halofit, standard precision’ agrees with the ‘Linear, standard precision’ on large scales (low k) as expected. Why is this also not true for the z = 2 plots? (i.e. for sufficiently low k, Halofit should not be modifying the matter power spectrum).”

vi) Couple typos in the notebook, “analytuic benchmark” —> “analytic benchmark”, and "next wet of plots" --> "next set of plots".  

vii) As a side note, in the paper, we keep defining A for every new observable, perhaps just define it once generically instead?

elisachisari:
i) Can we perhaps clarify the choice of 0.5-sigma for the target accuracy of the correlation functions? Ideally, it seems we would want the theory to be precise at the <0.1-sigma level?
-> Let's discuss this.

ii) In the notebook we have "accuracy = X" (where X is a real number) in multiple instances, which now refers to achieved precision. Can we perhaps go over the plots to make sure it is fair reflection of the results, e.g. \Delta Cl / Cl in cell 24 has significantly stronger precision than 0.01, while \Delta \chi / \chi in cell 28 has weaker precision than 1? In the former case, I understand that we want to round up, but perhaps make sure this is consistently applied?
-> I had another pass through the notebook, updated accuracy values and did the same in the table in the paper. I decreased "accuracy" for cell 24 (CMB lensing). We discussed the cell 28 case in person: here we are calculating |xi_ccl/xi_benchmark-1|. The reason why this value can be high is that xi_benchmark goes to 0 at large-scales. This is why the more meaningful comparison is the one in cell 27, where we compare |xi_ccl-xi_benchmark| to the expected LSST error bars.

iii) David writes: “yes, the correlation function benchmarks use the same cosmological parameters and P(k) to begin with, but all intermediate calculations (including C_ells) are different.” Can we please add this clarification in the opening text of that section (projected correlation function)?
-> Added.

iv) The precision of \script(H) seems to be 5 x 10^{-5}, not 5 x 10^{-3}. The precision in log \sigma^{-1} seems to ~8 x 10^{-4}, not 5 x 10^{-3}. And, the precision of \Delta \sigma seems to be 2 x 10^{-5}, not 10^{-4}. Since the mass function benchmarks only cover a single cosmology, please update Table 2.
-> All updated. Note sigmaM does cover CCL1-3. I checked in the C file.

v) For the "generalized validation..." section, it seems this wasn't addressed: “For the z = 0 plot, we find that the ‘Halofit, standard precision’ agrees with the ‘Linear, standard precision’ on large scales (low k) as expected. Why is this also not true for the z = 2 plots? (i.e. for sufficiently low k, Halofit should not be modifying the matter power spectrum).”
-> This was attributed to the CLASS switch that operates between the halofit and linear power spectrum. @philbull says: There is some subtle difference in how the linear power spectrum is retrieved depending on whether this switch is turned on or not, which is causing a discrepancy (discussion in Slack). But this is well below 10^{-4} so we think it's not worth investigating further at this point.

vi) Couple typos in the notebook, “analytuic benchmark” —> “analytic benchmark”, and "next wet of plots" --> "next set of plots".
-> Fixed.

vii) As a side note, in the paper, we keep defining A for every new observable, perhaps just define it once generically instead?
-> I think it's better to be explicit.

In general, I updated the draft and all the C tests to reflect achieved accuracy, meaning some of the tolerance values have gone down.

sjoudaki:
Thanks Elisa! Yes, let's discuss 0.5-sigma for the target accuracy. Minor items below:

i) For cell 28, I understand that xi_benchmark goes to 0 at large scales, but why do we have a line for accuracy at 0.01?

ii) In Table 2, it seems should be 6 x 10^{-7} for distances and 5 x 10^{-6} for growth? For "Cl clustering non-Limber", seems it's 1 x 10^{-2}? For cell 30, 3 x 10^{-2} seems fine, but for cell 29 could e.g. have 5 x 10^{-3} instead of 10^{-2}? Being a bit picky here, for 3d correlation (cell 39), seems it's 3 x 10^{-2}?

iii) in cell 35, perhaps change this sentence? "In the case with neutrinos, the result is an accuracy of 1E-3." (alternatively shift up the dashed accuracy line in cell 34 given the ratio for k > 15/Mpc) Above cell 40, there is buggy text: "...is the case of %%latex..." Equation below, k0 ---> k_0.

iv) As we list precisions for Pk_{BBKS} and Pk_{Eisenstein-Hu} in Table 2, perhaps include explicit plots for these in the notebook?

v) Since the linear to Halofit Pk ratio cases is off by orders of magnitude on large scales (z = 2 case), perhaps useful to also plot Halofit Pk / linear Pk for CCL alone, and for CLASS alone, as another sanity-check. (to be further discussed in person)

elisachisari:
i) For cell 28, I understand that xi_benchmark goes to 0 at large scales, but why do we have a line for accuracy at 0.01?
-> Removed.

ii) In Table 2, it seems should be 6 x 10^{-7} for distances and 5 x 10^{-6} for growth? 
-> I don't think so, it really is 5x10^{-7} for distances. Growth is 6x10^{-6}. If I impose 5x10^{-6}, model 4 of the checks fails (cell 14).
For "Cl clustering non-Limber", seems it's 1 x 10^{-2}? 
-> So this is not really cell 23. It's the fractional difference between the curves in cell 23 divided by cosmic variance. That's why it's 2x10^{-2}.
For cell 30, 3 x 10^{-2} seems fine, but for cell 29 could e.g. have 5 x 10^{-3} instead of 10^{-2}? Being a bit picky here, for 3d correlation (cell 39), seems it's 3 x 10^{-2}?
-> For the cosmicemu checks, we are bound by the accuracy of the emulator itself. These numbers are coming from the emulator paper and we cannot change them. We are not expecting to be more accurate than the emulator itself is.

iii) in cell 35, perhaps change this sentence? "In the case with neutrinos, the result is an accuracy of 1E-3." (alternatively shift up the dashed accuracy line in cell 34 given the ratio for k > 15/Mpc) 
-> Done.
Above cell 40, there is buggy text: "...is the case of %%latex..." Equation below, k0 ---> k_0.
-> Done.

iv) As we list precisions for Pk_{BBKS} and Pk_{Eisenstein-Hu} in Table 2, perhaps include explicit plots for these in the notebook?
-> Added.

v) Since the linear to Halofit Pk ratio cases is off by orders of magnitude on large scales (z = 2 case), perhaps useful to also plot Halofit Pk / linear Pk for CCL alone, and for CLASS alone, as another sanity-check. (to be further discussed in person)
-> Pinging @philbull

PR #370, opened 2018-04-13
Add ability to print C error messages from Python

philbull:
The Python wrapper normally suppresses error messages that happen in the C code (and prevents them from being fatal). Errors are then detected in the Python wrapper by checking the status flag, which then raises a Python exception.

Since the errors aren't fatal, the C code continues to run after the first exception is raised. As such, it's possible that more errors will be raised, overwriting what was there before. This makes it harder to debug the errors from Python, as the root cause of the issue may be somewhere completely different from where the most recent error was raised.

This PR allows users to enable a "debug" mode, by calling `ccl.debug_mode(True)` from Python. This makes the C code print all error messages, even though they're not fatal.


elisachisari:
@philbull could you document this new feature in the note and README?

tilmantroester:
The code looks good and works as advertised when I tested it. It works for the notebooks in the sense that the error messages get written to stderr, which isn't visible from the notebook by default. Since this change is only for debugging purposes, I think this is fine.
I assume the deleted part in ccl.i is no longer needed?

philbull:
Thanks folks, I added a note on `debug_mode` in `CONTRIBUTING.md`. I'm not sure this is something we want to document in the note, as it's pretty developer-focused, but happy to do that if you think it's worth it.

elisachisari:
Sounds good, I'm happy for this to be in CONTRIBUTING.md

Issue #372, opened 2018-04-17
Differences in Omega_n_rel between CCL and CLASS


tilmantroester:
At the moment CCL and CLASS seem to give different values (factor of ~2) for the energy density of relativistic neutrinos. 
Here's the snippet that should create the same cosmology but apparently doesn't:
```
import classy
import pyccl as ccl

class_params = {"output"    : "",
                "T_cmb"     : 2.725,
                "h"         : 0.7,
                "Omega_cdm" : 0.25,
                "Omega_b"   : 0.05,
                "Omega_k"  : 0.0,
                "N_ur"     : 3.0}
cosmo = classy.Class()
cosmo.set(class_params)
cosmo.compute()

ccl_params = ccl.Parameters(Omega_c=class_params["Omega_cdm"],
                            Omega_b=class_params["Omega_b"],
                            h=class_params["h"],
                            Omega_k=class_params["Omega_k"],
                            Neff=class_params["N_ur"],
                            A_s=2.0e-9,
                            n_s=1.0,
                            m_nu=0.0)

print("Omega_L: CLASS: {}, CCL: {}".format(cosmo.Omega_Lambda(), ccl_params.parameters.Omega_l))
print("Difference: {}".format(ccl_params.parameters.Omega_l-cosmo.Omega_Lambda()))
print("Fractional difference: {}".format(ccl_params.parameters.Omega_l/cosmo.Omega_Lambda()-1))
print()
print("Omega_g: CLASS: {}, CCL: {}".format(cosmo.Omega_g(), ccl_params.parameters.Omega_g))
print("Fractional difference: {}".format(ccl_params.parameters.Omega_g/cosmo.Omega_g()-1))
print()

Omega_ur = cosmo.get_background()["(.)rho_ur"]/cosmo.get_background()["(.)rho_crit"]
print("Omega_ur: CLASS: {}, CCL: {}".format(Omega_ur[-1], ccl_params.parameters.Omega_n_rel))
print("Difference: {}".format(ccl_params.parameters.Omega_n_rel-Omega_ur[-1]))
print("Fractional difference: {}".format(ccl_params.parameters.Omega_n_rel/Omega_ur[-1]-1))
```
This yields the following output for me:
```
Omega_L: CLASS: 0.6999152076713755, CCL: 0.6999327314693308
Difference: 1.75237979552767e-05
Fractional difference: 2.5037029862007287e-05

[...]

Issue #176, opened 2017-05-02
Test distances to recombination


slosar:
Current checks tests distances to z=5. It would be good to extend this to z=1150 and cross-check against CLASS for cosmologies with neutrinos. These distances enter in CMB theta parameter which is constrained to 1e-4 and quite important so one would expect people to use CCL in this regime and expect high accuracy.

c-d-leonard:
I did a cursory check of CCL vs CLASS at z about 1090 (taking a z for which background quantities were output by CLASS). I found that when calling ccl.background.comoving_angular_distance() at this z, the relative agreement with CLASS was only about 1e-3, so this needs work. This was for a cosmology with no massive neutrinos. To do this, I had to adjust A_SPLINE_MIN down to 0.0001.

philbull:
Hi @c-d-leonard, any progress on this? Do you know of any sensible tests we can add?

c-d-leonard:
I think that @damonge 's work on high-z may have solved this. We now have tests out to high z (ccl_test_distances_hiz.c for example). David is this solved?

damonge:
It might, but we haven't checked explicitly that our chi(z_LSS) matches CLASS's/CAMB's numbers for the same set of parameters

slosar:
Yes, let's do this -- this is the last thing I would like to see tested before I switch to CCL wholesale! :)

tilmantroester:
I think with PR #374 (and #390 ) having been merged, this can now be closed.

PR #374, opened 2018-04-27
Distance tests to z=1050 for massive neutrinos and curvature

tilmantroester:
This PR aims to address issue #176. It includes tests at 10 logarithmically spaced redshifts between 0.01 and 1050 for:

- no curvature, no massive neutrinos, 3 massless neutrinos
- positive curvature, no massive neutrinos, 3 massless neutrinos
- negative curvature, no massive neutrinos, 3 massless neutrinos
- no curvature, 1 massive neutrinos, 2 massless neutrinos
- no curvature, 3 massive neutrinos (2 different masses), no massless neutrinos
- no curvature, 3 massive neutrinos (3 different masses), no massless neutrinos
- no curvature, no massive neutrinos, 6 massless neutrinos
- negative curvature, 2 massive neutrinos, 4 massless neutrinos
- positive curvature, 3 massive neutrinos, 3 massless neutrinos

The distances are compared against a benchmark created with CLASS. The notebook to create the benchmark is included in the PR.

In order for the tests to pass, the calculation of Omega_nu_rel needs to be fixed (see issue #372), which currently breaks some tests (CMB lensing and 3D correlation function), however.

The new tests pass with tolerance of 1e-5 if the first redshift (z=0.01) is omitted, such that the tests are from z=0.03 to z=1050. For z=0.01, the tolerance needs to be relaxed to 1e-4.


slosar:
This is great! So, there was a bug in h in the neutrino branch that you now fixed?
One question: this z=0.01 doesn't matter, but it is annoying. It is also possible it is a class issue, rather than our issue. Can you see if it can get cured using either improving the number of points in our spline or class spline? Should be minimal change. Otherwise I'm happy with this to go forward. Great work!!

tilmantroester:
I changed the `back_integration_stepsize` parameter in CLASS to something smaller and now get 1e-6 agreement with CCL everywhere (which is also the precision setting for distance integrals in CCL, I think). The test still uses 1e-5, in case we add more challenging models later on.

src/ccl_core.c
```diff
@@ -243,16 +243,22 @@ void ccl_parameters_fill_initial(ccl_parameters * params, int *status)
...
+  double rho_crit = RHO_CRITICAL * SOLAR_MASS/pow(MPC_TO_METER, 3) * pow(params->h, 2);
+  params->Omega_g = rho_g/rho_crit;
   
   // Get the N_nu_rel from Neff and N_nu_mass
-  params->N_nu_rel = params->Neff - params->N_nu_mass * TNCDM * TNCDM * TNCDM * TNCDM / pow(4./11.,4./3.);
-  
-  //Get the relativistic neutrino Omega_nu. It's more efficient to get this in-line, to avoid computing the phase-space integral if not necessary.
-  double Tnu= (params->T_CMB) *pow(4./11.,1./3.); 
-  params-> Omega_n_rel = params->N_nu_rel* 8. * pow(M_PI,5) *pow((KBOLTZ/ HPLANCK),3)* KBOLTZ/(15. *pow( CLIGHT,3))* (8. * M_PI * GNEWT) / (3. * 100.*100.*1000.*1000. /MPC_TO_METER /MPC_TO_METER  * CLIGHT * CLIGHT)  * Tnu * Tnu * Tnu * Tnu *7./8.;
+  params->N_nu_rel = params->Neff - params->N_nu_mass * pow(TNCDM, 4) / pow(4./11.,4./3.);
```
c-d-leonard: I'm okay with chaning TNCDM * TNCDM * TNCDM * TNCDM to pow(TNCDM,4) if its more efficient - have you tested this? If its for readability at the expense of efficiency I would prefer to keep it as in, but others may disagree.

src/ccl_core.c
```diff
@@ -243,16 +243,22 @@ void ccl_parameters_fill_initial(ccl_parameters * params, int *status)
...
   
+  // Temperature of the relativistic neutrinos in K
+  double T_nu= (params->T_CMB) * pow(4./11.,1./3.); 
+  // in kg / m^3
+  double rho_nu_rel = params->N_nu_rel* 7.0/8.0 * 4. * STBOLTZ / pow(CLIGHT, 3) * pow(T_nu, 4);
+  params-> Omega_n_rel = rho_nu_rel/rho_crit;
+    
   // If non-relativistic neutrinos are present, calculate the phase_space integral.
```
c-d-leonard: I checked the math on this new expression vs the old one and I agree it the new Omega_n_rel expression solves the issues of the missing h^2 factor.

src/ccl_core.c
tilmantroester: I'm of the opinion that readability should always come before efficiency unless code has been shown to be a mission-critical bottleneck.
For this specific example, I'd be very surprised if the compiler doesn't optimise both expressions to the same machine code anyway. Since this is function is just called once per cosmology, efficiency shouldn't be an issue either.

tilmantroester:
I've now fixed the tests and moved the new tests to ccl_test_distances.py. Only the CMB lensing test was affected, the 3d correlation test failing was due to an interaction between the two tests. This has been addressed now but the way CTEST deals with clean-up should probably be revisited at some point.

tests/ccl_test_distances.py
```diff
@@ -257,7 +308,24 @@ def compare_distances_mnu_hiz(z, chi_bench,dm_bench, Omega_v, w0, wa, Neff_mnu,
     dm = ccl.distance_modulus(cosmo,a[a_not_one])
 
     assert_allclose(dm, dm_bench[a_not_one], atol=1e-3, rtol = DISTANCES_TOLERANCE_MNU)
+
+@decorators.slow
```
c-d-leonard: We probably don't need to label this as slow - usually calling distances is pretty fast? We only want to use this label for things that take a really long time to compute and significantly slow down running the test script.

PR #377, opened 2018-05-12
Struct and helper functions to track multiple parameter structures

philbull:
(In progress)
To support extensions to CCL, we need some way of handling new, non-standard sets of parameters. This PR proposes a way of handling extended parameter sets.

In addition to the core `ccl_parameters` object, each extension can define its own custom parameters struct, containing any new parameters that it needs. The extension should define functions to create and initialize these structs, similar to the functions that do this for `ccl_parameters` in `ccl_core.c`.

When an extension is enabled, the user can add the corresponding parameter struct to a `ccl_paramset_list`, using the `ccl_add_paramset()` function. (The standard `ccl_parameters` struct is also kept in this list). If multiple extensions are enabled, multiple structs can be added to the list. The list is then attached to a `ccl_cosmology` object. For example:
```
ccl_paramset_list pslist;
modgrav_params mg_params;
ccl_add_paramset(&pslist, "modgrav", &mg_params);
```

Each extension module can retrieve its own parameters struct from the list by using the `ccl_get_paramset()` function. This retrieves the paramset by name. The function returns a void pointer, which the extension module must cast to the correct type, e.g.
```
modgrav_params* mg_params;
mg_params = (modgrav_params*) ccl_get_paramset(&pslist, "modgrav", &status);
printf("MG parameter mu: %3.3f\n", mg_params->mu);
```
Functions are also provided to check if a paramset exists in the list, remove a paramset from the list, and print the names of all paramsets in the list.

At the moment, this PR only provides the new `ccl_paramset_list` struct and helper functions. Assuming that people are happy with this way of doing things, this PR should be further developed to plumb them in to CCL.

### DIFF
--- a/include/ccl_core.h
+++ b/include/ccl_core.h
@@ -122,7 +122,7 @@ typedef struct ccl_data{
 } ccl_data;
 
 /**
- * Sturct containing references to instances of the above structs, and boolean flags of precomputed values.
+ * Struct containing references to instances of the above structs, and boolean flags of precomputed values.
  */
 typedef struct ccl_cosmology
 {
--- a/include/ccl_paramset.h
+++ b/include/ccl_paramset.h
@@ -1,0 +1,31 @@
+/** @file */
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#pragma once
+// Max. no. of paramsets allowed in the list
+#define CCL_PARAMSET_LIST_SIZE 30
+
+// Status variables
+#define CCL_PARAMSET_NOT_FOUND 1
+#define CCL_PARAMSET_ALREADY_EXISTS 2
+
+typedef struct ccl_paramset_list {
+    void* paramset[CCL_PARAMSET_LIST_SIZE];
+    int num_sets;
+    char name[CCL_PARAMSET_LIST_SIZE][128];
+} ccl_paramset_list;
+
+
+int ccl_paramset_list_init(ccl_paramset_list* pslist);
+int ccl_has_paramset(ccl_paramset_list* pslist, char* name);
+void* ccl_get_paramset(ccl_paramset_list* pslist, char* name);
+void ccl_add_paramset(ccl_paramset_list* pslist, char* name, void* params, 
+                      int* status);
+void ccl_remove_paramset(ccl_paramset_list* pslist, char* name, int* status);
+void ccl_list_paramsets(ccl_paramset_list* pslist);
+
+#ifdef __cplusplus
+}
+#endif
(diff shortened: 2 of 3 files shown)


include/ccl_core.h
```diff
@@ -122,7 +122,7 @@ typedef struct ccl_data{
 } ccl_data;
 
 /**
- * Sturct containing references to instances of the above structs, and boolean flags of precomputed values.
+ * Struct containing references to instances of the above structs, and boolean flags of precomputed values.
  */
```
c-d-leonard: Did you intend to add the new paramslist object to ccl_cosmology, or is it correct that only a comment changed here?

c-d-leonard:
Hi @philbull , this looks great. I'm wondering how the new paramset_list object would get associated to a ccl_cosmology? Would it be a new component of ccl_cosmology? Otherwise it seems as though we would need to pass that around to all the functions as well.

c-d-leonard:
Also, it occurs to me that in addition to parameters, some extensions may require the additional data splines to be stored. For example, for the simple mu / Sigma modified gravity implementation we have discussed, it seems to me that we would need to store both the modified growth rate and the GR growth rate, in order to use the ratio of these in rescaling the GR power spectrum. Can this set-up allow for additional data elements like that?

include/ccl_core.h
philbull: I didn't intend to add it yet. That's the next step though.

philbull:
Options would be to (a) replace the existing `ccl_parameters` instance associated with `ccl_cosmology` with a paramset_list; or (b) append a paramset_list, in addition to a "vanilla" `ccl_parameters` instance. Both require a reasonably substantial overhaul of how parameters are accessed, so I think it deserves some further thought. We wouldn't pass it around as a separate object though; it should definitely be wrapped in a `ccl_cosmology` object.

I think this sort of structure could be used to hold splines as well. Perhaps I can add some way of classifying struct members as "data" or "parameter"? I would just worry about memory management in this case (everything is pointers!).

WeikangLin:
Hi @philbull, these looks great! 
For the above if I understand correctly, option (b) sounds better. For example (not sure it will work), an `extra_paramset` with type `ccl_paramset_list` can be first created using the above `ccl_add_paramset` function. An additional instance can be added to `ccl_cosmology`, which is empty if, e.g., MG is not considered. Then the access of the vanilla parameters should be the same, and the additional parameters are accessed via `cosmo->extra_paramset.param`?

If we implement MG as another power spectrum method, looks like this method needs to be followed by a specific parameter_create function. But this doesn't seem to be a big issue for only one extension to the vanilla at a time.

beckermr:
Dev has moved to python. Closing

PR #378, opened 2018-05-21
CCL paper

elisachisari:
With the renaming of the CCL branch, the [old pull request](https://github.com/LSSTDESC/CCL/pull/247) #247 was closed. I'm opening it again, now with he new branch. We should bear in mind the comment thread in the old one though.


elisachisari:
For reference, here are @timeifler's comments: 
https://www.dropbox.com/s/194um1za83qtnv2/ccl_paper.pdf?dl=0
Updates will follow.

PR #390, opened 2018-05-31
Additional CLASS distance tests

tilmantroester:
I added the CCL7-11 models to CLASS distance test to cover the same models described in the paper. I also added tests for distance moduli for the CLASS benchmarks. The agreement is still at 1e-6.


tilmantroester:
I agree that the current setup with a long list of mostly identical test_* functions isn't very neat and should probably be generated in some way. I just added the new tests in the same format as the existing tests in `ccl_test_distances.py`.

PR #391, opened 2018-05-31
Updates the travis ci encryption and adds documentation for PyPi deployement

EiffL:
This is a small PR that does the  following:
 - updates the version number in setup.py to v0.3.1
 - updates the encrypted pypi deploy key in travis.ci script (the problem was that special characters need to be escaped when encrypting the password)
 - adds a section in the CONTRIBUTING.md file to explain the procedure to follow when cutting a new version and deploying to PyPi


.travis.yml
```diff
@@ -1,82 +1,49 @@
...
-          python: 3.6
-          env: TOXENV=py36
-        - os: osx
-          language: generic
-          env: TOXENV=py27
-        - os: osx
-          language: generic
-          env: TOXENV=py36
-
-#addons:
```
fjaviersanchez: I think that this part was to deploy the paper stuff in travis but, I guess it's not necessary.

CONTRIBUTING.md
```diff
@@ -240,3 +242,31 @@ click in your build you will find more information about its status and a log
 describing the process. If your build errors or fails, you can scroll through
 the log to find out what went wrong. If your additions require new dependencies
 make sure that you include them in `.travis.yml`.
+
```
fjaviersanchez: Great!

.travis.yml
EiffL: got erased automatically by the travis script, and was commented out anyway so I figured it wasn't really necessary

PR #401, opened 2018-07-04
replaced sigma_8 -> sigma8

### LINKED ISSUE #399: Incorrect error for sigma8 naming convention
I recently got the following error from CCL:
`RuntimeError: Error CCL_ERROR_INCONSISTENT: ccl_power.c: ccl_cosmology_compute_power_emu(): sigma_8 is not defined; specify sigma_8 instead of A_s`

The problem is that the correct parameter name is `sigma8` with no `_`. I could not locate in what file the error message is written. This should be an easy fix.

damonge:
Fixes #399 as well as a couple other issues (bug in notebook and global pointer initialization). @tmcclintock , could you review?


tmcclintock:
Looks fine to me. Just curious why you added this line (and the one below it):
`+ccl_gsl_params * ccl_gsl=NULL; // Global variable`
I'm not sure why adding the `=NULL` was necessary.

damonge:
@tmcclintock : the `=NULL` lines are to initialize the global pointers. It's already pretty bad that we have global variables, but having them uninitialized is even worse. Especially since CCL checks explicitly for `ccl_gsl==NULL` to determine whether or not it has to allocate it and fill it up.

PR #402, opened 2018-07-06
3dcorrelation rsd

zdu863:
Added new function `ccl_correlation_3dRsd` to calculate the two-point correlation function in redshift space from power spectrum, using the linear theory given by Kaiser(1987). Also added the python wrapper and an example ipython notebook.


ellison-j:
@elisachisari, @dkirkby @slosar 

We have implemented functions to calculate the  redshift-space correlation function xi(s, \mu), it's average at constant s, xi(s), the multiploes xi_l(s), and xi(pi, sigma).

The examples/Correlation_3dRSd.ipynb notebook has example plots of the correlation functions.

The benchmark is based on comparing xi(s) with that calculated from CosmoMAD [https://github.com/damonge/CosmoMAD]. We use the power spectrum from CCL and pass it into CosmoMAD then compare the xi(s) produced by CosmoMAD to the CCL one.

We find a maximum difference in s^2 xi(s) of 0.185 for s = 0.1 - 250 Mpc which corresponds to 0.52% of the height of the BAO peak (s^2 xi(s) = 35.7). See plot in example/RSD_benchmark.ipynb notebook (but note that needs CosmoMAD to be installed to run). 

The functions are implemented in src/ccl_correlation.c.

We updated section 4.10 of the note describing what was done, but we have not added anything about the benchmarks yet. We will add that next.

David, Anze, would you be able to review this PR?

dkirkby:
Is there already a requirement written down for this?  If so, what is it based on?  Perhaps it would be useful to have a test case where the isotropic P(k) is distorted by something more than linear Kaiser, where higher order multipoles are important (e.g., a roll off in k_par).

ellison-j:
> Is there already a requirement written down for this? If so, what is it based on? Perhaps it would be useful to have a test case where the isotropic P(k) is distorted by something more than linear Kaiser, where higher order multipoles are important (e.g., a roll off in k_par).

I don't think there is a requirement that I know of, so we just quote the accuracy we find with respect to the benchmark.  We only implemented linear Kaiser for now as we have a benchmark for that (CosmoMAD). It could be extended later but we do not know of a calculation to benchmark against.

elisachisari:
Could you clarify what's being plotted on the first plot? Do you have a fractional difference plot as a function of s (in addition to the absolute one)? And what about a benchmark plot as a function of sigma and Pi? Would this be possible?

ellison-j:
Hi @elisachisari 

The first plot shows xi(pi, sigma).

The fractional difference plot is shown in the attachment
[RSD_plots_2.pdf](https://github.com/LSSTDESC/CCL/files/2617277/RSD_plots_2.pdf)
. Bear in mind of course that xi(s) passes through zero and so the fractional difference blows up to infinity when xi(s) = 0 as for xi(r).

Regarding a benchmark for xi(pi, sigma) we show a plot of the difference in s^2 xi(pi,sigma) between CCL and CosmoMAD. The maximum for pi, sigma out to 200 Mpc is ~ 1.25. This is about 3.5% of the BAO peak in s^2 xi(s) averaged over all directions. The maximum occurs at pi, sigma = (\pm 200, 0). For most of the region in (pi,sigma) the difference is much smaller than this.

@dkirkby: further to your question, the issue of going beyond linear was discussed in the CCL meeting and the conclusion was that due to the relatively large uncertainties in the photometric redshifts it is not needed at present. If you or @slosar have additional comments, please let us know. Thanks!

pyccl/ccl_correlation.i
```diff
@@ -39,4 +41,48 @@ void correlation_3d_vec(ccl_cosmology *cosmo,double a,
...
   ccl_correlation_3d(cosmo,a,nr,r,xi,0,NULL,status);
 }
+
+void correlation_multipole_vec(ccl_cosmology *cosmo,double a,double beta,
+			       int l,int ns,double *s,
+                               double *xis,int nxis,
+         		       int *status)
+{ 
+  assert(ns==nxis);
+
```
slosar: Do we need all this empty lines in this boilerplat code? (follow whatever is done elsewhere)

src/ccl_correlation.c
```diff
@@ -506,3 +509,419 @@ void ccl_correlation_3d(ccl_cosmology *cosmo, double a,
...
+  if (pk_arr == NULL) {
+    free(k_arr);
+    *status = CCL_ERROR_MEMORY;
+    strcpy(cosmo->status_message,
+           "ccl_correlation.c: ccl_correlation_multipole ran out of memory\n");
+    return;
+  }
+
+  for (i = 0; i < N_ARR; i++)
+    pk_arr[i] = ccl_nonlin_matter_power(cosmo, k_arr[i], a, status);
```
slosar: No, no. By default we calculate correlation function for liner power spectrum. Or, for BAO analysis you calculate it separately for wiggle and no-wiggle parts with wiggle part getting LPT damping... but this is beyond current scope. You should make it at minimum optional with default to linear.

src/ccl_correlation.c
```diff
@@ -506,3 +509,419 @@ void ccl_correlation_3d(ccl_cosmology *cosmo, double a,
...
+INPUT:  cosmology, scale factor a, beta (= growth rate / bias), 
+        multipole order l = 0, 2, or 4, number of s values, s values
+
+Multipole function result will be in array xi
+ */
+
+void ccl_correlation_multipole(ccl_cosmology *cosmo, double a, double beta,
+                               int l, int n_s, double *s, double *xi,
+                               int *status) {
+  int i, N_ARR;
```
slosar: Shouldn't this check if I have a splined version and if so use that?

src/ccl_correlation.c
```diff
@@ -506,3 +509,419 @@ void ccl_correlation_3d(ccl_cosmology *cosmo, double a,
...
+             "ccl_correlation.c: ccl_correlation_3dRsd_avgmu ran out of memory\n");
+      return;
+    }
+
+    ccl_correlation_multipole(cosmo, a, beta, 0, n_s, s, xi_arr0, status);
+    ccl_correlation_multipole(cosmo, a, beta, 2, n_s, s, xi_arr2, status);
+    ccl_correlation_multipole(cosmo, a, beta, 4, n_s, s, xi_arr4, status);
+    for (i = 0; i < n_s; i++)
+      xi[i] = xi_arr0[i] + xi_arr2[i] * 1./4 +
+              xi_arr4[i] * 9./64;
```
slosar: Huhm, are these factors correct? Perhaps yes, but I don't recognize the numbers.

src/ccl_correlation.c
ellison-j: The CCL function ccl_nonlinear_power will return either the linear or nonlinear power spectrum depending on how the cosmology object (cosmo) is set up by the user. If cosmo is set up with matter_power_spectrum=linear it will use the linear P(k). For the benchmarks we do use the nonlinear P(k), since we want to compare differences in s^2 xi(s) with the BAO peak.

src/ccl_correlation.c
ellison-j: Good point. We will do that.

ellison-j: It is not necessary. We only create the global spline (ccl_correlation_multipole_spline) since ccl_correlation_pi_sigma needs to call this many times.  ccl_correlation_multipole and  ccl_correlation_multipole_spline are not needed by the user, they are only used to calculate xi(s,mu), xi(pi,sigma) and xi(s).

src/ccl_correlation.c
ellison-j: Yes these are correct, but different from the coefficients in the expressions for xi(s,mu) that you are probably thinking of. This code is calculating xi(s) = average of xi(s,mu) at constant s. So here we calculate the multipoles  and multiply by the average of L_l(mu). So these coefficients are essentially the integrals of the Legendre polynomials.

src/ccl_correlation.c
slosar: So, for BAO, the accurate, well tested model that is used in BAO fitting code is to take linear power spectrum, split it into wiggle and smooth parth and then model non-linearties as sum of smooth correlation function plus zeldovich smoothing with free parameters on the wiggle plart (with sigmas usually marginalised over). For CCL we don't need all this baggage, this is just to say that non-linear xi is not really used in BAO. But fine, as long as the code can also return the linear xi if asked (by configuring cosmology object), I'm fine with that, definitely useful to have several options.

src/ccl_correlation.c
ellison-j: @slosar. Actually, we double checked this and found that this is the average over theta but we want the average over mu, so the second and third coefficients should be zero.  In this case (since we are using the linear approximation) xi(s) is just related to xi(r)  by xi(s) = (1 + 2/3 *beta + beta^2/5) xi(r). We will fix this.

ellison-j: @slosar Actually, we double-checked and these are the coefficients for averaging over theta but we want to average over mu. In this case the second and third coefficients are zero (since we are using the linear Kaiser approximation) and xi(s) is related to xi(r) by xi(s) = (1 + 2/3 beta + (beta^2)/5) xi(r).   We have fixed this now.

The agreement with the benchmark is now much better, see the plots attached.

The maximum difference in s^2 xi(s) is 0.058, which is 0.11% of the BAO peak height.

slosar: Yes, I recognize _these_ numbers!

elisachisari:
@zdu863 @ellison-j I had a quick look and here is some feedback:

- flake8 tests are failing. I think this is causing Travis to fail. Check out https://github.com/LSSTDESC/CCL/blob/master/CONTRIBUTING.md to see how to address this. 
- There are a few conflicts that need to be resolved. The pdf conflicts are expected, you'll have to overwrite those. But CMakeLists.txt, ccl_correlation.i and main.tex need to be inspected for conflicts. If you need help resolving this, please let me know.
- Is the benchmark code public? If so, we should point to it in the wiki (https://github.com/LSSTDESC/CCL/wiki/Benchmarks) and if not, we should try to make it public within CCL as with other codes here: https://github.com/LSSTDESC/CCL/tree/master/tests/benchmark/codes

ellison-j:
Hi @elisachisari. Thanks for the comments. Zilong is working on the flake8 errors and resolving the conflicts but needs help. I updated the wiki to add the benchmark info, and main.tex is now the master version with our RSD description added. Thanks.

PR #414, opened 2018-07-16
Comparison of Physics Constants

matthewkirby:
This is work that was done following a request of a reviewer.

> Section 3.1: GSL constants use CODATA 2006 and JPL/NASA. The ideal would be to always use an updated version of CODATA (2014 current one, and probably a new release next year). In NumComso, for example, we define the physical and astronomical constants using the last CODATA/2014 release + IAU/2015 + NIST/2016. I recommend to change the current approach. You should always use the same consistent set of constants and change it in CLASS, not the other way around.
>>Our current set of constants is documented in the CCL note. I think that as long as the differences are small (as the note suggests), we can stick to CLASS. But we need to take a look at these other sources because they aren’t among our current ones. 

CODATA/2014 and IAU/2015 constants were added to the comparison table in the note.


elisachisari:
Thanks @matthewkirby for making this comparison! I checked the NIST website and they use CODATA 2014, which is why we agree for that row. I clarified in the note that we choose to use CLASS constants in general, and GSL when those aren't available, but this might require further discussion.

elisachisari:
Here are the constants that @matthewkirby compiled: [external.txt](https://github.com/LSSTDESC/CCL/files/2201129/external.txt)
I changed them in `ccl_constants.h`, moving from CLASS+GSL to CODATA+IAU, to understand the impact on our tests. As a result, the `sigmaM` tests fail at 1.6e-4 level (compared to our tolerance of 1e-4). This is the only test that fails. 

I see several alternatives moving forward:
1. Leave the code as is, using CLASS+GSL, and acknowledge uncertainty in the constants at this level in the paper.
2. Incorporate CODATA+IAU and relax the tolerance on `sigmaM` tests. Acknowledge uncertainty in the constants at this level in the paper.
3. Incorporate a keyword in the `ccl_config` which lets the user change the set of constants they are using. There could be one keyword labeled "tests" and we can ensure tests are ran with CLASS+GSL, as in the current master (v0.3.4), by setting up the right configuration in all the `ccl_test*.c` files.

elisachisari:
Latest changes are in line with PR #423. This is ready for review.

doc/0000-ccl_note/main.tex
```diff
@@ -202,7 +202,11 @@ \section{Functionality}
 \subsection{Physical constants}
 
 \label{sec:constants}
-We have performed a comparison of the physical constants used in \ccl and included dependencies and external sources. See Table \ref{tab:constants} for percent differences of the constants between these sources.
+We have performed a comparison of the physical constants used in \ccl and included dependencies and external sources. See Table \ref{tab:constants} for percent differences of the constants between these sources. Notice that NIST adopts the CODATA 2014 values. Our final choice of constants for \ccl mainly relies on CODATA 2014 in as much as possible, except for ${\rm M}_\odot$, where we adopt the IAU 2015 value, and for the conversion between parsec and meters, where we take the PDG 2013 value.
```
philbull: "percent differences" -> "percentage differences"

philbull: A citation to CODATA would be worthwhile here.

elisachisari: Review addressed

PR #420, opened 2018-07-17
Adds support for versioning from git tags and fixes pip install 

### LINKED ISSUE #418: pyccl.__version__ undefined
To check which version of CCL is currently installed, it would be useful to be able to run `print(pyccl.__version__ )` but this is currently undefined.

EiffL:
This small PR adds support for versioning (in the setup.py file and pyccl.__version__) using the `setuptools_scm` package that automatically follows the git tags, no need to manually change the versions in the files anymore, versions will be updated anytime a new release gets made. Should fix #418 
Also fixes a small pip deployment bug which wasnt copying the CLASS files.


EiffL:
I'll go ahead and merge it, I've only modified the manifest.in file, which is not tested by these tests but I've manually tested that everything works with the pip install on the testing instance of the PyPi server: https://test.pypi.org/project/pyccl/

PR #423, opened 2018-07-18
Apply CODATA/IAU constants

elisachisari:
Update CCL constants to most recent literature values.


elisachisari:
Benchmarks have now been updated to meet the change in constants and this PR is ready for review. Travis is currently failing due to an OS X problem that is unrelated to this PR, so I think we should proceed with a review.

Issue #398, opened 2018-06-26
Exact neutrino masses


rmjarvis:
I wrote this when reviewing the current draft of the CCL paper, and it occurred to me that I probably ought to post it here as an issue, rather than just on the Confluence page.

The paper says that you use currently do some Taylor approximation for computing the 3 individual neutrino masses given a total mnu.  This seems to me to be gratuitously inaccurate, when Newton's method will give you the exact masses to numerical precision is just a few iterations.  Here is the code I wrote to do this.  (In python, but easy enough to translate to C if that's where this belongs.)

```
def split_neutrino_masses(mnu, inverted=False):
    """Return the three individual neutrino masses in either the normal hierarchy (default)
    or the inverted hierarchy (with inverted=True).
    
    Returns a tuple (m1, m2, m3)
    
    Note that these values are always returned with m1 < m2 < m2.
    However, in the inverted hierarchy, they are commonly labeled m3 < m1 < m2.
    """
    # delta m^2 from https://arxiv.org/pdf/1505.01891.pdf
    d21 = 7.5e-5
    if inverted:
        d31 = 2.448e-3
        d21 = d31 - d21
    else:
        d31 = 2.458e-3
     
    # Three equations:
    # m1 + m2 + m3 = mnu
    # m2^2 - m1^2 = d21  ->  m2 = (m1^2 + d21)^0.5
    # m3^2 - m1^2 = d31  ->  m3 = (m1^2 + d31)^0.5
    
    # Starting guess
    m1 = 0
    
    # Check that sum is consistent with d21, d31
    m2 = d21**0.5
    m3 = d31**0.5
    s = m1 + m2 + m3
    if mnu < s:
        raise ValueError("mnu may not be less than %f"%s)
    
    while abs(mnu-s) > 1.e-15:
        # Newton step
        dsdm1 = 1 + m1/m2 + m1/m3
        m1 -= (s-mnu)/dsdm1
        m2 = (m1**2 + d21)**0.5
        m3 = (m1**2 + d31)**0.5
        s = m1 + m2 + m3
        
    return m1,m2,m3
```

PR #432, opened 2018-07-27
exact splitting of neutrino masses for inverted and normal hierarchies

c-d-leonard:
This is a fix to issue #398 , a request from @rmjarvis that we do an exact numerical solution to the equations that split the sum of neutrino masses into individual masses in the normal and inverted hierarchies rather than using an approximation scheme. The implemented code is a C adaptation of his python example code in that issue.


src/ccl_core.c
```diff
@@ -386,47 +386,98 @@ ccl_parameters ccl_parameters_create(
...
+		  mnu_in[1] = 0.;
+		  mnu_in[2] = 0.;
+	  } else{
+	  
+	      mnu_in[0] = 0.; // This is a starting guess.
+	  
+	      double sum_check;
+	      // Check that sum is consistent
+	      mnu_in[1] = pow(DELTAM12_sq, 0.5);
+	      mnu_in[2] = pow(DELTAM13_sq_pos, 0.5);
```
rmjarvis: I think `sqrt(x)` should be (slightly) faster than `pow(x,0.5)` here and probably clearer.

src/ccl_core.c
```diff
@@ -386,47 +386,98 @@ ccl_parameters ccl_parameters_create(
...
+              dsdm1 = 1. + mnu_in[0] / mnu_in[1] + mnu_in[0] / mnu_in[2];
+              mnu_in[0] = mnu_in[0] - (sum_check - *mnu) / dsdm1;
+              mnu_in[1] = pow((mnu_in[0]*mnu_in[0] + DELTAM12_sq), 0.5);
+              mnu_in[2] = pow((mnu_in[0]*mnu_in[0] + DELTAM13_sq_pos), 0.5);
+              sum_check = mnu_in[0] + mnu_in[1] + mnu_in[2];
+          }
+	  }
 	  
+	  // Check if the user has provided a sum that is below the physical limit.
 	  if (mnu_in[0]<0 || mnu_in[1]<0 || mnu_in[2]<0){
```
rmjarvis: I think this should be impossible given the above check of `ccl_mnu_sum < sum_check`.  If you think it's possible to hit this branch, then you should make sure you have a unit test that does so.

src/ccl_core.c
```diff
@@ -386,47 +386,98 @@ ccl_parameters ccl_parameters_create(
...
+              mnu_in[1] = pow((mnu_in[0]*mnu_in[0] + DELTAM12_sq), 0.5);
+              mnu_in[2] = pow((mnu_in[0]*mnu_in[0] + DELTAM13_sq_neg), 0.5);
+              sum_check = mnu_in[0] + mnu_in[1] + mnu_in[2];
+          }
+	  
+      }
+      
+	  // Check if the user has provided a sum that is below the physical limit.
+	  if (mnu_in[0]<0 || mnu_in[1]<0 || mnu_in[2]<0){
 	    if (params.sum_nu_masses < 1e-14){
```
rmjarvis: Again, I don't think this is possible either.

src/ccl_core.c
c-d-leonard: Good point, I've changed this to sqrt(x).

src/ccl_core.c
c-d-leonard: You're right, I didn't think to remove this even though it's no longer possible given the check you mention. I've removed it now.

src/ccl_core.c
c-d-leonard: You're right, it's now removed.

c-d-leonard:
Thanks @slosar and @rmjarvis. In response to your question about unit tests Mike: yes, there are already unit tests for getting the three neutrino masses given a sum and a hierarchy definition. These can be found in CCL/tests/ccl_test_params_mnu.c. We test whether the returned masses are consistent with the input mass differences for each hierarchy. We could make the requirement for these to match more stringent, but we use 1e-4 as a default for checks throughout CCL so I don't want to divert from that just for this (it would seem to need to be implemented more broadly if we were going to do that ). 

Let me know if you are happy for us to merge this with these changes.

rmjarvis:
Looks fine except for the build failure, which doesn't seem related to this PR:
```
[ 15%] Swig source
/Users/travis/build/LSSTDESC/CCL/pyccl/ccl_constants.i:17: Error: Unable to find 'gsl/gsl_const_mksa.h'
make[3]: *** [pyccl/cclPYTHON_wrap.stamp] Error 1
make[3]: *** Deleting file `pyccl/cclPYTHON_wrap.stamp'
make[2]: *** [pyccl/CMakeFiles/gen_ccllib_cclPYTHON_wrap.dir/all] Error 2
make[1]: *** [pyccl/CMakeFiles/_ccllib.dir/rule] Error 2
make: *** [_ccllib] Error 2
  File "setup.py", line 65, in <module>
    'Topic :: Scientific/Engineering :: Physics'
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/command/build.py", line 127, in run
    self.run_command(cmd_name)
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/Users/travis/miniconda/envs/test-environment/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 17, in run
    raise Exception("Could not build CCL")
Exception: Could not build CCL
```

c-d-leonard:
Okay, that's weird. I know we've had problems with weird failures in Travis before ... @elisachisari does this look familiar to you? @EiffL any chance this is related to the new installation process?

EiffL:
sigh.... Yes, I was made aware of this yesterday, it seems to be a consequence of an update of the travis OSX image, I'm not entirely sure why, very annoying to debug as I need to wait for the build to start and reach that point to test fixes. I have been looking into it yesterday on the halo_model branch

rmjarvis:
FWIW, I wouldn't let that failure hold up this PR.  Especially if François is working on it on a different branch.

src/ccl_core.c
```diff
@@ -386,47 +386,80 @@ ccl_parameters ccl_parameters_create(
...
+	      mnu_in[1] = sqrt(DELTAM12_sq);
+	      mnu_in[2] = sqrt(DELTAM13_sq_pos);
+	      sum_check = mnu_in[0] + mnu_in[1] + mnu_in[2];
+	      if (ccl_mnu_sum < sum_check){
+		      *status = CCL_ERROR_MNU_UNPHYSICAL;
+          }
+      
+          double dsdm1;
+          // This is the Newton's method
+          while (fabs(*mnu - sum_check) > 1e-15){
```
tilmantroester: Would it make sense to add a counter and exit condition in case the target accuracy doesn't get reached for some reason? E.g., have it less than `ccl_gsl->N_ITERATION`.

rmjarvis: I don't think that is necessary.  This particular function is extremely smooth for all values of sum_check, and there is no regime where the derivative could possibly shoot off into infinity.  As a result, it rarely takes more than 4 iterations to reach machine precision.

rmjarvis: Actually, I guess the one thing that could possibly cause problems is if mnu >> 1.  It would still quickly converge to a relative accuracy of machine precision, but possibly not absolute accuracy.

So if you are concerned about that (not physical, but maybe chains could get there) case, then the easy fix would be to switch the check to compare agains `1e-15 * (*mnu)` to make it a relative accuracy test.

c-d-leonard:
I am going to go ahead and merge this now since the Travis checks are now passing.

PR #437, opened 2018-08-09
Rename K_MIN_DEFAULT to K_MIN

elisachisari:
This PR addresses one of the requests from the internal reviewers to rename the spline parameter K_MIN_DEFAULT to K_MIN. This is just a change of name, no changes in value or where this parameter comes in. C tests pass. The CCL note has been changed accordingly. The paper will be updated in its separate PR.


doc/0000-ccl_note/main.tex
```diff
@@ -511,11 +511,11 @@ \subsubsection{Spline parameters \& the INI file}
 \subsubsection{Extrapolation for the nonlinear power spectrum}
 \label{sec:NLextrapol}
 
-The computation of the power spectrum from {\tt CLASS} can be significantly sped up by extrapolating in the range $k>$~{\tt K$\_$MAX$\_$SPLINE} and $k<$~{\tt K$\_$MIN}. In this section, we describe the implementation of the extrapolation and the accuracy attained. These tests are performed in a flat $\Lambda$CDM cosmology with $\Omega_c=0.25$, $\Omega_b=0.05$, $A_s=2.1\times10^{-9}$, $h=0.7$ and $n_s=0.96$.
+The computation of the power spectrum from {\tt CLASS} can be significantly sped up by extrapolating in the range $k>$~{\tt K$\_$MAX$\_$SPLINE} and $k<$~{\tt K$\_$MIN$\_$SPLINE}. In this section, we describe the implementation of the extrapolation and the accuracy attained. These tests are performed in a flat $\Lambda$CDM cosmology with $\Omega_c=0.25$, $\Omega_b=0.05$, $A_s=2.1\times10^{-9}$, $h=0.7$ and $n_s=0.96$.
```
damonge: Is this K_MIN_DEFAULT or K_MIN_SPLINE? Or are we referring here to a different number?

src/ccl_power.c
```diff
@@ -1432,7 +1432,7 @@ static double ccl_power_extrapol_highk(ccl_cosmology * cosmo, double k, double a
     
 }
 
-/*------ ROUTINE: ccl_power_extrapol_hxighk ----- 
+/*------ ROUTINE: ccl_power_extrapol_lowk ----- 
 INPUT: ccl_cosmology * cosmo, a, k [1/Mpc]
 TASK: extrapolate power spectrum at low k
 */
```
rmjarvis: Maybe worth changing the `kmin` argument to this function to `kmin_spline`.  I believe this parameter is what you now call K_MIN_SPLINE in the paper.  

It took me a little bit of tracking to follow how your explanations in the paper translated into code.  But my conclusion (correct me if I'm wrong) is that this function's `kmin` parameter is really the value that you refer to as K_MIN_SPLINE now.  It's value is (as you describe in the paper) different for different use cases.  E.g. CLASS vs EMU vs linear will have different lower limits of the spline arrays.  Then these lower limits get saved as either `cosmo->data.k_min_lin` or `cosmo->data.k_min_nl`, which in turn get passed to this function as the kmin argument.  

I think this connection could be made more explicit by calling this `kmin_spline` here and where you set the different `k_min_*` values just add an in-line comment saying that this is the K_MIN_SPLINE value for this particular choice of power spectrum being splined.

doc/0000-ccl_note/main.tex
```diff
@@ -545,7 +547,7 @@ \subsubsection{Extrapolation for the linear power spectrum}
 
 With the implementation described in the previous section, the power spectrum splines are initialized up to {\tt K$\_$MAX$\_$SPLINE}. This is also true for the linear matter power spectrum, which is used within \ccl in particular to obtain $\sigma_8$ (see Eq. \ref{eq:sigR}). We have tested here how the procedure described in the previous section affects the convergence of the linear matter power spectrum. The result is shown in Figure \ref{fig:NLextrapol}. For some applications that use the linear power spectrum, the user might need to increase the value of {\tt K$\_$MAX$\_$SPLINE}.
 
-As in the previous section, the power spectrum at small wavenumber is extrapolated using a power-law. This extrapolation is performed below a fiducial value of {\tt K\_MIN\_DEFAULT}$=5\times 10^{-5}$.
+As in the previous section, the power spectrum at small wavenumber is extrapolated using a power-law. This extrapolation is performed below a fiducial value of {\tt K\_MIN\_SPLINE}={\tt K\_MIN=$=5\times 10^{-5}$ Mpc$^{-1}$.
```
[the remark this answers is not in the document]
elisachisari: This was outdated for the linear matter power spectrum. In practice we do the same as for the nonlinear one. I've now made that change.

src/ccl_power.c
elisachisari: Changed `kmin->kmin_spline` in `extrapol_lowk` and `kmax->kmax_spline` in `extrapol_highk`. The comment has been added.

doc/0000-ccl_note/main.tex
elisachisari: We discussed this in the telecon - these are different parameters. One sets the minimum K for integration, the other, for splining. K_MIN_SPLINE is not directly available to the user.

PR #440, opened 2018-08-13
installment 5 code review - mass func changes

villarrealas:
In progress changes for the Code Review of ccl_massfunc related files.

Currently addressed:
- Better specification about halo mass function where abbreviations are utilized.
- Moving compute_hmfparameters into static and removing it from the direct API.
- Fixing poor comment positioning/spacing
- Addressing memory leaks in spline calculation with considerable workflow change in ccl_compute_sigma
- Fixing kludgy math in compute_sigma calculation.

Not addressed yet:
- Python changes.


villarrealas:
I think we're actually going to open it up to all of installment 5, just fyi. Is there a way to remove the existing approval until the rest of Installment 5 gets in here?

rmjarvis:
OK.  Just ping me again for a review when that's ready.

elisachisari:
@villarrealas If you think this completes the requested changes for ccl_massfunc (incuding the python) and it is isolated from other items in installment 5, then I think you could go ahead and have a review and merge. It would be good to write a short statement in Mike's code review PR referencing what has been addressed, so we keep track of things.

villarrealas:
@elisachisari We had talked last CCL meeting about having separate PRs for each installment in order to minimize potential overlap. This would still make sense in this case, as the reviewed PR here does not cover any of the Python.

Alternatively, I could make changes w.r.t. Python as well, but regardless of the choice we should seek to be consistent about this throughout.

villarrealas:
(aside, I learned that I cannot rename this branch and forgot to mention this. If somebody wants to put installment 5 material into the `massfunc_codereview` branch, that would be fine. Assuming we don't feel that it is better to just work in necessary Python changes and submit, though I think now that other people are working on installment 1, that may not be advised.

villarrealas:
@elisachisari This is ready for review now; coverage has dropped due to integrating Master and is unrelated to the PR.

villarrealas:
No changes to anything relating to the code review, aside from one minor change to error handling (dedicated function vs strcpy uses now) that I am about to push. There is some new code related to the halo model sections that was not previously reviewed, I believe.

PR #450, opened 2018-09-12
Halo profiles 

KuanWang-Astro:


tmcclintock:
I'll be happy to review this once it's ready. Just let me know when!

include/ccl_haloprofile.h
```diff
@@ -0,0 +1,54 @@
+/** @file */
+#ifdef __cplusplus
+extern "C" {
```
rainwoodman: It would be nice if you can switch to the new include guards and use CCL_BEGIN_DECLS / CCL_END_DECLS following the example in 

https://github.com/LSSTDESC/CCL/blob/master/include/ccl_correlation.h

include/ccl_haloprofile.h
```diff
@@ -0,0 +1,54 @@
...
+   * @param a: scale factor normalised to a=1 today
+   * @param r: radius at which to calculate output
+   * @param status: Status flag: 0 if there are no errors, non-zero otherwise
+   * @return rho_r: mass density at given radius for an Einasto halo profile, in units of Msun/Mpc^{3}
+   */
+  double ccl_halo_profile_einasto(ccl_cosmology *cosmo, double c, double halomass, double massdef_delta_m, double a, double r, int *status);
+
+
+
+#ifdef __cplusplus
```
rainwoodman: this becomes CCL_END_DECLS

src/ccl_haloprofile.c
```diff
@@ -0,0 +1,134 @@
+#include "ccl_utils.h"
+#include "math.h"
+#include "stdio.h"
```
rainwoodman: Use include <> for stdlib headers and gsl.
Also, ccl.h already included most of the other ccl_headers. Follow the example in ccl_halomod.c.

KuanWang-Astro: Thank you! I'll look into it and make these changes in two or three days.

src/ccl_haloprofile.c
```diff
@@ -0,0 +1,279 @@
...
+#include <gsl/gsl_sf_expint.h>
+#include "ccl.h"
+#include "ccl_correlation.h"
+#include "ccl_massfunc.h"
+#include "ccl_halomod.h"
+#include "ccl_haloprofile.h"
+
+
+double r_delta(ccl_cosmology *cosmo, double halomass, double a, double odelta, int *status); //in Mpc comving. halomass in Msun, odelta integer.
+double ccl_sigmaM(ccl_cosmology *cosmo, double halomass, double a, int *status); //returns sigma from the sigmaM interpolation.
```
beckermr: This extra prototype should not be needed.

tests/ccl_test_haloprofile.c
```diff
@@ -0,0 +1,166 @@
...
+#include "ccl_haloprofile.h"
+#include "ccl_massfunc.h"
+#include "ccl_halomod.h"
+#include "ctest.h"
+#include <stdio.h>
+#include <math.h>
+#include <string.h>
+
+// Relative error tolerance in the halo profile density
+#define HALOPROFILE_TOLERANCE 1E-3
```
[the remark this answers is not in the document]
KuanWang-Astro: There is a difference in the critical density values in CCL and colossus which I use as benchmark, under the same cosmology. This enters through the calculation of halo radius and causes some discrepancy between my values and benchmark values, so we won't be able to get machine precision.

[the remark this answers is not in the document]
beckermr: Did you turn off radiation, setting Omega_g = 0?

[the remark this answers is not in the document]
tmcclintock: Yes. It is a units issue in colossus that we pointed out in [this issue](https://bitbucket.org/bdiemer/colossus/issues/12/inconsistent-unit-conversions). Machine precision will have to wait until colossus is patched.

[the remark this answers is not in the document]
beckermr: Yeah this issue is not large enough to cause what is going on here, right? The error there is 0.01% and the tolerance listed here is 1e-3.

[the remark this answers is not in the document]
tmcclintock: That was a choice made before the cause was identified. The tolerance could be tightened up to the difference in the units between the two codes.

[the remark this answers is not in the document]
beckermr: This makes no sense. We know how much the constant differs. We can test the code at much higher precision.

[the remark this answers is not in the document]
tmcclintock: Yea, didn't think of that. Good idea.

KuanWang-Astro:
The python tests are in there now but my results do not converge with benchmark data (from colossus) even though the relative tolerance is only set to 1e-3. The same tests pass in C. I believe this is due to 1. the different mean mass densities adopted in the two packages that we discussed earlier, and 2. some numerical issue with python, because no real calculation is done in the python wrapper. I haven't been able to locate the exact problem.

beckermr:
The probability that python has a numerical issue at 1-3 is very small. I'd double check the code.

KuanWang-Astro:
If we can find packages other than colossus for benchmark that would be great.

pennalima:
Hi Kuan,
You can compare it with NumCosmo.

KuanWang-Astro:
> Hi Kuan,
> You can compare it with NumCosmo.

I'll do that asap, thanks for reminding me!

src/ccl_haloprofile.c
```diff
@@ -0,0 +1,329 @@
...
+        }
+        else {
+            sigma_r[i] = 2.*rs*rhos*(1.-2.*atan(sqrt(fabs((1.-x)/(1.+x))))/sqrt(fabs(1.-x*x)))/(x*x-1.);
+        }
+    }
+    return;
+
+}
+
+//maths helper function for Einasto profile
```
tmcclintock: This comment is incorrect. The function below is for NFW profiles, not Einasto.

src/ccl_haloprofile.c
```diff
@@ -0,0 +1,329 @@
...
+    //haloradius: halo radius for mass definition
+    //rs: scale radius
+    double haloradius, rs;
+
+    haloradius = r_delta(cosmo, halomass, a, massdef_delta_m, status);
+    rs = haloradius/c;
+
+
+    //nu: peak height, https://arxiv.org/pdf/1401.1216.pdf eqn1
+    //alpha: Einasto parameter, https://arxiv.org/pdf/1401.1216.pdf eqn5
```
tmcclintock: I think the python-level comment for this function should contain the fact that alpha is set by the mass. This was a choice that you made, rather than letting the user pass in alpha.

src/ccl_haloprofile.c
```diff
@@ -0,0 +1,329 @@
...
+
+static double integrand_hernquist(double r, void *params){
+
+    Int_Hernquist_Par *p = (Int_Hernquist_Par *)params;
+    double rs = p->rs;
+
+    return 4.*M_PI*r*r/((r/rs)*(1.+r/rs)*(1.+r/rs)*(1.+r/rs));
+}
+
+//integrate einasto profile to get normalization of rhos
```
tmcclintock: comment is wrong

beckermr:
@tmcclintock Can you please look through this?

PR #451, opened 2018-09-13
Code Review: Installment 4

elisachisari:
This PR addresses the suggestions made in the code review (installment 4). 

Notice that the precision with which we pass parameters to CLASS has changed (it has been increased) and this we might need to run VARRIC again (@philbull) to see if this makes a difference.


elisachisari:
(The PR is not yet finished, I still need to make some changes on how the tests are run.)

beckermr:
You'll need to resolve the merge conflict but otherwise the changes look fine.

src/ccl_power.c
```diff
@@ -86,6 +86,8 @@ static void ccl_free_class_structs(ccl_cosmology *cosmo,
       return;
     }
   }
+
+  return;
```
rmjarvis: This is unnecessary.  (Although not a problem.)

PR #452, opened 2018-09-13
Add methods to dump and read parameters to/from YAML

joezuntz:
It would be great to have a nice robust serialization code for the Parameters object.

This is not that code - the parameters have a fixed order and changing it will break things.   But it does work to dump and load a ccl_parameter object to and from a fixed YAML format, which is human readable, and will work as a placeholder until something more useful is done.


joezuntz:
Apparently something I did only works on python 3 (or possibly with a different version of swig) - I'll check it out.

elisachisari:
Also, could you check the decrease in coverage and the conflict with `ccl_core.h`?

joezuntz:
I've added a new ipython notebook demonstrating this (Parameters.read_yaml is a class method; it doesn't need an actual instance of Parameters to be run - you use it like `Parameters.read_yaml(filename)`).

I've merged master in and fixed the issue.

I think there was only one line from this PR that wasn't tested, which was one of the error pathways.  I've added a test for that now, so let's see if that ups the coverage.  I think the other decreases may have been because I hadn't merged in master yet.

elisachisari:
Looks like coverage is still down. Could you take another look?

joezuntz:
I don't understand what versions coverall is comparing.  The changes marked in red here, for example:
https://coveralls.io/builds/19335596/source?filename=pyccl/core.py
aren't ones that are made in this PR.

joezuntz:
Can we just ignore the coverage thing and merge, assuming you're happy with the other things you mentioned?  It clearly isn't working here.

elisachisari:
Yes, I was waiting for the other tests to pass. Happy to merge this.

PR #459, opened 2018-09-19
Expose CCL functions as methods of the Cosmology object

philbull:
Mike J suggested that it would be more Pythonic to expose the CCL interface as methods on the `Cosmology` object rather than module-level functions. This PR implements a simple way of attaching module-level functions as methods to `Cosmology` objects.

(N.B. At the moment, only functions available in the `pyccl` root namespace will be added.)

### DIFF
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -1,10 +1,12 @@
 """The core functionality of ccl, including the core data types. This includes the cosmology and parameters objects used to instantiate a model from which one can compute a set of theoretical predictions.
 
 """
+import pyccl
 from pyccl import ccllib as lib
 import numpy as np
 from warnings import warn
 from pyccl.pyutils import check
+import sys, inspect, types
 
 # Configuration types
 transfer_function_types = {
@@ -148,7 +150,8 @@ class Parameters(object):
         """
         Creates a set of cosmological parameters.
 
-        .. note:: Although some arguments default to `None`, they will raise a ValueError inside this function if not specified, so they are not optional.
+        .. note:: Although some arguments default to `None`, they will raise a 
+        ValueError inside this function if not specified, so they are not optional.
         
         Args:
             Omega_c (float): Cold dark matter density fraction.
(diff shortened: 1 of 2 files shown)


joezuntz:
While I like the automation using inspect, in general I've found that these kind of magic functions tend to confuse people looking at the code more than they help.  Unless there's a really large number of these functions I'd suggest just hard-coding them as methods.

rmjarvis:
I actually don't like the inspect style here.  For several reasons:
1. As Joe said, it is potentially obfuscating for people who are scanning the code for a single function they want to look at more closely.
2. It doesn't attach the functions as methods of the class, just individual instances.  So `help(Cosmology.growth_factor)` doesn't work.
3. There is an unnecessary (and potentially non-trivial) runtime cost every time you make a new instance of the Cosmology class, which is something people should probably expect to not do any significant work, 
4. It keeps the C-style object-oriented code as the primary documented usage, rather than embracing the more pythonic method style.  (I.e. docs still treat `cosmo` as an argument, not as `self`.)

I'd much prefer each method be documented as though it were a method from the start, and they are just defined in other files as an organizational convenience.  For instance:
```
def growth_factor(self, a):  # self is a Cosmology instance here.
    """Growth factor.
    
    Args:
        a (float or array_like): Scale factor(s), normalized to 1 today.

    Returns:
        float or array_like: Growth factor.

    """
    return _vectorize_fn(lib.growth_factor,
                         lib.growth_factor_vec, self, a)

# Add this as a method of the Cosmology class
Cosmology.growth_factor = growth_factor
```

rmjarvis:
Well, I didn't actually install this version to test it.  But I don't see how you get any new methods in the class.  Just the instances.
I.e. I think `help(cosmo.growth_factor)` should work.  And `help(Cosmology().growth_factor)` should work.  But if I understand the code right, `help(Cosmology.growth_factor)` should give an AttributeError.

Here is my toy test that I think does essentially the same thing your inspect code does.
```
>>> class A(): pass
... 
>>> a = A()
>>> def foo(a): pass
... 
>>> import types
>>> setattr(a, 'foo', types.MethodType(foo, a))
>>> help(a.foo)  # works

>>> help(A.foo)
  File "<stdin>", line 1, in <module>
AttributeError: type object 'A' has no attribute 'foo'
```
Am I missing something?

philbull:
No, you're not missing anything -- this attaches methods at instantiation of the `Cosmology` object, so `help(Cosmology.some_function)` won't work (but the other ways do, sorry for the confusion).

philbull:
Deprecating this, as I believe an alternative solution has been proposed...

PR #464, opened 2018-09-28
ENH address CR installment 1

beckermr:
This PR addresses CR installment 1.


pyccl/background.py
```diff
@@ -11,174 +25,200 @@
...
 
+
 def comoving_angular_distance(cosmo, a):
     """Comoving angular distance.
 
-    .. note:: this quantity is otherwise known as the transverse comoving distance, and is NOT angular diameter distance or angular separation
-    
+    .. note:: This quantity is otherwise known as the transverse
+              comoving distance, and is NOT angular diameter
+              distance or angular separation. The comvoing angular distance
```
elisachisari: small typo here

.travis/install.sh
```diff
@@ -32,10 +32,10 @@ if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
     esac;
 
     source activate test-environment
-    pip install numpy nose coveralls
+    pip install numpy nose coveralls flake8
```
elisachisari: Will flake8 only run on travis? What about if we are doing development on our own systems? Do we have to run it manually?

pyccl/constants.py
```diff
@@ -1,20 +1,11 @@
...
-  DNDZ_NC, DNDZ_WL_CONS, DNDZ_WL_FID, DNDZ_WL_OPT, \
-  EPS_SCALEFAC_GROWTH, \
-  GNEWT, \
-  K_PIVOT, \
-  MPC_TO_METER, PC_TO_METER, RHO_CRITICAL, SOLAR_MASS, \
-  SplPar, SplPar_swigregister, \
-  Z_MAX_SOURCES, Z_MIN_SOURCES, \
-  CCL_CORR_FFTLOG, CCL_CORR_BESSEL, CCL_CORR_LGNDRE, CCL_CORR_GG, CCL_CORR_GL, CCL_CORR_LP, CCL_CORR_LM
+"""This file exposes constants present in CCL."""
+from .ccllib import (  # flake8: noqa
```
elisachisari: Why have some imports gone away here?

pyccl/core.py
```diff
@@ -1,10 +1,66 @@
...
+    - 'sum_equal': assume equal masses when converting the total mass to
+      individual masses
+
+emulator_neutrinos options
+  This parameter specifies how to handle inconsistencies in the treatment of
+  neutrinos between the Cosmic Emu (equal masses) and other models.
+    - 'strict': fail unless things are absolutely consistent
+    - 'equalize': redistribute the total mass equaly before using the Cosmic
+      Emu. This option may result in slight internal inconsistencies in the
+      physical model assumed for neutrinos.
```
elisachisari: This is great, thanks for adding these docs!

pyccl/core.py
```diff
@@ -276,252 +296,219 @@ def __getitem__(self, key):
...
-    
+
     def __setitem__(self, key, val):
         """Set parameter values by name.
 
         """
         raise NotImplementedError("Parameters objects are immutable; create a "
                                   "new Parameters() instance instead.")
-        """
-        try:
```
elisachisari: Where has this gone? Or was it unnecessary?

pyccl/core.py
```diff
@@ -276,252 +296,219 @@ def __getitem__(self, key):
...
                   'T_CMB', 'Omega_g', 'z_star', 'has_mgrowth']
-        
+
         # Get values of parameters
         vals = []
         for p in params:
             try:
                 v = getattr(self.parameters, p)
-            except:
+            except Exception:
```
elisachisari: What is the difference here between doing except and except Exception?

pyccl/core.py
```diff
@@ -276,252 +296,219 @@ def __getitem__(self, key):
...
-                raise KeyError( "'%s' is not a valid matter_power_spectrum "
-                                  "type. Available options are: %s" \
-                                 % (matter_power_spectrum, 
-                                    matter_power_spectrum_types.keys()) )
-            if baryons_power_spectrum not in baryons_power_spectrum_types.keys():
-                raise KeyError( "'%s' is not a valid baryons_power_spectrum "
-                                  "type. Available options are: %s" \
-                                 % (baryons_power_spectrum, 
-                                    baryons_power_spectrum_types.keys()) )
+                raise ValueError(
```
elisachisari: Why was this made a ValueError instead of a KeyError? Same a few lines below

pyccl/core.py
```diff
@@ -276,252 +296,219 @@ def __getitem__(self, key):
...
+                    "'%s' is not a valid matter_power_spectrum "
+                    "type. Available options are: %s"
+                    % (matter_power_spectrum,
+                       matter_power_spectrum_types.keys()))
+            if (baryons_power_spectrum not in
+                    baryons_power_spectrum_types.keys()):
+                raise ValueError(
```
elisachisari: Same as above

elisachisari:
There is also some decrease in coverage in pyutils.py that we should try to fix.

.travis/install.sh
beckermr: Yes. I run it on the fly in my editor.

pyccl/constants.py
beckermr: Mike suggested four to remove in CR. They are not used elsewhere.

pyccl/core.py
beckermr: It was commented code.

pyccl/core.py
beckermr: Mike explained this in CR. Basically, this allows some exceptions to no be caught in some situations which is better. I forget the exact ones. This is def the right (tm) way to do things in python though.

pyccl/core.py
beckermr: See mike's CR. ValueError is more pythonic. KeyErrors are reserved for dicts.

pyccl/core.py
beckermr: See above.

beckermr:
The decrease coverage is because I moved a few routines to core.py so we don't need to worry about it.

beckermr:
A few notes on procedure. 

1. We should be merging PRs faster. If you want to add docs, we can just open another PR after this one. There is no reason to force some changes to wait.
2. I think your review is more than thorough enough. I would rather us stick to one reviewer per PR to increase our dev speed then use yet more of our limited time bringing other reviewers into the mix.

pyccl/core.py
rmjarvis: In particular, `KeyboardInterrupt`.  `except:` will catch this, which is almost certainly not what you want.  `except Exception:` does not, since `KeyboardInterrupt` derived directly from `BaseException`, not `Exception`.

pyccl/correlation.py
```diff
@@ -19,20 +24,31 @@
...
+                                     function (in degrees).
+        corr_type (string): Type of correlation function. Choices:
+                            'GG' (galaxy-galaxy), 'GL' (galaxy-shear),
+                            'L+' (shear-shear, xi+),
+                            'L-' (shear-shear, xi-).
+        method (string, optional): Method to compute the correlation function.
+                                   Choices: 'Bessel' (direct integration over
+                                   Bessel function), 'FFTLog' (fast
+                                   integration with FFTLog), 'Legendre' (
+                                   brute-force sum over Legendre polynomials).
```
rmjarvis: This would read better with one choice per line starting with the method name at the front, rather than wrapping.
Likewise for corr_type, it would read better to have GG and GL each on their own line.

beckermr: Yes makes sense. I am going to merge and move this into an issue to be addressed later with other doc fixes.

Issue #466, opened 2018-09-28
pragma once is not supported by Cray C compiler


rainwoodman:
https://en.wikipedia.org/wiki/Pragma_once#cite_note-9 

I strongly suggest we get rid of these. I can sort this out as part of the include / cross object dependency clean up.

beckermr:
Looks like we just need the standard include guard. What do you mean by cross-object dependency clean up?

rainwoodman:
I meant some object files refer ro symbols and types in other obj files.
This information is currently descibed in both the header files and the c
files, and the description seemed to be incomplete hence the need to
multiple include a file.

I filed a PR to linearize most of this dependency in ccl.h. The include
sections look leaner after this.

rainwoodman:
Fxied in #467

PR #467, opened 2018-09-28
ENH: header include clean ups.

### LINKED ISSUE #463: include order in the C code
It appears we are including CCL headers ahead of ones in the stdlib in some of the C source. We should probably not be doing this (though most preprocessors will tell you if you are stomping on something.)

rainwoodman:
This PR addresses #463 

I would like to sort out #466 in this eventually.

Closes #463 

^ if you do this, github closes the issue automagically when the PR is merged @rainwoodman


beckermr:
Looks like the build failed. Happy to review once the tests pass!

rainwoodman:
My bad. Forgot to test the python extension locally. Easy fix.

examples/ccl_sample_pkemu.c
```diff
@@ -1,7 +1,7 @@
-#include "ccl.h"
 #include <math.h>
 #include <stdio.h>
-#include "ccl_params.h"
+#include <ccl_params.h>
+#include <ccl.h>
```
beckermr: Do we need to include ccl first here?

examples/ccl_sample_power.c
```diff
@@ -1,7 +1,7 @@
-#include "ccl.h"
 #include <math.h>
 #include <stdio.h>
-#include "ccl_params.h"
+#include <ccl_params.h>
+#include <ccl.h>
```
beckermr: Ditto on the order.

include/ccl.h
```diff
@@ -20,21 +20,17 @@ extern "C" {
...
-#include "ccl_config.h"
 #include "ccl_correlation.h"
 #include "ccl_massfunc.h"
 #include "ccl_neutrinos.h"
-  
-#include "ccl_placeholder.h"
 
-#ifdef __cplusplus
-}
 #endif
```
beckermr: Some people add `/*  __CCL_H_INCLUDED__ */` after the `#endif` to help make the code more readable. I am 50/50 on it but it can be nice.

include/ccl.h
```diff
@@ -20,21 +20,17 @@ extern "C" {
...
 #include "ccl_background.h"
-#include "ccl_config.h"
 #include "ccl_correlation.h"
 #include "ccl_massfunc.h"
 #include "ccl_neutrinos.h"
-  
-#include "ccl_placeholder.h"
 
-#ifdef __cplusplus
-}
```
beckermr: Does this file need the C++ wrapper macros?

include/ccl_defs.h
```diff
@@ -0,0 +1,21 @@
...
+ * */
+
+
+/* define marks for
+ * begin and end of declaration sections of header files for C++ compatibility */
+
+#ifdef __cplusplus
+
+#define CCL_BEGIN_DECLS extern "C" {
+#define CCL_END_DECLS }
```
beckermr: These macro names are confusing to me and generic. Maybe something that indicates they are for extern/C++ stuff?

rainwoodman: I learned of this trick / name from GTK:

https://code.woboq.org/gtk/include/errno.h.html#31

include/ccl.h
rainwoodman: It included ccl_defs.h that defined the macro.

examples/ccl_sample_power.c
rainwoodman: I left two files in different order to show that the order wasn't important. Such that when one day it becomes important one of the case will fail.

include/ccl.h
rainwoodman: I agree usually conditional compilations the matched end is good. In this case though, an important guard protected file must end with an endif, so it has to be the first ifdef at the top of file; there is no confusion. Plus it is less likely to miss change the name when the only two occurrences are right next to each other.

include/ccl_defs.h
rainwoodman: cdefs.h defines similar macros but I *think* the file is not "portable".

examples/ccl_sample_power.c
beckermr: Hmmmm. We should prob just specify a single order and stick with it. I don't think anyone will remember why this was done. 

Also, why are there two header files? Can't we just have a single ccl.h?

include/ccl.h
beckermr: I mean does the whole file need to have the starting and ending `CCL_*_DECLS` things?

include/ccl_defs.h
beckermr: OK.

examples/ccl_sample_power.c
rainwoodman: Either would be an API change How much are people concerned about this yet? I don't think this PR shall introduce API changes. I can switch to the same ordering in the examples for now.

include/ccl.h
rainwoodman: ccl.h doesn't declare anything thus doesn't need the pair.

I've added a dummy section in the file to clarify this.

examples/ccl_sample_power.c
beckermr: My understanding is that api changes are still ok but won’t be post v1

rainwoodman:
Then I would like to submit after this a more controversial PR to change
this to a single file include; and also prefix ccl/ to the header files.

What's the time line of v1?

beckermr:
IDK the timeline for v1. @elisachisari is the main dev here. I am just a hired hand. 

Feel free to make more PRs! The help is appreciated!

beckermr:
Also it looks like the tests are failing again. We really need to setup caching so they run faster. Recompiling all of the source for CLASS, GSL and FFTW each time is not really needed.

rainwoodman:
Thanks!

PR #481, opened 2018-10-03
fix for openHPC GSL rpm

jose-d:
there in ohpc GSL rpm is no GSL_ROOT_DIR but GSL_DIR. let's use it when it's available.
(tested on ohpc-release-1.3-1, centos7.5)

### DIFF
--- a/cmake/Modules/BuildGSL.cmake
+++ b/cmake/Modules/BuildGSL.cmake
@@ -3,6 +3,10 @@ include(ExternalProject)
 set(GSLVersion 2.4)
 set(GSLMD5 dba736f15404807834dc1c7b93e83b92)
 
+if (DEFINED ENV{GSL_DIR} AND NOT(DEFINED ENV{GSL_ROOT_DIR}))    #there in ohpc GSL rpm is no GSL_ROOT_DIR but GSL_DIR. let's use it when it's available
+  set(ENV{GSL_ROOT_DIR} $ENV{GSL_DIR})
+endif()
+
 find_package(GSL 2.1)
 
 # If GSL is not installed, lets go ahead and compile it


elisachisari:
hi @jose-d! Thanks for opening this PR. It would be great if you could join the CCL Slack channel #desc-ccl to introduce yourself and to interact with the team. Importantly, that's where we make all the telecon announcements.

beckermr:
I am confused by this PR. Can you say a bit more? I would have assumed that all we needed for GSL was to have the headers and libs on the right paths. What is this extra DIR used for?

jose-d:
@beckermr, TL/DR: we run  HPC system with openHPC software stack, providing us some well-known libraries like GSL. It seems that there is no clear view on topic what the "headers and libs on the right paths" means. In openHPC, by default, following gsl-related paths are exported:

```
[root@n21 ~]# env | grep -i gsl
...
LD_LIBRARY_PATH=/opt/ohpc/pub/libs/gnu7/gsl/2.4/lib:/opt/ohpc/pub/compiler/gcc/7.3.0/lib64
GSL_INC=/opt/ohpc/pub/libs/gnu7/gsl/2.4/include
PATH=/opt/ohpc/pub/libs/gnu7/gsl/2.4/bin:/opt/ohpc/pub/compiler/gcc/7.3.0/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/opt/puppetlabs/bin:/usr/bin/flashupdt:/usr/bin/syscfg:/usr/bin/sysinfo:/root/bin
GSL_LIB=/opt/ohpc/pub/libs/gnu7/gsl/2.4/lib
GSL_DIR=/opt/ohpc/pub/libs/gnu7/gsl/2.4
INCLUDE=/opt/ohpc/pub/libs/gnu7/gsl/2.4/include:/opt/ohpc/pub/compiler/gcc/7.3.0/include
...
[root@n21 ~]#
```

anyways, build of CCL will fail, because it uses cmake module findGSL, https://cmake.org/cmake/help/v3.2/module/FindGSL.html,  (src: https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindGSL.cmake) so it needs hint "GSL_ROOT_DIR" or it will failback to the standard OS directories.

So the question is, where to fix it. I am actually now not sure, if this should be fixed in cmake modules, in this project, or if we should just patch our site environment modules and workaround it for our users.

Any comments welcomed.

beckermr:
I see. So the GSL paths are in the standard `INCLUDE` and `LD_LIBRARY_PATH` you listed above. Do you know why the compiler is not picking up the files it needs from there?

I think generally it is best to keep the cmake build as system agnostic as possible. 

Here is another suggestion. We could add an option to specify the prefix to use for GSL as part of the cmake build options.

Finally, it seems like you could define `GSL_ROOT_DIR=GSL_DIR` locally before building.

rainwoodman:
I think adding an entry in CCL README for this simple workaround may be better than fixing FindGSL -- the purpose of `GSL_ROOT_DIR` is exactly for use cases like this.

jose-d:
> I think adding an entry in CCL README for this simple workaround may be better than modifying FindGSL -- the purpose of `GSL_ROOT_DIR` is exactly for use cases like this.
Agree. I'm closing this one and will check if I can create PR for the readme.

jose-d:
@rainwoodman : https://github.com/LSSTDESC/CCL/compare/master...jose-d:patch-1 - could be?

Issue #477, opened 2018-09-29
faster test suite


beckermr:
The test suite is pretty slow right now. @rmjarvis suggested a hack from GalSim [here](https://github.com/LSSTDESC/CCL/pull/433#discussion_r206374547) and [here](https://github.com/LSSTDESC/CCL/pull/433#discussion_r206374656).

tmcclintock:
For reference, the tests ran with `build/check_ccl` take much less time than `python setup.py test`. Not sure why though.

PR #484, opened 2018-10-04
Set up CI with Azure Pipelines

jkuruvilla:
This PR deals with setting up Azure Pipelines as the CI engine. Helps out with #477 

Pros: 
- Allows for parallel testing (upto 10 builds)
- Allows for better log monitoring 

Sample log for the build can be seen [here](https://dev.azure.com/jkuruvilla-ci/jkuruvilla-ci/_build/results?buildId=93&view=logs)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,79 @@
+jobs:
+- job: Linux
+  pool:
+    vmImage: ubuntu-16.04
+  
+  strategy:
+    matrix: 
+      Python27:
+        python.version: '2.7'
+      Python36:
+        python.version: '3.6'
+      Python37:
+        python.version: '3.7'  
+    maxParallel: 3
+    
+  steps:
+  - task: CondaEnvironment@1
+    inputs:
+      packageSpecs: python=$(python.version) numpy nose coveralls swig flake8
+      updateConda: True
+  - script: |
+      gcc --version
+      cmake --version
+      python --version
+    displayName: 'Checking installed versions'
+  - script: |
+      sudo apt-get install bison -y 
+    displayName: 'Install dependencies'
+  - script: flake8 pyccl
+    displayName: 'Flake8'
+  - script: python setup.py build
+    displayName: 'Install pyCCL'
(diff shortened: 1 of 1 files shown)


elisachisari:
hi @jkuruvilla! Thanks for opening this PR. It would be great if you could join the CCL Slack channel #desc-ccl to introduce yourself and to interact with the team. Importantly, that's where we make all the telecon announcements.

jkuruvilla:
@elisachisari I would be happy to do that. However since I am not a member of the LSSTDESC, I don't have access to the slack server. Let me know if you could invite me (if possible). You can see a brief introduction about myself [here](https://jkuruvilla.github.io/)

beckermr:
Thank you for the contribution! I really like azure CI (which is weird to me since microsoft but hey w/e).

One question. Travis runs the builds in parallel too. I am confused about the timing numbers you put above.

jkuruvilla:
> I really like azure CI (which is weird to me since microsoft but hey w/e).

Same for me! 

> One question. Travis runs the builds in parallel too. I am confused about the timing numbers you put above.

@beckermr  I think I may have messed up the timings. I will edit it out and can change the title too. But does travis build push and the pr build simultaneously?

beckermr:
I think for travis it depends on how much capacity they have? I don't know. 

I will say though that most of the CI time is compiling code and then some v slow tests.

We should probs switch to azure either way.

jkuruvilla:
> I will say though that most of the CI time is compiling code and then some v slow tests.
 
For example, here I installed swig through conda and seemed to have shaved off some time while building. I can try the same for gsl and fftw3, and see how it goes. Also do you know why macOS seems to take ~17 minutes for the nosetests while linux takes ~10 minutes?

beckermr:
IDK why the mac ones are slow, but my guess are two things:

1. The actual mac hardware might be older.
2. The build for osx doesn't use openmp right now.

Pulling some of the deps from conda is what I do locally. Def a good idea!

beckermr:
Hmmmm. Travis appears to be running all of the builds at once right now. I think given that, we might as well punt on azure for a bit. Is it ok if I close this and then we can revisit in the future?

A PR pulling more deps from conda would be good for the travis build.

jkuruvilla:
Okay sure thing, I can close it :+1:

PR #487, opened 2018-10-05
ENH raise an error if a C ccl cosmology object is passed

### LINKED ISSUE #475: make sure python layer only access the `Cosmology` object
@rmjarvis pointed out that the python layer should only access the python object, not the C counterpart.

beckermr:
This PR changes the code to raise an error if the C cosmology object is being used. 

Closes #475

### DIFF
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -656,7 +656,7 @@ def check(status, cosmo=None):
 
     # Get status message from Cosmology object, if there is one
     if cosmo is not None:
-        msg = _cosmology_obj(cosmo).status_message
+        msg = cosmo.cosmo.status_message
     else:
         msg = ""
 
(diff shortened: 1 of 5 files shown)


pyccl/core.py
```diff
@@ -670,17 +670,16 @@ def check(status, cosmo=None):
...
+        ccl_cosmo (:obj:`ccl.cosmology`): The CCL C cosmology object.
     """
-    if isinstance(cosmo, lib.cosmology):
-        return cosmo
-    elif isinstance(cosmo, Cosmology):
+    if isinstance(cosmo, Cosmology):
         return cosmo.cosmo
     else:
-        raise TypeError("Invalid Cosmology or ccl_cosmology object.")
+        raise TypeError("Invalid Cosmology object!")
```
rmjarvis: I guess I'm not sure that this function is even required anymore.  Why not just `s/_cosmology_obj(cosmo)/cosmo.cosmo/g`

beckermr: Yeah I had the same thought. I am new to the code base and so I didn't want to mess with the internals of the code too much. I am happy to try. It just made me nervous.

beckermr: This also raises an error with some extra information which could be viewed as an advantage. /shrug

beckermr:
OK @rmjarvis. This PR should be ready for another look!

Issue #425, opened 2018-07-19
Typo in "Lensing angular power spectrum" example notebook


EiffL:
@cwwalter highlighted a typo in the [Lensing angular power spectrum](https://github.com/LSSTDESC/CCL/blob/master/examples/Lensing%20angular%20power%20spectrum.ipynb) at cell 7 which crashes the python kernel with following error:
```
Assertion failed: (nz_ba == nba), function cl_tracer_new_wrapper, file /private/var/folders/4t/66c5hggs5_1f191_9f2dljh80000gn/T/pip-build-w625qxgq/pyccl/build/pyccl/cclPYTHON_wrap.c, line 4184.
```
it looks to me that it's only a typo and that cell 7 should use `bias_ia2` in the call to `ccl.ClTracerLensing`.

rmjarvis:
I would add that you should be checking for whatever caused the assertion error in the C layer at the python layer and raise a suitable exception.  Crashing the python kernel is bad form.

Indeed this typo is probably a good unit test to add with `assert_raises`.

beckermr:
Ahhh. This is fixed in the CR 1 Installment PR I have open now. Sorry for the confusion on my end @rmjarvis.

beckermr:
Update: It is not fixed in the PR above. I have made a new issue on #479 to address these asserts.

PR #489, opened 2018-10-05
ENH remove asserts in the SWIG interface

### LINKED ISSUE #479: remove/refactor asserts in the pyccl interface
It appears we have asserts in the pyccl interface files that kill the python interpreter. This was brought up in #425 by @rmjarvis. I am making a new issue here since this appears to be a big job.

beckermr:
The asserts in the SWIG interface kill the python interpreter. This is bad form. 

Closes #479


beckermr:
OK @rmjarvis. This PR is ready for some quick CR. I am honestly 50/50 on having the checks here. It is definitely safer, but CCL only ever calls ccllib via functions which always do the right thing. So technically these checks are not needed. /shrug

tests/ccl_test_distances.py
```diff
@@ -111,139 +112,139 @@ def Neff_from_N_ur_N_ncdm(N_ur, N_ncdm):
...
     """
-    Read the file containing all the radial comoving distance benchmarks 
+    Read the file containing all the radial comoving distance benchmarks
     (distances are in Mpc/h)
     """
     # Load data from file
     dat = np.genfromtxt(join(path,"benchmark/chi_model1-5.txt")).T
-    assert(dat.shape == (6,6))
-    
+    assert_(dat.shape == (6,6))
```
beckermr: I found these weird uses of assert so I went ahead and fixed them. Not related to this PR. Sorry. Bad form on my part.

tests/ccl_test_growth.py
```diff
@@ -26,7 +26,7 @@ def read_growth_lowz_benchmark_file():
     """
     # Load data from file
     dat = np.genfromtxt(join(dirname(__file__),"benchmark/growth_model1-5.txt")).T
-    assert(dat.shape == (6,6))
+    assert_(dat.shape == (6,6))
```
beckermr: Ditto here.

pyccl/ccl_background.i
```diff
@@ -5,103 +5,124 @@
 %}
 
 // Enable vectorised arguments for arrays
-%apply (double* IN_ARRAY1, int DIM1) {(double* a, int na)};
-%apply (double* IN_ARRAY1, int DIM1) {(double* chi, int nchi)};
-%apply (double* ARGOUT_ARRAY1, int DIM1) {(double* output, int nout)};
+%apply (double *IN_ARRAY1, int DIM1) {(double *a, int na)};
+%apply (double *IN_ARRAY1, int DIM1) {(double *chi, int nchi)};
+%apply (int DIM1, double *ARGOUT_ARRAY1) {(int nout, double *output)};
```
rmjarvis: I have to say, I'm not a fan of this style change.  I'm not usually one to impose style preferences, but it doesn't seem like a great idea to change a bunch of existing code to a different (and imo worse) style.

To me `double* a` is significantly more legible than `double *a`.  `double*` is the type - pointer to a double.  So I always find the original syntax to be clearer about what the code means.  I.e. `a` is a pointer to double.  Rather than `a` when dereferenced (`*a`) is a double, which is kind of what the latter syntax means.  

It ends up in the same place (as it has to, since white space is not significant in C), but the former syntax more direct to me.

However, it's probably worth hearing from some other CCL coders about this to settle on a consistent style.  Both forms were in the existing code, and it is probably worth being consistent at least, rather than having both styles interspersed.

pyccl/ccl_background.i
```diff
@@ -5,103 +5,124 @@
...
 
 %include "../include/ccl_background.h"
 
+/* The python code here will be executed before all of the functions that
+   follow this directive. */
+%feature("pythonprepend") %{
+    if numpy.shape(a) != (nout,):
+        raise CCLError("Input shape for `a` must match `(nout,)`!")
+%}
+
```
rmjarvis: This is an interesting hack.  It's not how I would have done it.  I would have added checks in the python code directly before passing into the C functions.  I'm not sure what all the implications are about doing it this way instead though.

The only possible downside I can think of is what the traceback looks like if a user hits this.  I guess the line number of the error will show the C function call?  I'm not sure if that's going to be harder for the user to figure out how to correct their code.  Probably it's fine, but worth checking how that looks to make sure you're not making obfuscated tracebacks.

pyccl/ccl_massfunc.i
```diff
@@ -7,55 +7,61 @@
...
-                    int* status)
-{
-    assert(nout == nm);
-    for(int i=0; i < nm; i++){
+void massfunc_vec(
+        ccl_cosmology * cosmo,
+        double a, double odelta,
+        double *halo_mass, int nm,
+        int nout, double *output,
+        int* status) {
```
rmjarvis: All of these style changes also seem a bit gratuitous.  I usually prefer indenting to the opening `(` unless the function name is REALLY long, so you want to avoid cramming things into the right margin.  But that doesn't seem to be the case here.   

So again, I'd probably like some of the CCL coders to weigh in on this if they care and pick something consistent for the style of these function names.

tests/ccl_test_distances.py
rmjarvis: Not sure what was wrong with the original line here.  `assert` and `numpy.assert_` are practically synonyms.  I don't see any reason to prefer the latter.

pyccl/ccl_background.i
beckermr: Happy to switch back the location of the `*`.

I swapped the order of the size and pointer in the SWIG apply statements for good reason. SWIG by default for output arrays keeps only the first of the two parameter names and expects you to pass in the size of the output. Thus when we list the size first, it generates a function prototype in python that looks sensible.

pyccl/ccl_massfunc.i
beckermr: Happy to change it back. It was a huge hodge podge of spaces and tabs before which is why I went for it. I don't care how it actually looks. ;P

tests/ccl_test_growth.py
beckermr: Yeah so assert is not a function in python 3 and the rest of the code used `assert_`. I was trying to stick to one thing. Happy to change it back. 🤷‍♀️

rmjarvis: It's not a function, but `assert(a==b)` still works, so you don't need to care about that if you like the parentheses.

pyccl/ccl_background.i
beckermr: I will check the traceback. I added it here mostly to keep the error checking in the swig interface intact.

beckermr: BTW @rmjarvis. The python code always calls the C code through the functions here: https://github.com/LSSTDESC/CCL/blob/master/pyccl/pyutils.py#L67. These functions always pass the right size. So there is not a place to put these checks in the python code as you suggest. 

Given this, they may not even be needed. Thoughts?

pyccl/ccl_massfunc.i
beckermr: I changed these back for the most part. Some fixes of tabs and spaces remain.

pyccl/ccl_background.i
beckermr: Also, the backtrace looks fine to me.

```python
In [1]: import pyccl                                                                                                                                                                                    

In [2]: pyccl.ccllib.growth_factor_vec?                                                                                                                                                                 
Signature: pyccl.ccllib.growth_factor_vec(cosmo, a, nout, status)
Docstring: growth_factor_vec(cosmology cosmo, double * a, int nout, int * status)
File:      ~/Desktop/CCL/pyccl/ccllib.py
Type:      function

In [3]: cosmo = ccl.Cosmology(Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96)                                                                                                                
---------------------------------------------------------------------------
NameError                                 Traceback (most recent call last)
<ipython-input-3-ffb1d6c5f82d> in <module>
----> 1 cosmo = ccl.Cosmology(Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96)

NameError: name 'ccl' is not defined

In [4]: cosmo = pyccl.Cosmology(Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96)                                                                                                              

In [5]: pyccl.ccllib.growth_factor_vec(cosmo, [0.0, 1.0, 2.0], 2, 0)                                                                                                                                    
---------------------------------------------------------------------------
CCLError                                  Traceback (most recent call last)
<ipython-input-5-72f9c8633d66> in <module>
----> 1 pyccl.ccllib.growth_factor_vec(cosmo, [0.0, 1.0, 2.0], 2, 0)

~/Desktop/CCL/pyccl/ccllib.py in growth_factor_vec(cosmo, a, nout, status)
    647 
    648     if numpy.shape(a) != (nout,):
--> 649         raise CCLError("Input shape for `a` must match `(nout,)`!")
    650 
    651 

CCLError: Input shape for `a` must match `(nout,)`!

In [6]:  
```

Note, as I described above, there are no higher level calls that are not wrapped in a `_vectorize_fn_*` function.

beckermr: Note also the nice function signature from switching the order of the argout size and pointer. :)

rmjarvis: Well, the main point was to avoid problems like [this](#425).  As long as the code doesn't let the user accidentally crash the interpreter, I don't much care how you go about it.

Although, now that I'm looking, I don't see the typo version as a unit test.  It should be.  You should check that the typo that used to be in the notebook raises an exception rather than crashing the interpreter.

beckermr: Sure I can add the unit test!

PR #490, opened 2018-10-05
GGL checks

elisachisari:
This PR will (eventually) add GGL tests to Cls and correlation functions. For the time being, I have modified `ccl_test_cls.c` to generate a fake benchmark file and cross-check against it. This file corresponds to the cross-correlation of number counts in bin 1 and sources in bin 2 both for the analytic and histo case. One needs to swap the "fake" file for the new one. 

The case of the correlation function benchmark still needs to be addressed.


elisachisari:
This PR has been merged into #506, so I'm closing it.

PR #493, opened 2018-10-10
ENH remove the Parameters object, repr and pickling

beckermr:
This PR 

1. removes the parameters object 
2. makes sure the Cosmology object can pickle
3. makes sure the Cosmology object has a eval-able repr.


src/ccl_core.c
```diff
@@ -887,7 +887,7 @@ ccl_parameters ccl_parameters_read_yaml(const char * filename, int *status)
 
   double mnu[3] = {0.0, 0.0, 0.0};
   if (N_nu_mass>0){
-    *status |= (0==fscanf(f, "mnu = ["));
+    *status |= (0==fscanf(f, "mnu: ["));
```
beckermr: We were not writing valid YAML IIUIC.

beckermr:
Sorry for all of the white space changes. My editor...

beckermr:
I added one more commit. It let's the user specify `Omega_g` directly. This is a common hack where at late times one wants to neglect radiation in the Hubble function. It is used extensively in the benchmarks, tests and examples. I added a warning in the docs that using this might result in a non-physical model.

beckermr:
This PR is good to go @rmjarvis.

pyccl/core.py
```diff
@@ -303,9 +308,10 @@ def read_yaml(cls, filename):
         return cls(**inits)
 
     def _build_config(
-            self, transfer_function, matter_power_spectrum,
```
rainwoodman: Isn't it a bit dangerous that an internal facing function takes default values and optional arguments? Shouldn't the internal code always know what to passin? In addition, the docstring didn't say what None stands for.

pyccl/core.py
```diff
@@ -174,10 +175,12 @@ class Cosmology(object):
         n_s (:obj:`float`): Primordial scalar perturbation spectral index.
         Omega_k (:obj:`float`, optional): Curvature density fraction.
             Defaults to 0.
-        N_nu_rel (:obj:`float`, optional): Effective number of massless
+        Omega_g (:obj:`float`, optional): Density in relativistic species
+            except massless neutrinos. The default of `None` corresponds
+            to setting this from the CMB temperature. Note that if a non-`None`
+            value is given, this may result in a physically inconsistent model.
```
rainwoodman: Please explain inconsistency -- it means inconsistency with Tcmb or not flat or inconsistency with Neff?

pyccl/core.py
beckermr: The internal function is only ever called by us so I’m not worried about the defaults. 

As for a doc string, I would think the name of the function is sufficient.

rainwoodman: Sure. It was just a random thought that popped up when I saw this. Now I realized my concept flow was 'optional arg' => 'syntax sugar' => 'public api', then I blindly inverted it.

beckermr:
The updates will happen in #494 . Are we good to merge @rmjarvis ?

rmjarvis:
> The updates will happen in #494 . Are we good to merge @rmjarvis ?

Not my call, I don't think.  One of the principal CCL devs should probably decide whether to approve the API changes or not.  Hence my review being a comment, not an approval.

beckermr:
@elisachisari care to weigh in here?

src/ccl_core.c
philbull: Do we have a unit test for this?

tests/ccl_test_core.py
```diff
@@ -9,115 +11,92 @@ def test_parameters_valid_input():
...
+    assert_no_warnings(
+        ccl.Cosmology,
+        Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9, n_s=0.96,
+        z_mg=None)
+    assert_no_warnings(
+        ccl.Cosmology,
+        Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9, n_s=0.96,
+        df_mg=None)
 
-def test_parameters_spelling():
```
philbull: Why was this removed?

tests/ccl_test_core.py
```diff
@@ -251,5 +248,59 @@ def test_cosmology_output():
...
+        cosmo2 = pickle.load(fp)
+
+    assert_(
+        ccl.comoving_radial_distance(cosmo, 0.5) ==
+        ccl.comoving_radial_distance(cosmo2, 0.5))
+
+
+def test_cosmology_repr():
+    """Check that we can make a Cosmology object from its repr."""
+    import pyccl  # noqa: F401
```
philbull: Cryptic comment?

src/ccl_core.c
beckermr: Yes. I tagged you below.

beckermr:
@philbull I plan to do the doc, example and other changes in other PRs. This is a big change and one giant PR is a lot.

beckermr:
I'm gonna merge if we have no other things.

pyccl/core.py
```diff
@@ -59,7 +59,7 @@
       physical model assumed for neutrinos.
 """
 import numpy as np
-from warnings import warn
```
elisachisari: Are we not using this anymore?

pyccl/core.py
```diff
@@ -164,45 +164,231 @@ class Parameters(object):
...
+        h (:obj:`float`): Hubble constant divided by 100 km/s/Mpc; unitless.
+        A_s (:obj:`float`): Power spectrum normalization. Exactly one of A_s
+            and sigma_8 is required.
+        sigma8 (:obj:`float`): Variance of matter density perturbations at
+            an 8 Mpc/h scale. Exactly one of A_s and sigma_8 is required.
+        n_s (:obj:`float`): Primordial scalar perturbation spectral index.
+        Omega_k (:obj:`float`, optional): Curvature density fraction.
+            Defaults to 0.
+        Omega_g (:obj:`float`, optional): Density in relativistic species
+            except massless neutrinos. The default of `None` corresponds
```
elisachisari: In general, we should edit the CCL note in doc/0000_ccl_note/main.tex to cover these changes. But specifically this warning should be given there. Perhaps we could check for inconsistency and give a warning in the code itself?

pyccl/core.py
```diff
@@ -164,45 +164,231 @@ class Parameters(object):
...
+            inits['mnu_type'] = 'list'
+
+        return cls(**inits)
+
+    def _build_config(
+            self, transfer_function=None, matter_power_spectrum=None,
+            baryons_power_spectrum=None,
+            mass_function=None, halo_concentration=None,
+            emulator_neutrinos=None):
+        """Build a ccl_configuration struct"""
```
elisachisari: Can we have a bit more explanation of what this does?

pyccl/core.py
```diff
@@ -275,304 +461,107 @@ def __init__(self, Omega_c=None, Omega_b=None, h=None, A_s=None, n_s=None,
...
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__ = state
+        # we removed the C data when it was pickled, so now we unpickle
+        # and rebuild the C data
+        self._build_cosmo()
+
+    def __repr__(self):
+        """Make an eval-able string."""
```
elisachisari: Give an example here within the comment?

tests/ccl_test_core.py
```diff
@@ -134,52 +113,68 @@ def test_parameters_mgrowth():
...
+        ValueError, ccl.Cosmology,
+        Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9, n_s=0.96,
+        z_mg=zarr,
+        df_mg=dfarr[1:])
+    assert_raises(
+        ValueError, ccl.Cosmology,
+        Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9, n_s=0.96,
+        z_mg=zarr,
+        df_mg=np.column_stack((dfarr, dfarr)))
+
```
elisachisari: Have added a check for the new Omega_g implementation?

tests/ccl_test_core.py
```diff
@@ -251,5 +248,59 @@ def test_cosmology_output():
     assert_(cosmo.has_power() is True)
 
 
+def test_cosmology_pickles():
```
elisachisari: What is this testing?

pyccl/core.py
beckermr: Yep.

tests/ccl_test_distances.py
```diff
@@ -368,7 +363,7 @@ def compare_distances_mnu(z, chi_bench,dm_bench, Omega_v, w0, wa, Neff, mnu):
 
     assert_allclose(dm, dm_bench[a_not_one], atol=1e-3, rtol = DISTANCES_TOLERANCE_MNU)
 
-@decorators.slow
+@numpy.testing.dec.slow
```
elisachisari: Have you checked slow/fast tests still run after this change?

pyccl/core.py
beckermr: Just to make sure I understand, you want me to warn the user if Omega_g is zero but T_CMB is not? This is a very common use case for late-time LSS computations. I am a bit worried this would be too noisy for an option people only opt into. 

However, if you still want to go ahead with the warning I am happy to do so!

beckermr: Sounds good. Do you mean the DESC note, or somewhere else in the docs? I have added an extra note in the docstring as well.

tests/ccl_test_distances.py
beckermr: yes it appears to work.

pyccl/core.py
beckermr: i'm making an issue to update the note and add this comment.

beckermr: https://github.com/LSSTDESC/CCL/issues/502

beckermr:
OK @elisachisari. I think I got everything. I made an issue to update the note, docs etc. We have an open PR for the examples/benchmarks etc. 

BTW, once all of that is good to go, we should consider bumping the version. However, we might want to wait until the final PRs from the code review have gone in before bumping the version. 

I also think we may want to take some time to do another review of the API-only before we get to v1. We are going to be stuck with our decisions and so making sure we are happy with them makes sense. 

Are we good to merge?

elisachisari:
I'd be keen on modifying the note already here. (Also, I realized, the README has examples that need to be modified.) Why not bumping the version with this PR as well? We can do v0.9.1.

beckermr:
The note is already out of date from when we merged the changes to the tracers. I don't want to have to touch it twice given my time constraints. If we can merge this so I can get the examples done that would be great.

elisachisari:
I really think we should bump the version with this, and have the docs in the note. Why not leave this open till the note and examples are updated?

beckermr:
Yeah so bumping the version to 0.9.1 will be misleading. This would indicate a microversion change with API compatibility with 0.9.0. However we have broken the API with this PR. Thus for our version numbers to make any sense, we need to wait and bump to 1.0.0 all at once. 

Blocking other PRs that depend on this one to update the note doesn't make a ton of sense either. It limits how much work we can get done in parallel.

We really should merge and then continue fixing the other things. This way someone else can pick up the note edits next week if they so desire. I will be traveling next week and so really want to have the examples updated before the sprint week.

rmjarvis:
People really shouldn't be using the master branch for anything other than development.  If people find their code breaks because they update to a newer version of master, then this provides an opportunity to tell them they shouldn't do that and to use `v0.9` instead.

PR #496, opened 2018-10-13
ENH simpler tracers

### LINKED ISSUE #471: rename the `CLTracer` class and its subclasses
@rmjarvis suggested renaming these to `Tracer`, `NumberCounts`, etc.

### LINKED ISSUE #472: one and only one way to do things
Right now the python interface has a lot of arguments that are redundant, polymorphic, or multiple ways to do the same thing. This is a lot of work to maintain and makes the API more confusing. Let's simplify and save ourselves headaches down the road.

For example,

```python
tracer_types = {
    'nc':               const.CL_TRACER_NC,
    'number_count':     const.CL_TRACER_NC,
    'wl':               const.CL_TRACER_WL,
    'lensing':          const.CL_TRACER_WL,
    'weak_lensing':     const.CL_TRACER_WL,
    'cmbl':             const.CL_TRACER_CL,
    'cmb_lensing':      const.CL_TRACER_CL,
}
```

We don't need to maintain all of the different options above and it is easier for people to remember one thing instead of having to guess which variants of 'weak_lensing' are allowed.

Other things I am seeing along these lines:

1. We probably should force users to specify the array of redshifts separately for each aspect of a tracer.
2. ~We should force people to make a `Parameters` object or just roll all of that functionality into the `Cosmology` object. I vote for the latter actually.~ (done)
3. The `has_magnification` and `mag_bias` keyword arguments are redundant. 
4. The `has_intrinsic_alignment` keyword can be inferred from the other settings on the tracer object.

beckermr:
Closes #471
Closes #472

### DIFF
--- a/pyccl/cls.py
+++ b/pyccl/cls.py
@@ -36,7 +36,7 @@ class _Tracer(object):
     """
     def __init__(self, *args, **kwargs):
         raise NotImplementedError(
-            "A `Tracer` object cannot be used directly. Use one of "
+            "A `_Tracer` object cannot be used directly. Use one of "
             "`NumberCounts`, `WeakLensing` or `CMBLensing` instead.")
 
     def _build_tracer(
@@ -286,7 +286,7 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell,
 
     Args:
         cosmo (:obj:`Cosmology`): A Cosmology object.
-        cltracer1, cltracer2 (:obj:`Tracer`): Tracer objects, of any kind.
+        cltracer1, cltracer2 (:obj:`_Tracer`): Tracer objects, of any kind.
         ell (float or array_like): Angular wavenumber(s) at which to evaluate
             the angular power spectrum.
         l_limber (float) : Angular wavenumber beyond which Limber's
--- a/pyccl/cls.py
+++ b/pyccl/cls.py
@@ -28,7 +28,7 @@ class Tracer(object):
     """A tracer of the matter density field.
 
     .. note:: This class cannot be used directly. Use one of
-              :obj:`NumberCounts`, :obj:`WeakLensing` or :obj:`CMBLensing`
+              :obj:`NumberCountsTracer`, :obj:`WeakLensingTracer` or :obj:`CMBLensingTracer`
               instead.
 
     This class contains all information describing the transfer functon of
@@ -37,7 +37,7 @@ class Tracer(object):
     def __init__(self, *args, **kwargs):
         raise NotImplementedError(
             "A `Tracer` object cannot be used directly. Use one of "
-            "`NumberCounts`, `WeakLensing` or `CMBLensing` instead.")
+            "`NumberCountsTracer`, `WeakLensingTracer` or `CMBLensingTracer` instead.")
 
     def _build_tracer(
             self, cosmo, tracer_type, has_rsd=False,
@@ -183,7 +183,7 @@ class Tracer(object):
                 lib.cl_tracer_free(self.cltracer)
 
 
-class NumberCounts(Tracer):
+class NumberCountsTracer(Tracer):
     """A Tracer for galaxy number counts (galaxy clustering).
 
     Args:
@@ -214,7 +214,7 @@ class NumberCounts(Tracer):
             ia_bias=None, red_frac=None)
 
 
-class WeakLensing(Tracer):
+class WeakLensingTracer(Tracer):
     """A Tracer for weak lensing shear (galaxy shapes).
 
     Args:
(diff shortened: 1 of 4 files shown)


beckermr:
All the tests pass. I skipped them since I was just putting in doc changes.

pyccl/cls.py
```diff
@@ -3,121 +3,106 @@
...
+    'mag_win': const.CCL_CLT_WM,
 }
 
 # Define symbolic 'None' type for arrays, to allow proper handling by swig
 # wrapper
 NoneArr = np.array([])
 
 
-class ClTracer(object):
+class _Tracer(object):
```
rmjarvis: I get why you made this a leading underscore, since you don't want users to instantiate it.  But there are docs that specify ``:obj:`Tracer` ``, so it's not really private in the sense of being merely an implementation detail. 

I'd probably lean towards leaving this as a regular public name without the underscore.  Users might have use of writing the class name directly, so you don't necessarily want to scare them into thinking the base class name is liable to change without notice.  (As "private" classes and methods are subject to do.)

pyccl/cls.py
```diff
@@ -145,12 +130,12 @@ def get_internal_function(self, cosmo, function, a):
...
-                    'rfz', 'red_fraction': red fraction
-                    'baz', 'a_bias': intrinsic alignment bias
-                    'wL', 'window_lensing': weak lensing window function
-                    'wM', 'window_magnif': magnification window function
+                    'bias': bias
+                    'mag_bias': magnification bias
+                    'red_frac': red fraction
+                    'ia_bias': intrinsic alignment bias
+                    'lensing_win': weak lensing window function
+                    'mag_win': magnification window function
```
rmjarvis: I like a lot of the changes to the names here (in addition to the class name changes).  But I suspect this breaks quite a few of the jupyter notebooks that are in the repo `examples` directory.  You should probably fix them to conform to the new API as part of this PR.

beckermr: I made a second pr for this. I fixed everything there. I was going to do it in two passes.

beckermr: Happy to put those changes back here. It is just a lot.

beckermr: Ahhh Just saw this. Great!

beckermr: I tagged you on the PR with the changes to the examples. I reran all of them too.

pyccl/cls.py
beckermr: Good point. `isinstance` checks come to mind actually. Will change.

beckermr:
@damonge Care to review?

pyccl/cls.py
```diff
@@ -3,121 +3,105 @@
...
-tracer_types = {
-    'nc':               const.CL_TRACER_NC,
-    'number_count':     const.CL_TRACER_NC,
-    'wl':               const.CL_TRACER_WL,
-    'lensing':          const.CL_TRACER_WL,
-    'weak_lensing':     const.CL_TRACER_WL,
-    'cmbl':             const.CL_TRACER_CL,
-    'cmb_lensing':      const.CL_TRACER_CL,
-}
-
```
damonge: Out of curiosity: why did you get rid of these? I'd have thought that users being able to use informative strings rather than constants would be friendlier.

pyccl/cls.py
```diff
@@ -3,121 +3,105 @@
...
-    'window_lensing':   const.CCL_CLT_WL,
-    'wM':               const.CCL_CLT_WM,
-    'window_magnif':    const.CCL_CLT_WM,
+    'dndz': const.CCL_CLT_NZ,
+    'bias': const.CCL_CLT_BZ,
+    'mag_bias': const.CCL_CLT_SZ,
+    'red_frac': const.CCL_CLT_RF,
+    'ia_bias': const.CCL_CLT_BA,
+    'lensing_win': const.CCL_CLT_WL,
+    'mag_win': const.CCL_CLT_WM,
```
damonge: Or, otherwise, why did you not get rid of these?

pyccl/cls.py
beckermr: No longer needed. You can’t instantiate a tracer directly.

pyccl/cls.py
beckermr: These are still used as arguments for a function.

pyccl/cls.py
beckermr: You have to use a subclass instead and so I am using the constants directly.

beckermr:
@damonge Tests should be fixed. Back to you!

beckermr:
At minus -0.1%? Let's pass on this. We have too much to do to worry about this right now.

PR #500, opened 2018-10-18
Code review - installment 4

elisachisari:
Further changes were needed to address installment 4. I'm re-opening this branch to address them.


src/ccl_power.c
```diff
@@ -489,71 +488,80 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
-    free(y2d_lin);
-    free(z);
   }
 
 }
 
-/* BCM correction */
-// See Schneider & Teyssier (2015) for details of the model.
+/* BCM correction 
+   See Schneider & Teyssier (2015) for details of the model.
```
rmjarvis: In ccl_power.h, Teyssier is misspelled as Teyrrsier.  Should fix that.

src/ccl_power.c
```diff
@@ -489,71 +488,80 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+	  cosmo->data.p_nl = log_power_nl;
+	}
       }
-
+      
+      ccl_free_class_structs(cosmo, &ba,&th,&pt,&tr,&pm,&sp,&nl,&le,init_arr,status);
+      free(x);
+      free(y2d_lin);
+      free(a);
       free(y2d_nl);
```
rmjarvis: I find all these frees all over the place hard to follow.  I still recommend switching over to the pattern that Tom has implemented in #507.  It makes it much clearer that everything is properly freed, since you do them all only once just before returning (also only once).

(Fine though if the plan is to do more refactorings like that in other PRs.)

src/ccl_power.c
```diff
@@ -660,94 +668,97 @@ static eh_struct *eh_struct_new(ccl_parameters *params)
...
+  return f*tkEH_0(eh->keq,k,1,eh->betac)+ 
+    (1-f)*tkEH_0(eh->keq,k,eh->alphac,eh->betac); //Returns Eq 17
 }
 
 static double jbes0(double x)
 {
   double jl;
-  double ax=fabs(x);
-  double ax2=ax*ax;
+  double ax2=fabs(x)*fabs(x);
```
rmjarvis: fabs isn't necessary here.  This could just be `ax2 = x*x`.

src/ccl_power.c
```diff
@@ -1164,10 +1180,11 @@ static void ccl_cosmology_compute_power_emu(ccl_cosmology * cosmo, int * status)
...
 		  if (cosmo->params.N_nu_mass==3){
-			  double diff1 = pow((cosmo->params.mnu[0] - cosmo->params.mnu[1]) * (cosmo->params.mnu[0] - cosmo->params.mnu[1]), 0.5);
-			  double diff2 = pow((cosmo->params.mnu[1] - cosmo->params.mnu[2]) * (cosmo->params.mnu[1] - cosmo->params.mnu[2]), 0.5);
-			  double diff3 = pow((cosmo->params.mnu[2] - cosmo->params.mnu[0]) * (cosmo->params.mnu[2] - cosmo->params.mnu[0]), 0.5);
-			  if (diff1>1e-12 || diff2>1e-12 || diff3>1e-12){
+			  //double diff1 = pow((cosmo->params.mnu[0] - cosmo->params.mnu[1]) * (cosmo->params.mnu[0] - cosmo->params.mnu[1]), 0.5);
+			  //double diff2 = pow((cosmo->params.mnu[1] - cosmo->params.mnu[2]) * (cosmo->params.mnu[1] - cosmo->params.mnu[2]), 0.5);
+			  //double diff3 = pow((cosmo->params.mnu[2] - cosmo->params.mnu[0]) * (cosmo->params.mnu[2] - cosmo->params.mnu[0]), 0.5);
+			  //if (diff1>1e-12 || diff2>1e-12 || diff3>1e-12){
+			  if (cosmo->params.mnu[0] != cosmo->params.mnu[1] || cosmo->params.mnu[0] != cosmo->params.mnu[1] || cosmo->params.mnu[1] != cosmo->params.mnu[2]){
```
rmjarvis: You have `mnu[0] != mnu[1]` here twice.

include/ccl_emu17.h
```diff
@@ -4,6 +4,7 @@
 #define A_MIN_EMU 1./3.
 #define K_MAX_EMU 5.0
 #define K_MIN_EMU 1.0000000474974513E-003
+#define NK_EMU 351
```
rmjarvis: In ccl_emu.c, you should use `NK_EMU` now in place of `nmode` (which is currently set to 351 near the top).

src/ccl_power.c
elisachisari: I changed `ccl_cosmology_compute_power_class` following the example of #507. Let me know if this is what you meant, then I can refactor the rest.

src/ccl_power.c
```diff
@@ -436,124 +436,134 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+
+    //If error, store status, we will free later
+    if(newstatus) {
       *status = CCL_ERROR_CLASS;
       ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error computing CLASS power spectrum\n");
+    }
 
-      ccl_free_class_structs(cosmo, &ba,&th,&pt,&tr,&pm,&sp,&nl,&le,init_arr,status);
+    //If not proceed
+    if(!newstatus){
```
rmjarvis: This should be back at the main level and start another `if (!*status)` now.

src/ccl_power.c
```diff
@@ -436,124 +436,134 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+      }
 
-    // At the moment KMIN can't be less than CLASS's kmin in the nonlinear case.
-    if (kmin<(exp(sp.ln_k[0]))) {
-      *status = CCL_ERROR_CLASS;
-      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): K_MIN is less than CLASS's kmin. Not yet supported for nonlinear P(k).\n");
+      //If not, proceed
+      if(!pwstatus){
+	cosmo->data.p_lin = log_power;
+      }
```
rmjarvis: This bit also would be back at the main level as another `if (!*status)`.

src/ccl_power.c
```diff
@@ -436,124 +436,134 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+    
+  } else { //Linear power spec failed, so we return without proceeding to nonlinear.
+
+    free(x);
+    free(a);
+    free(y2d_nl);
+    free(y2d_lin);
+    ccl_free_class_structs(cosmo, &ba,&th,&pt,&tr,&pm,&sp,&nl,&le,init_arr,status);
+    return;
+  }
```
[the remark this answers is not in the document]
rmjarvis: I think you probably see the pattern now, so I'll refrain from annotating the rest of the function, but the basic pattern is:

start with *status = 0

if (!*status) {  // no errors yet.
    Do something that might cause an error;
    If (there was an error) *status = ERROR_CODE;
}

.. repeat as many times as needed

// At the end clean up any memory
free stuff that was temporary within this function
if (status) {
   free stuff that we aren't returning because of the error
}

return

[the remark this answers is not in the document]
elisachisari: Ok, thanks. Please check the new structure of that function. It's proving more difficult for the other ones and I'm not sure it's necessary to get all these changes in right now. Could we open an issue for the other power spectrum methods?

Issue #501, opened 2018-10-18
refactor code in ccl_background.c


beckermr:
@rmjarvis pointed out an opportunity to refactor code in ccl_background.c

See here: https://github.com/LSSTDESC/CCL/pull/433#discussion_r206747504

PR #507, opened 2018-10-22
ccl_background function rewrite

tmcclintock:
In the code review, @rmjarvis suggested that the workflow in some of the C portions of the code should be rewritten in an effort to avoid memory leaks. Specifically, instead of calling `free()` on allocated variables as errors were discovered, we should instead make all calls to `free()` and all return statements at the end of the function. As @damonge pointed out on slack, this can lead to a proliferation of extra `if(*status){..}` type of statements.

This PR rewrote the `ccl_cosmology_compute_distances()` in `ccl_background.c` in Mike's style. IMO, the readability went up, despite the extra conditionals, and the memory is guaranteed to be freed by the end (since there is only one return statement). 

A reviewer should probably not look at the "unified" view and instead the "split" view of this file, if you read the code on Github. In fact, looking at the source file itself might be more helpful.

I ran my own comparison between comoving distances computed before/after my changes, and also ran the tests in `build/check_ccl`.


philbull:
Lots of errors here...

tmcclintock:
Hey @philbull, my apologies, but I don't know what to do with the coveralls and CI responses. You mention there are lots of errors, but I don't know what I am looking for. Can you advise me on what I should do with this? Sorry for being a n00b.

beckermr:
Looking at the CI, it appears that the osx builds completed but not the linux ones. This is rather odd. I rekicked everything just in case.

beckermr:
@tmcclintock @philbull This PR is failing the build because you are exceeding the log length for travis. GSL is printing a bunch of errors over and over again. This needs to be fixed.

beckermr:
This is the warning `WARNING: ccl_background.c: a_of_chi(): GSL error: problem with user-supplied function`.

elisachisari:
There will be some conflicts with installment 2 of the code review, about to be merged.

tmcclintock:
The Travis failures are only w.r.t. flake8, and all the actual tests pass. I'll try to merge master onto this branch so that there won't be any conflicts with installment 2 of the code review.

elisachisari:
Code review installment 2 now merged, so it's a good time to bring in the changes in master again.

rmjarvis:
I'd recommend disabling the flake8 errors that are causing this to fail.
* F401 is dumb.  There are lots of reasons to import things into a module and not use them directly.  (Since other code will use them within that namespace.)
* E501 is archaic at best.  Who keeps their editor at 80 characters anymore?
* W605 is inappropriate for parsing doc strings.  These are escape sequences in valid Sphinx markup.  You don't want to be having to put noqa all  around your doc strings that use these things.

So really, you should just turn these off in the flake8 tests.  Just change the flake8 line in `.travis.yml` to
```
flake8 --ignore=F401,E501,W605 pyccl
```

tmcclintock:
I should have linked this earlier, but this PR addresses #501.

The most recent commit merges master (post-code review installment 2) into this branch.

tmcclintock:
I realized I forgot to ignore those flake8 rules and it's probably going to break again, unless the updates to master fix them.

src/ccl_background.c
```diff
@@ -345,157 +345,168 @@ TASK: if not already there, make a table of comoving distances and of E(a)
...
+  a     = malloc(sizeof(double)*na);
+  //Allocate space for GSL root finders
   const gsl_root_fdfsolver_type *T=gsl_root_fdfsolver_newton;
   gsl_root_fdfsolver *s=gsl_root_fdfsolver_alloc(T);
-  for(int i=1;i<na-1;i++) {
-    // we are using the previous value as a guess here to help the root finder
-    // as long as we use small steps in a this should be fine
-    a_of_chi(y[i],cosmo, status, &a0,s);
-    a[i]=a0;
+  gsl_spline * achi=gsl_spline_alloc(A_SPLINE_TYPE, na);
```
rmjarvis: This bit needs to be moved down into the below `if (!*status)` block.

src/ccl_background.c
```diff
@@ -345,157 +345,168 @@ TASK: if not already there, make a table of comoving distances and of E(a)
...
-    gsl_spline_free(chi);
-    *status = CCL_ERROR_ROOT;
-    ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_distances(): a(chi) root-finding error \n");
-    return;
+  //Check for messed up chi conditions
+  if (!*status){
+    if((fabs(chi_a[0]-chi0)>1e-5) || (fabs(chi_a[na-1]-chif)>1e-5)) {
+      *status = CCL_ERROR_LINSPACE; 
+      ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_distances(): Error creating linear spacing in chi\n");
+    }
```
rmjarvis: This could be combined with the above status check:
```
if (a == NULL || chi_a==NULL) {
    *status = CCL_ERROR_MEMORY;
    ...
} else if (fabs(chi_a[0]-chi0)>1e-5) || (fabs(chi_a[na-1]-chif)>1e-5)) {
    *status = CCL_ERROR_LINSPACE;
    ...
}
```

src/ccl_background.c
```diff
@@ -345,157 +345,168 @@ TASK: if not already there, make a table of comoving distances and of E(a)
...
+  // Calculate a(chi)
+  if (!*status){
+    a[0]=a0; a[na-1]=af;
+    for(int i=1;i<na-1;i++) {
+      // we are using the previous value as a guess here to help the root finder
+      // as long as we use small steps in a this should be fine
+      a_of_chi(chi_a[i],cosmo, status, &a0, s);
+      a[i]=a0;
+    }
+    gsl_root_fdfsolver_free(s);
```
rmjarvis: This doesn't belong here.  Should put this with the other frees below that happen regardless of status.

tmcclintock:
@beckermr @elisachisari I have made the updates to this branch that @rmjarvis requested, however I don't want to commit them yet. At the moment I cannot compile CCL. I don't know if it's related to updating to the new version of CLASS, but I get this error when I use `make` from the `build` directory:
```
[ 33%] Performing build step for 'CLASS'
clang: warning: -O4 is equivalent to -O3 [-Wdeprecated]
In file included from ../tools/growTable.c:5:
In file included from ../include/growTable.h:8:
../include/common.h:3:10: fatal error: 'stdio.h' file not found
#include "stdio.h"
...
```

I have exported both the `CCL_PARAM_FILE` and `CLASS_PARAM_DIR`, so I am at a loss as to what the issue is. Any ideas?

beckermr:
Yeah that is your compiler not finding the system headers. Idk why this happening. Maybe nuke the build dir and start over. 

Also, @rmjarvis, turning off flake8 errors globally is a bad idea. We can comments that tell flake8 to ignore certain errors in certain spots. The reason to keep F401 on is that many times an unused import is a sign of a bug. The line length one is better solved by adding an increased line length to the flake8 invocation.

tmcclintock:
@rmjarvis The most recent commits addressed all of your requested changes. The update passed all travis checks, so I think we are ready to merge. Let me know if you want me to make any other edits, and thanks again for your review!

Issue #512, opened 2018-10-25
Passing bias=None to NumberCountsTracer crashed python kernel


tilmantroester:
This snippet crashes my python kernel (python 3.6, CCL master):

```import numpy as np
import pyccl as ccl

ccl_cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7, n_s=1.0, sigma8=0.8)

z = np.linspace(0, 4, 100)
nofz = np.exp(-0.5*(z-0.3)**2/0.15**2)
nc_tracer = ccl.NumberCountsTracer(cosmo=ccl_cosmo, dndz=(z,nofz), bias=None, has_rsd=False)
```

`bias` is currently a required argument, contrary to the documentation.

beckermr:
Ack. My bad. Putting in a patch now.

beckermr:
So the bias is required. I changed the docs.

beckermr:
The fact that the interpreter crashes is fun.

beckermr:
You have to pass the bias. None is. I longer listed as a valid value. We can do more error checking I suppose but I don’t think we can account for everything. Happy to discuss more.

tilmantroester:
If passing `None` (or something else) crashes the interpreter, that should probably be caught.

beckermr:
Sure. Feel free to make a PR. I won't be able to get to it.

rmjarvis:
If something crashes the interpreter, you definitely want to check for that in python before calling the SWIG-ed C function.  That should really never happen.

PR #515, opened 2018-10-25
Guard against ill-formed arguments when creating Tracer objects

tilmantroester:
This addresses issue #512 . When passing something that's not a tuple for `dndz` (affects all tracers) and `bias` (only affects number counts) the python interpreter crashes because the C layer seg faults. The reason is that there are no guards to check that there are actually arrays with elements in the tuple. The C layer tries to build splines in any case and crashes when there's nothing to build.

This PR adds some guards for the `dndz` and `bias` cases but ideally `cl_tracer` should check its arguments before trying to access arrays.

### DIFF
--- a/pyccl/cls.py
+++ b/pyccl/cls.py
@@ -3,6 +3,8 @@ from . import constants as const
 from .core import check
 import numpy as np
 
+import collections
+
 # Same mapping for non-Limber integration methods
 nonlimber_methods = {
     'native': const.CCL_NONLIMBER_METHOD_NATIVE,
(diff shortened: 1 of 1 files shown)


pyccl/cls.py
```diff
@@ -89,6 +91,20 @@ def _build_tracer(
...
+        # but this should probably be checked for at the C level.
+        if tracer_type in [const.CL_TRACER_WL, const.CL_TRACER_NC, const.CL_TRACER_CL]:
+            if not isinstance(dndz, collections.Iterable) or \
+               len(dndz) != 2 or \
+               not (isinstance(dndz[0], np.ndarray) and isinstance(dndz[1], np.ndarray)):
+                raise ValueError("dndz needs to be a tuple of two arrays.")
+        if tracer_type in [const.CL_TRACER_NC]:
+            if not isinstance(bias, collections.Iterable) or \
+               len(bias) != 2 or \
+               not (isinstance(bias[0], np.ndarray) and isinstance(bias[1], np.ndarray)):
```
beckermr: This should accept any iterable, not just numpy arrays.

pyccl/cls.py
```diff
@@ -89,6 +91,20 @@ def _build_tracer(
                 "must be specified.")
         has_intrinsic_alignment = red_frac is not None
 
+        # Passing None for certain arguments causes segmentation faults at the
+        # moment. The following checks try to guard against these instances
+        # but this should probably be checked for at the C level.
+        if tracer_type in [const.CL_TRACER_WL, const.CL_TRACER_NC, const.CL_TRACER_CL]:
+            if not isinstance(dndz, collections.Iterable) or \
+               len(dndz) != 2 or \
+               not (isinstance(dndz[0], np.ndarray) and isinstance(dndz[1], np.ndarray)):
```
beckermr: ditto here.

pyccl/cls.py
tilmantroester: Fixed.

PR #516, opened 2018-10-26
Disable GIT_SHALLOW for cmake < 3.10.

tilmantroester:
Using `GIT_SHALLOW` in `ExternalProject_Add` makes the cloning a lot faster but seems to break for older cmake versions. This PR attempts to address that. 

Requires testing from people with old cmake versions!

### DIFF
--- a/cmake/Modules/BuildAngpow.cmake
+++ b/cmake/Modules/BuildAngpow.cmake
@@ -1,5 +1,12 @@
 include(ExternalProject)
 
+# Old versions of cmake don't seem to play nice with the GIT_SHALLOW option
+if(${CMAKE_VERSION} VERSION_LESS "3.10.0")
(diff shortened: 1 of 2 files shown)


damonge:
@tilmantroester works for me. Thanks a lot!

damonge:
@tilmantroester this doesn't seem to work for me (v3.5), but it works if I remove `GIT_SHALLOW ${SHALLOW_GIT_CLONE}` altogether.

tilmantroester:
> @tilmantroester this doesn't seem to work for me (v3.5), but it works if I remove `GIT_SHALLOW ${SHALLOW_GIT_CLONE}` altogether.

@damonge Could you try again? It should hide the whole `GIT_SHALLOW` business when cmake < 3.10 now.

c-d-leonard:
This works for me now.

PR #524, opened 2018-10-29
DOC updated the note

### LINKED ISSUE #502: update CCL note for new API changes
Several API changes have been put in the code:

- redid API of the CCL tracers (#496)
- deprecated the Parameters object (#493)

We need to update the DESC note and any other docs. The examples, benchmarks etc are being handled in another PR. 

Also we need to add docs on setting Omega_g to zero per this [convo](https://github.com/LSSTDESC/CCL/pull/493#discussion_r226700626).

beckermr:
This PR updates the note for the new API changes.

Closes #502.


beckermr:
Ack. I'm sorry for missing stuff. Will fix!

beckermr:
@elisachisari The C parameters object still exists so the "creating a cosmology" section should still reference it. I removed the constructors that we no longer maintain. Finally, I went through the readme. 

Back to you!

Issue #511, opened 2018-10-24
Generalized Limber integrator


damonge:
OK, so to get the angular power spectrum between two projected fields a and b, in the Limber approximation you need to integrate:
![codecogseqn](https://user-images.githubusercontent.com/8809493/47445411-c12d7d80-d7b0-11e8-8db5-31218e6f8f24.gif)
where the Deltas are "transfer functions" associated with each projected field (basically think of them as their radial kernels) and P_ab is the 3D power spectrum between whatever 3D quantities a and b are related to.

The idea would be to do the following:
- [x] Create a generalized power spectrum structure that represents the P_ab above.
- [ ] Create a generalized transfer function structure that represents the Deltas. This will be a generalization of the ClTracer structures we currently have.
- [ ] Implement typical use cases in CCL for both structures and ways for users to set those from python.

This is based on discussions with @elikrause , @jablazek and @tilmantroester during the Edinburgh sprint week, and they can add more info here.

jablazek:
This looks correct to me. my guess is that there are a finite number of Deltas that will appear, except it very strange situations. Of course, allowing that to be general as well is helpful.

damonge:
@jablazek @elikrause @tilmantroester I've had a first stab at point number 1 above in PR #526 
Have a look there and let me know what you think.

damonge:
Part 1 of this just got merged

PR #526, opened 2018-11-01
2D power spectrum structures

damonge:
Addresses part of #511 , associated with the generalized power spectrum structures

I'd suggest deferring the merging of this until all PRs related to the CCL paper have been merged.

At the pyhton level, the generalized structures are defined and described in [this file](https://github.com/LSSTDESC/CCL/blob/power_spectrum_structs/pyccl/p2d.py) (look at the docstrings for the time being).

The interface of these objects with the Limber integrator is described [here](https://github.com/LSSTDESC/CCL/blob/2e0c1f02f30a86a77ca4abb364293674f1b43f0d/pyccl/cls.py#L297).


tmcclintock:
Looked at the docstring. This would be a good first docstring to note whether distances are comoving or not. For instance, here, the wavenumbers are comoving, so at [this line](https://github.com/LSSTDESC/CCL/blob/50b02858d284b049c81eef0d31b889f940d47f97/pyccl/p2d.py#L25) we could write `(in units of Mpc^-1; comoving)` instead of just `(in units of Mpc^-1)`.

damonge:
@tmcclintock you'd have to be kind of evil to ask for not-comoving k, but granted, we can document that better. I wonder if there's 100 other places where the same thing happens.

tmcclintock:
@damonge of course :), but I think back to a younger me who wondered at one time "does CLASS return k in comoving or physical coordinates?" and spent an hour finding the answer. Clarity never hurt, IMO.

damonge:
No, definitely. Will do!

src/ccl_p2d.c
```diff
@@ -0,0 +1,182 @@
...
+	  *status=CCL_ERROR_SPLINE_EV;
+	  return -1;
+	}
+	pk_post+=pd*dlk*dlk*0.5;
+      }
+    }
+  }
+  else if(is_lok) {
+    pk_post=pk_pre;
+    pk_post=pk_pre;
```
elikrause: (duplicate line)

src/ccl_p2d.c
```diff
@@ -0,0 +1,182 @@
...
+    if(psp->extrap_order_hik>0) {
+      double pd;
+      double dlk=lk-lk_ev;
+      spstatus=gsl_spline2d_eval_deriv_x_e(psp->pk,lk_ev,a_ev,NULL,NULL,&pd);
+      if(spstatus) {
+	*status=CCL_ERROR_SPLINE_EV;
+	return -1;
+      }
+      pk_post+=pd*dlk;
+      if(psp->extrap_order_hik>1) {
```
elikrause: I should know this - is extrap_order_* restricted to {0,1,2}?

src/ccl_p2d.c
damonge: Well spotted

src/ccl_cls.c
```diff
@@ -887,27 +887,39 @@ static void get_k_interval(ccl_cosmology *cosmo,CCL_ClWorkspace *w,
...
   double result=0,eresult;
   double lkmin,lkmax;
+  ccl_p2d_t *psp_use;
   gsl_function F;
   gsl_integration_workspace *w=gsl_integration_workspace_alloc(ccl_gsl->N_ITERATION);
 
+  if(psp==NULL) {
+    if (!cosmo->computed_power) ccl_cosmology_compute_power(cosmo, status);
+    // Return if compilation failed
+    if (!cosmo->computed_power) return NAN;
```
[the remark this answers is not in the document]
damonge: It is handled by the python status check. However, to make the process faster I've changed things a bit so that, the first time this happens, it goes straight back to python without trying to compute anything else.

src/ccl_p2d.c
damonge: Correct. This was documented in the C docstrings, and I've now added documentation in the python functions too.

damonge:
@elikrause thanks a lot for your review!

I've added some extra documentation and warnings about power spectrum sampling (in the python docstrings).

About ccl_sigmaR etc.: now CCL stores the linear and non-linear power spectra as two internal `ccl_p2d_t` structures, which are evaluated whenever the functions `ccl_linear_matter_power` or `ccl_nonlin_matter_power` are called, so those functions will still work as before. I have not added functionality to modify the internal matter power spectrum structures manually from python, but I don't know if this is something we wanted to do (i.e. the main use of the Pk2D objects now is to pass them to the Limber integrator, but not to mass functions etc.).

elikrause:
> About ccl_sigmaR etc.: now CCL stores the linear and non-linear power spectra as two internal `ccl_p2d_t` structures, which are evaluated whenever the functions `ccl_linear_matter_power` or `ccl_nonlin_matter_power` are called, so those functions will still work as before. 

Yes, I agree that they should still work as before as it simply calls the z=0 power spectrum, and factoring in the growth. I believe this assumption in current variance calculation + halo model routines could lead to inconsistencies if a 2D power spectrum is supplied by the user, but then only used at z=0 and rescaled by growth factor in the halo calculations.

damonge:
@elikrause thanks a lot!

The user is not able to alter the internal linear and non-linear matter power spectra. They are just able to create new Pk2D objects and to pass them to the limber integrator. So I think this shouldn't lead to inconsistencies.

Now that you mention it, however, the fact that we extrapolate with linear growth the z=0 power spectrum for halo quantities is kind of problematic for e.g. cosmologies with neutrinos, so that may be something we should fix in a different PR...

elikrause:
@damonge 
> The user is not able to alter the internal linear and non-linear matter power spectra. They are just able to create new Pk2D objects and to pass them to the limber integrator. So I think this shouldn't lead to inconsistencies.

Thanks, it seems to me that we agree on the workings of CCL, just call it differently. It might be useful to comment somewhere in the description that Pk2D is *only* used by the 2-pt Limber integrator, not in the rest of the code. At least I briefly had the idea that Pk2D might be a method to do variance/halo model computations where scale and redshift dependence don't factorize. I agree now that this should be a separate PR.

I approve, but as this is a substantial PR, a second review might be useful.

include/ccl_emu17.h
```diff
@@ -11,11 +11,12 @@ CCL_BEGIN_DECLS
  * Emulator power spectrum
  * Obtain P(k,z) [Mpc^3] for a given set of input parameters.
  * @param xstarin vector of input parameters for the emulator, including redshift. 
+ * @param nkemu number of elements to which Pkemu has been allocated.
```
beckermr: What does this mean?

include/ccl_params.h
```diff
@@ -44,6 +44,11 @@ typedef struct ccl_spline_params {
 
 extern ccl_spline_params * ccl_splines;
 
+int ccl_get_pk_spline_na(void);
+int ccl_get_pk_spline_nk(void);
+void ccl_get_pk_spline_a_array(int ndout,double* doutput,int *status);
+void ccl_get_pk_spline_lk_array(int ndout,double* doutput,int *status);
```
beckermr: Why are these prototypes here and not in the 2D power spectrum header file?

src/ccl_power.c
```diff
@@ -406,194 +443,125 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
-    ccl_free_class_structs(cosmo, &ba,&th,&pt,&tr,&pm,&sp,&nl,&le,init_arr,status);
-    
-    // free the parser
-    parser_free(&fc);
-    return;
+  if (*status == 0) {
+    if (parser_free(&fc)== _FAILURE_) {
+      *status = CCL_ERROR_CLASS;
+      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error freeing CLASS parser\n");
+    }
```
beckermr: I don't see the place where the parser is freed if status is non-zero. Did we reintroduce this memory leak?

include/ccl_emu17.h
damonge: It's the size of the next argument (`Pkemu`)

src/ccl_power.c
damonge: Right, the whole file has been refactored (that's why I didn't push to solving this earlier, since I thought this would be fixed when this got merged). This is now done in https://github.com/LSSTDESC/CCL/blob/3a3f5e9a2194465312e03b7782134ea0876eb07c/src/ccl_power.c#L446

include/ccl_params.h
damonge: These functions are basically a patch to grab the internal CCL spline parameters and make them visible to python (essentially this gets superseded by your PR, I think). Because they just access the contents of `ccl_spline_params`, I think it's more appropriate for them to be in ccl_params.h

include/ccl_emu17.h
beckermr: Ahhh. The doc string confused me.

include/ccl_params.h
beckermr: Got it. In my PR to make the spline parameters changeable, I moved these structs to `ccl_core.h` and deleted the `ccl_params.h` file. Could we move these prototypes and the function defs to `ccl_core.h` and `ccl_core.c` to avoid the merge conflicts?

pyccl/cls.py
```diff
@@ -316,18 +328,18 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell,
...
+            cosmo, clt1, clt2, psp, l_limber, l_logstep,
             l_linstep, [ell], 1, status)
         cl = cl_one[0]
     elif isinstance(ell, np.ndarray):
         # Use vectorised function
         cl, status = lib.angular_cl_vec(
-            cosmo, clt1, clt2, l_limber, l_logstep,
+            cosmo, clt1, clt2, psp, l_limber, l_logstep,
             l_linstep, ell, ell.size, status)
     else:
```
beckermr: While you are here, using `np.atleast_1d` will cover all three of these cases in this if else block

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+# TODO choices about interpolation/extrapolation
+
+
+class Pk2D(object):
+    """A power spectrum class holding the information needed to reconstruct an
+    arbitrary function of wavenumber and scale factor.
+    """
+    def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
+                 is_logp=True, interp_order_lok=1, interp_order_hik=2):
+        """Constructor for Pk2D objects.
```
beckermr: Doc strings for the init go in the class level doc string, not on the init.

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+                  above the maximum of the splines (use 0, 1 or 2).
+            is_logp (boolean): if True, pkfunc/pkarr return/hold the natural
+                  logarithm of the power spectrum. Otherwise, the true value
+                  of the power spectrum is expected.
+        """
+        status = 0
+        if(pkfunc is None):  # Initialize power spectrum from 2D array
+            # Make sure input makes sense
+            if (a_arr is None) or (lk_arr is None) or (pk_arr is None):
+                raise TypeError("If you do not provide a function, "
```
beckermr: This should be a ValueError.

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+            pkflat = pk_arr.flatten()
+            # Check dimensions make sense
+            if (len(a_arr)*len(lk_arr) != len(pkflat)):
+                raise ValueError("Size of input arrays is inconsistent")
+        else:  # Initialize power spectrum from function
+            # Check that the input function has the right signature
+            try:
+                pkfunc(k=np.array([1E-2, 2E-2]), a=0.5)
+            except Exception:
+                raise TypeError("Can't use input function")
```
beckermr: This is also a ValueError.

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+        Returns:
+            float or array_like: value(s) of the power spectrum.
+        """
+        status = 0
+        if cosmo is not None:
+            cospass = cosmo.cosmo
+        else:
+            raise NotImplementedError("Currently we need a cosmology to "
+                                      "extrapolate growth")
+            cospass = None
```
beckermr: This block means you have to pass a cosmology even though the parameter argument is optional.

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+        elif isinstance(k, np.ndarray):
+            f, status = lib.p2d_eval_multi(self.psp, np.log(k), a, cospass,
+                                           k.size, status)
+        else:
+            f, status = lib.p2d_eval_multi(self.psp, np.log(k), a, cospass,
+                                           len(k), status)
+        check(status, cosmo)
+
+        return f
+        raise NotImplementedError("Not implemented yet")
```
beckermr: Remove this line here.

pyccl/p2d.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if isinstance(k, int):
+            k = float(k)
+        if isinstance(k, float):
+            f, status = lib.p2d_eval_single(self.psp, np.log(k), a, cospass,
+                                            status)
+        elif isinstance(k, np.ndarray):
+            f, status = lib.p2d_eval_multi(self.psp, np.log(k), a, cospass,
+                                           k.size, status)
+        else:
```
beckermr: again `np.atleast_1d` might cover some of these cases.

src/ccl_emu17.c
```diff
@@ -138,21 +138,30 @@ static void emuInit() {
 } // emuInit()
 
 // Actual emulation
-void ccl_pkemu(double *xstar, double **ystar, int* status, ccl_cosmology* cosmo) {
-    
+void ccl_pkemu(double *xstar, int sizeofystar, double *ystar, int* status, ccl_cosmology* cosmo)
+{
```
beckermr: Why the interface change here?

src/ccl_p2d.c
```diff
@@ -0,0 +1,181 @@
...
+  int is_loz= a>psp->amax;
+
+  if(is_loz) { //Are we above the interpolation range in a?
+    *status=CCL_ERROR_SPLINE_EV;
+    return -1;
+  }
+  else if(is_hiz) { //Are we below the interpolation range in a?
+    if(psp->extrap_linear_growth==ccl_p2d_no_extrapol) {
+      *status=CCL_ERROR_SPLINE_EV;
+      return NAN;
```
beckermr: Why does one statement return -1 and the other NaN?

src/ccl_p2d.c
```diff
@@ -0,0 +1,181 @@
...
+
+double ccl_p2d_t_eval(ccl_p2d_t *psp,double lk,double a,ccl_cosmology *cosmo,
+		      int *status)
+{
+  double a_ev=a;
+  int is_hiz= a<psp->amin;
+  int is_loz= a>psp->amax;
+
+  if(is_loz) { //Are we above the interpolation range in a?
+    *status=CCL_ERROR_SPLINE_EV;
```
beckermr: Do you not need an extrapolation check here to like the one below

src/ccl_p2d.c
```diff
@@ -0,0 +1,181 @@
...
+  if(is_hik) //Are we above the interpolation range in k?
+    lk_ev=psp->lkmax;
+  else if(is_lok) //Are we below the interpolation range in k?
+    lk_ev=psp->lkmin;
+
+  //Evaluate spline
+  int spstatus=gsl_spline2d_eval_e(psp->pk,lk_ev,a_ev,NULL,NULL,&pk_pre);
+  if(spstatus) {
+    *status=CCL_ERROR_SPLINE_EV;
+    return -1;
```
beckermr: Is this -1 or NaN?

src/ccl_params.c
```diff
@@ -0,0 +1,66 @@
...
+    d=ccl_log_spacing(ccl_splines->K_MIN,ccl_splines->K_MAX,ndout);
+    if(d==NULL)
+      *status=CCL_ERROR_MEMORY;
+  }
+  if(*status==0) {
+    for(int ii=0;ii<ndout;ii++)
+      doutput[ii]=log(d[ii]);
+  }
+  free(d);
+}
```
beckermr: Let's move these functions to `ccl_core.c`

tests/ccl_test_correlation_3d.c
```diff
@@ -83,9 +83,9 @@ static void compare_correlation_3d(int i_model,struct corrs_3d_data * data)
   nr=linecount(f)-1; rewind(f);
 
   // tolerence on abs difference in r^2 xi(r) for the range r = 0.1 - 100 Mpc (40 points in r) for z=0,1,2,3,4,5
-  double CORR_TOLERANCE1[6] = {2.84e-2, 3.57e-3, 1.37e-3, 1.26e-3, 2.0e-4, 1.8e-4};
+  double CORR_TOLERANCE1[6] = {4.0e-2, 1.0e-2, 5.0e-3, 4.0e-3, 1.0e-3, 8.0e-4};
   // tolerence on abs difference in r^2 xi(r) for the range r = 50 - 250 Mpc (100 points in r) for z=0,1,2,3,4,5
-  double CORR_TOLERANCE2[6] = {3.15e-2, 9.87e-3, 5.04e-3, 3.05e-3, 9.98e-4, 7.14e-4};
+  double CORR_TOLERANCE2[6] = {4.0e-2, 1.0e-2, 5.0e-3, 4.0e-3, 1.0e-3, 8.0e-4};
```
beckermr: Should we be concerned about these tolerances going up?

tests/ccl_test_p2d.c
```diff
@@ -0,0 +1,172 @@
...
+  ASSERT_DBL_NEAR(-1.,pk);
+  status=0;
+
+  //Now put some sensible numbers within the redshift and k range
+  pk=ccl_p2d_t_eval(psp,lktest,atest,NULL,&status); 
+  ASSERT_TRUE(status==0);
+  ASSERT_DBL_NEAR(1,pk/pk_model_analytical(exp(lktest),atest));
+
+  //Evaluate at very low z and see if it checks out
+  double alo=0.02;
```
beckermr: Lowz is a ~ 1.

tests/ccl_test_p2d.c
```diff
@@ -0,0 +1,172 @@
...
+  ASSERT_DBL_NEAR(1,pk/pk_model_analytical(exp(lktest),alo));
+
+  //Evaluate at very high k and see if it checks out
+  double lkhi=data->lk_arr[data->n_k-1]*1.1;
+  pk=ccl_p2d_t_eval(psp,lkhi,atest,NULL,&status);
+  ASSERT_TRUE(status==0);
+  ASSERT_DBL_NEAR(1,pk/pk_model_analytical(exp(lkhi),atest));
+
+  //Evaluate at very high k and see if it checks out
+  double lklo=data->lk_arr[0]*1.1;
```
beckermr: Do you need to divide by 1.1 here?

tests/ccl_test_p2d.py
```diff
@@ -0,0 +1,108 @@
...
+def all_finite(vals):
+    """
+    Returns True if all elements are finite (i.e. not NaN or inf).
+    """
+    return np.all( np.isfinite(vals) )
+
+def check_p2d_init():
+    """
+    Check that background and growth functions can be run.
+    """
```
beckermr: Why is this function empty?

src/ccl_power.c
```diff
@@ -406,194 +443,123 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
-  //If error, store status, we will free later
-  if (a==NULL|| x==NULL || y2d_lin==NULL || y2d_nl==NULL) {
-    *status = CCL_ERROR_SPLINE;
-    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): memory allocation error\n");
+  double kmin,kmax,ndecades,amin,amax,ic;
+  int nk,na,s;
+  double *x=NULL, *z=NULL, *y2d_lin=NULL, *y2d_nl=NULL;
+  if (*status == 0) {
+    //CLASS calculations done - now allocate CCL splines
+    kmin = 2*exp(sp.ln_k[0]);
```
beckermr: Why the factor of 2 here?

src/ccl_power.c
```diff
@@ -406,194 +443,123 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+    kmax = ccl_splines->K_MAX_SPLINE;
```
beckermr: Why are we not pulling kmax from the CLASS spectra?

src/ccl_power.c
beckermr: Or should the min be the CCL value?

src/ccl_power.c
```diff
@@ -406,194 +443,123 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+  if(*status==0)
+    cosmo->data.p_lin=ccl_p2d_t_new(na,z,nk,x,y2d_lin,1,2,ccl_p2d_cclgrowth,1,NULL,0,ccl_p2d_3,status);
 
+  if(*status==0) {
+    // At the moment KMIN can't be less than CLASS's kmin in the nonlinear case.
+    if (kmin<(exp(sp.ln_k[0]))) {
+      *status = CCL_ERROR_CLASS;
+      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): K_MIN is less than CLASS's kmin. Not yet supported for nonlinear P(k).\n");
+    }
   }
```
beckermr: Why are testing this if we set it above?

src/ccl_power.c
```diff
@@ -873,120 +836,98 @@ static void ccl_cosmology_compute_power_eh(ccl_cosmology * cosmo, int * status)
...
-    free(eh); free(x); free(y); free(a); free(y2d);
-    gsl_spline2d_free(log_power_lin);
-    gsl_spline2d_free(log_power_nl);
-    *status = CCL_ERROR_SPLINE;
-    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_eh(): Error creating log_power_nl spline\n");
-    return;
+  
+  if(*status==0) {
+    if(cosmo->config.baryons_power_spectrum_method==ccl_bcm)
+      correct_bcm(cosmo,na,z,nk,x,y2d,status);
```
beckermr: are we applying BCM to the linear power spectrum here?

src/ccl_power.c
```diff
@@ -1037,130 +976,100 @@ static void ccl_cosmology_compute_power_bbks(ccl_cosmology * cosmo, int * status
...
-      cosmo->data.p_nl = log_power_nl;
-    } else {
-      gsl_spline2d_free(log_power_nl);
-      *status = CCL_ERROR_SPLINE;
-      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_bbks(): Error creating log_power_nl spline\n");
-    }
+  
+  if(*status==0) {
+    if(cosmo->config.baryons_power_spectrum_method==ccl_bcm)
+      correct_bcm(cosmo,na,z,nk,x,y2d,status);
```
beckermr: Again we should not be applying the BCM to the linear P(k).

src/ccl_power.c
```diff
@@ -1238,116 +1135,150 @@ static void ccl_cosmology_compute_power_emu(ccl_cosmology * cosmo, int * status)
...
+    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_emu(): sigma8 is not defined; specify sigma8 instead of A_s\n");
+    return;
   }
-  
-  if (!*status){
+
+  // Prepare to run CLASS for linear scales
+  ccl_fill_class_parameters(cosmo,&fc,parser_length, status);
+
+  if (*status != CCL_ERROR_CLASS)
```
beckermr: Random question. Does the emulator not have a linear prediction to go with it?

src/ccl_power.c
```diff
@@ -1238,116 +1135,150 @@ static void ccl_cosmology_compute_power_emu(ccl_cosmology * cosmo, int * status)
...
-      *status = CCL_ERROR_CLASS;
-      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error freeing CLASS parser\n");
+  double kmin,kmax,ndecades,amin,amax,ic;
+  int nk,na,s;
+  double *x=NULL,*z=NULL,*y2d_lin=NULL;
+  double *y=NULL,*logx=NULL,*xstar=NULL,*zemu=NULL,*y2d_nl=NULL;
+  if (*status == 0) {
+    //CLASS calculations done - now allocate CCL splines
+    kmin = 2*exp(sp.ln_k[0]);
+    kmax = ccl_splines->K_MAX_SPLINE;
```
beckermr: Same questions about the lower K limit.

src/ccl_power.c
```diff
@@ -1358,47 +1289,46 @@ static void ccl_cosmology_compute_power_emu(ccl_cosmology * cosmo, int * status)
...
-	ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_emu(): Error obtaining emulator predictions\n");
+      ccl_pkemu(xstar,NK_EMU,y, status, cosmo);
+      ccl_check_status(cosmo, status);
+      if (y == NULL) {
+	*status=CCL_ERROR_MEMORY;
+	ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_emu(): memory allocation error\n");
+	break;
       }
       for (int i=0; i<NK_EMU; i++){
 	logx[i] = log(mode[i]);
```
beckermr: I cannot find where `mode` is defined. Is this a global var? Does it need a more descriptive name?

src/ccl_emu17.c
damonge: Before this routine set `*ystar` to `NULL` to mark when it failed, which could create memory leaks. This is now handled through the `status` flag, and I pass the size of `ystar` to add an extra check that it has been correctly allocated.

tests/ccl_test_correlation_3d.c
damonge: These were set originally to **exactly** the tolerances that were achieved at the time. When the P(k) splines change, they move up and down slightly, so I've just made them all larger to allow future changes in the P(k) part from affecting this test. 
We're trying to test the P(k) -> xi(r) code here, so this test should be robust against changes in the P(k) calculation.

src/ccl_power.c
damonge: I'd point you to Elisabeth or Elisa (they implemented the CLASS interface, and this particular line hasn't been changed in this PR). I imagine this is to make sure that we don't try to evaluate things below the minimum k used by CLASS.

src/ccl_power.c
damonge: Again, Elisabeth and Elisa are the right people to ask here.

src/ccl_power.c
damonge: We only apply it to whatever array is gonna end up in the non-linear power spectrum spline

src/ccl_power.c
damonge: Agreed :-)

src/ccl_power.c
damonge: Nope, we use CLASS

src/ccl_power.c
damonge: Fixed in last commit

src/ccl_power.c
damonge: No idea. This is a leftover from the previous code. Removed in my last commit

src/ccl_power.c
damonge: `include/ccl_emu17_params.h`. It is a global variable indeed (which hasn't been added in this PR)

damonge:
@beckermr I've made the following modifications to (mostly) ccl_power.c:
- Simplified the code so that BBKS and EH power spectra are now computed with the same method (so we don't have to repeat all the splining code)
- Simplified the emulator pk calculator to avoid recalculating a bunch of things and to avoid rewriting the CLASS splining code.
- Simplified the CLASS interface a bit
- Changed some of the variable names to make them more transparent
- Fixed the bug in the BBKS transfer function noted in #591 

I'll move on to solving all conflicts with the master branch and then to simplifying the C_ell interface to address #553

include/ccl_params.h
damonge: Done in my last commit

src/ccl_params.c
damonge: Done in my last commit

pyccl/cls.py
damonge: Done in last commit. Note that there's a lot of these in most of the python functions.

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Made non-optional

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed in last commit

beckermr:
@damonge. Looking great! Let’s address #553 in another PR.

src/ccl_power.c
damonge: Done in last commit

src/ccl_p2d.c
damonge: We do not have a method to extrapolate below the maximum redshift of the spline. The logic is that it may be safe to assume linear growth at high-redshift, but not at low redshift. So in this case we just return with an error.

src/ccl_p2d.c
damonge: Homogenized to NAN

src/ccl_p2d.c
damonge: Homogenized to NAN

tests/ccl_test_p2d.c
damonge: Correct! Fixed

tests/ccl_test_p2d.py
damonge: Sorry, some residuals from copy-paste. Removed

tests/ccl_test_p2d.c
damonge: Yes, sorry, this was meant to read high z

damonge:
OK, great @beckermr !
I think I've addressed all of the comments, so happy to have another review whenever you have time

src/ccl_p2d.c
beckermr: got it.

include/ccl_params.h
beckermr: These prototypes are still here. Let's delete the params.h file and just use core.h.

tests/ccl_test_p2d.c
beckermr: cool. Let's change the comment in the code.

include/ccl_core.h
```diff
@@ -11,6 +11,37 @@
...
+  double amin,amax; /**< Edges in a*/
+  int extrap_order_lok; /**< Order of extrapolating polynomial in log(k) for low k (0, 1 or 2)*/
+  int extrap_order_hik; /**< Order of extrapolating polynomial in log(k) for high k (0, 1 or 2)*/
+  ccl_p2d_extrap_growth_t extrap_linear_growth;  /**< Extrapolation type at high redshifts*/
+  int is_log; /**< Do I hold the values of log(P(k,a))?*/
+  double (*growth)(double); /**< Custom extrapolating growth function*/
+  double growth_factor_0; /**< Constant extrapolating growth factor*/
+  gsl_spline2d *pk; /**< Spline holding the values of P(k,a)*/
+} ccl_p2d_t;
+
```
beckermr: Whoops. I think we might have gotten confused between various comments. These should be in `ccl_ps2d.h` right? I only meant the ones you moved below.

pyccl/ccl_p2d.i
```diff
@@ -0,0 +1,52 @@
...
+%inline %{
+ccl_p2d_t *set_p2d_new_from_arrays(double* lkarr,int nk,
+				   double* aarr,int na,
+				   double* pkarr,int npk,
+				   int order_lok,int order_hik,
+				   int is_logp,
+				   int *status)
+{
+  ccl_p2d_t *psp=ccl_p2d_t_new(na,aarr,nk,lkarr,pkarr,
+			       1,2,ccl_p2d_cclgrowth,
```
beckermr: You didn't pass the extrapolation orders here.

pyccl/cls.py
```diff
@@ -292,6 +293,8 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell,
         cltracer1, cltracer2 (:obj:`Tracer`): Tracer objects, of any kind.
         ell (float or array_like): Angular wavenumber(s) at which to evaluate
             the angular power spectrum.
+        p_of_k_a (Pk2D object or None): 3D Power spectrum to project. If None,
```
beckermr: We should use the correct sphinx syntax to link things here.

pyccl/cls.py
```diff
@@ -308,27 +311,27 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell,
...
-            l_linstep, ell, ell.size, status)
-    else:
-        # Use vectorised function
-        cl, status = lib.angular_cl_vec(
-            cosmo, clt1, clt2, l_limber, l_logstep,
-            l_linstep, ell, len(ell), status)
+    cl, status = lib.angular_cl_vec(
+        cosmo, clt1, clt2, psp, l_limber, l_logstep,
+        l_linstep, ell_use, ell_use.size, status)
+    if ell_use.size==1:
```
beckermr: We can't test like this since a 1d array will always get coerced to a scalar. You can use `np.isscalar` and test `ell` directly.

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
+from . import ccllib as lib
+
+from .core import check
+import numpy as np
+
+# TODO choices about interpolation/extrapolation
```
beckermr: This comment can be removed.

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
...
+        cosmo (:obj:`Cosmology`): Cosmology object. The cosmology object
+             is needed in order if `pkfunc` is not `None`. The object is
+             used to determine the sampling rate in scale factor and
+             wavenumber.
+    """
+    def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
+                 is_logp=True, interp_order_lok=1, interp_order_hik=2, cosmo=None):
+
+        status = 0
+        if(pkfunc is None):  # Initialize power spectrum from 2D array
```
beckermr: no parens needed here

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
...
+        status = 0
+        cospass = cosmo.cosmo
+
+        if isinstance(k, int):
+            k = float(k)
+        if isinstance(k, float):
+            f, status = lib.p2d_eval_single(self.psp, np.log(k), a, cospass,
+                                            status)
+        else:
+            k_use=np.atleast_1d(k)
```
beckermr: space around `=` here.

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
...
+                                           k_use.size, status)
+        check(status, cosmo)
+
+        return f
+
+    def __del__(self):
+        """Free memory associated with this Pk2D structure
+        """
+        if hasattr(self, 'has_psp'):
+            if self.has_psp:
```
beckermr: you might want to check directly on psp argument too? I'm not sure.

src/ccl_cls.c
```diff
@@ -451,7 +451,7 @@ static void clt_init_wL(CCL_ClTracer *clt,ccl_cosmology *cosmo,
       ccl_cosmology_set_status_message(cosmo, "ccl_cls.c: clt_init_wL(): error computing lensing window\n");
     }
   }
-
+
```
beckermr: i'd be nice to avoid these whitespace changes.

src/ccl_p2d.c
```diff
@@ -0,0 +1,181 @@
...
+      psp->pk=NULL;
+    }
+    if(psp->pk==NULL)
+      *status = CCL_ERROR_MEMORY;
+  }
+
+  if(*status==0) {
+    s2dstatus=gsl_spline2d_init(psp->pk,lk_arr,a_arr,pk_arr,nk,na);
+    if(s2dstatus)
+      *status = CCL_ERROR_SPLINE;
```
beckermr: If we fail here, we should probably free the memory associated with `psp->pk`.

tests/ccl_test_correlation_3d.c
beckermr: Right. This test is pretty useless TBH.

src/ccl_power.c
```diff
@@ -33,17 +70,15 @@ static void ccl_free_class_structs(ccl_cosmology *cosmo,
   if(init_arr[i_init--]) {
     if (spectra_free(sp) == _FAILURE_) {
       *status = CCL_ERROR_CLASS;
-      ccl_cosmology_set_status_message(
-        cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS spectra:%s\n", sp->error_message);
+      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS spectra:%s\n", sp->error_message);
```
beckermr: These long lines were changed to be easier to read for a reason. Let's put them back.

src/ccl_power.c
```diff
@@ -33,17 +70,15 @@ static void ccl_free_class_structs(ccl_cosmology *cosmo,
...
       return;
     }
   }
 
   if(init_arr[i_init--]) {
     if (transfer_free(tr) == _FAILURE_) {
       *status = CCL_ERROR_CLASS;
-      ccl_cosmology_set_status_message(
-        cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS transfer:%s\n", tr->error_message);
+      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS transfer:%s\n", tr->error_message);
```
beckermr: Ditto here.

src/ccl_power.c
```diff
@@ -75,8 +110,7 @@ static void ccl_free_class_structs(ccl_cosmology *cosmo,
   if(init_arr[i_init--]) {
     if (thermodynamics_free(th) == _FAILURE_) {
       *status = CCL_ERROR_CLASS;
-      ccl_cosmology_set_status_message(
-        cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS thermo:%s\n", th->error_message);
+      ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_free_class_structs(): Error freeing CLASS thermo:%s\n", th->error_message);
```
beckermr: Ditto here.

src/ccl_power.c
```diff
@@ -180,56 +214,48 @@ static void ccl_run_class(ccl_cosmology *cosmo,
 
   if(input_init(fc,pr,ba,th,pt,tr,pm,sp,nl,le,op,errmsg) == _FAILURE_) {
     *status = CCL_ERROR_CLASS;
-    ccl_cosmology_set_status_message(
-      cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS input:%s\n", errmsg);
+    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS input:%s\n", errmsg);
```
beckermr: Ditto here.

src/ccl_power.c
```diff
@@ -180,56 +214,48 @@ static void ccl_run_class(ccl_cosmology *cosmo,
...
     return;
   }
   if (background_init(pr,ba) == _FAILURE_) {
     *status = CCL_ERROR_CLASS;
-    ccl_cosmology_set_status_message(
-      cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS background:%s\n", ba->error_message);
+    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS background:%s\n", ba->error_message);
```
beckermr: and here

src/ccl_power.c
```diff
@@ -180,56 +214,48 @@ static void ccl_run_class(ccl_cosmology *cosmo,
...
     return;
   }
   init_arr[i_init++]=1;
   if (thermodynamics_init(pr,ba,th) == _FAILURE_) {
     *status = CCL_ERROR_CLASS;
-    ccl_cosmology_set_status_message(
-      cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS thermodynamics:%s\n", th->error_message);
+    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS thermodynamics:%s\n", th->error_message);
```
beckermr: and here

src/ccl_power.c
```diff
@@ -180,56 +214,48 @@ static void ccl_run_class(ccl_cosmology *cosmo,
...
     return;
   }
   init_arr[i_init++]=1;
   if (perturb_init(pr,ba,th,pt) == _FAILURE_) {
     *status = CCL_ERROR_CLASS;
-    ccl_cosmology_set_status_message(
-      cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS pertubations:%s\n", pt->error_message);
+    ccl_cosmology_set_status_message(cosmo, "ccl_power.c: ccl_cosmology_compute_power_class(): Error running CLASS pertubations:%s\n", pt->error_message);
```
beckermr: I;m not going to mark the rest. If the code cannot fit into a 13 inch laptop screen in regular sized font on github, then it needs to be broken over multiple lines.

src/ccl_power.c
```diff
@@ -407,277 +431,130 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
-    return;
   }
 
-  ccl_fill_class_parameters(cosmo,&fc,parser_length, status);
+  if(*status==0)
+    ccl_fill_class_parameters(cosmo,&fc,parser_length, status);
 
-  if (*status != CCL_ERROR_CLASS)
+  if (*status==0) {
+    init_parser=1;
```
beckermr: The way this is setup if we fail to fill the class parameters, the parser will not be freed.

src/ccl_power.c
beckermr: @elisachisari @elikrause can you answer the question here?

src/ccl_power.c
```diff
@@ -407,277 +431,130 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
...
+    kmin = 2*exp(sp.ln_k[0]);
+    kmax = cosmo->spline_params.K_MAX_SPLINE;
+    //Compute nk from number of decades and N_K = # k per decade
+    ndecades = log10(kmax) - log10(kmin);
+    nk = (int)ceil(ndecades*cosmo->spline_params.N_K);
+    amin = cosmo->spline_params.A_SPLINE_MINLOG_PK;
+    amax = cosmo->spline_params.A_SPLINE_MAX;
+    na = cosmo->spline_params.A_SPLINE_NA_PK+cosmo->spline_params.A_SPLINE_NLOG_PK-1;
+
+    // The x array is initially k, but will later
```
beckermr: comment out of date.

src/ccl_power.c
```diff
@@ -13,6 +13,43 @@
...
+static void correct_bcm(ccl_cosmology *cosmo,
+			int na,double *a_arr,int nk,double *lk_arr,double *pk2d,
+			int *status)
+{
+  for(int ii=0;ii<na;ii++) {
+    double a=a_arr[ii];
+    for(int jj=0;jj<nk;jj++) {
+      double k=exp(lk_arr[jj]);
+      double fbcm=ccl_bcm_model_fka(cosmo,k,a,status);
+      pk2d[jj+nk*ii]+=log(fbcm);
```
beckermr: So the `correct_bcm` method applies corrections to its pk2d input.

src/ccl_power.c
beckermr: Wow ok. So actually if you specify EH98 for the transfer function, you get that for the non-linear and linear power?

beckermr: @damonge pinging you to make sure you see this.

include/ccl_core.h
damonge: so a `ccl_cosmology` contains `ccl_p2d_t` structs, and at the same time the functions that use `ccl_p2d_t` structs also use ccl_cosmology objects. If I put the definition of `ccl_p2d_t` in `ccl_p2d.h` I get a circular dependence between `ccl_core.h` and `ccl_p2d.h`, which is very annoying

beckermr: ah shit. well maybe dump the extra header file or just leave it.

beckermr:
flake8 is going to choke on this python. I suggest you run that locally and clean it up

src/ccl_cls.c
damonge: sorry, my emacs sometimes is a bit stupid. Fixed

src/ccl_p2d.c
damonge: mmh, this will be freed up when `psp` is freed up, right?

src/ccl_power.c
damonge: correct, but we only pass pk2d holding non-linear power spectra to it.

src/ccl_power.c
damonge: updated

src/ccl_power.c
damonge: Argh, sorry. Good catch. This was meant to go in the previous block

src/ccl_power.c
damonge: Done

src/ccl_power.c
damonge: Done

src/ccl_power.c
damonge: Done

src/ccl_power.c
damonge: Done

src/ccl_power.c
damonge: Done

src/ccl_power.c
damonge: Ditto too :-)

src/ccl_power.c
damonge: I've broken up several of them.

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed

pyccl/p2d.py
damonge: Fixed

pyccl/cls.py
damonge: Fixed

pyccl/cls.py
damonge: Fixed

pyccl/ccl_p2d.i
damonge: Argh, stupid me. Fixed

damonge:
@beckermr : thanks again and back to you!

src/ccl_p2d.c
beckermr: True. Not clear to me the caller will bother if there is an error, but you are correct.

src/ccl_power.c
beckermr: still needs to be changed. I see a refresh at the top. I will check back in a bit.

src/ccl_p2d.c
damonge: I don't see how this would lead to a leak (I may be missing something): psp is freed up from python as usual, and this function makes sure that psp->pk is either NULL or has been allocated.

src/ccl_power.c
damonge: Mmh, I've changed this in the last commit

damonge:
>The drama over flake8 is really not needed. 

Don't mind me, I just enjoy moaning

>One question. Is it possible to mix the EH98 model for the linear power and the CosmicEMU for the non-linear power?

Currently no, I'm afraid.

I think I fixed the leak, but let me know if there are still issues.

src/ccl_p2d.c
beckermr: yep all true - mostly thinking about usage here. resolving

src/ccl_power.c
beckermr: huh? This line is still rather long.

src/ccl_power.c
beckermr: discussed below.

beckermr:
OK. LGTM!

damonge:
Great! I'll wait for travis, and maybe we'll do it live during the telecon later

src/ccl_power.c
damonge: Sorry, didn't see this comment:

>So actually if you specify EH98 for the transfer function, you get that for the non-linear and linear power?

Yes, that's correct

beckermr:
Yup. This is going to make @philbull's PR out of date, yet again. Merging PRs faster and not letting them linger will help with this.

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
...
+             hold the values of the natural logarithm of the power
+             spectrum, depending on the value of `is_logp`. If `pkfunc`
+             is not None, then `a_arr`, `lk_arr` and `pk_arr` are ignored.
+             However, either `pkfunc` or all of the last three array must
+             be non-None. Note that, if you pass your own Pk array, you
+             are responsible of making sure that it is sufficiently well
+             sampled (i.e. the resolution of `a_arr` and `lk_arr` is high
+             enough to sample the main features in the power spectrum).
+             For reference, CCL will use bicubic interpolation to evaluate
+             the power spectrum at any intermediate point in k and a.
```
jablazek: Any choice on interpolation?

pyccl/p2d.py
```diff
@@ -0,0 +1,142 @@
...
+        interp_order_lok (int): extrapolation order to be used on k-values
+             below the minimum of the splines (use 0, 1 or 2).
+        interp_order_hik (int): extrapolation order to be used on k-values
+             above the maximum of the splines (use 0, 1 or 2).
```
jablazek: extrapolation is done fitting a polynomial. Is this in P-k-a space, or logP-logk-loga space? Or is this determined by 'islogP'?

pyccl/p2d.py
damonge: No, bicubic is enforced. We could revisit later on

pyccl/p2d.py
damonge: It's determined by `islogP` precisely to avoid that problem. I'll document better.

damonge:
@jablazek : Angular power spectra now take in an arbitrary P(k,a) as a Pk2D object (or use the default non-linear matter power spectrum if you pass `None`). The next step will be to define generic "transfer functions" or "tracers" (haven't decided on the name yet) that can then be used to compute any C_ell you want. That'll be in a separate PR though

pyccl/p2d.py
jablazek: So if I want log-log interpolation to be done, I have to pass in all log values at the start? This is fine, but should be well documented. (It's the sort of thing I have messed up in the past.)

damonge: Is that not clear enough from the description of `is_logp` ?

jablazek: What is expected is indeed clear. The fact that log interpolation can only be done this way is less clear.

damonge: OK, just clarified that.

PR #531, opened 2018-11-23
Cosmic shear and GGL benchmarks including intrinsic alignments

sjoudaki:
New benchmark files added generated from CosmoLSS, and expanded the c unit tests + jupyter notebook. This provides a self-consistent set of comparisons for shear-shear, shear-intrinsic, intrinsic-intrinsic, galaxy-shear, and galaxy-intrinsic for both histogram and analytic n(z), for all tomographic bin combinations.


include/ccl_constants.h
```diff
@@ -115,7 +115,8 @@
 /**
  * Default relative precision if not otherwise specified
  */
-#define GSL_EPSREL 1E-4
+#define GSL_EPSREL 2.5E-5
```
damonge: I'm worried about making CCL slow just to pass this test. Could we make this a tunable parameter through the params.ini file? I'm happy to implement this for you.

tests/ccl_test_correlation.c
```diff
@@ -1,13 +1,15 @@
 #include "ccl.h"
+#include "../include/ccl_params.h"
```
beckermr: Is this include actually needed? I thought this header was in ccl.h?

Issue #539, opened 2019-01-05
CCL consistently segfaults (through CLASS) at NERSC


damonge:
If you install CCL at NERSC following the instructions in https://confluence.slac.stanford.edu/display/~heather/Building+CCL+at+NERSC (which are quite outdated), it segfaults whenever it tries to compute the CLASS power spectrum.

I've managed to track this down to somewhere within `evolver_ndf15`, in `tools/evolver_ndf15`, but then I ran out of stamina.

damonge:
Alright, further on this: the issue seems to be caused by the `-ffast-math` compiler flag used by CLASS by default, coupled with the version of gcc that NERSC's `cc` uses.

I'll investigate whether these options would be easily tuneable through cmake, but more savvy people (@EiffL ?) may have other suggestions...

damonge:
Currently addressed by PR #540 . Will close when merged.

rainwoodman:
I am curious -- were you using the gcc 7.3 (module list gcc will show)?

EiffL:
I looked at the PR, it looks good, but I'm very confused as to why this is happening, it used to work fine at NERSC. I tried to log in and try myself today, only to be reminded about this MFA business that I have had no patience for so far... I have a bunch of deadlines coming up, not sure I'll reset my NERSC access in a least a week time-scale.
I would usually recommend pushing this as an issue upstream (to the CLASS repo), but if this issue is a show stopper for you, I'm ok with merging your PR.

damonge:
This has been solved by #540 , so I'll close it, but I wonder if this has been communicated to the CLASS bunch

PR #540, opened 2019-01-06
Avoid -ffast-math when compiling CLASS

damonge:
In some systems (critically NERSC) the `-ffast-math` compiler flag used to compile CLASS will cause segfaults. This PR adds a `cmake` flag to avoid this issue.

While addressing this, I have also realized that the NERSC instructions on the wiki are outdated, as is [this page](https://confluence.slac.stanford.edu/display/~heather/Building+CCL+at+NERSC). I'll update them after this gets merged.

This PR addresses #539

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -187,6 +187,10 @@ export CC=gcc
 export CLASS_PARAM_DIR=your_ccl_path/CCL/build/extern/share/class/
 ```
 or add this to your `.bashrc`.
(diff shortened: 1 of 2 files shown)


beckermr:
Fwiw linking libraries compiled with fast-math to python / numpy has undesireable side effects. I found this out when building the conda forge package for camb. I suspect on the whole this is a good thing.

PR #544, opened 2019-01-24
Compiles CCL for version of  python used to call pip or setup.py

### LINKED ISSUE #541: Problem with python3 install
I am currently having trouble installing `pyccl` using `python3`, but it works fine with `python2`. I feel that this is probably a problem with my `python`, rather than with `ccl`, but I figured I'd open an issue and hope that someone took pity on me:

If I do: 

`python2 setup.py install`, then leave the `ccl` directory and then do `python2 -c 'import pyccl'` I get no errors and everything works fine.

However, if I do:

`python3 setup.py install`, then leave the `ccl` directory and then do `python3 -c 'import pyccl'` I get `ModuleNotFoundError: No module named '_ccllib'`. So, it certainly installs and is able to find the library, but for some reason it doesn't work. The full error is below. Any help would be much appreciated.

```
Mead:Mead/Physics> python3 -c 'import pyccl'
  File "/usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/ccllib.py", line 20, in swig_import_helper
    return importlib.import_module(mname)
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 670, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 583, in module_from_spec
  File "<frozen importlib._bootstrap_external>", line 1043, in create_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
ImportError: dlopen(/usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/_ccllib.so, 2): Symbol not found: _PyCObject_Type
  Referenced from: /usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/_ccllib.so
  Expected in: flat namespace
 in /usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/_ccllib.so

During handling of the above exception, another exception occurred:

  File "<string>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/__init__.py", line 21, in <module>
    from . import ccllib as lib
  File "/usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/ccllib.py", line 23, in <module>
    _ccllib = swig_import_helper()
  File "/usr/local/lib/python3.7/site-packages/pyccl-1.0.1.dev2+g2b4f3cf2-py3.7.egg/pyccl/ccllib.py", line 22, in swig_import_helper
    return importlib.import_module('_ccllib')
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ModuleNotFoundError: No module named '_ccllib'
```

EiffL:
This PR aims fixes #541 .  So far, pyccl was compiled for the default python interpreter found in the environment. In some cases, one may want to compile pyccl for a different version, like :
```
$ python2 setup.py build
```
when python3 is the default interpreter. 

With this PR, CMake will accept a new command line option `PYTHON_VERSION` which is used internally when looking for the python interpreter. This could be used in a call to cmake like so:
```
$ cmake .. -DPYTHON_VERSION=2
```
but in practice, the setup.py will take care of providing this option. So, all that is needed is to run `python2 setup.py build` like before.

### DIFF
--- a/pyccl/CMakeLists.txt
+++ b/pyccl/CMakeLists.txt
@@ -5,7 +5,7 @@ include(BuildSWIG)
 include(UseSWIG)
 #set (UseSWIG_TARGET_NAME_PREFERENCE STANDARD)
 
-find_package(PythonLibsNew)
+find_package(PythonLibsNew ${PYTHON_VERSION})
 find_package(NumPy)
 include_directories(BEFORE ${PYTHON_INCLUDE_DIRS} ${NUMPY_INCLUDE_DIRS})
 
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,9 @@ class build(_build):
     """Specialized Python source builder."""
     def run(self):
         call(["mkdir", "-p", "build"])
-        if call(["cmake", "-H.", "-Bbuild"]) != 0:
+        v = sys.version_info
+        if call(["cmake", "-H.", "-Bbuild" ,
+                 "-DPYTHON_VERSION=%d.%d.%d"%(v.major, v.minor, v.micro)]) != 0:
             raise Exception("Could not run CMake configuration. Make sure CMake is installed !")
         if call(["make", "-Cbuild", "_ccllib"]) != 0:
             raise Exception("Could not build CCL")


alexander-mead:
Okay, this works for me. Previously, my problem was that I could not easily install the `python3` version of `CCL` without invoking the evil little hack of @damonge. Now, I get a working `python3` install if I do:

```
cmake .. -DPYTHON_VERSION=3
make; make install
python3 setup.py install
```

If I leave out the `-DPYTHON_VERSION=3` then `python3 setup.py install` incorrectly uninstalls a `python2` version in my python3 library file structure (I think) as it did before, so I *need* to pass `-DPYTHON_VERSION=3` to cmake. 

Conversely, if I include `-DPYTHON_VERSION=3` and then try `python setup.py install` then I think I get a `python3` version in my python2 library file structure and I get errors when trying to run `python2 -c 'import pyccl'`.

This is fine for my purposes at the moment, so thank you very much for doing this @EiffL. My feeling is that in future is that people may wish to have `python2` and `python3` versions installed on their system simultaneously, so finding some way of allowing this could be useful. Also, I suppose it would be useful to have a way of preventing libraries being installed in the incorrect locations (python2 version in python3 library tree and visa versa) and at the moment the errors are not terribly enlightening as to what the problem actually is.

EiffL:
Great, thanks @alexander-mead for testing it!
Yes, so this is expected behavior, if you run `python setup.py build` with an empty `build` folder, the `setup.py` will take care of adding the correct -DPYTHON_VERSION flag, but if  cmake is ran manually before, `setup.py` will not rerun the configuration step.

So the only way it will fail is in the following case:
  - The user manually runs cmake without the proper PYTHON_VERSION flag before running the `setup.py`. 
  - AND the user runs `setup.py` with a non default interpreter interpreter.

PR #557, opened 2019-02-18
remove global parameter settings and some headers; remove ini files; set stuff in python

### LINKED ISSUE #504: move ccl_gsl and ccl_splines to be allocated once per cosmology
Right now these are global state. We should associate them with each cosmology. Then we can make them modifiable from python. We can also include the physical constants in here too. 

See these comments:
- https://github.com/LSSTDESC/CCL/pull/433#discussion_r206719385
- https://github.com/LSSTDESC/CCL/pull/433#discussion_r206677013
- https://github.com/LSSTDESC/CCL/pull/433#discussion_r206719627

Note that the last one is asking for a new integration parameter to be added to the structures above.

### LINKED ISSUE #555: deprecate the params.ini files
We don't need these.

### LINKED ISSUE #560: centralize halo model GSL integration parameters
Right now the GSL integration parameters for the halo model are in a header and not adjustable.

beckermr:
This PR

1. removes most of the global params 
2. removes the ini files
3. allows most of the integration parameters to be set from python
4. moves the halo model integration parameters into the main settings

There are a few exceptions/details to be aware of
- the integration parameters for the neutrino phase space integrals are now hard coded
- the physical constants can only be set globally for the entire library 

These things are limitations of the C API essentially, so if we deprecate it, we can gain more flexibility.

~Finally, I needed to make a PR on angpow in order to merge this one. I have hard coded my branch, but it needs to be set back to a new released version of angpow.~

Closes #555 
Closes #504 
Closes #560


cmake/Modules/BuildAngpow.cmake
```diff
@@ -5,12 +5,12 @@ if(${CMAKE_VERSION} VERSION_GREATER "3.10.0")
...
 endif()
 
-set(AngpowTag v0.4.1)
+set(AngpowTag patch-1)
 
 # Downloads and compiles Angpow
 ExternalProject_Add(ANGPOW
         PREFIX ANGPOW
-        GIT_REPOSITORY https://github.com/LSSTDESC/Angpow4CCL.git
+        GIT_REPOSITORY https://github.com/beckermr/Angpow4CCL.git
```
beckermr: Needs to be hacked back to original state once the PR gets merged.

beckermr:
OSX tests pass on my mac. Not sure why the CI is messing up. Trying again.

beckermr:
OK @damonge. This PR is ready for review. It has a lot of breaking changes to the python interface.

cmake/Modules/BuildAngpow.cmake
beckermr: This PR https://github.com/LSSTDESC/Angpow4CCL/pull/6 is the one we are blocked on. IDK why the CCL interface is in the upstream library. I would have thought that would be maintained here.

beckermr: This is done!

src/ccl_redshifts.c
```diff
@@ -171,109 +170,90 @@ struct norm_params {
...
   // Set up parameters for the pz part of the intermediary integral.
   pz_val_p.z_true = z;
   pz_val_p.status = p->status;
   pz_val_p.pz_information = p-> pz_information;
-  
-  // Do the intermediary integral over the model relating  photo-z to true-z	
-  gsl_integration_cquad_workspace * workspace = gsl_integration_cquad_workspace_alloc(ccl_gsl->N_ITERATION);
+
+  // Do the intermediary integral over the model relating  photo-z to true-z
+  gsl_integration_cquad_workspace * workspace = gsl_integration_cquad_workspace_alloc(RDSH_GSL_N_ITERATION);
```
damonge: Same for these: why are these not part of the new structs?

damonge: OK, I think I get it: none of these functions take a cosmology, and since we're thinking about scrapping them, it's not worth complicating our lives further. Correct?

tests/ccl_test_correlation.c
```diff
@@ -58,9 +58,9 @@ static void compare_corr(char *compare_type,int algorithm,struct corrs_data * da
 
   double epsrel_save;
   if(!strcmp(compare_type,"histo")) { //This is needed for the histogrammed N(z) in order to pass the IA tests
-    epsrel_save=ccl_gsl->INTEGRATION_LIMBER_EPSREL;
-    ccl_gsl->INTEGRATION_LIMBER_EPSREL=2.5E-5;
-    ccl_gsl->INTEGRATION_EPSREL=2.5E-5;
+    epsrel_save = cosmo->gsl_params.INTEGRATION_LIMBER_EPSREL;
```
damonge: Since these are now reinitialized for each cosmology, I think we don't need to save them...

tests/ccl_test_correlation.c
```diff
@@ -784,8 +784,8 @@ static void compare_corr(char *compare_type,int algorithm,struct corrs_data * da
   ccl_cosmology_free(cosmo);
   ccl_cl_workspace_free(wyl);
   if(!strcmp(compare_type,"histo")) {
-    ccl_gsl->INTEGRATION_EPSREL=epsrel_save;
-    ccl_gsl->INTEGRATION_LIMBER_EPSREL=epsrel_save;
+    cosmo->gsl_params.INTEGRATION_EPSREL = epsrel_save;
```
damonge: ...and then reset them

src/ccl_neutrinos.c
```diff
@@ -10,20 +10,39 @@
 #include <gsl/gsl_roots.h>
 
 #include "ccl.h"
-#include "ccl_params.h"
-
 
 // Global variable to hold the neutrino phase-space spline
 gsl_spline* nu_spline=NULL;
```
tilmantroester: Should this be moved to the cosmology object with the other splines?

beckermr: Good question! There are few issues at play here. 

1. Right now we have to create objects in the following order. First you make a parameters structure and then you make a cosmology structure. At the time the spline above is built, the cosmology structure is not yet made. Thus putting this spline in it is a huge pain right now. We could do a larger refactor of the API here, but I would want a separate PR for that.

2. As far as I know, this spline doesn't actually need to be remade per cosmology. It holds an integral that is evaluated at different points for different cosmologies IIUIC. 

Thus I have left this code in place.

src/ccl_redshifts.c
beckermr: Yes this is correct!

src/ccl_neutrinos.c
```diff
@@ -38,28 +57,18 @@ gsl_spline* calculate_nu_phasespace_spline(int *status) {
...
-                        "ccl_neutrinos.c: Failed to read config file.");
-    *status = CCL_ERROR_MISSING_CONFIG_FILE;
-    return NULL;
-  }
-  
+
   int stat=0, gslstatus;
   gsl_integration_cquad_workspace * workspace = \
-                gsl_integration_cquad_workspace_alloc(ccl_gsl->N_ITERATION);
+                gsl_integration_cquad_workspace_alloc(GSL_N_ITERATION_NU);
```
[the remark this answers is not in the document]
beckermr: See my comments below. https://github.com/LSSTDESC/CCL/pull/557#discussion_r259576607

src/ccl_core.c
```diff
@@ -7,149 +7,228 @@
...
+  GSL_INTEG_GAUSS41                    // HM_INT_METHOD
+  };
+
+#undef GSL_EPSREL
+#undef GSL_N_ITERATION
+#undef GSL_INTEGRATION_GAUSS_KRONROD_POINTS
+#undef GSL_EPSREL_SIGMAR
+#undef GSL_EPSREL_DIST
+#undef GSL_EPSREL_GROWTH
+#undef GSL_EPSREL_DNDZ
```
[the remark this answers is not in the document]
beckermr: I wanted to preserve what we were doing with these settings. If we don't think this is worth it, then we can scrap them.

tests/ccl_test_correlation.c
beckermr: removed.

tests/ccl_test_correlation.c
beckermr: fixed

beckermr:
Alright. I think I addressed the comments and answered questions. Back to you @damonge !

src/ccl_neutrinos.c
damonge: Sounds good to me and we can revisit at a later PR

elisachisari:
This is a big change. I love that we will be able to set parameters from python. We've been in need of this for a while. I think this PR should be accompanied by changes in the CCL note, jupyter notebooks and C/python examples. Could we please make sure those make it into this PR as well? In addition, it would be nice to have a jupyter notebook example solely dedicated to exemplify how to pass on different integration paramaters.

beckermr:
@elisachisari I am going to do those in another PR to keep the sizes down.

beckermr:
OK. Docs updated. I cannot get the note to compile with my local latex distribution. I updated the main tex file so I'm not going to waste my time debugging an overly complicated latext setup.

beckermr:
Running make update pulled a bunch of dead links and wiped everything out, so that sucks.

beckermr:
OK. Got it to work.

PR #567, opened 2019-02-20
improved error reporting for C(ell)

beckermr:
This PR ensures we pass out an error message if C(ell) computations fail.


beckermr:
Thank you!

PR #574, opened 2019-02-23
ENH add context managers

### LINKED ISSUE #573: enable context manager for memory management of CCL Cosmology objects
Right now, the C memory held by CCL `Cosmology` objects is not freed until the garbage collector actually runs and calls the `__del__` method. This could happen sooner or could happen later. We don't control much about the python garbage collector.

We should add and recommend that in python people use a context manager API to enforce the timely freeing of memory like this 

```python
with Cosmology(...) as cosmo:
    # do things
```

This involves setting the proper methods on the `Cosmology` object.

beckermr:
This PR adds context managers. Closes #573

### DIFF
--- a/tests/ccl_test_core.py
+++ b/tests/ccl_test_core.py
@@ -303,5 +303,15 @@ def test_cosmology_repr():
         ccl.comoving_radial_distance(cosmo3, 0.5))
 
 
+def test_cosmology_context():
+    with ccl.Cosmology(
+            Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9, n_s=0.96,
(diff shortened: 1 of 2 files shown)


rainwoodman:
I see. If you delete the cosmo object, all members are automatically disabled.

tests/ccl_test_core.py
```diff
@@ -303,5 +303,21 @@ def test_cosmology_repr():
         ccl.comoving_radial_distance(cosmo3, 0.5))
 
 
+def test_cosmology_context():
+    with ccl.Cosmology(
```
philbull: Could do with docstring explaining what this test does

beckermr: Done.

PR #579, opened 2019-02-24
DEP remove redshifts module

### LINKED ISSUE #556: remove the redshifts module
I brought this up before and I am bringing it up again. We should remove the redshifts module. These integrations are very easy to do in python. We don't need them here.

beckermr:
This PR removes the redshifts module.

Closes #556


examples/Photo-z example.ipynb
```diff
@@ -1,326 +0,0 @@
-{
```
philbull: I'm fine with removing the core functionality, but I don't think we should lose the examples. Since calculating photo-z pdfs is important for correlation function calculations, we should show an alternative way of doing it now that the built-in CCL functions have been removed.

beckermr:
We absolutely do not need to maintain documentation on how to do numerical integration. This is what google is for. People are more than smart enough to figure this out on their own. 

If you want to maintain such documentation, feel free to do so in another PR!

philbull:
@beckermr Is there an example elsewhere that covers this use-case then? We want something that new users can copy-paste and noodle around with. There are also subtleties like normalisation etc. that may not be completely obvious.

beckermr:
Really people know how to find this stuff. 

https://www.google.com/search?q=numerical+integration+python&rlz=1CDGOYI_enUS590US590&oq=numerical+integration+python&aqs=chrome..69i57.6493j0j7&hl=en-US&sourceid=chrome-mobile&ie=UTF-8

philbull:
I can't seem to find the right emoji, so this will have to do: https://i.imgur.com/iWKad22.jpg

The point is that other parts of the docs rely on having a photo-z distribution that's been calculated from an input dN/dz, which was previously done using the CCL functions that have now been removed. For example, take a look at `examples/3x2demo.ipynb` or `examples/SLAC Feb2018 Demo.ipynb`. These are things that are intended as walkthrough tutorials for very new users.

Of course I think those new users can figure it out for themselves given a bit of time, and of course I'm not suggesting we have a pure numerical integration tutorial... What I am suggesting is that we provide a snippet that gets the appropriate selection function for them, like before, so they don't have to bother with it themselves while following tutorials that are ostensibly about something else. No need for that snippet to be in `examples/Photo-z example.ipynb`, but we should at least fix those other notebooks.

Also, I see nothing wrong with our docs providing a few copy-pasteable examples for very common use cases like this, even if they aren't particularly CCL specific. We shouldn't go nuts of course, but if tens of percents of our users are going to have to write something like this when they first start out with CCL, then perhaps we can make their lives a bit easier at little cost to ourselves?

beckermr:
Yeah this all too much long term maintenance. I am happy to adjust existing tutorials but that is all that is needed.

beckermr:
OK @philbull. The documentation update is finished.

c-d-leonard:
So I get the point of removing this, but I would like to raise the issue that during the code review with @rmjarvis we had a pretty extensive discussion about whether to keep or remove this where we decided to keep it, and this resulted in me doing a really non-negligible amount of work on it. This was like 4 months ago. I would obviously not have done all that work if we were going to deprecate it. So how do we avoid this happening? Obviously "I spent time on this" is not a reason not to deprecate something but I think this speaks to a general development workflow issue.

elisachisari:
I'm likewise not sure about this deprecation. We have multiple examples of how to produce observables with CCL that rely on this redshift module. They are being used widely, including tutorials that we've given at the DE School.  I understand the concern about maintenance, but I think that will be better addressed by deprecating the C API. In comparison, removing this functionality will not gain us much and it will instead create a lot of problems for current and potential CCL users.

beckermr:
I know @c-d-leonard. I 

My personal feelings are:

1. We should be very spartan about what we support in CCL. This is why I suggested removing this module in the first place. 

2. I redid the examples. See the code the in the PR. It is literally hundreds of lines of code that have been removed.

3. Think of all of the time we could have saved from when we started writing this code, through various API fixes, bugs (which I myself spent time on) and code review. Redoing the examples took me maybe 30 minutes using tools in scipy, which is all we would have had to spend in the first place had we made better decisions. 

This deprecation is about saving us time and giving us more capacity to focus on core parts of the library.

rmjarvis:
I don't have a strong opinion about whether this functionality should stay in CCL or not.  But to add some perspective for Danielle in particular, I have many, many times removed huge swaths of my own code from things.  Code which often took my days or weeks of work (or more!) to build in the first place, which I later realized was unnecessary or less efficient or maybe less accurate than some other alternative.  Removing it should be considered a big accomplishment that you found a more streamlined way to support the given functionality (assuming that is indeed the case, which I guess Matt is arguing is true here).  There is no gain to be had from dwelling on the sunk cost of the effort you expended in the past.

c-d-leonard:
You guys are right, I was just being grumpy, I'm fine with you merging this.

beckermr:
I'm grumpy most of the time so no worries there. :) 

Once we have an approved review I'm happy to merge!

elisachisari:
If this is to be removed, we need to check the CCL note in `doc/` and make sure we are consistent there as well.

beckermr:
OK @elisachisari. I have refreshed the documentation and removed all of the references to the old code.

beckermr:
Bump here. @elisachisari @philbull We good to merge?

philbull:
(Note: This is an API change and will require a version bump and changelog entry at some point)

beckermr:
Absolutely @philbull. The changelog entry is in this PR already. We will bump versions when we decide to release v2.

PR #521, opened 2018-10-29
Update CCL to compile against v2.7.1 of CLASS by default

philbull:
* Retains compatibility with pre-2.7.0 CLASS by setting the `CLASS_VERSION_PRE27` environment variable at build time (CLASS broke its power spectrum API between versions)
 * Adds small correction to redshifts in CLASS power spectrum calls to prevent interpolation errors due to rounding
 * Adds note to `README` on this
 * Minor whitespace and code style changes in and around affected functions in `ccl_power.c`

Issue #454, opened 2018-09-14
Test 71 fails: Error computing CLASS power spectrum


msyriac:
I'm following the Development Workflow instructions. I get no errors during compilation.

The following is the output of `build/check_ccl`

```
TEST 62/80 corrs:analytic_bessel 0.000000 % [OK]
TEST 63/80 corrs:analytic_fftlog 0.000000 % [OK]
TEST 64/80 corrs_3d:model_3 [OK]
TEST 65/80 corrs_3d:model_2 [OK]
TEST 66/80 corrs_3d:model_1 [OK]
TEST 67/80 bcm:model_1 [OK]
TEST 68/80 emu:model_1 [OK]
TEST 69/80 emu_nu:model_1 [OK]
TEST 70/80 power_nu_nl:model_3 [OK]
TEST 71/80 power_nu_nl:model_2 ERROR 1032: ccl_power.c: ccl_cosmology_compute_power_class(): Error computing CLASS power spectrum
```

Any ideas why this test fails?

c-d-leonard:
@EiffL @msyriac , I am not reproducing this behaviour and I'm not sure where it's from. What I can say is that I had similar issues if I tried to use the development workflow without ever installing the library. Apparently you have to install once to set certain environment variables, so if you just follow the development workflow right off the bat you can have issues. I think we should document this more carefully in the readme.

philbull:
Could this be related to issue #518?

beckermr:
fixed

PR #580, opened 2019-02-24
Refactored: Use CLASS 2.7 as default

philbull:
This PR replaces #521, which had diverged too much from `master`. It enables CLASS 2.7 support in CCL by default, but still allows linking against CLASS 2.6. It also adds docstrings for various power spectrum methods in the C code.

Note that this PR disables the CLASS Halofit non-linear computation for models with w!=-1. CLASS uses an internal spline-based calculation to rescale the LambdaCDM Halofit results if asked to compute the Halofit power spectrum for w!=-1 models. This effectively hangs when called by CCL, as the value of `z_max_pk` used by CCL is too large and CLASS doesn't handle it properly. Instead of allowing different values of `z_max_pk` for different power spectrum methods (which would soon get messy), CCL just raises an error if Halofit is used in w!=-1 cosmologies. (This behaviour can be revisited later if needed though.)


CMakeLists.txt
```diff
@@ -48,7 +49,15 @@ project(ccl VERSION 0.2.1)
...
       include(BuildCLASS)
     endif(EXTERNAL_CLASS_PATH)
-
+    
+    # Check whether an older version of CLASS is being used
+    if(CLASS_VERSION_PRE27)
+      message("Using an old (pre-v2.7.0) version of CLASS")
+      message("  CLASS_VERSION_PRE27: ${CLASS_VERSION_PRE27}")
+    else(CLASS_VERSION_PRE27)
+      message("Assuming CLASS v2.7.0 or later")
```
beckermr: The class devs clearly do not know about semantic versioning. Why is 2.6 to 2.7 an API change?!

src/ccl_power.c
```diff
@@ -402,6 +458,20 @@ static void ccl_cosmology_compute_power_class(ccl_cosmology * cosmo, int * statu
   struct file_content fc;
 
   ErrorMsg errmsg; // for error messages
+  
+  // Check if w!=-1 and halofit is enabled (raise error if so). CLASS uses a 
+  // spline to rescale the LambdaCDM halofit output if w!=-1, but this takes 
+  // an extremely long time if z_max_pk is set to a large value (which it is 
+  // by default in CCL). Instead of using a different z_max_pk (actually, 
+  // pk_eq_z_max) for different power spectrum methods, we just raise an error 
+  // for this method if w!=-1.
```
beckermr: This seems like a serious restriction. Why can't we use Takahashi et al with CLASS here? Should we not rely on CLASS to compute halofit at all maybe? This might make out lives easier.

tests/ccl_test_correlation_3d.c
```diff
@@ -165,3 +166,4 @@ CTEST2(corrs_3d,model_3) {
   int model=3;
   compare_correlation_3d(model,data);
 }
+*/
```
beckermr: Why are we removing this test?

tests/ccl_test_correlation_3dRSD.c
```diff
@@ -163,3 +164,4 @@ CTEST2(corrs_3dRSD,model_3) {
   int model=3;
   compare_correlation_3dRSD(model,data);
 }
+*/
```
beckermr: This one too.

tests/ccl_test_power.py
```diff
@@ -87,7 +87,10 @@ def loop_over_params(transfer_fn, matter_power, lin, raise_errs):
     print(">>> (%s; %s)" % (transfer_fn, matter_power))
 
     # Loop over parameters
-    for i in [0,2]: #w0_vals.size):
+    for i in [0,2]:
+        if matter_power=='halofit' and (w0_vals[i] != -1. or wa_vals[i] != 0.):
+            continue # Halofit disabled for w != -1
+
```
beckermr: This seems like a pretty serious handicap.

tests/ccl_test_power_nu.c
```diff
@@ -271,3 +273,4 @@ CTEST2(power_nu_nl,model_3) {
   int model=3;
   compare_power_nu_nl(model,data);
 }
+*/
```
beckermr: Again more commented out tests without an explanation.

CMakeLists.txt
philbull: INDEED!

tests/ccl_test_correlation_3d.c
philbull: This test is one of the ones that relies on the now-"forbidden" halofit cosmologies

tests/ccl_test_power_nu.c
philbull: As above.

src/ccl_power.c
philbull: That might be a good option. The CLASS version of halofit has some slightly hacky bits, like the spline rescaling that is causing these problems...

elisachisari:
Installation of this branch works for me, but then I've never had problems with the CCL version using CLASS 2.6. I think we need @c-d-leonard to comment on whether this solves her issues.

beckermr:
We should hold off on extensive testing of this branch until the halofit issues are fixed.

elisachisari:
Do we think this PR might solve this issue: https://github.com/LSSTDESC/CCL/issues/454?

c-d-leonard:
I downloaded and ran the tests fine in this branch.

However, I agree with Matt that requiring w=-1 to get a halofit P(k) is a pretty serious limitation since this is something that e.g. was used in the DES Y1 extensions analysis - we at least need support for some kind of nonlinear with w!=-1.

tests/ccl_test_power.py
elisachisari: So the previous bersion of CLASS does not have this limitation?

philbull:
I think we should enable CLASS 2.7 support first, and then address the Halofit + wCDM issue. The idea behind moving to CLASS 2.7 was to fix a load of spline-handling/bounds error bugs inside CLASS that were causing issues with MCMC runs. I know that doing this would technically be a feature regression, but the Halofit handling needs more thought than just implementing a basic fix here.

So, I agree that we want wCDM Halofit enabled soon, but I'm not sure the solution needs to be in this PR. Matt suggested implementing Halofit ourselves (or using a non-CLASS method) for example, which I like as a solution because it puts it on the same footing as our CosmicEmu nonlinear-P(k) method (which uses CLASS for the linear P(k) only).

Alternatively, if we want to make the CLASS Halofit work properly, we need to change the z_max for that method to be a much lower value. We do have a different z_max for CosmicEmu than other methods, so there is precedent. But I'd rather implement it in a separate branch so I can do some refactoring to make the "different z_max" handling more elegant.

beckermr:
Mmmk. Feel free to copy paste from my own halofit here: https://github.com/beckermr/cosmocalc/blob/master/src/nonlinear_powspec.c

Let’s put that in, then we can revisit this PR.

c-d-leonard:
I'm fine with doing this for now and worrying about halofit support for wCDM in a different PR.

beckermr:
At the rate we complete PRs this will leave master in a pretty bad state for a while. Let's get it done right.

philbull:
OK, so the way to do this right is a full refactor of `ccl_power.c`, along the lines we discussed a couple of weeks ago. It doesn't make sense to just tack Matt's halofit implementation onto this PR without restructuring things fully (this would be another method that uses CLASS for the linear power spectrum, something which we agreed should be factored out into a more generic function; the previous PR did this but got stale).

This would change the purpose of the PR from "get CLASS 2.7 working, but with some temporary limitations" to "fundamentally change ccl_power *and* add new halofit code *and* get CLASS 2.7 working". Is that really necessary? I'm just not seeing any upside to turning this into a monster PR instead of merging as-is and doing the refactor separately. At least then we solve a slew of CLASS-related bugs.

beckermr:
At the phone con @philbull, we had decided to wait on upgrading class until the other refactoring was done in separate PRs. So I think we can simply put this PR on ice until that is done.

philbull:
OK, which refactoring is that? My memory is hazy (or perhaps I wasn't on that telecon).

beckermr:
Adding halofit and separating the linear and non-linear power spectrum computations.

beckermr:
Whoops! wrong buttong 🤦‍♂️

beckermr:
CLASS integration will be moved to python where the API instabilities can be handled more easily. Closing.

PR #582, opened 2019-02-25
BUG remove the global state stored in the RSD splines

### LINKED ISSUE #569: global state depends on cosmology
It appears that we have some splines which are stored globally but depend on cosmology in the correlation function module.

```
// Global variable to hold the redshift space correlation multipole function spline
SplPar* xir_spline[3]={NULL,NULL,NULL};
```

beckermr:
This PR closes #569. 

Before this PR, splines of RSD correlation functions were shared between cosmologies and one would have to free them by hand. Now this is done automagically in the code and no global state is shared.


beckermr:
OK @damonge. Back to you!

Issue #552, opened 2019-02-18
Neutrino masses error handling


damonge:
Just running:
```
cosmo = ccl.Cosmology(Omega_c=0.27, Omega_b=0.045, h=0.67, 
                      sigma8=0.8, n_s=0.96, w0=-1., wa=-0.0,
                      m_nu=0.05)
```
gets into an infinite loop that you can only `Ctrl+C` out of. This will happen whenever you pass m_nu<0.06.

1. Why is this happening? It is a bit weird that m_nu=0 is fine, but the gap between 0 and 0.06 is not. In principle people should be able to put in whatever positive neutrino masses they want, even if they're unphysical (and I'd like to understand what "unphysical" means exactly - maybe @c-d-leonard knows?).
2. In any case, it should not be possible to fall into that infinite loop. Looking at the [code](https://github.com/LSSTDESC/CCL/blob/e32b466e3583408a01835a74360ddd6a94a89ea3/src/ccl_core.c#L357), I think it should be very easy to sort that out.

c-d-leonard:
Hmm, that's not good. I can answer (1) - "unphysical" refers to the fact that the sum of neutrino masses should, from the mass squared differences from particle physics, be either the inverted hierarchy sum (which is I forget but around 0.1) or the normal hierarchy sum (which is 0.059). So unless there is something unexpected and horribly wrong with the sum of mass squared constraints, the sum can't be below 0.059.

But regardless, we shouldn't fall into an infinite loop. I will take a look at this (or if someone else wants to, feel free, but I guess it's my fault so I can fix it).

damonge:
OK, but leaving observational constraints aside. Is there any physical reason why neutrino masses shouldn't add up to e.g. 0.03? I think we should allow users to have such values (e.g. I could imagine wanting to make a plot that shows constraints from cosmology regardless of any other measurements).

c-d-leonard:
Good point. I'm not 100% sure. We should probably check with someone who is more of a neutrino theory expert - @slosar do you have any ideas?

slosar:
So, the reason is that the code tries to satisfy both the sum of masses and mass square difference from oscillation data. It can't find a solution and then it goes into an infinite loop (there is some sort of newton minimizer than jarvis, horrified at my analytical approximation implemented in 10 mins :).
When you pass zero mass it just assume threes massless neutrinos.
[...]

PR #605, opened 2019-03-07
Fix bug with low sum of neutrino masses

c-d-leonard:
There was a bug which meant that when defining a cosmology within the normal or inverted hierarchy convention for neutrinos and with a mass below the respective physical limit in each case (~0.059 and ~0.1) you would fall into an infinite loop in ccl_parameters_create when trying to split the sum of masses according to the hierarchy instead of exiting gracefully with an error.

This PR fixes the bug and avoids the infinite loop but doesn't manage to pass an informative error to python because the problem arises in ccl_parameters_create, where there is not yet a cosmology object and therefore no way to pass the error message back to python. I'm leaving this for later (maybe if we deprecate the ccl_parameters_create function in C).

See Issue #552 for discussion.


src/ccl_core.c
```diff
@@ -424,19 +425,26 @@ ccl_parameters ccl_parameters_create(
...
 		  mnu_in[2] = 0.;
-	  } else{
-
+	  } 
+    //else if(*mnu>1e-15 && *mnu<0.06){
+    //*status = CCL_ERROR_LOWMNU;
+    //ccl_raise_exception(*status, "The neutrino mass is too low to satisfy \
+    the mass difference measurements in the case of normal hierarchy.\
+    In this low mass case, set mnu_type to sum_inverted or sum_equal.");
+    //  }
```
beckermr: Let's remove this extra code.

c-d-leonard:
Thanks, good call, I'll make those changes.

pyccl/core.py
```diff
@@ -579,10 +579,22 @@ def _build_parameters(
...
+                np.sqrt(2.55E-3)) and (m_nu[0]>1e-15)):
+                raise ValueError("if mnu_type= sum, we are using the "
+					                 "normal hierarchy and so m_nu must "
+					                 "be less than (~)0.0592")
+            elif (mnu_type=='sum_inverted' and 
+                  m_nu[0]<( np.sqrt(2.43e-3 - 7.62e-5)+ np.sqrt(2.43e-3))
+                  and (m_nu[0]>1e-15)):
+                raise ValueError("if mnu_type= sum_inverted, we are using the "
+					                 "inverted hierarchy and so m_nu must "
+					                 "be less than (~)0.0978")
```
beckermr: Oy. flake8 is going to be mean to you. :(

c-d-leonard: I did go run flake8 locally immediately after pushing that commit :P it should be okay now.

c-d-leonard:
The current status is: I added a check from the python that the sum of neutrino masses is above the minimum allowed value in the given hierarchy, so we now get an informative error in python if that is done wrong. I also added a test in python to make sure that error gets raised.

I changed the C so that if you call ccl_parameters_create directly from C and you make the same kind of too-low choice for a sum of mnu given your hiearchy, we exit with an informative error. I manually checked that this behaves correctly if I do this, but I'm not sure if there's an elegant way to add a test in C to the suite because of the lack of cosmo status message (there's no cosmology object yet when we're still setting up the parameters object, so we can't pass it back out). What we want the C to do is just exit gracefully, which it is doing - but I don't know how to test it's doing that without just stopping the test suite.

beckermr:
Right. At this point we either:

1. Significantly change the C API.
2. Deprecate the C API so that we are sure the test from Python is sufficient.
3. Figure out some other magic that IDK.

My motivation to deprecate the C API is because of shit like this that is super annoying and time consuming.

beckermr:
OTOH, the python part is already done! It is a sign from the heavens!

damonge:
@beckermr can we forget about the unit test and revisit once we have deprecated the C API?
Since this is a more general problem (testing errors while creating parameters), should we open an issue for that and move on?

beckermr:
Seems fine to me.

PR #611, opened 2019-03-19
ENH adding our own halofit

### LINKED ISSUE #599: implement halofit
Right now IIUIC, when people request EH98 for the transfer function, they are not getting halofit back out for the non-linear power spectrum. I think this is pretty weird TBH. We should have halofit for this.

beckermr:
This PR adds our own halofit.

Closes #599 

To do:
 - [x] rebase on master after #610 
 - [x] finish halofit implementation

### DIFF
--- a/src/ccl_halofit.c
+++ b/src/ccl_halofit.c
@@ -17,6 +17,9 @@
 */
 
 // I snuck a private function from backgrounc.c into here. :P
+// The default routines create a full spline but we just need one value.
+// This may seem like over optimizing, but in testing initializing halofit
+// is an order of magnitude or more slower if we don't do this.
 void compute_chi(double a, ccl_cosmology *cosmo, double * chi, int * stat);
 
 static double zdrag_eh(ccl_parameters *params) {
@@ -36,6 +39,8 @@ struct hf_model_match_data {
 };
 
 static ccl_cosmology *create_w0eff_cosmo(double w0eff, ccl_cosmology *cosmo, int *status) {
+  // create a cosmology with the same parameters as the input except w0-wa. Instead
+  // the cosmology is created with w0 = w0eff.
   ccl_parameters params_w0eff;
   double norm_pk;
   double mnu[3];
@@ -66,6 +71,9 @@ static ccl_cosmology *create_w0eff_cosmo(double w0eff, ccl_cosmology *cosmo, int
 }
 
 static double w0eff_func(double w0eff, void *p) {
+  // function used to compare the distance to the CMB in a test cosmology to
+  // to the value in the original cosmology
+  // returns chi_eff - chi
   struct hf_model_match_data *hfd = (struct hf_model_match_data*)p;
   ccl_cosmology *cosmo_w0eff = NULL;
   double chi_drag_w0eff, tmp, zdrag_w0eff;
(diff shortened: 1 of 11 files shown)


beckermr:
bump here @damonge

src/ccl_halofit.c
```diff
@@ -0,0 +1,941 @@
...
+/*
+ * We use the remapping procedure in https://arxiv.org/pdf/1601.07230.pdf
+ * to map w0-wa models onto wa == 0 models.
+ * The functions below implement this procedure.
+*/
+
+// I snuck a private function from backgrounc.c into here. :P
+void compute_chi(double a, ccl_cosmology *cosmo, double * chi, int * stat);
+
+static double zdrag_eh(ccl_parameters *params) {
```
damonge: I wonder if we should put this in ccl_eh.c and call it from there to avoid repeating code

src/ccl_halofit.c
```diff
@@ -0,0 +1,941 @@
...
+}
+
+struct hf_model_match_data {
+  double chi_drag;
+  double a;
+  ccl_cosmology *cosmo;
+  int *status;
+};
+
+static ccl_cosmology *create_w0eff_cosmo(double w0eff, ccl_cosmology *cosmo, int *status) {
```
damonge: Could we just write a couple of sentences describing what these functions are doing?

tests/ccl_test_correlation_3d.c
```diff
@@ -84,9 +84,9 @@ static void compare_correlation_3d(int i_model,struct corrs_3d_data * data)
 
   // FIXME: these are not real standards
   // tolerence on abs difference in r^2 xi(r) for the range r = 0.1 - 100 Mpc (40 points in r) for z=0,1,2,3,4,5
-  double CORR_TOLERANCE1[6] = {4.0e-2, 1.0e-2, 5.0e-3, 4.0e-3, 1.0e-3, 8.0e-4};
+  double CORR_TOLERANCE1[6] = {5.0e-2, 5.0e-2, 5.0e-2, 5.0e-2, 5.0e-2, 5.0e-2};
```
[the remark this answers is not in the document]
beckermr: or just remove it...

src/ccl_halofit.c
```diff
@@ -0,0 +1,941 @@
...
+#include "ccl.h"
+
+/*
+ * We use the remapping procedure in https://arxiv.org/pdf/1601.07230.pdf
+ * to map w0-wa models onto wa == 0 models.
+ * The functions below implement this procedure.
+*/
+
+// I snuck a private function from backgrounc.c into here. :P
+void compute_chi(double a, ccl_cosmology *cosmo, double * chi, int * stat);
```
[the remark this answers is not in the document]
beckermr: This is used to compute some stuff related to halofit for w0-wa. Using the standard API is waaay to slow since it builds a full spline.

beckermr:
There is a test for the outputs. See `ccl_test_power_nu.c` It reads in benchmark data (presumably from CLASS) and compares to halofit from CCL in a number of cases.

src/ccl_halofit.c
beckermr: I thought of this, but decided against it. Basically, there is no reason to couple the two modules together. Also any other approximation for zdrag would work fine here too. I just happened to pull the EH one, but we could have used CLASS.

beckermr:
OK @damonge I added some extra notes. Back to you!

src/ccl_halofit.c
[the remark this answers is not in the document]
damonge: I see now, you need to generate new cosmologies. OK, thanks

src/ccl_halofit.c
damonge: Another option would be to put it in background.c and make it visible throughout the whole library. Things like this may come in handy at some point for BAO likelihoods. We can reassess then if something similar comes up.

Issue #593, opened 2019-03-04
neutrino parameterization


beckermr:
from @viventi

>In the neutrino sector the parametrization of the massless neutrinos is done in terms of Neff. The problem here is that, if Neff is smaller than the contribution of the massive neutrinos when ultra-relativistic, you end up with a negative params->N_nu_rel and, consequently, negative Omega (ccl_species_ur_label). In your tests 34/94 and 39/94 you actually use a combination of parameters that implies in negative params->N_nu_rel (the contribution from the ultra relativistic massive neutrinos is slightly larger than Neff=3 used). It may be a good idea to use the number of massless neutrinos and compute Neff from it and the ultra-relativistic massive ones (that was our choice in NumCosmo and I think it is the same as CLASS). Moreover, in a MCMC analysis the use of Neff as parameter can be trick, one would have to use a complicated prior to avoid params->N_nu_rel < 0.

beckermr:
@c-d-leonard @philbull @slosar 

Please look closely at this. If we need to fix this, we need to do it before releasing v2

slosar:
I don't know exactly how this is parameterized? Normally, you specify (Neff, mnu) for each species + Neff_nonrel. So the total Neff is then the sum of those Neffs. Here, if I understand correctly, we specify sum mnu and Neff total assuming each species is deltaNeff of  1 or what? @c-d-leonard

c-d-leonard:
@beckermr @slosar Sorry for the delayed reply, I'm looking into this now.

I can say that the reason we selected Neff at the time as the parameter was because people who had feelings about sampling (e.g. @joezuntz is one person I remember talking to) indicated it would be important to be able to sample over Neff.

I'll post more soon once I remind myself what's going on here.

joezuntz:
We will probably need to sample over Neff, yes, to match other analyses at the very least.  But as long as we can translate that into whatever parametrization CCL wants then it doesn't matter what that is exactly.

slosar:
But can someone just write exactly who it is parameterized? Having negative energy densities appear in some parts of your parameter space does seem like not good.

vitenti:
In most codes (I know) they use N_ur as a parameter and Neff as a derived parameter in the MCMC. Is there any reason to sample Neff directly?

c-d-leonard:
To answer @slosar, the way it is parameterized is that the user passes:

- Neff

[...]

PR #620, opened 2019-04-12
Add check to ensure N_nu_rel>0

c-d-leonard:
This PR addresses Issue #593, which pointed out that we had no check to prevent Neff / mnu combinations which result in N_nu_rel<0, and in fact some of our tests were for cosmologies with this problem.

I added a check in C and in Python. Annoyingly, because this once again is a problem that arises during parameter creation, there is no cosmology object yet and so there is no way to pass an error message back to python from C. Because we need to know how many neutrinos with non-negligible mass there are to compute N_nu_rel and check if it's greater than zero, that means I had to reproduce the calculation to split mnu (passed as sum) into masses within core.py. This is not a great solution and should be seen as a stopgap measure while we figure out how to solve that larger error passing issue.

Currently, to run a chain, the user would still be required to place a complicated prior on Neff which depends on mnu to ensure N_nu_rel stays above zero. We could consider switching to sampling in N_nu_rel to prevent this, but that is beyond the scope of this PR.


beckermr:
Maybe we should wait on this pr until after the next telecon. If everyone agrees dump the c api then this gets simpler

PR #626, opened 2019-04-18
DEP remove all accelerators

### LINKED ISSUE #594: deprecate use of GSL spline accelerators
from @vitenti

> When comparing NumCosmo distances with CCL's we found that once in a while CCL was giving really weird results. After investigating it, we saw that it had evaluated the spline to compute the Omega of massive neutrinos and the knot used was 950 (the neutrino spline has 1000 knots by default). Then just after the spline for h_over_h0 was evaluated, since the accelerator contained the last knot used (950), it started the search by evaluating the xi[950], however, the spline for h_over_h0 (cosmo->data.E) has only 499 knots (by default) and for this reason xi[950] is actually out-of-bounds and end up reading a random place in memory. This confuses the binary search and the spline evaluation giving a random result.

beckermr:
This PR removes all GSL accelerators so we can implement threading more easily. 

Closes #594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@
 - Added `has_density` and `has_shear` flags to ClTracers (#614)
 - Simplified C_ell unit tests (#614)
 - Fixed a small bug in the w_tophat expression and increased precision (#607)
+- Deprecated the use of GSL spline accelerators (#626)
 
 ## Python library
 - Improved error reporting for `angular_cl` computations (#567).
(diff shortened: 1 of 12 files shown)


beckermr:
Looking at the CI run times for the C benchmarks, I am seeing a small performance degradation of ~3% in run time. Given that even simple threading will make parts of the library at least 2x faster, I think we should not worry about this.

beckermr:
The python tests run in the same amount of time. So I do think we are good to go on performance here.

beckermr:
@elisachisari I have removed the GSL accelerators and everything runs at basically the same speed. I actually remembered to update the changelog on the first go! Yay! Can you take a look here?

damonge:
Alright, done. Press merge whenever you want.

Issue #627, opened 2019-04-19
Proposal for generalized tracers


damonge:
Here's a proposal for the implementation of generalized tracers (the second task in issue #511). Comments from @elisachisari @elikrause @jablazek @tilmantroester @beckermr @slosar and/or @joezuntz would be useful. Apologies for the long issue!

### Setup and definitions
Start by interpreting a tracer as a given projected quantity in the sky:
<a href="https://www.codecogs.com/eqnedit.php?latex=\dpi{100}&space;u(\hat{\bf&space;n})=\int&space;d\chi\,W_u(\chi)\,U(\chi\hat{\bf&space;n},t(\chi))" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\dpi{100}&space;u(\hat{\bf&space;n})=\int&space;d\chi\,W_u(\chi)\,U(\chi\hat{\bf&space;n},t(\chi))" title="u(\hat{\bf n})=\int d\chi\,W_u(\chi)\,U(\chi\hat{\bf n},t(\chi))" /></a>
here <a href="https://www.codecogs.com/eqnedit.php?latex=\dpi{100}&space;u(\hat{\bf&space;n})" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\dpi{100}&space;u(\hat{\bf&space;n})" title="u(\hat{\bf n})" /></a> is the projected quantity, <a href="https://www.codecogs.com/eqnedit.php?latex=\dpi{150}&space;U({\bf&space;x},t)" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\dpi{100}&space;U({\bf&space;x},t)" title="U({\bf x},t)" /></a> is the corresponding 3D quantity and <a href="https://www.codecogs.com/eqnedit.php?latex=\dpi{100}&space;W_u(\chi)" target="_blank"><img src="https://latex.codecogs.com/gif.latex?\dpi{100}&space;W_u(\chi)" title="W_u(\chi)" /></a> is what I'll call the tracer's *radial kernel*. This setup should be enough to describe the vast majority of 2D quantities we'll be interested in in DESC and future CMB/radio experiments (see one exception below).

[...]

PR #629, opened 2019-04-20
Moved Cell tests C->python

damonge:
I've moved the C_ell tests from C to python. This will be necessary once the work in #627 is done, since none of the stuff currently tested in C will live in C. Thoughts @elisachisari @beckermr ? If you agree with this I'll move the correlation function tests too, since those will become obsolete as well.


damonge:
> We need to be making proper use of parametrized tests here. See the pytest documentation.

Sorry @beckermr , I don't know what this means. I had a quick look at [this](https://docs.pytest.org/en/latest/parametrize.html) but I don't know exactly what you're looking for.

beckermr:
So what I mean here is that there is a lot of repeated code testing all of the different cases. The parametrized tests allow pytest to separate out each test and have the added benefit that if one fails it will still run the rest of them.

damonge:
@beckermr alright, how about now?

beckermr:
Looks great! One more thing. We should preserve the link between the benchmarks themselves and the code that generated them. Right now the files were moved, but we won't be able to easily track this all down later. 

IDK how exactly to do this. Maybe @elisachisari has some thoughts here?

damonge:
We could keep the files in both the `benchmarks/data` and the `tests/benchmarks` directories until  all tests have been migrated, but the size of the repo will increase

beckermr:
Hmmmm. Can we move the scripts over? We can tell flake8 to ignore them using the exclude argument. It might be good organize the files with the benchmark data into directories instead of a grab bag of stuff all in one spot.

damonge:
The codes that generated the benchmarks live here:
https://github.com/LSSTDESC/CCL/tree/master/tests/benchmark/codes

They're all there with a README that explains what to do in order to reproduce things. So we could move that whole directory over into `benchmarks`, but then the tests that currently reside in `tests` will be disconnected from them. We could also move things gradually, but maintaining two separate READMEs while this happens is going to be a pain and probably lead to inconsistencies in the end result...

beckermr:
Ahhhh OK. Let's make an issue to move it and leave it in place for now then.

damonge:
OK, great. Do you mean move the benchmark files back to where they were before?

beckermr:
We can move the files, but we should make an issue to move the code.

damonge:
OK, good. I'll make an issue then. Should I add this to the CHANGELOG? Not sure this qualifies as code change.

Anything else?

beckermr:
We can punt on the changelog and make one entry later.

Issue #280, opened 2017-11-10
Add conventional IA model bias with the usual normalization


elisachisari:
This is a reminder for myself and @jablazek that when possible, we should add a b_I(z) function in CCL for the fiducial NLA model with the usual normalisation convention. See discussion in https://github.com/LSSTDESC/WLPipe/issues/26

damonge:
For clarification: the way this is implemented in CCL is as follows (I'm quoting from https://arxiv.org/pdf/1506.02671.pdf, eqs. 4.14, 5.2):

1. In the large-ell limit, the IA kernel is basically proportional to W_IA = p_red*b_I, where p_red(z) is the redshift distribution of the galaxy population that aligns, and b_I(z) is some alignment bias function.
2. Writing p_red(z) = p_tot(z)*f_red(z), where f_red is the "red" fraction, this is just W_IA = p_tot(z)*f_red(z)*b_I(z). Currently users need to pass both b_I and f_red to CCL (as was initially suggested by @elisachisari and @jablazek).
3. According to Eq. 5.2 of the paper above, b_I can be related to the more common A_I as: b_I = -A_IOmega_MC_1*rho_crit/D(z)

If you confirm that this is really the parametrization that you want I could easily implement the modification in the code.

EiffL:
@elisachisari @jablazek Quick question, there are 2 options:
  - Replacing the current implementation so that everywhere in the code b_I(z) gets replaced by A_I(z), thus changing the API, in particular the input parameters of `ccl_cl_tracer_new`
  - Give the user the option of providing b_I(z) or A_I(z) (for instance at the level of the python interface), then convert internally everything to b_I(z), thus preserving the current internal implementation.

Which one would you prefer ? Is option 1 ok ?

elisachisari:
I would prefer the second option, so we can marginalize directly over b_I if we want without using A_I.

damonge:
I'd suggest implementing this once the angpow branch has been merged (which should be soon).

EiffL:
Ok great, happy to implement it, just ping me when the angpow branch is
ready then.

jablazek:
I agree that you should be able to specify {A(z), f_fred(z)} or b_I(z), the latter of which would override the assumed 1/Growth(z) scale dependence. Is that the baseline plan?

[...]

PR #630, opened 2019-04-23
Generalizing Cl tracers

### LINKED ISSUE #511: Generalized Limber integrator

### LINKED ISSUE #639: Simplify IAs
Right now, when creating a tracer with intrinsic alignments, we need to pass two arrays: the alignment amplitude and the fraction of aligned galaxies. In the L-NLA model these two quantities are completely degenerate, and in fact people usually fit for a single amplitude that is the product of these two quantities.

I'd like to simplify the current interface so IAs (with the L-NLA model) only take a single array of amplitudes as input. If people agree, then I'll do this in the ongoing PR #630 .

Pinging @jablazek @elisachisari

damonge:
Closes #511 
Closes #639 

This will generalize the current treatment of tracers as described in #627 
Tasks to accomplish:

- [x] Generalize current 2D power spectra to arbitrary functions of k and a. These could be factorizable, and can be extrapolated in redshift in a different way than power spectra.
- [x] Implement generic tracer structure in C. Main ingredients are: transfer function (a 2D function as above), radial kernel (a 1D spline), ell-prefactor (a discrete set of choices), Bessel derivative order (discrete set of choices).
- [x] Make it possible to combine tracers at the C level.
- [x] Hook the generalized tracers with the Limber integrator.
- [x] Make the new structure visible from python.
- [x] Make some specific tracer constructors for useful cases in python. At least they should cover the current set of tracers: NumberCounts (linear bias, RSDs, magnification), WeakLensing (with IAs), CMBLensing.


damonge:
Some work still needs to be done:

- [ ] Make sure that previous benchmarks pass (they do at the C level, so should do in python too)
- [ ] Make new unit tests that validate the fully general tracers
- [ ] Document everything!

So not fully ready for review, but ready for eyeballing if anyone is interested.

damonge:
Alright, this is now ready for review. Tagging all relevant people @elisachisari @beckermr @jablazek @elikrause @slosar @jeremyneveu @jecampagne . It'd be ideal if at least a subset of you can have a look at this.

To facilitate the review, here's a summary of the changes:
- Generalized tracers fully implemented as described in section 4.9 of the CCL note and in #627 .
- The C implementation is coded up in `src/ccl_tracers.c` and `include/ccl_tracers.h`.
- The C implementation is pure infrastructure. This means that there's no explicit functionality to compute number counts / weak lensing / CMB lensing tracers in C, only the infrastructure for the generalized ones. The convenience functions for the specific tracers now live in python.
- Tracers and power spectra are linked in `src/ccl_cls.c`, which has also seen significant modifications.
- I've had to disengage angpow entirely, so currently we don't have non-Limber functionality. This is temporary only until @jeremyneveu @jecampagne can provide the interface. I would propose that this happen in a different PR after this one gets merged. When writing the angpow interface, I believe the contents of section 4.9 in the note and the C docstrings should be sufficiently informative, and I'm more than happy to help here.
- The python implementation lives in `pyccl/tracers.py` and `pyccl/ccl_tracers.i`. You can see the new constructors for `NumberCountsTracer`s etc. in there, and how they all inherit from a general `Tracer` class that is now accessible to users.
- The only change to the python API is that I've deprecated the red fraction argument for intrinsic alignments, as discussed in #639 .
- I've adapted all existing benchmark checks and they all pass.
- I've created unit tests for the generalized tracers in `benchmarks/test_tracers.py`. Since now we can pass whatever P(k) and transfer functions we want, these tests are based on matching power-law versions of these to their corresponding analytic expectation.

Alright, I think that's it!

damonge:
OK, to try and entice people to review this, I've created a notebook that describes the functionality of the generalized tracers: https://github.com/LSSTDESC/CCL/blob/generalized_tracers/examples/GeneralizedTracers.ipynb

beckermr:
Sorry for the delay. I will be looking tomorrow.

jablazek:
will look next week. totally swamped this week - sorry for the delay.

slosar:
Again, terribly sorry, but cannot look before after collab meeting by when others will be able to do it. (but if not happy to look)

pyccl/tracers.py
```diff
@@ -0,0 +1,405 @@
...
+            alignments. Defaults to None.
+    """
+    def __init__(self, cosmo, dndz, has_shear=True, ia_bias=None):
+        self.trc = []
+        if has_shear:
+            # Kernel
+            kernel_l = get_lensing_kernel(cosmo, dndz)
+            self.add_tracer(cosmo, kernel=kernel_l,
+                            der_bessel=-1, der_angles=2)
+        if ia_bias is not None:  # Has magnification bias
```
beckermr: Comment is wrong here.

pyccl/tracers.py
```diff
@@ -0,0 +1,405 @@
...
+    magnification.
+
+    Args:
+        cosmo (:obj:`Cosmology`): Cosmology object.
+        has_rsd (bool): Flag for whether the tracer has a
+            redshift-space distortion term.
+        dndz (tuple of arrays): A tuple of arrays (z, N(z))
+            giving the redshift distribution of the objects. The units are
+            arbitrary; N(z) will be normalized to unity.
+        bias (tuple of arrays): A tuple of arrays (z, b(z))
```
beckermr: I'm wondering about the API here. Is there a reason to not require the bias be given?

pyccl/tracers.py
```diff
@@ -0,0 +1,405 @@
...
+    A `Tracer` object will in reality be a list of different such
+    tracers that get combined linearly when computing power spectra.
+    Further details can be found in Section 4.9 of the CCL note.
+    """
+    def __init__(self):
+        """By default this `Tracer` object will contain no actual
+        tracers
+        """
+        # Do nothing, just initialize list of tracers
+        self.trc = []
```
beckermr: We should either make this "private" via prefixing with `_` or make it a public attribute that is documented. I think private is better so it would `self._trc`.

pyccl/tracers.py
```diff
@@ -0,0 +1,405 @@
...
+            self.add_tracer(cosmo, kernel=kernel_i, transfer_a=t_a,
+                            der_bessel=-1, der_angles=2)
+
+
+class CMBLensingTracer(Tracer):
+    """A Tracer for CMB lensing.
+    Args:
+        cosmo (:obj:`Cosmology`): Cosmology object.
+        z_source (float): Redshift of source plane for CMB lensing.
+        nsamples (int): number of samples over which the kernel
```
beckermr: Mark as optional with default in the docstring.

pyccl/tracers.py
```diff
@@ -0,0 +1,405 @@
...
+            The kernel is quite smooth, so usually O(100) samples
+            is enough.
+    """
+    def __init__(self, cosmo, z_source, n_samples=100):
+        self.trc = []
+        kernel = get_kappa_kernel(cosmo, z_source, n_samples)
+        self.add_tracer(cosmo, kernel=kernel, der_bessel=-1, der_angles=1)
+
+
+def check_returned_tracer(return_val):
```
beckermr: This function should probably have an `_` in the front of the name so it is private.

tests/ccl_test_pyccl_interface.py
```diff
@@ -617,20 +619,9 @@ def check_cls_nu(cosmo):
     assert_( all_finite(ccl.angular_cl(cosmo, nc1, lens1, ell_arr)) )
     assert_( all_finite(ccl.angular_cl(cosmo, nc1, lens2, ell_arr)) )
 
-    # Check get_internal_function()
-    a_scl = 0.5
-    a_lst = [0.2, 0.4, 0.6, 0.8, 1.]
-    a_arr = np.linspace(0.2, 1., 5)
-    assert_( all_finite(nc1.get_internal_function(cosmo, 'dndz', a_scl)) )
-    assert_( all_finite(nc1.get_internal_function(cosmo, 'dndz', a_lst)) )
-    assert_( all_finite(nc1.get_internal_function(cosmo, 'dndz', a_arr)) )
```
beckermr: So we lost this method? Not a big deal, but we should think about adding it back. I'd be nice to be able to access to the values of the kernels used by the C layer. Maybe make an issue for this?

pyccl/ccl_tracers.i
```diff
@@ -0,0 +1,182 @@
...
+    (double* chi_s, int nchi),
+    (double* wchi_s, int nwchi),
+    (double* lk_s, int nlk),
+    (double* a_s, int na),
+    (double* tka_s, int ntka),
+    (double* tk_s, int ntk),
+    (double* ta_s, int nta)}
+//%apply (double* IN_ARRAY1, int DIM1) {
+//    (double* ell, int nell),
+//    (double* aarr, int na)};
```
beckermr: These commented lines can be removed.

damonge:
This is great, @beckermr , thanks a lot! I'll address these ASAP.

Something very strange is happening where the linux tests are passing but all the C_ell tests are failing on OSX. If I look at the pytest error messages, it's failing when comparing the benchmark C_ells with the calculated ones, but this is not a matter of floating-point precision, both arrays are different by several orders of magnitude (which, again, doesn't happen on linux...).

I don't have easy access to a mac, but I wonder if you have any intuition why this might be happening.

src/ccl_f1d.c
```diff
@@ -0,0 +1,61 @@
...
+{
+  ccl_f1d_t *spl=malloc(sizeof(ccl_f1d_t));
+  if(spl==NULL)
+    return NULL;
+
+  spl->spline=gsl_spline_alloc(gsl_interp_cspline,n);
+  int parstatus=gsl_spline_init(spl->spline,x,y,n);
+  if(parstatus) {
+    gsl_spline_free(spl->spline);
+    return NULL;
```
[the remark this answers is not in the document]
damonge: Correct! Done.

include/ccl_cls.h
```diff
@@ -2,266 +2,42 @@
...
-			     CCL_ClTracer *clt1,CCL_ClTracer *clt2,
-			     ccl_p2d_t *psp,double l,int * status);
-
-
+			     ccl_cl_tracer_collection_t *trc1,
+			     ccl_cl_tracer_collection_t *trc2,
+			     ccl_f2d_t *psp,double l,int *status);
 /**
- * Computes non-Limber power spectrum for two different tracers
+ * Computes Limber power spectrum for two different tracers at a given ell.
```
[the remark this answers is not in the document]
damonge: Fixed. I left that function there to guide the development of the angpow wrapper. Currently it just throws an error if called.

include/ccl_tracers.h
```diff
@@ -0,0 +1,188 @@
+/** @file */
+
+#ifndef __CCL_CLTRACERS_H_INCLUDED__
+#define __CCL_CLTRACERS_H_INCLUDED__
+
+CCL_BEGIN_DECLS
+
+#define CCL_FRAC_RELEVANT 5E-4
```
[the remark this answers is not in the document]
damonge: Added!

include/ccl_tracers.h
```diff
@@ -0,0 +1,188 @@
...
+  ccl_f1d_t *kernel; //Radial kernel.
+  double chi_min; //Minimum radial comoving distance for this tracer.
+  double chi_max; //Maximum radial comoving distance for this tracer.
+} ccl_cl_tracer_t;
+
+/**
+ * Constructor for a ccl_cl_tracer_t. See CCL note for a description of how tracers are used to compute power spectra.
+ * @param cosmo Cosmology structure.
+ * @param der_bessel Bessel function derivative order (-1, 0, 1 or 2).
+ * @param der_angles ell-dependent prefactor type (0, 1 or 2)
```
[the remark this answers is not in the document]
damonge: yes, done!

src/ccl_tracers.c
```diff
@@ -0,0 +1,516 @@
...
+{
+  ccl_cl_tracer_collection_t *trc=malloc(sizeof(ccl_cl_tracer_collection_t));
+  if(trc==NULL)
+    *status=CCL_ERROR_MEMORY;
+
+  if(*status==0) {
+    trc->n_tracers=0;
+    //Currently this is hard-coded.
+    //It should be enough for any practical application with minimal memory overhead 
+    trc->ts=malloc(100*sizeof(ccl_cl_tracer_t *));
```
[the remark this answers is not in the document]
damonge: Yes. Done in ccl_tracers.h

src/ccl_tracers.c
```diff
@@ -0,0 +1,516 @@
...
+    }
+  }
+
+  ccl_f1d_t_free(nz_f);
+}
+
+//3 H0^2 Omega_M / 2
+static double get_lensing_prefactor(ccl_cosmology *cosmo,int *status)
+{
+  double hub=cosmo->params.h*ccl_h_over_h0(cosmo,1.,status)/ccl_constants.CLIGHT_HMPC;
```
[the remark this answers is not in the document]
damonge: This is a remnant from the previous code and I was too lazy to check. I think you're right, and the tests seem to agree. Removed!

src/ccl_tracers.c
```diff
@@ -0,0 +1,516 @@
...
+				 is_factorizable, //is factorizable
+				 extrap_order_lok, //extrap_order_lok
+				 extrap_order_hik, //extrap_order_hik
+				 ccl_f2d_constantgrowth, //extrap_linear_growth
+				 is_fka_log, //is_fka_log
+				 NULL, //growth (function)
+				 1, //growth_factor_0 -> will assume constant transfer function
+				 0, //growth_exponent
+				 ccl_f2d_3, //interp_type
+				 status);
```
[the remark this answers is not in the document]
damonge: Thanks!

pyccl/tracers.py
damonge: Done!

pyccl/tracers.py
damonge: Right now, if bias is `None`, we do not include the density contribution. This can be useful if, for example, you want a magnification-only tracer.

pyccl/tracers.py
damonge: Done

pyccl/tracers.py
damonge: Done

pyccl/tracers.py
damonge: Done

tests/ccl_test_pyccl_interface.py
damonge: It should be easy to add `eval_kernel`, `eval_transfer` and `eval_f_ell` functions here. I'll try to get this done in the next few commits and otherwise will open an issue.

pyccl/ccl_tracers.i
damonge: Done

beckermr:
I'll checkout the tests. you might try rebuilding without the cache on circle ci too.

tilmantroester:
I just ran the tests on my machine (macOS 10.14.5, via `python setup.py test`) and didn't encounter any issues.

damonge:
Thanks a lot @tilmantroester 
I'm really puzzled then about what's happening...

beckermr:
Yes thank you @tilmantroester! IDK how we can fix this if we cannot reproduce this locally. Let me dig into the CI env since I think I was the last person to touch it.

elisachisari:
I can reproduce the failure locally on my Mac.

benchmarks/test_cls.py
```diff
@@ -43,20 +43,26 @@ def set_up(request):
...
+        _, tmp_a2 = np.loadtxt(dirdat + "ia_amp_histo_2.txt", unpack=True)
     else:
         raise ValueError("Wrong Nz type " + nztyp)
     bz = np.ones_like(pz1)
 
+    # Renormalize the IA amplitude to be consistent with A_IA
+    D1 = ccl.growth_factor(cosmo, 1./(1+z1))
+    D2 = ccl.growth_factor(cosmo, 1./(1+z2))
+    a1 = - tmp_a1 * D1 / (5e-14 * ccl.physical_constants.RHO_CRITICAL * cosmo['Omega_m'])
+    a2 = - tmp_a2 * D2 / (5e-14 * ccl.physical_constants.RHO_CRITICAL * cosmo['Omega_m'])
```
beckermr: A question to demonstrate my ignorance. Do we include Omega_nu here? Is that already in Omega_m?

damonge: I think, if we had to worry about neutrinos here, we should also worry about the fact that for neutrinos the growth factor is scale dependent. @elisachisari  will know more though.

This is the reason why I don't like the cosmology dependence of the IA bias, but I agree we should use whatever is more standard.

beckermr: Right. My second question is just one of convention. Does ccl include omega nu in omega_m? This is a redshift dependent statement I suppose since they can be relativistic etc.

elisachisari: Good point! I think we don't include Omega_nu in Omega_m. @c-d-leonard or @sjoudaki would know. We probably include it in the redshift-dependent density functions though.
Regarding IA, I think what we want with this renormalization is to be consistent with the literature, where I haven't seen anyone include the neutrino component so far. I can document the caveats in the note, but it's probably not worth modifying the expression for now.

damonge: I agree with this 100%

Omega_m does not contain neutrinos ([here](https://github.com/LSSTDESC/CCL/blob/6b8f0af672098e55c151b53c9a53789de46d7176/src/ccl_core.c#L340)), but documenting this doesn't need to be done in this PR, I think.

sjoudaki: Briefly chatted with David and Elisa. I think we agree that in principle the massive neutrino component should be included in Omega_m in the equation above, but that its effect is small (2% in shear-intrinsic for 1 eV neutrino) and that it can be resolved later in the issue Anze opened (#647).

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Unreleased
 
 ## C library
+- Changed IA bias normalization to be consistent with A_IA=1 (#630).
+- Implemented generalized models for tracers (#630)
```
beckermr: This is a python change too, right?

benchmarks/test_cls.py
beckermr: resolving this for now.

pyccl/tracers.py
```diff
@@ -0,0 +1,409 @@
...
+                            der_bessel=-1, der_angles=2)
+        if ia_bias is not None:  # Has intrinsic alignments
+            z_a, tmp_a = _check_array_params(ia_bias)
+            # Kernel
+            kernel_i = get_density_kernel(cosmo, dndz)
+            # Normalize so that A_IA=1
+            D = growth_factor(cosmo, 1./(1+z_a))
+            # Transfer
+            rho_m = lib.cvar.constants.RHO_CRITICAL * cosmo['Omega_m']
+            a = - tmp_a * 5e-14 * rho_m / D
```
beckermr: We want a code comment explaining this magic number/formula with a reference to a paper.

damonge:
@elisachisari @sjoudaki : now the tests won't pass because of this silly error (remember, we're calling `growth_factor` now when implementing IAs because of the new parametrization):
```
CCLError: Error 1040: ccl_background.c: ccl_cosmology_compute_growth(): Support for the growth rate in cosmologies with massive neutrinos is not yet implemented.
```

Are you happy with me transforming this error into a warning in the code or even removing the error altogether? Otherwise we need to change our tests to avoid IAs for cosmologies with neutrinos.

damonge:
And I'd like to again voice my discontent with this parametrization (even though I know it's the standard, and it's pointless to fight against it :-( )

damonge:
Also, unless we fix this in the code, rather than the tests, it will come back to bite us in the bottom when we ever want to use CCL to do any analysis with neutrinos

pyccl/tracers.py
damonge: I'm actually not sure where this is coming from, @elisachisari 

If I go by Eq. 5.2 of https://arxiv.org/pdf/1506.02671.pdf, I believe this should be:
```
a = - tmp_a * cosmo['Omega_m'] * 0.0134 / D
```

Where does the 5e-14 come from?

CHANGELOG.md
damonge: Yes, and the IAs actually happens in python. Just fixed this.

pyccl/tracers.py
elisachisari: This is the original value of C_1 from SuperCOSMOS. C1*rho_crit=0.0134 in many works, but we have our own value of rho_crit we can use.

damonge: Aaah, I see. OK, should we just add a comment with a citation to https://arxiv.org/abs/astro-ph/0009499 then?

elisachisari: Yes, adding docs for this now

sjoudaki:
Chatting with @damonge it seems the growth function includes massive neutrinos, just not the scale dependence. As we currently don't account for scale-dependence in the IA equation anyway, David's suggestion seems fine, i.e. for the error message to be turned into a warning.

elisachisari:
I'm happy with the suggestion for the error to be turned into a warning. I am also open to discussing reverting to the old normalization if others find this preferable, in which case we should document this in the open issue regarding IA amplitude #280.

damonge:
Sorry @elisachisari , I was just nagging above. I think it makes sense to keep whatever convention is the most common one (i.e. what you've implemented).

About the growth issue, I would rather fix that in a different PR. For now what I've done is avoid IAs in the tests when neutrinos are present, and all tests are passing now.

damonge:
OK, now everything is passing (including coveralls), and Elisa has addressed the magic numbers issue.
We haven't received comments from other interested parties (@jablazek ?), but I have discussed this in detail with them during the collaboration meeting. So @beckermr , do you think we're ready to merge?

beckermr:
LGTM on merging. We need to make issues about any remaining tasks. I think this is just the growth function error?

beckermr:
https://github.com/LSSTDESC/CCL/issues/648

damonge:
Thanks!

jablazek:
Hi all. Sorry - have continued traveling this week, and just looking at this now (post merge). In general, looks great. On the issue of IA normalization, I think CCL just should at least allow the standard convention to be used (even if it is annoying). With the new generalized tracer structure, where would be best to put in a different z-dependence? Should that go in specifying the IA amplitude itself (an array of A(z) values) or in the generalized transfer functions?

damonge:
Hey @jablazek . No worries. @elisachisari has implemented the standard IA normalization in the custom IA tracer, so I think we're good on that front.

PR #632, opened 2019-04-23
Move all C benchmarks to "benchmarks" directory

### LINKED ISSUE #631: Arrange benchmark checks and unit tests
Move all C benchmark checks, their data and the associated generating code to the `benchmarks` directory. We'll work on translating them into python there. All new benchmarks will need to go there from now on, and should be written in python.

damonge:
All new benchmarks will be implemented there, and in python.
We will gradually work our way through translating all relevant C benchmark checks into python and move all C unit tests back to the `tests` directory.

I also did some general cleanup of the old `tests` folder

Closes #631


damonge:
Done!

PR #637, opened 2019-05-28
BUG use proper pk everywhere for RSD transfer function

### LINKED ISSUE #635: BUG RSD + number counts ignores input power spectrum
It appears the RSD + number counts transfer function is ignoring the input power spectrum. See `ccl_cls.c` around line 722

beckermr:
This PR makes sure to use the proper power spectrum for all of the RSD window functions. 

closes #635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,11 @@
 - Separated Limber and Non-Limber C_ell calculations (#614)
 - Added `has_density` and `has_shear` flags to ClTracers (#614)
 - Simplified C_ell unit tests (#614)
-- Changed TCMB to T_CMB everywhere (#615)	
+- Changed TCMB to T_CMB everywhere (#615)
 - Fixed a small bug in the w_tophat expression and increased precision (#607)
 - Deprecated the use of GSL spline accelerators (#626)
+- Fixed bug in use of the input power spectrum for the RSD part of the number
+  counts transfer function (#637)
 
 ## Python library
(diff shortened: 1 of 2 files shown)


beckermr:
Note this bug is only on master since it involves the new 2d power spectrum interpolation structures.

damonge:
Looks good. Just to clarify that the "bug" was there on purpose for a reason: you can pass any P(k) you want as the python Pk2D object, and it doesn't have to be the matter power spectrum (e.g. you can use this to pass a galaxy HOD power spectrum, which is what we're doing in the LSS HSC analysis). In that case, you want the power spectrum that enters the RSD term to still be the matter power spectrum.

However, the whole thing is inconsistent right now anyway, because with the current formulation you end up multiplying everything by the Pk2D in the end. Since this will all be irrelevant once the generalized tracers are in place (working on it), I'm happy for this to be merged right away.

beckermr:
Ack. Missed the physics here. Should we close this issue?

beckermr:
OK. I am going to close this since it is my stupidity.

damonge:
wait, did you not merge this? I think it would have been fine to do so!

beckermr:
Yeah I decided against it. If you are going to change the code anyways, there is no reason.

damonge:
OK, as long as this doesn't prevent someone's analysis right now, I'm happy

beckermr:
I noticed this while studying the code for adding openmp.

Issue #536, opened 2018-12-12
Kernel crashes for low mass values in ccl.massfunc.


nikfilippas:
Kernel exits with an undefined CCL error for masses < 1e6 and for a=1.
Kernel crashes for masses < 1e6 and for any scale factor other than 1.

Try:
ccl.massfunc(cosmo, 1e5, 1)
ccl.massfunc(cosmo, 1e5, 0.999)

elisachisari:
This is expected. CCL only provides the mass function above 1e6, so it is natural it would crash below. We should implement a graceful exit for this, but I think the models we use for the halo mass function are only valid above this threshold. Tagging @villarrealas to chime in.

damonge:
Yes, I think the main problem here is the ungraceful crashing. In general I think error reporting could be improved in the future.

beckermr:
Yeahhhhhhh. This is really bad

```python
In [5]: ccl.massfunc(COSMO, 1e5, 1, 200)                                                                                                                                                                
---------------------------------------------------------------------------
CCLError                                  Traceback (most recent call last)
<ipython-input-5-ced40bedf01c> in <module>
----> 1 ccl.massfunc(COSMO, 1e5, 1, 200)

~/miniconda3/envs/anl/lib/python3.6/site-packages/pyccl/massfunction.py in massfunc(cosmo, halo_mass, a, overdensity)
     16     """
     17     return _vectorize_fn4(lib.massfunc,
---> 18                           lib.massfunc_vec, cosmo, halo_mass, a, overdensity)
     19 
     20 

~/miniconda3/envs/anl/lib/python3.6/site-packages/pyccl/pyutils.py in _vectorize_fn4(fn, fn_vec, cosmo, x, a, d, returns_status)
    251 
    252     # Check result and return
--> 253     check(status, cosmo_in)
    254     return f

~/miniconda3/envs/anl/lib/python3.6/site-packages/pyccl/core.py in check(status, cosmo)
    680     # Check for unknown error
    681     if status != 0:
--> 682         raise CCLError("Error %d: %s" % (status, msg))

CCLError: Error 1: 

In [6]: ccl.massfunc(COSMO, 1e5, 0.1, 200)                                                                                                                                                              
Segmentation fault: 11
```

elisachisari:
To solve this issue, we simply need to add a check that the user does not ask for a mass below or above our supported range.

beckermr:
This is fixed. Closing.

PR #644, opened 2019-07-19
Added error for mass function when mass is out of range

elisachisari:
Addresses issue #536. Ready to review.

### DIFF
--- a/src/ccl_massfunc.c
+++ b/src/ccl_massfunc.c
@@ -225,7 +225,7 @@ TASK: Outputs fitting function for use in halo mass function calculation;
     ccl_angulo (arxiv 1203.3216 )
     ccl_watson (arxiv 1212.0095 )
     ccl_shethtormen (arxiv 9901122)
-*/
+-*/
 static double massfunc_f(ccl_cosmology *cosmo, double halomass, double a, double odelta, int *status)
 {
   double fit_A, fit_a, fit_b, fit_c, fit_d, fit_p, overdensity_delta;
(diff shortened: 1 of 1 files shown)


src/ccl_massfunc.c
```diff
@@ -573,6 +573,12 @@ double ccl_massfunc(ccl_cosmology *cosmo, double halomass, double a, double odel
 	  ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_growth(): Support for the halo mass function in cosmologies with massive neutrinos is not yet implemented.\n");
 	  return NAN;
   }
+
+  if (halomass > 1e17 || halomass < 1e6){
```
damonge: I think these should be set to `spline_params.LOGM_SPLINE_MIN` and `spline_params.LOGM_SPLINE_MAX`
@villarrealas do you agree?

villarrealas: I forgot that we were making those user accessible. I agree that this change should be made before merging.

elisachisari: Agree! I'll change it

src/ccl_massfunc.c
```diff
@@ -574,7 +574,7 @@ double ccl_massfunc(ccl_cosmology *cosmo, double halomass, double a, double odel
 	  return NAN;
   }
 
-  if (halomass > 1e17 || halomass < 1e6){
+  if (log10(halomass) > cosmo->spline_params.LOGM_SPLINE_MAX || log10(halomass) < cosmo->spline_params.LOGM_SPLINE_MIN){
```
damonge: Could we take log10(halomass) above and assign it to a variable to avoid having to calculate it twice?

elisachisari: Done!

Issue #651, opened 2019-07-30
don't run class twice


beckermr:
It looks like we are running class twice when running cosmologies with sigma8 as opposed to A_s. I don't think we need to do this, but we should make sure and if so, take it out.

PR #652, opened 2019-07-30
#python4lyfe

### LINKED ISSUE #292: Make CLASS optional?
One short-term fix to the CLASS distribution/installation issue could be to make calls to CLASS optional, e.g. through a preprocessor directive: This would allow casual CCL users to check out the functionality of CCL (except for getting power spectra from CLASS), and while keeping the full functionality for users who already have CLASS installed/are willing to install it for CCL.

In practice, it would be possible to modify ccl_power.c such that include "class.h", and routines from "ccl_free_class_structs" through "ccl_cosmology_compute_power_class" are optionally included, with an alternate version for "ccl_cosmology_compute_power_class" that raises an exception if CCL attempts to call CLASS when it is not linked.

This would be very simple to implement on the C level. @damonge, do you know whether it would be an easy modification to autotools to check whether CLASS is installed, and set the preprocessor directive accordingly?

Please comment if you think this is bad style, or can think of a better short-term solution!

### LINKED ISSUE #642: ENH integrate CLASS in python
Per the discussion on the telecon. Related to #226

beckermr:
This PR changes the code to call CLASS from python.

Closes #642 
Closes #292 

@philbull @damonge for viz


beckermr:
I have to move all of the C benchmarks to python before this. Blarg. :(

benchmarks/test_power_nu.py
```diff
@@ -3,7 +3,7 @@
 
 import pyccl as ccl
 
-POWER_NU_TOL = 1.0E-3
+POWER_NU_TOL = 2.0E-2
```
beckermr: Gotta figure this out...

pyccl/boltzmann.py
```diff
@@ -0,0 +1,136 @@
...
+        params["A_s"] = A_s_fid
+
+    model = None
+    try:
+        model = classy.Class()
+        model.set(params)
+        model.compute()
+
+        # if we are using sigma8, we rerun the code with a new A_s
+        # FIXME - I don't think rerunning class is needed...
```
damonge: I think you're right if now we're computing our own halofit.

pyccl/boltzmann.py
```diff
@@ -0,0 +1,136 @@
...
+            model.compute()
+
+        # Set k and a sampling from CCL parameters
+        nk = lib.get_pk_spline_nk(cosmo.cosmo)
+        na = lib.get_pk_spline_na(cosmo.cosmo)
+        status = 0
+        a_arr, status = lib.get_pk_spline_a(cosmo.cosmo, na, status)
+        check(status)
+
+        # FIXME - getting the lowest CLASS k value from the python interface
```
damonge: If we're now querying from classy, do we really need do do this?

CMakeLists.txt
```diff
@@ -47,27 +47,15 @@ set(EXTRA_DIST_DIRS)
...
-# otherwise download and compile CLASS
-if(EXTERNAL_CLASS_PATH)
-    message("Using external install of CLASS")
-    set(CLASS_LIBRARY_DIRS ${EXTERNAL_CLASS_PATH})
-    set(CLASS_INCLUDE_DIRS ${EXTERNAL_CLASS_PATH}/include)
-    set(CLASS_LIBRARIES -lclass)
-    set(CLASS_EXTERNAL True)
-else(EXTERNAL_CLASS_PATH)
-    include(BuildCLASS)
-endif(EXTERNAL_CLASS_PATH)
```
damonge: Is it now not possible to couple to a custom version of CLASS?

beckermr: It is possible. You simply need to modify the CLASS package upstream and install it.

pyccl/boltzmann.py
beckermr: Awesome! I've opened another issue. I will address this there. #651

pyccl/boltzmann.py
beckermr: Yeah, unfortunately, we do. The benchmarks fail because CCL uses extrapolation. So if we set the min k value too high, we don't pass because the benchmark data was *not* interpolated, whereas as CCL's prediction is interpolated. Thus we need to set it low enough to match the old CCL way of doing things. I could not find any way to get the actual value out of the CLASS python wrapper, so I hardcoded something reasonable.

beckermr:
I agree on the docs. I am making a separate set of PRs for the docs. I will update the README.md, but I am going to leave the rest of them below since I will be changing them soon.

beckermr:
OK @damonge. I've made the requested changes. Back to you!

pyccl/boltzmann.py
damonge: OK, I see. This is kind of silly, since at small k we know that P_k~k^ns almost exactly, and this would be much better than extrapolating.

beckermr: agreed but here we are - we can redo the benchmarks I suppose

PR #653, opened 2019-07-31
REF python benchmarks and tests

### LINKED ISSUE #410: Unit Tests to Check Basic Consistency Thoroughly
Recently, an issue was brought up that most of `ccl.massfunc` can have very non-intuitive failures in `pyccl`, which has been narrowed down to issues in the `ccl_compute_cosmology_power` call.

Specifically:

`cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7, n_s=0.96, sigma8=0.8)`
leads to a nan result for functions such as `ccl.massfunc` while

`cosmo=ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7, n_s=0.96, sigma8=0.8, transfer_function='bbks')`
does not.

This is currently not captured in our existing tests due to the fact that the benchmarks for the halo mass function (and apparently anything else calling sigmaR, which I believe is where this pops up) does not utilize our default 'class' transfer function. While the resulting error handling is another concern entirely, I think this does raise an issue: this behavior is not captured at all in any of our test procedures.

I suggest that a good hack week problem would be to generate a large number of unit tests - not to compare against benchmarks, but to make certain that common function calls do not outright crash.  For example, just making sure that varying transfer function method or matter power spectrum method does not lead to nans *without* appropriate error messaging. I think it is reasonable that some combinations are unsupported, but then a clear error message should be available to acknowledge this. In cases where it should work, we do not necessarily have benchmark values, but testing that it returns a sane value (aka not nan or infinity) seems straightforward enough.

These would have to be filed under slow tests (however that works), but it would be worth running before uploading things to be a pip install or the like.

### LINKED ISSUE #470: better unit tests in file `test/ccl_test_pyccl_interface.py`
@rmjarvis pointed out that the unit tests in this file are pretty poor. We should beef them up!

### LINKED ISSUE #477: faster test suite

### LINKED ISSUE #559: move to pytest and package tests with code
`nose` is basically deprecated at this point (https://nose.readthedocs.io/en/latest/#note-to-users)

We should move to pytest which will help us mark tests as slow and package the tests with the code for downstream users to test their installations.

### LINKED ISSUE #595: clean up tests
from @vitenti

> The use of global variables in ccl_test_distances.py makes the script hard to read and error prone. For example, the variables Neff and Neff_mnu are defined globally early on in the script. Then, we have them being used in different contexts:
> - Neff as an argument to other function. It is used in compare_distances, compare_distances_hiz;
> - Neff as an argument of compare_distances_mnu, which shadows the global Neff, but then the argument is ignored and the global Neff_mnu is used;
> - Neff_mnu as an argument of compare_distances_mnu_hiz, which shadows the global Neff_mnu, but then it is ignored and the global Neff is used;
> - Neff as an argument of compare_class_distances which has a default value and it is correctly used
> 
> It was time consuming to read the script and disentangle all variables to see which parametrizations are actually tested.

beckermr:
This PR refactors all of the benchmarks and tests to be in python where possible and to use pytest.

Features and results:
1. The core unit tests in the code now run in ~30 seconds instead of ~5 minutes or so. The entire CI build is at 10 minutes, down from 20 or more.
2. The benchmarks are now all in python.
3. We can now measure Python source code coverage with the benchmarks too.
4. With the new unit tests and the benchmarks, source coverage is now at ~96%. 
5. The unit tests are now better and test more than simply if the code runs where easily feasible.

Closes #595 
Closes #559 
Closes #470 
Closes #477 
Closes #410


benchmarks/test_correlation_MG.py
```diff
@@ -7,7 +7,7 @@
 
 @pytest.fixture(scope='module', params=['fftlog', 'bessel'])
 def corr_method(request):
-    errfacs = {'fftlog': 0.2, 'bessel': 0.2}
+    errfacs = {'fftlog': 0.22, 'bessel': 0.22}
```
beckermr: This tolerance had to go up just a little. NBD.

damonge:
I'll try to get this reviewed between today and tomorrow. Sorry for the delay @beckermr

benchmarks/test_bbks.py
```diff
@@ -0,0 +1,39 @@
+import numpy as np
+import pyccl as ccl
+import pytest
+
+BBKS_TOLERANCE = 1.0e-5
```
elisachisari: Have you preserved all tolerances from the C code?

benchmarks/test_correlation.py
```diff
@@ -17,10 +17,9 @@ def set_up(request):
     dirdat = os.path.dirname(__file__) + '/data/'
     cosmo = ccl.Cosmology(Omega_c=0.30, Omega_b=0.00, Omega_g=0, Omega_k=0,
                           h=0.7, sigma8=0.8, n_s=0.96, Neff=0, m_nu=0.0,
-                          w0=-1, wa=0, transfer_function='bbks',
+                          w0=-1, wa=0, T_CMB=2.7, transfer_function='bbks',
```
elisachisari: Was T_CMB added in this PR?

benchmarks/test_correlation_MG.py
```diff
@@ -18,11 +18,10 @@ def set_up(request):
     logA = 3.05  # log(10^10 A_s)
     cosmo = ccl.Cosmology(Omega_c=0.12/h0**2, Omega_b=0.0221/h0**2, Omega_k=0,
                           h=h0, A_s=np.exp(logA)/10**10, n_s=0.96, Neff=3.046,
-                          m_nu=0.0, w0=-1, wa=0, mu_0=0.1, sigma_0=0.1,
+                          m_nu=0.0, w0=-1, wa=0, T_CMB=2.7255,
```
elisachisari: Has @c-d-leonard confirmed this is the value of T_CMB used for this benchmark?

pyccl/core.py
```diff
@@ -312,12 +312,19 @@ class Cosmology(object):
...
+        T_CMB (:obj:`float`): The CMB temperature today. The default of
+            ``None`` uses the global CCL value in
+            ``pyccl.physical_constants.T_CMB``.
         bcm_log10Mc (:obj:`float`, optional): One of the parameters of the
             BCM model. Defaults to `np.log10(1.2e14)`.
         bcm_etab (:obj:`float`, optional): One of the parameters of the BCM
             model. Defaults to 0.5.
         bcm_ks (:obj:`float`, optional): One of the parameters of the BCM
             model. Defaults to 55.0.
+        mu_0 (:obj:`float`, optional): One of the parameters of the mu-Sigma
```
elisachisari: Why is this showing up as a new line here? Didn't the MG PR bring it in already?

pyccl/pyutils.py
```diff
@@ -27,44 +27,44 @@ def debug_mode(debug):
...
-        # Use vectorised function
-        if returns_status:
-            f, status = fn_vec(x, len(x), status)
-        else:
-            f = fn_vec(x, len(x))
-
-    # Check result and return
-    check(status)
-    return f
+# def _vectorize_fn_simple(fn, fn_vec, x,
```
elisachisari: Why has all this been commented out?

benchmarks/test_halomod.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
elisachisari: There is a test_halomod and a test_halomodel. It seems one does the benchmarking and the other the unit test. It might be useful to have this clarification in the name of the file. E.g., test_halomodel_unit and test_halomodel_benchmark. This would also make it more evident to developers that they need to provide both type of tests.

benchmarks/test_bbks.py
beckermr: Yes except the on I specifically called out in my comments.

benchmarks/test_correlation.py
beckermr: Yes. We were not updating in properly when mucking with the value for the parameters.

benchmarks/test_correlation_MG.py
beckermr: I copied it out of the benchmarks that were in C, so yes it is correct.

pyccl/core.py
beckermr: Nope this was missed.

pyccl/pyutils.py
beckermr: We don't use it and it is decreasing the test coverage making it less accurate.

benchmarks/test_halomod.py
beckermr: I am not following. Unit tests live with the code and benchmarks live in the benchmarks folders. Their names being similar doesn't show up when you browse the file tree. Is there something I am missing?

pyccl/core.py
elisachisari: So how was it possible to call an MG cosmology from python then?

pyccl/pyutils.py
elisachisari: Should we remove it then?

elisachisari:
Tests/Benchmark checks are running without problems in my OSX system.

pyccl/core.py
beckermr: This is a change to the doc string, not the object signature.

pyccl/pyutils.py
beckermr: I'm ok either way. There was a comment above the function that motivated keeping it around for people to look at. IDK where that came from, but this is why I went with commenting it out instead of deleting.

beckermr:
There are still two C-level unit tests in the code. I did not move those to python since they made more sense being in C.

elisachisari:
Yes, C tests also working.

beckermr:
OK @elisachisari. I added minimal changes to the docs for the new dev workflow. I want to defer major changes for when we update the docs for v2.

PR #661, opened 2019-08-03
ENH openmp, C cleanup, simpler C error handling

### LINKED ISSUE #634: ENH add OpenMP threading to the code
There are a few places where this would be helpful.

- [x] computation of the angular Cls (timing tests on benchmarks show nearly perfect scaling)
- [x] ~splining of the halofit parameters~ (not worth it turns out)
- [x] ~computation of the lensing kernels~ (scratch that - doesn't save enough time).
- [x] computation of the various correlation functions via Bessel integration

### LINKED ISSUE #662: simplify C layer warnings etc
We have a lot of unused/poorly used (and confusing) infrastructure for warnings and errors at the C layer. I think we should simplify to a single, consistently used debug mode toggle that always prints C layer errors/warnings to stderr.

### LINKED ISSUE #663: check memory allocation
We are not checking memory allocation everywhere.

### LINKED ISSUE #664: fix memory allocation in ccl_neutrinos.c
mnu is allocated and then not freed in certain cases. Relevant to issue #663 .

### LINKED ISSUE #667: manage internal data with python
The use of openmp has opened a can of worms on how to manage the internal data associated with the cosmology object. I think the best solution here is to manage it from python, placing the required init calls there. This way any C function can simply assume the data is there and ready to use. Any errors with the init won't ever get past the Python error that is raised. This will require us to put the appropriate init statements in the python wrapper. All such statements needed by a given invocation of the C code will need to be there as well. The c layer can check and return an error if an init statement has not been called.

beckermr:
This PR introduces OpenMP and a bunch of simplifications to the C library. I've also fixed several instances of unchecked memory allocation.

Closes #634 
Closes #662 
Closes #663 
Closes #664
Closes #667


src/ccl_power.c
```diff
@@ -431,46 +434,69 @@ TASK: compute power spectrum
...
 {
-
-  if ( (cosmo->config.transfer_function_method != ccl_boltzmann_class) && (fabs(cosmo->params.mu_0)>1e-14 || fabs(cosmo->params.sigma_0)>1e-14)){
-	  *status = CCL_ERROR_NOT_IMPLEMENTED;
-	  strcpy(cosmo->status_message,"ccl_power.c: ccl_cosmology_compute_power(): The power spectrum in the mu / Sigma modified gravity parameterisation is only implemented with the ccl_boltzmann_class power spectrum method.\n");
-	  return;
+  if ((cosmo->config.transfer_function_method != ccl_boltzmann_class &&
+       cosmo->config.transfer_function_method != ccl_transfer_none) &&
+      (fabs(cosmo->params.mu_0) > 1e-14 || fabs(cosmo->params.sigma_0) > 1e-14)) {
+    #pragma omp atomic write
```
beckermr: This should be removed.

include/ccl_cls.h
```diff
@@ -33,11 +35,11 @@ double ccl_angular_cl_limber(ccl_cosmology *cosmo,
...
 void ccl_angular_cls_nonlimber(ccl_cosmology *cosmo,
-			       ccl_cl_tracer_collection_t *trc1,
-			       ccl_cl_tracer_collection_t *trc2,
-			       ccl_f2d_t *psp,
-			       int nl_out,int *l_out,double *cl_out,
-			       int *status);
+         ccl_cl_tracer_collection_t *trc1,
+         ccl_cl_tracer_collection_t *trc2,
+         ccl_f2d_t *psp,
+         int nl_out,int *l_out,double *cl_out,
```
tmcclintock: Add spaces to the arguments.

beckermr: bug @damonge about this ;P

beckermr: I've fixed as much as I can but I did not get everything.

tmcclintock: Sounds good. I spotted a bunch of other potential formatting improvements, but figured you didn't want to hear about them.

beckermr: Right. I've been looking for a good C linter but have not found one.

benchmarks/test_cls.py
```diff
@@ -1,11 +1,14 @@
...
 import pyccl as ccl
+import time
 import os
 import pytest
 
 
 @pytest.fixture(scope='module', params=['analytic', 'histo'])
 def set_up(request):
+    t0 = time.time()
+
```
damonge: Do you intend to leave these here or were you gonna remove the timings?

include/ccl_cls.h
damonge: I do couldn't care less. I personally prefer no spaces and consistent indentation (the latter is actually quite nightmarish right now).

pyccl/cls.py
```diff
@@ -27,6 +27,9 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell, p_of_k_a=None,
             :math:`C_\\ell`, for the pair of tracers, as a function of
             :math:`\\ell`.
     """
+    # we need the distances for the integrals
+    cosmo.compute_distances()
```
damonge: So it is gonna be kind of annoying if there's no way for the cosmo object to realize that it needs to compute stuff at the C level. Can you clarify why this is necessary?

pyccl/tracers.py
```diff
@@ -292,6 +298,9 @@ class NumberCountsTracer(Tracer):
     def __init__(self, cosmo, has_rsd, dndz, bias, mag_bias=None):
         self._trc = []
 
+        # we need the distance functions at the C layer
+        cosmo.compute_distances()
+
```
damonge: Do you also need to initialize growth if you need RSDs? Maybe it happens automatically.

pyccl/tracers.py
```diff
@@ -292,6 +298,9 @@ class NumberCountsTracer(Tracer):
     def __init__(self, cosmo, has_rsd, dndz, bias, mag_bias=None):
         self._trc = []
 
+        # we need the distance functions at the C layer
+        cosmo.compute_distances()
```
damonge: Why do you initialize this explicitly here? I thought you were doing that in the kernel calls already.

benchmarks/test_cls.py
beckermr: They were very useful for testing the openmp. Happy to remove them if you’d like.

pyccl/cls.py
beckermr: With threading at the c level it is basically impossible to properly sync the global state in the cosmology object. By moving all of that to python, we can avoid this. Additionally future dev will require that all global state be initialized at the python level before calling into C. So this won’t be annoying as much as just the way things work. There are right now a few exceptions to this in the code. These will be removed over time.

pyccl/tracers.py
beckermr: It depends on the ordering of the calls. This call is a no-op if the init has already been done, so no harm here.

pyccl/tracers.py
beckermr: Where is the growth function called? 

Note that calling the growth function in python does the init but it does not happen in C.

pyccl/cls.py
beckermr: To be a bit more clear, the idea is that when coding in C with threads, you never have to worry about the cosmology state being change by another thread. You do have to test for the data, but if it is not there you simply set an error and move on.

benchmarks/test_cls.py
damonge: no, just checking it was intentional

pyccl/cls.py
damonge: ok, I'm sure you're right. Let me look at the C code and I'll get back to this one.

pyccl/tracers.py
damonge: Line 321, but you've already answered this, so all good

src/ccl_background.c
```diff
@@ -434,153 +452,176 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
-  }
-
-  //Check for messed up scale factor conditions
-  if (!*status){
-    if ((fabs(a[0]-cosmo->spline_params.A_SPLINE_MINLOG)>1e-5) ||
-	(fabs(a[na-1]-cosmo->spline_params.A_SPLINE_MAX)>1e-5) ||
-	(a[na-1]>1.0)) {
-      *status = CCL_ERROR_LINSPACE;
-      ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_distances(): Error creating first logarithmic and then linear spacing in a\n");
+  #pragma omp master
```
damonge: I thought these functions were only gonna ever be called from python (i.e. always outside of openmp loops). If that's the case, why the pragma?

src/ccl_background.c
```diff
@@ -590,210 +631,258 @@ TASK: if not already there, make a table of growth function and growth rate
...
-			      cosmo->params.nz_mgrowth);
-
-    if(chistatus) {
-      free(a);
-      free(df_arr);
-      gsl_spline_free(df_z_spline);
-      *status = CCL_ERROR_SPLINE;
-      ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_growth(): Error creating Delta f(z) spline\n");
-      return;
+  #pragma omp master
```
damonge: same here

src/ccl_background.c
```diff
@@ -901,13 +995,15 @@ double ccl_comoving_angular_distance(ccl_cosmology * cosmo, double a, int* statu
...
     return NAN;
   }
   else {
     if (!cosmo->computed_distances) {
-      ccl_cosmology_compute_distances(cosmo, status);
-      ccl_check_status(cosmo, status);
+      *status = CCL_ERROR_DISTANCES_INIT;
+      ccl_cosmology_set_status_message(
+        cosmo,
+        "ccl_background.c: ccl_comoving_angular_distance(): distance splines have not been prcomputed!");
```
damonge: typo

src/ccl_background.c
```diff
@@ -955,17 +1051,18 @@ double ccl_distance_modulus(ccl_cosmology * cosmo, double a, int* status)
...
-    ccl_check_status(cosmo,status);
     return NAN;
   } else {
     if (!cosmo->computed_distances) {
-      ccl_cosmology_compute_distances(cosmo, status);
-      ccl_check_status(cosmo, status);
+      *status = CCL_ERROR_DISTANCES_INIT;
+      ccl_cosmology_set_status_message(
+        cosmo,
+        "ccl_background.c: ccl_distance_modulus(): distance splines have not been prcomputed!");
```
damonge: typo

src/ccl_background.c
```diff
@@ -1029,27 +1128,31 @@ void ccl_scale_factor_of_chis(ccl_cosmology * cosmo, int nchi, double chi[], dou
...
     return NAN;
   }
   else {
     if (!cosmo->computed_growth){
-      ccl_cosmology_compute_growth(cosmo, status);
-      ccl_check_status(cosmo, status);
+      *status = CCL_ERROR_GROWTH_INIT;
+      ccl_cosmology_set_status_message(
+        cosmo,
+        "ccl_background.c: ccl_growth_factor(): growth factor splines have not been prcomputed!");
```
damonge: typo

pyccl/cls.py
beckermr: I did some experimenting with allowing this to happen. The issue comes when you have to sync the data. Basically you need a mechanism to 1) lock the Cosmo structure so you can modify it and 2) block the execution of other threads until the data is updated. I am sure we could work out how to do this, but it would be complicated and bug prone. By forcing the init to only happen from python when the code is single threaded, we effectively use the GIL to take care of both of these things. It also makes the programming model a lot simpler.

src/ccl_background.c
beckermr: Hold overs from previous versions of this. Happy to remove. They don’t do any harm here.

beckermr:
anaconda inc broke the osx compilers :/

should be fixed soon

damonge:
OK, back on this today. Sorry for the hiatus.

src/ccl_cls.c
```diff
@@ -18,202 +18,271 @@ typedef struct{
...
-  if(gslstatus == GSL_EROUND) {
-    ccl_raise_gsl_warning(gslstatus, "ccl_cls.c: ccl_angular_cl_limber(): "
-			  "Default GSL integration failure, attempting backup method.");
-    gsl_integration_cquad_workspace *w_cquad= \
-      gsl_integration_cquad_workspace_alloc(cosmo->gsl_params.N_ITERATION);
-    size_t nevals=0;
-    gslstatus=gsl_integration_cquad(&F, lkmin, lkmax, 0,
-				    cosmo->gsl_params.INTEGRATION_LIMBER_EPSREL,
-				    w_cquad, &result, &eresult, &nevals);
+  #pragma omp parallel private(lind, clastatus, ipar, lkmin, lkmax, result, \
```
damonge: I usually find it clearer to just declare all private variables inside the parallel `{ }` and never use the `private` clause.
Not asking to modify anything, but rather if you have any thoughts.

src/ccl_cls.c
```diff
@@ -18,202 +18,271 @@ typedef struct{
...
   return k*pk*d1*d2;
 }
 
-double ccl_angular_cl_limber(ccl_cosmology *cosmo,
-			     ccl_cl_tracer_collection_t *trc1,
-			     ccl_cl_tracer_collection_t *trc2,
-			     ccl_f2d_t *psp,double l,int *status)
-{
-  //Figure out which power spectrum to use
+void ccl_angular_cl_limber(ccl_cosmology *cosmo,
```
damonge: should we rename this to `angular_cls_limber`?

src/ccl_cls.c
```diff
@@ -18,202 +18,271 @@ typedef struct{
...
+      ipar.cosmo = cosmo;
+      ipar.trc1 = trc1;
+      ipar.trc2 = trc2;
+      ipar.psp = psp_use;
+      ipar.status = &clastatus;
+      F.function = &cl_integrand;
+      F.params = &ipar;
+    }
+
+    #pragma omp for
```
damonge: Have you tested if `schedule(dynamic)` helps here? E.g. high-ell and low-ell integrals take different amounts of time.

src/ccl_correlation.c
```diff
@@ -194,81 +193,103 @@ static double corr_bessel_integrand(double l,void *params)
...
+
+    w = gsl_integration_workspace_alloc(cosmo->gsl_params.N_ITERATION);
+
+    if (w == NULL) {
+      local_status = CCL_ERROR_MEMORY;
+    }
+    F.function = &corr_bessel_integrand;
+    F.params = &cp;
+
+    #pragma omp for
```
damonge: same comment about `schedule(dynamic)` here. I bet this actually matters for the bessel integrator, since for larger thetas the bessel functions oscillate faster. This may be a moot point though, since I doubt we'll use the bessel integrator in practice.

src/ccl_correlation.c
```diff
@@ -382,7 +402,7 @@ static void ccl_tracer_corr_legendre(ccl_cosmology *cosmo,
       wtheta[i]=0;
```
damonge: there's no parallelization here. Is this intentional? As above, I doubt we'll use this in practice, and even if we do, it's probably not super-high priority that we parallelize this now. Just making sure this is not an oversight.

src/ccl_emu17.c
```diff
@@ -35,111 +35,119 @@ static double lamws[2][28];
...
+  gsl_vector_free(b);
 
+  return status;
 } // emuInit()
 
 // Actual emulation
-void ccl_pkemu(double *xstar, int sizeofystar, double *ystar, int* status, ccl_cosmology* cosmo)
-{
+void ccl_pkemu(double *xstar, int sizeofystar, double *ystar, int* status, ccl_cosmology* cosmo) {
+  #pragma omp master
```
damonge: Do we need this?

src/ccl_emu17.c
```diff
@@ -149,174 +157,195 @@ void ccl_pkemu(double *xstar, int sizeofystar, double *ystar, int* status, ccl_c
...
+      // Convert to P(k)
+      for(i=0; i<NK_EMU; i++) {
+        ystar[i] = ystar[i] - 1.5*log10(mode[i]) + log10(2) + 2*log10(M_PI);
+        ystar[i] = pow(10, ystar[i]);
+      }
     }
+
+    gsl_spline_free(zinterp);
+  }
+  #pragma omp flush
```
damonge: connected to the previous one

src/ccl_background.c
```diff
@@ -314,20 +324,29 @@ void compute_chi(double a, ccl_cosmology *cosmo, double * chi, int * stat)
...
-    &F, a, 1.0, 0.0, cosmo->gsl_params.INTEGRATION_DISTANCE_EPSREL, workspace, &result, NULL, NULL);
-  *chi=result/cosmo->params.h;
-  gsl_integration_cquad_workspace_free(workspace);
 
-  if (gslstatus != GSL_SUCCESS) {
-    ccl_raise_gsl_warning(gslstatus, "ccl_background.c: compute_chi():");
-    *stat = CCL_ERROR_COMPUTECHI;
+  workspace = gsl_integration_cquad_workspace_alloc(cosmo->gsl_params.N_ITERATION);
+
+  if (workspace == NULL) {
```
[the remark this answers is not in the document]
beckermr: I have an issue open already. #663 

I cannot claim I fixed all of them, but I did get quite a few and read through each source file. So I think we can call this good for now.

src/ccl_cls.c
beckermr: I don't have strong thoughts either way here. This is how I've always done things so I just continued to do them that way. I did not know that OpenMP was smart enough to understand the intent if you declared vars inside the parallel block!

damonge: OK, sounds good. It's smart enough as long as you use `default(none)`, which I think is always necessary for safety anyway.

src/ccl_background.c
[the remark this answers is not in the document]
damonge: OK, great, thanks. I missed that issue.

src/ccl_correlation.c
beckermr: I've been trying to add openmp only where it seems to help. I don't think I chcked this section of code, so this was def an oversight. I will take a look.

damonge: ok, great. As I said, zero priority, I think.

src/ccl_emu17.c
beckermr: nope - good catch!

src/ccl_background.c
```diff
@@ -590,210 +627,254 @@ TASK: if not already there, make a table of growth function and growth rate
...
+
+  if (*status == 0) {
+    growth = gsl_spline_alloc(cosmo->spline_params.A_SPLINE_TYPE, na);
+    fgrowth = gsl_spline_alloc(cosmo->spline_params.A_SPLINE_TYPE, na);
+
+    if (growth == NULL || fgrowth == NULL) {
+      *status = CCL_ERROR_MEMORY;
       ccl_cosmology_set_status_message(
-        cosmo, "ccl_background.c: ccl_cosmology_compute_growth(): integral for MG growth factor didn't converge\n");
+          cosmo, "ccl_background.c: ccl_cosmology_compute_distances(): ran out of memory\n");
```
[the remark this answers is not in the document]
beckermr: yes.

src/ccl_cls.c
beckermr: yup

beckermr:
@damonge I have made this issue: https://github.com/LSSTDESC/CCL/issues/672

We agreed to not make issues for the memory stuff.

src/ccl_correlation.c
beckermr: This is done.

beckermr:
Alright, @damonge. All of the comments have been addressed. Back to you!

PR #676, opened 2019-09-04
ENH all the warnings

### LINKED ISSUE #648: error to warning for linear growth function and neutrinos
Right now the linear growth function throws an error for cosmologies with massive neutrinos. We should make this a warning.

### LINKED ISSUE #675: ENH warnings for inconsistent models
We agreed that CCL should throw warnings when the models are not consistent. This issue is a start at listing all of the ways the models in CCL are inconsistent.

- [x] halofit is not consistent with the modified growth cosmologies (both `dfg` and `mu-Sigma`)
- [x] the growth function is not consistent for cosmologies with neutrinos
- [x] the BCM model is not consistent with the mass function models
- [x] the BCM model is not consistent with the halo concentration models
- [x] the BCM model is not consistent with the halo bias parametrizations
- [x] neutrino cosmologies are broadly inconsistent with the mass function models
- [x] neutrino cosmologies are broadly inconsistent with the halo concentration models
- [x] neutrino cosmologies are broadly inconsistent with the halo bias models
- [x] neutrino cosmologies are not consistent with the BCM model 
- [x] neutrino cosmologies are not consistent with any of the modified growth models
- [x] non-flat models are not consistent with some of the projected statistic computations (see the CCL note)
- [x] none of the linear power spectra we have are consistent with the effective linear power spectrum assumed by the emulator at high-precision
- [ ] cosmological parameters outside halofit
- [ ] cosmological parameters outside ranges expected by CLASS or CAMB

@joezuntz @slosar for viz

beckermr:
This PR does all of the warnings. 

Closes #675 
Closes #648


beckermr:
@joezuntz here is what I have

1. I am using a custom registry for each cosmology object. This means we will reprint warnings as new models are made. 

2. I have put the warnings inside the init methods for the method data. This centralizes them in one spot. Every function that returns something (e.g., `compute_growth`) calls one of these methods. Thus the warnings are lazy and print as needed.

3. I have protected the init functions with a check to se if they have been called before. This eliminates duplicate warnings in ipython. 

The custom registry is not needed if we want the warnings to print only once globally. It is the init protection in 3 that eliminates the duplicates.

joezuntz:
Do we definitely want this to warn again for each new Cosmology object?  One common use case for this would be an MCMC where we will construct many instances in a long sequence.

beckermr:
Happy to warn once, but it won’t work in ipython.

joezuntz:
Could we not make the warning registry a class variable instead of an instance variable?

beckermr:
That’s a good idea!

beckermr:
Ok so the class attribute doesn't help because ipython is resetting the warning filters version. :/

beckermr:
Alright the warnings are done. I did not do all of the suggestions in the issue because some were not well defined (e.g., parameter ranges for halofit/CLASS).

beckermr:
I've also punted on including tests since these are non-critical and seem more likely to change in the future.

beckermr:
@joezuntz @slosar this is ready for review: https://github.com/LSSTDESC/CCL/pull/676

pyccl/core.py
```diff
@@ -737,27 +738,65 @@ def __repr__(self):
...
         status = 0
         status = lib.cosmology_compute_distances(self.cosmo, status)
         check(status, self)
 
     def compute_growth(self):
         """Compute the growth function."""
+        if self.has_growth:
+            return
+
+        if np.sum(self._params_init_kwargs['m_nu']) > 0:
```
joezuntz: Is the (practical) limit for this really zero, or could we e.g. reasonable accurately compute it for the minimal mass?  That would seem likely to be a common use case.

pyccl/core.py
```diff
@@ -789,7 +857,25 @@ def compute_nonlin_power(self):
 
     def compute_sigma(self):
         """Compute the sigma(M) and mass function splines."""
+        if self.has_sigma:
+            return
+
         # we need these things before building the mass function splines
+        if np.sum(self._params_init_kwargs['m_nu']) > 0:
```
joezuntz: This combination has come up enough - and is useful enough to users - that I'm wondering if there should be a (presumably read-only) property for it?

setup.py
```diff
@@ -1,38 +1,49 @@
 #!/usr/bin/env python
 from setuptools.command.build_py import build_py as _build
+from setuptools.command.develop import develop as _develop
```
joezuntz: Are these changes to setup.py supposed to be part of this PR?  They look unconnected.

beckermr: Yes.  I made them while debugging. Happy to open a new pr if it really matters to you.

pyccl/core.py
beckermr: Good idea! I've switched to use the number of massive neutrinos that CCL computes internally.

pyccl/core.py
beckermr: Yup. Using the number of massive neutrinos that CCL uses internally addresses this.

beckermr:
OK @joezuntz, I think I have addressed everything. Back to you!

setup.py
joezuntz: No, that's fine, just wanted to check.  I don't know much about setuptools.

PR #684, opened 2019-10-01
check Nnurel is not negative

c-d-leonard:
This PR adds a check to make sure that you can't accidentally create a cosmology with a combination of mnu and Neff which gives negative number of relativistic neutrinos in the early universe. 

To accomplish this in a way where we get an appropriate error in python, we move the code which takes a sum of neutrino masses and splits it into three masses according to the requested convention from C into python, in `core.py`.

I also add a test in `test_parameters.py` which checks if this error gets thrown when appropriate, and remove two tests from `test_distances.py` which were using cosmologies which had negative relativistic neutrinos by mistake.

(The computer I am using doesn't have flake8 and I don't have admin right ie can't install it, so this will probably initially fail the travis tests on those grounds.)

Addresses Issue #593


pyccl/ccl_core.i
```diff
@@ -31,11 +31,11 @@ ccl_parameters parameters_create_nu(
                         double Neff, double w0, double wa, double h,
                         double norm_pk, double n_s, double bcm_log10Mc,
                         double bcm_etab, double bcm_ks, double mu_0,
-                        double sigma_0, int mnu_is_sum,
-                        double* m_nu, int n_m, int* status)
+                        double sigma_0, double* m_nu, int n_m, int* status)
 {
+    printf("n_m=%d\n", n_m);
```
beckermr: This needs to be removed

pyccl/core.py
```diff
@@ -236,11 +236,18 @@
     'equalize': lib.emu_equalize
 }
 
-m_nu_types = {
+"""m_nu_types = {
     'list': lib.mnu_list,
     'normal': lib.mnu_sum,
     'inverted': lib.mnu_sum_inverted,
     'equal': lib.mnu_sum_equal,
+}"""
```
beckermr: We can remove this old code. No reason to keep it.

pyccl/core.py
```diff
@@ -236,11 +236,18 @@
...
+
+m_nu_types = {
+    'list': 0,
+    'normal': 1,
+    'inverted': 2,
+    'equal': 3,
```
beckermr: Are these numeric codes used anywhere?

pyccl/core.py
```diff
@@ -374,6 +381,7 @@ def _build_cosmo(self):
         # We have to make all of the C stuff that goes into a cosmology
         # and then we make the cosmology.
         self._build_parameters(**self._params_init_kwargs)
+
```
beckermr: Please remove white space changes.

pyccl/core.py
```diff
@@ -562,6 +571,9 @@ def _build_parameters(
                     "of neutrino masses, only with a sum." % m_nu_type)
             elif m_nu_type is None:
                 m_nu_type = 'list'  # False
+            mnu_list = [0]*3
+            for i in range(0,3):
+                mnu_list[i] = m_nu[i]
```
beckermr: What is the purpose of this block of code?

pyccl/core.py
```diff
@@ -574,19 +586,88 @@ def _build_parameters(
...
+        # Fill an array with the non-relativistic neutrino masses
+        if N_nu_mass>0:
+            mnu_final_list=[0]*N_nu_mass
+            relativistic = [0]*3
+            for i in range(0, N_nu_mass):
+                for j in range(0,3):
+                    if (mnu_list[j]>0.00017 and relativistic[j]==0):
+				        relativistic[j]=1
+				        mnu_final_list[i] = mnu_list[j]
+				        break
```
beckermr: Why are we recording `relativistic` here? Is that used anywhere?

src/ccl_core.c
```diff
@@ -533,7 +547,7 @@ ccl_parameters ccl_parameters_create(
 	  params.m_nu[0] = 0.;
   }
   // Free mnu_in
-  if (mnu_in != NULL) free(mnu_in);
+  if (mnu_in != NULL) free(mnu_in);*/
```
beckermr: Please remove old commented code.

beckermr:
Also yeah, flake8 is going to barf. You can do a user install of flake8 easily with pip. You could also put a conda env in your home environment.

pyccl/core.py
c-d-leonard: Removed

pyccl/ccl_core.i
c-d-leonard: removed

src/ccl_core.c
beckermr: you must have forgotten to push a file or something. The code is still in the diff.

include/ccl_core.h
```diff
@@ -307,13 +307,13 @@ typedef struct ccl_cosmology {
...
 // Label for whether you are passing a pointer to a sum of neutrino masses or a pointer to a list of 3 masses.
-typedef enum ccl_mnu_convention {
+/*typedef enum ccl_mnu_convention {
   ccl_mnu_list = 0,   // you pass a list of three neutrino masses
   ccl_mnu_sum = 1,  // sum, defaults to splitting with normal hierarchy
   ccl_mnu_sum_inverted = 2, //sum, split with inverted hierarchy
   ccl_mnu_sum_equal = 3, //sum, split into equal masses
   // More options could be added here
-} ccl_mnu_convention;
+} ccl_mnu_convention;*/
```
beckermr: Dead code still in the repo.

pyccl/core.py
c-d-leonard: If m_nu is passed as a sum, m_nu is a single value and it gets split into 3 values which are held in mnu_list below. If m_nu is passed as a list, this bit of code just puts those individual listed values in mnu_list so that below we can work with mnu_list regardless of the way m_nu is initially provided. I agree it's clunky though - better ideas?

beckermr: Ahhhhh. I did not follow. This is totally fine. No need for a change here. Thanks!

beckermr:
You might add a changelog entry, but NBD if not.

c-d-leonard:
Changelog entry added. Once the travis tests have rerun I will push the big green button. Thanks!

PR #685, opened 2019-10-01
ENH make CAMB the default Boltzmann code

### LINKED ISSUE #683: default to CAMB
Now that we have CCL in conda, we can safely default to CAMB to make it work with a boltzmann code right out of the box!

beckermr:
This PR makes CAMB the default Botlzmann code. The rationale is as follows. In CCL v2, we can now ship CCL via conda because we have changed CCL from linking to CLASS in C to using the CLASS python wrapper. Further, because CAMB is in conda, we can automatically install CAMB with CCL when installing the conda package. Thus to make CCL work with a Boltzmann code without changing the default params, we are switching the default to CAMB. Otherwise, seeming simple invocations of CCL would fail because CLASS is not installed.

Closes #683


elisachisari:
Could we please summarize the discussion on the Slack here for reference?

beckermr:
Done @elisachisari !

beckermr:
I only had to change the tests that explicitly used the power spectrum from class to compare and only that one with neutrinos. Otherwise things are great. I changed both power spectrum benchmarks to be consistent however.

Issue #668, opened 2019-08-07
Halo theory in CCL


tmcclintock:
# Halo theory in CCL

(Tagging relevant people: @elisachisari @damonge @villarrealas @pennalima @aimalz @beckermr @alexander-mead @BenediktDiemer @cavestruz @matthewkirby @elikrause @KuanWang-Astro @zdu863 @c-d-leonard. I’m sure I forgot someone that will want to see this post. My bad.)

This issue is meant to subsume issues #656, #655, and #291. It is also relevant for #344, #235, #275, and #219, so maybe close those too.

As of now, halos are missing from CCL. This is a problem because a huge fraction of the community would benefit from these theory tools. Both cluster science and HOD models are unable to use CCL without these routines. The upside is that most of halo theory is very easy to write down!

This issue contains (my vision for) an outline for how a full halo model should be implemented in CCL. It includes where we need to make changes to the existing code structure, what new files we need to create and build tests for, what theory routines should go into CCL for now, and what pieces of code I think we should toss out/rewrite. I also point out routines that I think should exist elsewhere (namely Firecrown). At the end I also gave a wishlist for things that would be nice but tricky to implement (like how to generally accommodate emulators).

If you don’t spend most of your day thinking about halos and want some background reading, my suggestion is to look at [Frank van den Bosch’s lectures](https://campuspress.yale.edu/astro610/). Lectures 4, 6, 8, 9, 10, and 13 would cover everything I mention here. The seminal review paper on halo theory is [Cooray & Sheth (2002)](https://arxiv.org/abs/astro-ph/0206508).

## An aside: halo theory =/= cluster science

Cluster science is distinct from halo theory because the former includes a selection function. A ton of effort outside of CCL and DESC is spent understanding cluster (and galaxy) selection functions. Numerically, dealing with selection functions amounts to doing extra integrals over PDFs of your observables and any other quantities you need to consider to link your observable with halo mass. For instance, with redMaPPer clusters, to accurately model some observable (forgetting about redshift for a second) you have to integrate over a fixed bin of observed richness, an infinite integral over true richness, and an infinite integral over halo mass.

The selection function can be arbitrarily more complicated than that, so CCL should probably not try to support cluster science directly. Those routines are probably best left to Firecrown, and as CCL developers we should look to provide highly-optimized versions of functions of halo mass so that those higher-level routines won’t be bogged down.

# What I am NOT talking about

[...]

PR #688, opened 2019-10-11
General halo model

### LINKED ISSUE #636: No bias for Tinker 2008 halo mass function
Hi all, 

When I try computing the mass function and the halo bias for the Tinker 2008 mass function (i.e. specifying `mass_function='tinker'`) I get the following error message from CCL:

`pyccl.errors.CCLError: Error CCL_ERROR_MF: ccl_massfunc.c: ccl_halo_b1(): No b(M) fitting function implemented for mass_function_method: 1`

Would anyone know how to fix this?

Thank you very much!
All the best,
Andrina

### LINKED ISSUE #655: mass function enhancements
Some queries/comments from @BenediktDiemer

- A number of routines appear in counter-intuitive places (e.g., general routines such as ccl_dlninvsig_dlogm() are hidden away in the MF module).
- Concentration is hidden inside the halo model unit, though those are also not exactly the same things.
- Personally, it seems strange to me that c/MF are fixed when the cosmology is fixed, but I'm sure there are reasons for that?
- The naming scheme for c/MF models is somewhat inconsistent: The scheme is incoherent, e.g. "tinker" and "tinker10". There are some errors, e.g. "bhattacharya11" should be "bhattacharya13" (I think)

### LINKED ISSUE #656: Consistency in mass definitions
We got this from @BenediktDiemer

- Overdensity mass definitions are not implemented in a coherent, functional manner
- Since all c/MF models refer to a particular mass definition, it's important to implement those cleanly. Currently, only one overdensity is passed, making it hard to understand for the user whether this refers to mean, critical etc.
- Checks on whether two mass definitions are the same are performed with lines such as `if (odelta == Dv_BryanNorman(cosmo, a, status))`. This is very dangerous, as such float comparisons can easily fail. For example, I computed Delta_vir using the same function in Colossus, and got an error about non-matching mass definitions.
- The mass definitions assumed in some models need to be checked. For example, I think Angulo+ used FOF, not M200m as in CCL.
- It would be good to add conversion between mass definitions, given that each c/MF model predicts only one or a few (typically).
- I'm not sure the implemented concentration models are best for cosmology. Both the Bhattacharya and Duffy models are calibrated on older simulations with limited mass range, calibrated at a particular cosmology, and/or valid only for particular redshifts. Power-law fits such as Duffy become demonstrably wrong at low masses. These are undesirably features in a cosmology code where cosmology is definitely going to vary! I realize that I have a bias here given that I've proposed some concentration models myself, but I think it'd be critical to have at least one model that automatically takes variations with cosmology into account (such as Bullock+01, Ludlow+16, Child+18, or our models).

### LINKED ISSUE #657: Bug in Angulo MF
From @BenediktDiemer :
 the formula for the Angulo+ MF model is wrong (including in their paper).

damonge:
Addressed #668 

To-do list
- [x] Mass definition class
- [x] Mass function and halo bias classes
~~ Halo profile classes~~
~~ Halo model Pk functionality~~

closes #655
closes #656
closes #657
closes #636


damonge:
Extra things to do:

- [ ] Mass definitions for FOF and virial
- [ ] More mass functions and halo biases

damonge:
I've decided to make this PR less ambitious, and just cover mass functions, mass definitions and mass biases. Profiles and power spectra will be done later on.

I'd appreciate some feedback on the API, which is showcased in this notebook: https://github.com/LSSTDESC/CCL/blob/general_halo_model/examples/HaloMassFunctions.ipynb

Tagging a few people: @elisachisari @tmcclintock @beckermr

benchmarks/ccl_test_mconv.c
```diff
@@ -0,0 +1,46 @@
+#include "ccl.h"
```
beckermr: This test should be in python. The C tests are not reliably run and are basically deprecated.

benchmarks/test_massdef.py
```diff
@@ -0,0 +1,53 @@
+import os
```
beckermr: These look like unit tests. Those now go in the main code. The basic distinction is whether or not the test reads data from disk. If it does, it is a benhcmark, otherwise it is a unit test.

pyccl/__init__.py
```diff
@@ -60,6 +57,26 @@
...
+    sigmaM, MassFunc, MassFuncPress74,
+    MassFuncSheth99, MassFuncJenkins01,
+    MassFuncTinker08, MassFuncTinker10,
+    MassFuncWatson13, MassFuncAngulo12,
+    MassFuncDespali16, MassFuncBocquet16)
+from .hbias import (
+    HBiasFunc, HBiasFuncSheth99,
+    HBiasFuncSheth01, HBiasFuncTinker10,
+    HBiasFuncBhattacharya11)
+
```
beckermr: We are adding a lot to the global namespace here. Maybe we should keep this stuff in some other module, like `pyccl.halos` or something?

beckermr:
I'll add more comments once the ones above are addressed.

pyccl/__init__.py
beckermr: Also HBias -> HaloBias. Three letters don't matter when you need to type Tinker etc.

tmcclintock:
@damonge that notebook looks great! I think it is very self explanatory and intuitive (to me, at least). One question though - is it possible to use a definition like 200m *without* having an M-c relation? I'm looking at cell [3] in particular. Additionally, is it possible to specify arbitarary c(M) definitions?

@matthewkirby you should take a look at this, since you are developing a similar tool in parallel.

damonge:
@tmcclintock yes, that's definitely possible. You just won't be able to translate into other mass definitions (if you try, you'll get an error message).

Is @matthewkirby working on this for CCL?

benchmarks/ccl_test_mconv.c
damonge: I've added a python test that does the same thing (`pyccl/tests/test_massdef.py`).

benchmarks/test_massdef.py
damonge: I think only the first one was a unit test (the other ones are actually all reading from disk). I've moved this one to `pyccl/tests/test_massdef.py`

pyccl/__init__.py
damonge: I've put all of this into a `halos` module, and renamed a few things to make it more transparent.

damonge:
Done! Back to you @beckermr 
BTW, my intention is to implement a couple more concentration-mass relations before I'm happy to merge this, so you can wait until then if you want to have another go. I'll try to do that today.

benchmarks/ccl_test_mconv.c
beckermr: OK. Let's remove the C test.

benchmarks/test_massdef.py
beckermr: great!

pyccl/__init__.py
beckermr: yay!

beckermr:
I left one more comment. Ping me again when you are ready for final review.

tmcclintock:
@damonge that's too bad. That singular functionality is what *would* let emulators work for any mass definition. That is, train an emulator to model \rho(r) under one definition. As long as that density profile is properly normalized, you can solve (via root finding) for what the mass and radius are at other definitions.

damonge:
@tmcclintock OK, I see. How about this: for the time being we only implement the simplest mass conversion assuming NFW. I'll make the current c(M) API a bit more flexible.
Then, once general profiles are properly implemented (in the next PR), we revisit the idea of having a more general mass conversion interface.

As a side note: in the interactions I've had with the Simons Observatory clusters people, their opinion is that, in any proper cosmological analysis, you shouldn't rely on mass conversions at all, and rather that you should make sure to use mass functions/biases/c(M)/profiles that have been calibrated to whatever mass definition you're using. This is the reason why at this stage I'm not so worried about implementing a very thorough mass conversion framework.

tmcclintock:
@damonge that's fair and I agree with those people.

damonge:
OK, this is now ready for review @beckermr .
I've added a more generic concentration-mass API and a few bells and whistles.
After this is merged, it should solve issues:
#655 
#656 
#657 
#636 
and half of #668

damonge:
Oh, forgot to say, please have a look at the notebook again (which showcases the new bells and whistles)

damonge:
Hey guys @beckermr @tmcclintock ! Let me know if you have time to review this one.

benchmarks/test_hbf.py
```diff
@@ -0,0 +1,47 @@
...
+
+# Redshifts
+zs = np.array([0., 0.5, 1.])
+
+
+def test_hbf_tinker10():
+    d_hbf = np.loadtxt(dirdat + 'hbf_tinker10.txt',
+                       unpack=True)
+    mf = ccl.halos.HaloBiasTinker10(cosmo)
+    m = d_hbf[0]
```
beckermr: We only test one mass? Ditto for the ones below.

benchmarks/test_hmf.py
```diff
@@ -0,0 +1,96 @@
...
+import numpy as np
+import pyccl as ccl
+
+# Set cosmology
+cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, Omega_g=0, Omega_k=0,
+                      h=0.7, sigma8=0.8, n_s=0.96, Neff=0, m_nu=0.0,
+                      w0=-1, wa=0, T_CMB=2.7255,
+                      transfer_function='eisenstein_hu')
+# Read data
+dirdat = os.path.dirname(__file__) + '/data/'
```
beckermr: again `os.path.join`

benchmarks/test_hmf.py
```diff
@@ -0,0 +1,96 @@
...
+
+# Redshifts
+zs = np.array([0., 0.5, 1.])
+
+
+def test_hmf_despali16():
+    hmd = ccl.halos.MassDef('vir', 'critical')
+    mf = ccl.halos.MassFuncDespali16(cosmo, hmd)
+    d_hmf = np.loadtxt(dirdat + 'hmf_despali16.txt', unpack=True)
```
beckermr: Here you want `os.path.join` too. I missed several of these in the other test file. All should be changed.

benchmarks/test_hmf.py
```diff
@@ -0,0 +1,96 @@
...
+    m = d_hmf[0]
```
beckermr: Why one mass?

benchmarks/test_massdef.py
```diff
@@ -0,0 +1,77 @@
...
+    # Diemer 200 critical
+    cs_200c_dih = hmd_200c_di.get_concentration(cosmo, Ms, 1.)
+    assert np.all(np.fabs(cs_200m_dh/cs_200m_d-1) < 1E-6)
+    assert np.all(np.fabs(cs_200c_dh/cs_200c_d-1) < 1E-6)
+    assert np.all(np.fabs(cs_200m_bh/cs_200m_b-1) < 3E-2)
+    assert np.all(np.fabs(cs_200c_bh/cs_200c_b-1) < 3E-2)
+    assert np.all(np.fabs(cs_200c_ph/cs_200c_p-1) < 3E-2)
+    assert np.all(np.fabs(cs_200c_dih/cs_200c_di-1) < 3E-2)
+    assert np.all(np.fabs(cs_vir_kh/cs_vir_k-1) < 3E-2)
+    assert np.all(np.fabs(cs_vir_bh/cs_vir_b-1) < 3E-2)
```
beckermr: Why are some of these tolerances so high? This is evaluating a function right? We at least need a comment about it.

benchmarks/test_massdef.py
```diff
@@ -0,0 +1,77 @@
...
+import pyccl as ccl
+
+cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, Omega_g=0, Omega_k=0,
+                      h=0.7, sigma8=0.8, n_s=0.96, Neff=0, m_nu=0.0,
+                      w0=-1, wa=0, T_CMB=2.7255,
+                      transfer_function='eisenstein_hu')
+dirdat = os.path.dirname(__file__) + '/data/'
+Ms, Rs_200m, Rs_500c, Ms_500c = np.loadtxt(dirdat + 'mdef_bm.txt',
+                                           unpack=True)
+dc = np.loadtxt(dirdat + 'conc_bm.txt', unpack=True)
```
beckermr: more os.path.join

benchmarks/test_sigmaM.py
```diff
@@ -32,6 +32,6 @@ def test_sigmaM(model, w0, wa):
 
     for i in range(data.shape[0]):
         m = data[i, 0] / cosmo['h']
-        sm = ccl.sigmaM(cosmo, m, 1)
+        sm = ccl.halos.sigmaM(cosmo, m, 1)
```
beckermr: This is an API break. I am ok with that, but it means we are going to be shipping v3 on master.

include/ccl_mass_conversion.h
```diff
@@ -0,0 +1,23 @@
...
+ * Get concentration for a new mass definition.
+ * @param cosmo Cosmological parameters
+ * @param delta_old overdensity parameter in the input mass definition.
+ * @param nc number of concentration values to translate.
+ * @param c_old input concentration.
+ * @param delta_new overdensity parameter in the output mass definition.
+ * @param c_new output concentration.
+ * @param status Status flat. 0 if everything went well.
+ */
+void ccl_get_new_concentration(ccl_cosmology *cosmo,
```
beckermr: The name of this function is not great. This is a conversion utility right? Maybe it should be called something like that?

include/ccl_massfunc.h
```diff
@@ -22,8 +22,7 @@ void ccl_cosmology_compute_sigma(ccl_cosmology *cosmo, int *status);
  */
 void ccl_cosmology_compute_hmfparams(ccl_cosmology *cosmo, int *status);
 
-// TODO: smooth_mass is not really correct in this function, tho it makes sense in compute_sigma
```
beckermr: Did we address this todo or is it no longer relevant?

include/ccl_massfunc.h
```diff
@@ -35,8 +34,7 @@ void ccl_cosmology_compute_hmfparams(ccl_cosmology *cosmo, int *status);
  */
 double ccl_massfunc(ccl_cosmology *cosmo, double smooth_mass, double a, double odelta, int *status);
 
-// TODO: status flag unused here
```
beckermr: Ditto here.

pyccl/__init__.py
```diff
@@ -34,9 +34,6 @@
 # BCM stuff
 from .bcm import bcm_model_fka
 
-# Halo mass function
-from .massfunction import massfunc, massfunc_m2r, sigmaM, halo_bias
-
```
beckermr: Again more API breaks. Just an FYI. I know I suggested it.

pyccl/ccl_mass_conversion.i
```diff
@@ -0,0 +1,33 @@
...
+%apply (int DIM1, double* ARGOUT_ARRAY1) {(int nout, double* output)};
+
+%feature("pythonprepend") %{
+    if numpy.shape(c_in) != (nout,):
+        raise CCLError("Input shape for `c` must match `(nout,)`!")
+%}
+
+%inline %{
+
+  void get_new_concentration_vec(ccl_cosmology *cosmo,
```
beckermr: The name of this file doesn't match the corresponding one in the C code. This is a convention we have been sticking to that helps other navigate the code base.

pyccl/ccl_mass_conversion.i
```diff
@@ -0,0 +1,33 @@
...
+			       double delta_old,
+			       double* c_in, int nc,
+			       double delta_new,
+			       int nout, double *output,
+			       int *status) {
+  *status=0;
```
beckermr: Why is the status being reset? i don't remember doing this in other swig interface files but I could be wrong.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
...
+        raise CCLError("Input shape for `halo_mass` must match `(nout,)`!")
+%}
+
+%inline %{
+
+void sigM_vec(ccl_cosmology * cosmo, double a,
+	      double *logM, int nM,
+	      int nout, double* output, int *status)
+{
+  if (cosmo->params.N_nu_mass>0){
```
beckermr: I think this should be a warning, consistent with how we handle the growth function.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
...
+	  ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_growth(): Support for the sigma(M) function in cosmologies with massive neutrinos is not yet implemented.\n");
+	  return;
+  }
+
+  // Check if sigma has already been calculated
+  if (!cosmo->computed_sigma) {
+    *status = CCL_ERROR_SIGMA_INIT;
+    ccl_cosmology_set_status_message(
+      cosmo,
+      "ccl_massfunc.c: ccl_sigmaM(): linear power spctrum has not been computed!");
```
beckermr: Typo.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
...
+
+  for(int i=0; i<nM; i++) {
+    double lgsigmaM;
+    int gslstatus = gsl_spline_eval_e(cosmo->data.logsigma, logM[i], NULL, &lgsigmaM);
+    if(gslstatus) {
+      ccl_raise_gsl_warning(gslstatus, "ccl_massfunc.c: ccl_sigmaM():");
+      *status |= gslstatus;
+      return;
+    }
+    output[i] = pow(10,lgsigmaM)*ccl_growth_factor(cosmo, a, status);
```
beckermr: We spline log10?

beckermr: This is an odd choice.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
+%module ccl_massfunc
```
beckermr: There is a bit too much going on in this interface file. Probably best to move most of the action to C.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
...
+    if(gslstatus) {
+      ccl_raise_gsl_warning(gslstatus, "ccl_massfunc.c: ccl_sigmaM():");
+      *status |= gslstatus;
+      return;
+    }
+    output[i] = pow(10,lgsigmaM)*ccl_growth_factor(cosmo, a, status);
+  }
+}
+
+void dlnsigM_dlogM_vec(ccl_cosmology * cosmo,
```
beckermr: Same comments here on warnings etc.

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,101 @@
...
+		       double *logM, int nM,
+		       int nout, double* output, int *status)
+{
+  if (cosmo->params.N_nu_mass>0){
+	  *status = CCL_ERROR_NOT_IMPLEMENTED;
+	  ccl_cosmology_set_status_message(cosmo, "ccl_background.c: ccl_cosmology_compute_growth(): Support for the sigma(M) function in cosmologies with massive neutrinos is not yet implemented.\n");
+	  return;
+  }
+
+  // Raise an error if we have mu / Sigma modifcation to gravity turned on
```
beckermr: This should be a warning.

pyccl/concentration.py
```diff
@@ -0,0 +1,317 @@
...
+        if np.isscalar(M):
+            c = c[0]
+        return c
+
+
+class ConcentrationDiemer15(Concentration):
+    name = 'Diemer15'
+
+    def __init__(self, mdef=None):
+        super(ConcentrationDiemer15, self).__init__("Diemer15",
```
beckermr: I'm confused as to why the classes have a name attribute and we pass the same string to the init.

pyccl/concentration.py
```diff
@@ -0,0 +1,317 @@
...
+    Args:
+        name (str): a name for this c(M) object.
+        mass_def (:obj:`MassDef`): a mass definition object that fixes
+            the mass definition used by this c(M)
+            parametrization.
+    """
+    name = 'default'
+
+    def __init__(self, name, mass_def=None):
+        self.name = name
```
beckermr: Huh? Doesn't this override the class-level attribute?

pyccl/concentration.py
```diff
@@ -0,0 +1,317 @@
...
+
+        # Compute power spectrum slope
+        R = mass2radius_lagrangian(cosmo, M_use)
+        lk_R = np.log10(2.0 * np.pi / R * self.kappa)
+        lkmin = np.amin(lk_R-0.05)
+        lkmax = np.amax(lk_R+0.05)
+        logk = np.arange(lkmin, lkmax, 0.01)
+        lpk = np.log10(linear_matter_power(cosmo, 10**logk, a))
+        interp = InterpolatedUnivariateSpline(logk, lpk)
+        n = interp(lk_R, nu=1)
```
beckermr: This block of operations seems expensive. A simple finite diff would be ok maybe? Or I bet we could pull this out of the C-level splines we have. Usually I would not care, but a routine like this will sit deep inside loops over halo masses etc and so is more performance sensitive.

pyccl/concentration.py
```diff
@@ -0,0 +1,317 @@
...
+                self.A = 5.71
+                self.B = -0.084
+                self.C = -0.47
+
+    def concentration(self, cosmo, M, a):
+        M_pivot_inv = cosmo.cosmo.params.h * 5E-13
+        return self.A * (M * M_pivot_inv)**self.B * a**(-self.C)
+
+
+def concentration_from_name(name):
```
beckermr: Do we need this factor function? Seems like the names are in the class names so that we are not saving anyone any effort?

pyccl/concentration.py
```diff
@@ -0,0 +1,317 @@
+from . import ccllib as lib
```
beckermr: A general comment here. Do we want all of this code at the Python level? I understand we are python-first, but my guess is that eventually we want to use these routines to do halo model computations. It is very likely that those computations will happen mostly in C for performance reasons. In this case, I think we would want the code callable from C. This is possible with Python OFC but it would mean we are stuck with the GIL and so would not have an opportunity to use OpenMP.

pyccl/halos.py
```diff
@@ -0,0 +1,31 @@
+# Halo concentration
```
beckermr: As long as we are breaking the API, we should make this a legit submodule to clean up the code a bit. I realize this might break some of the swig build, but we should cross that bridge sooner rather than later maybe.

pyccl/concentration.py
damonge: So here's my thinking: for halo-model computations we need to solve a few integrals over mass. All of these routines allow you to compute some of the functions of mass that go into those integrals. The idea then is that, when we write the general halo model module, these routines will generate these functions sampled over a range of log(Mass), we'll pass those to C, and C will do the integrals with GSL (and using openMP where needed).
Kind of the same rationale behind the generalized P(k)s right now. does this make sanse?

pyccl/concentration.py
damonge: Meh, I thought this would be a convenient function to have. Saves O(10) characters each time, and it's also nice for the mass definitions (since you attach a c(M) to them just by passing the name of the c(M) parametrization).

pyccl/halos.py
damonge: OK, what would make this a legit submodule?

pyccl/concentration.py
damonge: Ah damn. Yes, you're right. I've actually removed the object-level name, since I'm not using that one for anything

benchmarks/test_hbf.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+import pyccl as ccl
+
+# Set cosmology
+cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, Omega_g=0, Omega_k=0,
+                      h=0.7, sigma8=0.8, n_s=0.96, Neff=0, m_nu=0.0,
+                      w0=-1, wa=0, T_CMB=2.7255,
+                      transfer_function='eisenstein_hu')
+# Read data
+dirdat = os.path.dirname(__file__) + '/data/'
```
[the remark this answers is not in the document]
damonge: done

benchmarks/data/codes/mdef_colossus.py
```diff
@@ -0,0 +1,39 @@
...
+                                        0.0, model = 'prada12')
+cs_200c_di = concentration.concentration(Ms * h, '200c',
+                                         0.0, model = 'diemer15')
+Ms_500c, _, _ = mass_defs.changeMassDefinition(Ms * h, cs_200m_d,
+                                               0., '200m', '500c')
+Ms_500c /= h
+
+np.savetxt("../mdef_bm.txt",
+           np.transpose([Ms, Rs_200m, Rs_500c, Ms_500c]))
+np.savetxt("../conc_bm.txt",
```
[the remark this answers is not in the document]
damonge: done

benchmarks/data/codes/hmf_bm.py
```diff
@@ -0,0 +1,45 @@
...
+    m_arr=np.geomspace(1E11,1E15,9) 
+    z_arr=np.array([0., 0.5, 1.])
+    d_out = []
+    d_out.append(m_arr)
+    normfac = h**3 * np.log(10)
+    for z in z_arr:
+        d_out.append(bias.haloBias(m_arr * h, z = z,
+                                   mdef = mdef,
+                                   model = model))
+    np.savetxt("../hbf_" + model + ".txt", np.transpose(d_out))
```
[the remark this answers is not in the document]
damonge: done

benchmarks/data/codes/hmf_bm.py
```diff
@@ -0,0 +1,45 @@
...
+    z_arr=np.array([0., 0.5, 1.])
+    d_out = []
+    d_out.append(m_arr)
+    normfac = h**3 * np.log(10)
+    for z in z_arr:
+        d_out.append(mass_function.massFunction(m_arr * h, z,
+                                                mdef = mdef,
+                                                model = model,
+                                                q_out = 'dndlnM') * normfac)
+    np.savetxt("../hmf_" + model + ".txt", np.transpose(d_out))
```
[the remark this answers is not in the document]
damonge: done

benchmarks/data/conc_bm.txt
```diff
@@ -0,0 +1,3 @@
+1.000000000000000000e+12 1.103998398537084391e+01 6.236406169708174296e+00 9.666178823786799157e+00 6.431031972382390549e+00 9.860271653196956976e+00 8.269952993684261955e+00 7.970717249470773424e+00 8.095358770694428330e+00
```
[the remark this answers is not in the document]
damonge: OK, done. I think most of our benchmark files have this tough.

pyccl/concentration.py
beckermr: Ahhhh great. This sounds good to me!

pyccl/halos.py
beckermr: You make a directory `pyccl/halos` and then put the code in files within that dir. Make sure to add an `__init__.py` to the directory. You will need to adjust the `setup.py` and any swig wrappers to make sure it builds right.

benchmarks/data/conc_bm.txt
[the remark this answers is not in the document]
beckermr: They do! However, it is never too late to start doing things better!

pyccl/concentration.py
damonge: Mmmh, so what's happening here is essentially finite differences (except through a spline). Also, for a relatively large array of input masses, this should be computationally more efficient than finite differences around each of those.
I buy the point about making C return the spline derivatives (since this is probably useful for other purposes too), but we don't have infrastructure for that yet. I've left a #TODO and will open an issue.

pyccl/halos.py
damonge: ok, sounds good. I may ask for help with this, since I always get lost with `setup.py`

benchmarks/test_hbf.py
damonge: nope, `d_hbf[0]` is the first column of the data file, which contains to a bunch of masses.

benchmarks/test_hmf.py
damonge: (see above)

benchmarks/test_sigmaM.py
damonge: I'm OK with this too, but I'm also happy to move `sigmaM` out of `halos`

pyccl/concentration.py
beckermr: Can we actually test that doing a bunch of masses at once is faster this way?

include/ccl_mass_conversion.h
damonge: changed to `convert_concentration`

benchmarks/test_hbf.py
beckermr: Ahhhhh great. So confusing!

benchmarks/test_sigmaM.py
beckermr: Hmmmmmm. As long as we don't create a circular import, keeping it out of halos actually makes a ton of sense. I'd vote to keep sigmaM where it is and not move it into halos.

damonge: Is this just because of the API break? Because we're gonna be breaking it anyway

include/ccl_massfunc.h
damonge: ok, this was a naming issue. This appeared as `smooth_mass` in the `.h` and ass `halomass` in the `.c`. I've just homogenized to `halomass`.

include/ccl_massfunc.h
damonge: This is just not true anymore

pyccl/__init__.py
damonge: Yes, this is what I meant. This is not how we will compute mass functions anymore (sigmaM is still the same though)

benchmarks/test_sigmaM.py
beckermr: No, I like it more aesthetically, but I am ok either way.

pyccl/ccl_mass_conversion.i
damonge: yes, sorry, this is completely superfluous. Fixed

pyccl/concentration.py
tmcclintock: Is this an evaluation of the derivative of sigma(R) with respect to mass?

damonge: Actually, I misspoke. This will only be faster if the array of masses you pass are sampled faster than delta(log10(mass))=0.03. Happy to change it to a finite differences (I just checked the tests pass either way).

benchmarks/test_sigmaM.py
damonge: ok, we could put this into power.py (that's where `sigmaR` lives)

benchmarks/test_massdef.py
damonge: I copy-pasted a lot. Let me check what the actual tolerances should be. In any case, there's always a ~5-10% error on all of these parametrizations due to uncertainties in halo definition, AFAIK.

pyccl/concentration.py
beckermr: Sounds good. Let's go with the simpler python code for now. We can leave the todo to expose the C level spline derivs for another time.

beckermr: simpler python code == finite differences

pyccl/concentration.py
damonge: :partyparrot:

benchmarks/test_sigmaM.py
beckermr: Yup. Let's do that.

benchmarks/test_massdef.py
damonge: Alright, I've made these as tight as possible. I can't really get mass functions to agree to better than 1%. This is mostly due to the prefactor dlogsigma/dlogM, which is quite sensitive to the power spectrum parametrization. This is only in the high-mass end and at large redshifts. Again, I think 1% in the mass function is not something to worry about, given existing uncertainties (and expected statistical uncertainties for halos of mass 1E15 Msun at z=1).

benchmarks/test_sigmaM.py
damonge: Done

pyccl/ccl_sigM.i
damonge: I agree (this is legacy from the C implementation). I've updated this consistently in C

pyccl/ccl_sigM.i
damonge: Done

pyccl/ccl_sigM.i
damonge: same answer as above.

pyccl/ccl_sigM.i
damonge: as above

pyccl/ccl_sigM.i
damonge: removed this (this is already checked at the python level)

pyccl/ccl_sigM.i
damonge: These are all actually checked when calling `compute_sigma` from python (which is called within `sigmaM`), so there's no need for them. Removed

damonge:
OK @beckermr . I've addressed everything other than making this a proper module. I'll work on that tomorrow, but just in case you want to have another look at the rest.

damonge:
Alright, @beckermr , halos is now a submodule. I didn't really have to tinker with the swig build, but let me know if you would have done things differently.

beckermr:
You will need to adjust the setup.py here: https://github.com/LSSTDESC/CCL/blob/master/setup.py#L60

Probably best to use find_packages

damonge:
@beckermr I did change that already (your link was to master, not this branch). Do you mean that you'd rather have find_packages than the current listing of submodules?

beckermr:
ahhhh sorry. Yes, find_packages is easier to maintain.

damonge:
alright, done!

damonge:
ping! @beckermr @tmcclintock ?

beckermr:
I'm looking today but this PR is big so it will take some time.

pyccl/ccl_sigM.i
beckermr: You marked this as removed but I still see the code. Did you forget to push?

pyccl/ccl_sigM.i
```diff
@@ -0,0 +1,59 @@
...
+		       int nout, double* output, int *status)
+{
+  // Check if sigma has already been calculated
+  if (!cosmo->computed_sigma) {
+    *status = CCL_ERROR_SIGMA_INIT;
+    ccl_cosmology_set_status_message(
+      cosmo,
+      "ccl_massfunc.c: ccl_sigmaM(): linear power spctrum has not been computed!");
+    return;
+  }
```
beckermr: Ditto here for code I think you meant to delete.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+                raise ValueError("c(M) relation " + self.name +
+                                 " is not compatible with mass definition" +
+                                 " Delta = %s, " % (mass_def.Delta) +
+                                 " rho = " + mass_def.rho_type)
+            self.mdef = mass_def
+        else:
+            self._default_mdef()
+        self._setup()
+
+    def _default_mdef(self):
```
beckermr: Why do we need a default? I think requiring an input from the user makes sense here.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+        """ Assigns a default mass definition for this object if
+        none is passed at initialization.
+        """
+        self.mdef = MassDef('fof', 'matter')
+
+    def _setup(self):
```
beckermr: If you remove the default mass def, then we can simplify Concentration to be an ABC and not even have a constructor.

beckermr: or this function.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+
+        Returns:
+            float or array_like: concentration.
+        """
+        M_use = self._get_consistent_mass(cosmo,
+                                          np.atleast_1d(M),
+                                          a, mdef_other)
+
+        c = self.concentration(cosmo, M_use, a)
+        if np.isscalar(M):
```
beckermr: This is not safe. See https://docs.scipy.org/doc/numpy/reference/generated/numpy.isscalar.html in the notes section.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+        self.kappa = 1.0
+        self.phi_0 = 6.58
+        self.phi_1 = 1.27
+        self.eta_0 = 7.28
+        self.eta_1 = 1.56
+        self.alpha = 1.08
+        self.beta = 1.77
+
+    def _check_mdef(self, mdef):
+        if (np.fabs(mdef.Delta - 200.) > 1E-4) and \
```
beckermr: I tend to agree with the comments we got that this is not safe. I expect the failures are rare, but I feel like we should have a better solution

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+
+        sig = sigmaM(cosmo, M_use, a)
+        delta_c = 1.68647
+        nu = delta_c / sig
+
+        floor = self.phi_0 + n * self.phi_1
+        nu0 = self.eta_0 + n * self.eta_1
+        c = 0.5 * floor * ((nu0 / nu)**self.alpha +
+                           (nu / nu0)**self.beta)
+        if np.isscalar(M):
```
beckermr: again not safe.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+                self.B = 0.54
+                self.C = -0.35
+
+    def concentration(self, cosmo, M, a):
+        gz = growth_factor(cosmo, a)
+        status = 0
+        delta_c, status = lib.dc_NakamuraSuto(cosmo.cosmo, a, status)
+        sig = sigmaM(cosmo, M, a)
+        nu = delta_c / sig
+        return self.A * gz**self.B * nu**self.C
```
beckermr: no test for scalars here?

beckermr:
Also I have not yet looked at the full thing. Smaller PRs would make this faster. :/

pyccl/halos/concentration.py
damonge: Hmmh, I think I disagree. There are parametrizations that are only valid for one particular mass definition (or for a couple of them). Having a default mass definition just makes your life easier, I think. I'm happy to remove this if you prefer.

pyccl/halos/concentration.py
damonge: I'll wait for your response to the above. I think it makes sense to have a base constructor to avoid repetition of check mass definition -> setup

pyccl/halos/concentration.py
damonge: The instances of this we got comments about where things like `if Delta != 200` or even worse `if Delta != ccl.delta_vir(blah)`, which made some of our mass functions completely unusable. I think if someone passes 200.0001 or 199.9999, they actually meant 200. I did set up specific Delta values as strings for `fof` and `vir`. I could set up one for `200`, since it's a common one...

pyccl/halos/concentration.py
damonge: Humm, I don't think we need that here. All functions of `M` here operate on scalars or arrays equally.

pyccl/halos/concentration.py
damonge: Great, thanks. Done. We had a couple other instances of this from before.

pyccl/halos/concentration.py
damonge: Done

pyccl/ccl_sigM.i
damonge: (somehow I lost your previous comment, not sure what happened there)
This function is useful here to translate the output of `ccl_sigmaM` into a numpy array.
I've removed the check for the linear power spectrum, since that one is superfluous.

pyccl/halos/concentration.py
beckermr: Hmmm ok. I don’t follow your argument here but I’m happy to defer to you.

damonge:
Thanks a lot @beckermr ! I'm sorry the PR is so large. I couldn't figure out a better way to factor things out, since all of this was very tightly intertwined. Any modifications on any of this functionality should be a lot simpler now, I think.

On the existing HMF C API: sorry I forgot to respond to this one. I've left the code there because it's needed by the existing halo model power spectrum. The goal is to implement halo profiles properly and then reimplement the halo model power spectrum in a generalized way (so that you can also compute HODs etc.). Unless you want me to make this PR larger I don't see a better way right now than leaving the code in for the time being!

pyccl/halos/concentration.py
beckermr: Ahhh right.

pyccl/halos/concentration.py
beckermr: We could force casts to int to solve this.

beckermr: We could force casts to int to solve this.

pyccl/halos/concentration.py
damonge: OK, thanks. For clarification, what I mean is that, for instance, the Klypin11 concentration model only works for Delta_vir masses, so forcing people to pass that mass definition every time they want to use it is kind of clunky. 
BTW, I just realized I need to add a lot of documentation to the concentration stuff. I'll do that now.

pyccl/halos/concentration.py
damonge: so like `if int(mdef.Delta) != 200`?

pyccl/halos/concentration.py
beckermr: Alright! Sounds good!

pyccl/ccl_sigM.i
beckermr: I just meant the check, not the whole function.

beckermr:
Right of course. I wanted to get a sense of the plan here, not to have it addressed necessarily.

pyccl/ccl_sigM.i
damonge: Ah, ok. Yes, you're right. I've removed both instances.

pyccl/cls.py
```diff
@@ -75,7 +75,7 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell, p_of_k_a=None,
     cl, status = lib.angular_cl_vec(
         cosmo, clt1, clt2, psp, l_limber,
         ell_use, ell_use.size, status)
-    if np.isscalar(ell):
+    if np.ndim(ell) == 0:
```
beckermr: Good catch!

pyccl/halos/concentration.py
beckermr: yeah we could do it like this. Better yet to force all delta to be ints on the mdef object.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+        nu0 = self.eta_0 + n * self.eta_1
+        c = 0.5 * floor * ((nu0 / nu)**self.alpha +
+                           (nu / nu0)**self.beta)
+        if np.ndim(M) == 0:
+            c = c[0]
+
+        return c
+
+
+class ConcentrationBhattacharya13(Concentration):
```
beckermr: Each subclass will need a class-level doc string. Otherwise readthedocs will be ugly. Copy-paste is the best!

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+        return self.c0 + (self.c1 - self.c0) * \
+            (np.arctan(self.al * (x - self.x0)) / np.pi + 0.5)
+
+    def imin(self, x):
+        return self.i0 + (self.i1 - self.i0) * \
+            (np.arctan(self.be * (x - self.x1)) / np.pi + 0.5)
+
+    def concentration(self, cosmo, M, a):
+        sig = sigmaM(cosmo, M, a)
+        om = cosmo.cosmo.params.Omega_m
```
beckermr: This will be OmegaM with massive neutrinos if they are non-zero. IDK if we care or not. Simply pointing it out.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,308 @@
...
+        self.be = 7.386
+        self.x1 = 0.526
+        self.cnorm = 1. / self.cmin(1.393)
+        self.inorm = 1. / self.imin(1.393)
+
+    def cmin(self, x):
+        return self.c0 + (self.c1 - self.c0) * \
+            (np.arctan(self.al * (x - self.x0)) / np.pi + 0.5)
+
+    def imin(self, x):
```
beckermr: Any functions that are not external interfaces should be prefixed by `_` (e.g., `_cmin`). This is a python convention that helps keep APIs clean and clear.

pyccl/halos/hbias.py
```diff
@@ -0,0 +1,289 @@
...
+        sigM, status = lib.sigM_vec(cosmo.cosmo, a, logM,
+                                    len(logM), status)
+        check(status)
+
+        b = self.get_bsigma(cosmo, sigM, a)
+        if np.ndim(M) == 0:
+            b = b[0]
+        return b
+
+    def get_bsigma(self, cosmo, sigM, a):
```
beckermr: This routine and the one above it have force a specific functional form of the halo bias into the API. Do we want to be able to generalize to things like emulators later?

pyccl/halos/hmfunc.py
```diff
@@ -0,0 +1,647 @@
...
+        rho = (lib.cvar.constants.RHO_CRITICAL *
+               cosmo['Omega_m'] * cosmo['h']**2)
+        f = self.get_fsigma(cosmo, sigM, a, 2.302585092994046 * logM)
+        mf = f * rho * dlns_dlogM / M_use
+
+        if np.ndim(M) == 0:
+            mf = mf[0]
+        return mf
+
+    def get_fsigma(self, cosmo, sigM, a, lnM):
```
beckermr: Same comment on API and functional forms and emulators.

pyccl/halos/concentration.py
damonge: alas, but there are mass functions that work for a continuous range of Deltas...
If you're OK with checking for ints in the specific cases of 200 and 500, I'll implement that.

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+        cosmo (:obj:`Cosmology`): A Cosmology object.
+        M (float or array_like): halo mass in units of M_sun.
+
+    Returns:
+        float or array_like: lagrangian radius in Mpc
+    """
+    return (M / (4.18879020479 * rho_x(cosmo, 1, 'matter')))**(1./3.)
+
+
+def convert_concentration_py(cosmo, c_old, Delta_old, Delta_new):
```
beckermr: Why is this routines name subscripted with `py`?

pyccl/halos/concentration.py
damonge: Done!

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+        # c(M) relation
+        if c_m_relation is None:
+            self.concentration = None
+        else:
+            self._concentration_init(c_m_relation)
+
+    def __eq__(self, other):
+        """ Allows you to compare two mass definitions
+        """
+        return (self.Delta == other.Delta) and \
```
beckermr: Here we are doing exact equality again on delta. See discussion above for how to potentially handle this.

pyccl/halos/concentration.py
damonge: I think none of these have been calibrated with neutrinos, but that doesn't mean that we should exclude neutrinos here. @tmcclintock may know!

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+                c_this = self.get_concentration(cosmo, M, a)
+                R_this = self.get_radius(cosmo, M, a)
+                om_new = omega_x(cosmo, a, m_def_other.rho_type)
+                D_new = m_def_other.get_Delta(cosmo, a) * om_new
+                c_new = convert_concentration_py(cosmo, c_this, D_this, D_new)
+                R_new = c_new * R_this / c_this
+                return m_def_other.get_mass(cosmo, R_new, a)
+
+
+class MassDef200mat(MassDef):
```
beckermr: `200mat` or `200m`? I think the second is more common?

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+    Args:
+        c_m (string): concentration-mass relation.
+    """
+    def __init__(self, c_m='Duffy08'):
+        super(MassDef200mat, self).__init__(200,
+                                            'matter',
+                                            c_m_relation=c_m)
+
+
+class MassDef200crit(MassDef):
```
beckermr: `200crit` or `200c`? Again the second is more common.

pyccl/halos/hbias.py
damonge: Yeah, I thought about this. Not sure, but I think in that case those potential subclasses should just overload the `get_halo_bias` method.

pyccl/halos/hmfunc.py
damonge: Same as above.

pyccl/halos/massdef.py
damonge: I think this is fine. We really want to check for exactly equal Deltas in this case (we want to see if two mass definitions are actually the same one, so I would enforce them being actually the same). If the Deltas differ by a bit, then there will be a mass conversion, but this shouldn't matter for practical purposes.

pyccl/halos/concentration.py
beckermr: I agree with both of your statements.

pyccl/halos/hbias.py
beckermr: Would those sub classes be expect to implement `get_bsigma` even if they had overloaded `get_halo_bias`? If not, then `get_bsigma` should be named `_get_bsigma` so it is not part of the API.

pyccl/halos/hmfunc.py
beckermr: Same comment about the public API as above as well!

pyccl/halos/massdef.py
damonge: No good reason. I thought there was a `convert_concentration` swig function, but there isn't. Changed!

pyccl/halos/massdef.py
damonge: Done

pyccl/halos/massdef.py
damonge: Done

damonge:
Yes, I'll make a few of these (and hopefully improve our coveralls).

damonge:
Most or all of these subclasses are tested in benchmarks, BTW, but I'd like to make a couple unit test checking for inconsistent mass definitions etc.

pyccl/halos/concentration.py
damonge: Done

pyccl/halos/hbias.py
damonge: Good point. Done

pyccl/halos/hmfunc.py
damonge: Done as above

beckermr:
The smoke tests serve a different purpose than the benchmarks. They are a very fast check that the code still runs that developers can use to quickly verify changes. The entire unit test suite runs in about 30 seconds.

damonge:
Alright, I've implemented a batch of unit tests @beckermr

pyccl/tests/test_concentration.py
```diff
@@ -0,0 +1,32 @@
...
+# are defined for FoF halos.
+MDEF = ccl.halos.MassDef('fof', 'matter')
+
+
+@pytest.mark.parametrize('cM_class', CONCS)
+def test_cM_subclasses_smoke(cM_class):
+    cM = cM_class()
+    for m in MS:
+        c = cM.get_concentration(COSMO, m, 0.9)
+        assert np.all(np.isfinite(c))
```
beckermr: You should check that the output is the right shape/type here too.

pyccl/tests/test_hbias.py
```diff
@@ -0,0 +1,29 @@
...
+MS = [1E13, [1E12, 1E15], np.array([1E12, 1E15])]
+MFOF = ccl.halos.MassDef('fof', 'matter')
+
+
+@pytest.mark.parametrize('bM_class', HBFS)
+def test_bM_subclasses_smoke(bM_class):
+    bM = bM_class(COSMO)
+    for m in MS:
+        b = bM.get_halo_bias(COSMO, m, 0.9)
+        assert np.all(np.isfinite(b))
```
beckermr: Again check the right shape/type

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,38 @@
...
+MVIR = ccl.halos.MassDef('vir', 'critical')
+MDFS = [MFOF, MFOF, MVIR, MFOF, MFOF]
+
+
+@pytest.mark.parametrize('nM_class', HMFS)
+def test_nM_subclasses_smoke(nM_class):
+    nM = nM_class(COSMO)
+    for m in MS:
+        n = nM.get_mass_function(COSMO, m, 0.9)
+        assert np.all(np.isfinite(n))
```
beckermr: Again shape/type checks.

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+@pytest.mark.parametrize('nM_pair', zip(HMFS_SO, MDFS))
+def test_nM_mdef_raises(nM_pair):
```
beckermr: Do we want to test that it breaks for deltas that don't match too?

pyccl/halos/concentration.py
beckermr: This is probably fine.

pyccl/tests/test_hbias.py
damonge: Done

pyccl/tests/test_hmfunc.py
damonge: Done

pyccl/tests/test_hmfunc.py
damonge: I've added a few of these

pyccl/tests/test_concentration.py
damonge: Done

pyccl/halos/concentration.py
damonge: OK, done.

pyccl/tests/test_hmfunc.py
```diff
@@ -29,10 +32,72 @@ def test_nM_subclasses_smoke(nM_class):
...
 @pytest.mark.parametrize('nM_pair', zip(HMFS_SO, MDFS))
 def test_nM_mdef_raises(nM_pair):
     nM_class, mdef = nM_pair
     with pytest.raises(ValueError):
         nM_class(COSMO, mdef)
+
+
+@pytest.mark.parametrize('nM_class', [ccl.halos.MassFuncTinker08,
+                                      ccl.halos.MassFuncTinker10])
+def test_nM_mdef_raises2(nM_class):
```
beckermr: This needs a better name.

damonge: done

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,338 @@
...
+        if np.ndim(M) == 0:
+            c = c[0]
+        return c
+
+
+class ConcentrationDiemer15(Concentration):
+    """ Concentration-mass relation by Diemer & Kravtsov 2015
+    (arXiv:1407.4730). This parametrization is only valid for
+    S.O. masses with Delta = 200-critical.
+    """
```
beckermr: This doc string needs the Args part and everything else. Ditto for all classes.

pyccl/halos/concentration.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def _check_mdef(self, mdef):
+        if isinstance(mdef.Delta, str):
+            return True
+        elif (int(mdef.Delta) != 200) and \
+             (mdef.rho_type != 'critical'):
+            return True
+        return False
+
+    def concentration(self, cosmo, M, a):
```
beckermr: All public methods need doc strings.

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+        elif isinstance(c_m_relation, str):
+            # Grab class
+            conc_class = concentration_from_name(c_m_relation)
+            # instantiate with this mass definition
+            self.concentration = conc_class(mdef=self)
+        else:
+            raise ValueError("c_m_relation must be `None`, "
+                             " a string or a `Concentration` object")
+
+    def get_Delta(self, cosmo, a):
```
beckermr: A lot of these methods need unit tests.

pyccl/halos/massdef.py
```diff
@@ -0,0 +1,247 @@
...
+    The lagrangian radius is defined as that enclosing
+    the mass of the halo assuming a homogeneous Universe.
+
+    Args:
+        cosmo (:obj:`Cosmology`): A Cosmology object.
+        M (float or array_like): halo mass in units of M_sun.
+
+    Returns:
+        float or array_like: lagrangian radius in Mpc
+    """
```
beckermr: The use of lagrangian here confuses me. Also is the output in comoving units? I think yes.

pyccl/halos/concentration.py
damonge: Done

pyccl/halos/concentration.py
damonge: Done

pyccl/halos/hmfunc.py
```diff
@@ -0,0 +1,653 @@
...
+    def _get_fsigma(self, cosmo, sigM, a, lnM):
+        return self.A * np.exp(-np.fabs(-np.log(sigM) + self.b)**self.q)
+
+
+class MassFuncTinker08(MassFunc):
+    """ Implements mass function described in arXiv:0803.2706.
+
+    Args:
+        cosmo (:obj:`Cosmology`): A Cosmology object.
+        mass_def (:obj:`MassDef`): a mass definition object.
```
beckermr: So one quirk in the API here is that some mass functions do not take a mass def. I can imagine this causing weird things for people downstream. Thoughts on this?

Maybe each should have a default and we allow a new one but error check it?

pyccl/halos/massdef.py
damonge: I think the nomenclature is standard.
https://bdiemer.bitbucket.io/colossus/lss_peaks.html#lss.peaks.lagrangianR
(i.e. it's the comoving radius of the halo in the initial condition).
Added comoving

pyccl/halos/hmfunc.py
damonge: I can add a mass_def argument for those that only have one allowed mass definition for homogeneity. They won't work unless you pass the correct mass definition (which will be the default value). Make sense?

beckermr: Yep. This keeps the interface uniform which is the goal. Errors are user problems.

beckermr: Yep. This keeps the interface uniform which is the goal. Errors are user problems.

pyccl/halos/massdef.py
damonge: done

pyccl/halos/hmfunc.py

damonge:
Alright, coverage is now back where it was!

beckermr:
Great! I’ll be going through the changes again tomorrow!

pyccl/halos/hmfunc.py
```diff
@@ -0,0 +1,673 @@
...
+            this parametrization accepts SO masses with
+            Delta = 200 (matter, critical) and 500 (critical).
+            If `None`, Delta = 200 (matter) will be used.
+    """
+    name = 'Bocquet16'
+
+    def __init__(self, cosmo, mass_def=None, hydro=True):
```
beckermr: Is with hydro the right default here?

beckermr: Also no doc string for hydro

pyccl/tests/test_concentration.py
```diff
@@ -0,0 +1,57 @@
...
+        assert np.shape(c) == np.shape(m)
+
+
+def test_cM_from_string_raises():
+    with pytest.raises(ValueError):
+        ccl.halos.concentration_from_name('Duffman08')
```
beckermr: LLOL

beckermr: However, I don't want to offend anyone with this. Maybe `blah` or something?

pyccl/tests/test_concentration.py
```diff
@@ -0,0 +1,57 @@
...
+@pytest.mark.parametrize('cM_class', CONCS)
+def test_cM_subclasses_smoke(cM_class):
+    cM = cM_class()
+    for m in MS:
+        c = cM.get_concentration(COSMO, m, 0.9)
+        assert np.all(np.isfinite(c))
+        assert np.shape(c) == np.shape(m)
+
+
+def test_cM_duffy_extra():
```
beckermr: What is this test testing? Maybe a better name?

pyccl/tests/test_hbias.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+@pytest.mark.parametrize('bM_pair', zip(HBFS, MDFS))
+def test_bM_mdef_raises(bM_pair):
+    bM_class, mdef = bM_pair
+    with pytest.raises(ValueError):
+        bM_class(COSMO, mdef)
+
+
+def test_bM_Tinker10_extra():
```
beckermr: better test name here

pyccl/tests/test_hbias.py
```diff
@@ -0,0 +1,65 @@
...
+    bM = bM_class(COSMO)
+    for m in MS:
+        b = bM.get_halo_bias(COSMO, m, 0.9)
+        assert np.all(np.isfinite(b))
+        assert np.shape(b) == np.shape(m)
+
+
+def test_bM_from_string_raises():
+    with pytest.raises(ValueError):
+        ccl.halos.halo_bias_from_name('Tanker10')
```
beckermr: Again LLOL, but maybe let's not possibly offend Tinker.

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@pytest.mark.parametrize('nM_class', [ccl.halos.MassFuncTinker08,
+                                      ccl.halos.MassFuncTinker10])
+def test_nM_mdef_bad_delta(nM_class):
+    with pytest.raises(ValueError):
+        nM_class(COSMO, M100)
+
+
+def test_nM_despali_extra():
```
beckermr: better test name

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,116 @@
...
+    nM = ccl.halos.MassFuncDespali16(COSMO,
+                                     ellipsoidal=True)
+    for m in MS:
+        n = nM.get_mass_function(COSMO, m, 0.9)
+        assert np.all(np.isfinite(n))
+        assert np.shape(n) == np.shape(m)
+
+
+@pytest.mark.parametrize('mdef', [MFOF, M200m])
+def test_nM_watson_extra(mdef):
```
beckermr: better test name

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,116 @@
...
+        assert np.all(np.isfinite(n))
+        assert np.shape(n) == np.shape(m)
+    for m in MS:
+        n = nM.get_mass_function(COSMO, m, 0.1)
+        assert np.all(np.isfinite(n))
+        assert np.shape(n) == np.shape(m)
+
+
+@pytest.mark.parametrize('with_hydro', [True, False])
+def test_nM_bocquet_extra(with_hydro):
```
beckermr: better test name

pyccl/tests/test_hmfunc.py
```diff
@@ -0,0 +1,116 @@
...
+    nM = nM_class(COSMO)
+    for m in MS:
+        n = nM.get_mass_function(COSMO, m, 0.9)
+        assert np.all(np.isfinite(n))
+        assert np.shape(n) == np.shape(m)
+
+
+def test_nM_from_string_raises():
+    with pytest.raises(ValueError):
+        ccl.halos.mass_function_from_name('Tanker08')
```
beckermr: again LLOL but change this

pyccl/tests/test_massdef.py
```diff
@@ -0,0 +1,116 @@
...
+
+    hmd = ccl.halos.MassDef('vir', 'critical')
+    assert np.isfinite(hmd.get_Delta(COSMO, 1.))
+
+    hmd = ccl.halos.MassDef('fof', 'critical')
+    with pytest.raises(ValueError):
+        hmd.get_Delta(COSMO, 1.)
+
+
+def test_get_mass():
```
beckermr: this test could easily check the actual value

pyccl/tests/test_massdef.py
```diff
@@ -0,0 +1,116 @@
...
+    hmd = ccl.halos.MassDef(200, 'critical')
+    for R in [1., [1., 2.], np.array([1., 2.])]:
+        m = hmd.get_mass(COSMO, R, 1.)
+        assert np.all(np.isfinite(m))
+        assert np.shape(m) == np.shape(R)
+
+
+def test_get_radius():
```
beckermr: Ditto here for checking the actual value

pyccl/tests/test_massdef.py
```diff
@@ -0,0 +1,116 @@
...
+    hmdb = ccl.halos.MassDef(200, 'critical')
+    with pytest.raises(RuntimeError):
+        hmd.translate_mass(COSMO, 1E12,
+                           1., hmdb)
+
+
+@pytest.mark.parametrize('scls', [ccl.halos.MassDef200m,
+                                  ccl.halos.MassDef200c,
+                                  ccl.halos.MassDefVir])
+def test_subclasses(scls):
```
beckermr: What is this test doing? Smoke test for subclasses?

pyccl/tests/test_massdef.py
```diff
@@ -0,0 +1,116 @@
...
+    hmdb = ccl.halos.MassDef200c()
+    for M in [1E12, [1E12, 2E12],
+              np.array([1E12, 2E12])]:
+        m = hmd.translate_mass(COSMO, M,
+                               1., hmdb)
+        assert np.all(np.isfinite(m))
+        assert np.shape(m) == np.shape(M)
+
+
+def test_translate_mass_raises():
```
beckermr: Why does this one raise an error, but the test above it passes? I don't see an obvious difference.

src/ccl_mass_conversion.c
```diff
@@ -0,0 +1,112 @@
+#include <stdio.h>
```
beckermr: If you wanted to put some comments on the formulae in this file, that would be nice. I am not going to hold up this PR because of that though.

pyccl/halos/hmfunc.py
damonge: I think `hydro = True` is the more accurate version (this is also the default in Colossus).
Docs added.

pyccl/tests/test_concentration.py
damonge: Ah, no fun!
Kidding, changed.

pyccl/tests/test_concentration.py
damonge: It mostly tests that you can use this parametrization with other input parameters. Changed to `duffy_smoke`

pyccl/tests/test_hbias.py
damonge: same as duffy_extra (changed  to Tinker10_smoke)

pyccl/tests/test_hbias.py
damonge: changed

pyccl/tests/test_hmfunc.py
damonge: changed to smoke

pyccl/tests/test_hmfunc.py
damonge: changed to smoke

pyccl/tests/test_hmfunc.py
damonge: changed to smoke

pyccl/tests/test_hmfunc.py
damonge: changed

pyccl/tests/test_massdef.py
damonge: this is tested explicitly as a benchmark here: https://github.com/LSSTDESC/CCL/blob/24b17f46673fc63f95e7a569179dc6abf3854070/benchmarks/test_massdef.py#L42

pyccl/tests/test_massdef.py
damonge: tested here: https://github.com/LSSTDESC/CCL/blob/24b17f46673fc63f95e7a569179dc6abf3854070/benchmarks/test_massdef.py#L35

pyccl/tests/test_massdef.py
damonge: yes. Added "smoke"

pyccl/tests/test_massdef.py
damonge: these mass definitions are not associated to a c(M) relation, so it's not possible to translate between them.

src/ccl_mass_conversion.c
damonge: Added some explanation

damonge:
Alright, thanks a lot!
I'll look at changelog and docs in a bit and ping you when everything is ready

damonge:
Alright, done @beckermr !

beckermr:
The spacing on the docs changed in some spots. Do they build correctly?

damonge:
They do. I only get this warning:
```
WARNING: failed to reach any of the inventories with the following issues:
WARNING: intersphinx inventory ' https://docs.python.org/3/objects.inv' not fetchable due to <class 'requests.exceptions.InvalidSchema'>: ('intersphinx inventory %r not fetchable due to %s: %s', ' https://docs.python.org/3/objects.inv', <class 'requests.exceptions.InvalidSchema'>, InvalidSchema(...))
b
```
Not sure this is the cause of anything though...

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 ## Python library
+- Improved implementation of halo model quantities (n(M), b(M), c(M)) (#668, #655, #656, #657, #636).
```
beckermr: Ack. This needs to go above v2. Going to change myself.

damonge:
Great, should we get a quick look from @tmcclintock ?

beckermr:
See the slack discussion on deprecation versus and API change.

beckermr:
Also pls squash and merge!

beckermr:
If you want to wait for @tmcclintock feel free. I am ok merging.

damonge:
awesome

beckermr:
Pls back this out and squash!

damonge:
ah crap, I forgot about that

PR #692, opened 2019-10-21
BUG fix PyPi release by adding extra files to the sdist

beckermr:
This PR adds some extra files to the sdist in order to fix the PyPi release. IDK how we did not catch this before. :/

PR #693, opened 2019-10-22
Switch to token authentification for pypi upload

EiffL:
The travis script currently attempts  (and fails) to upload new tagged releases to pypi, because we have a weird password that we never manage to correctly encrypt with travis.

So, this PR replaces the current PyPi authentication method (username+password) by a token-based method. I created a token for the PyPi CCL project, encrypted it, and replaced the old encrypted password.

So this should work, just need to decide whether everyone is ok with automated pushed on newly tagged release. Otherwise, we should comment out/scrap the deploy section of the travis script


beckermr:
I say we comment out automated releases. Also can you approve #692 ?

EiffL:
@beckermr  Alright, I commented the deploy section, if we ever change our mind, it's just a matter of uncommenting it, everything is otherwise well configured.

PR #701, opened 2019-11-13
MAINT clean house after #688

beckermr:
This PR cleans up after #688 

The history will still be full of muck, but we can cherry pick 465d4c5 later when we want to push to the v2 branch


beckermr:
Closing in favor of just living with things as they are.

PR #703, opened 2019-11-13
Bringing back old mass function functions for bckw compat.

### LINKED ISSUE #702: Halo backwards compatibility
We need to create functions to compute mass functions and mass bias that have the same syntax used in v2 but use the new functionality for backwards compatibility. We will mark them as deprecated and remove them altogether in v3

damonge:
Closes #702

### DIFF
--- a/pyccl/__init__.py
+++ b/pyccl/__init__.py
@@ -13,7 +13,7 @@ if environ.get("CLASS_PARAM_DIR") is None:
     environ["CLASS_PARAM_DIR"] = path.dirname(path.abspath(__file__))
 
 from . import ccllib as lib
-from . import core, constants, background, power, halomodel, pk2d, haloprofile, halos
+from . import core, constants, background, power, halomodel, pk2d, haloprofile, halos, massfunction
 
 # Core data structures
 from .core import Cosmology
@@ -33,6 +33,9 @@ from .power import linear_matter_power, nonlin_matter_power, sigmaR, \
 
 # BCM stuff
 from .bcm import bcm_model_fka
+
+# Old halo mass function
+from .massfunction import massfunc, halo_bias, massfunc_m2r
 
 # Cl's and tracers
 from .tracers import Tracer, NumberCountsTracer, WeakLensingTracer, CMBLensingTracer, \
--- a/pyccl/massfunction.py
+++ b/pyccl/massfunction.py
@@ -7,11 +7,13 @@ from .power import sigmaM  # noqa
 @deprecated(hal.MassFunc)
 def massfunc(cosmo, halo_mass, a, overdensity=200):
     """Halo mass function, dn/dlog10M.
+
     Args:
         cosmo (:obj:`Cosmology`): Cosmological parameters.
         halo_mass (float or array_like): Halo masses; Msun.
         a (float): scale factor.
         overdensity (float): overdensity parameter (default: 200)
+
     Returns:
         float or array_like: Halo mass function; dn/dlog10M.
     """
@@ -35,12 +37,14 @@ def massfunc(cosmo, halo_mass, a, overdensity=200):
 
 @deprecated(hal.HaloBias)
 def halo_bias(cosmo, halo_mass, a, overdensity=200):
-    """Tinker et al. (2010) halo bias
+    """Halo bias
+
     Args:
         cosmo (:obj:`Cosmology`): Cosmological parameters.
         halo_mass (float or array_like): Halo masses; Msun.
         a (float): Scale factor.
         overdensity (float): Overdensity parameter (default: 200).
+
     Returns:
         float or array_like: Halo bias.
     """
@@ -63,9 +67,11 @@ def massfunc_m2r(cosmo, halo_mass):
     """Converts smoothing halo mass into smoothing halo radius.
     .. note:: This is R=(3M/(4*pi*rho_m))^(1/3), where rho_m is the mean
     matter density.
+
     Args:
         cosmo (:obj:`Cosmology`): Cosmological parameters.
         halo_mass (float or array_like): Halo masses; Msun.
+
     Returns:
         float or array_like: Smoothing halo radius; Mpc.
     """
--- a/pyccl/massfunction.py
+++ b/pyccl/massfunction.py
@@ -1,5 +1,6 @@
 from . import halos as hal
 from .pyutils import deprecated
+from .errors import CCLError
 
 
 @deprecated(hal.MassFunc)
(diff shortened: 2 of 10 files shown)


damonge:
Alright @beckermr , this is ready for review

pyccl/__init__.py
```diff
@@ -34,6 +34,9 @@
 # BCM stuff
 from .bcm import bcm_model_fka
 
+# Old halo mass function
+from .massfunction import massfunc, halo_bias, massfunc_m2r
```
beckermr: You are missing sigmaM here.

pyccl/massfunction.py
```diff
@@ -0,0 +1,68 @@
...
+        cosmo (:obj:`Cosmology`): Cosmological parameters.
+        halo_mass (float or array_like): Halo masses; Msun.
+        a (float): Scale factor.
+        overdensity (float): Overdensity parameter (default: 200).
+    Returns:
+        float or array_like: Halo bias.
+    """
+    mdef = hal.MassDef(overdensity, 'matter')
+    mf_par = cosmo._config_init_kwargs['mass_function']
+    if mf_par == 'tinker10':
```
beckermr: Halo bias supports shethtormen too https://github.com/LSSTDESC/CCL/blob/v2.0.1/src/ccl_massfunc.c#L489

pyccl/massfunction.py
```diff
@@ -0,0 +1,68 @@
...
+        bf = hal.HaloBiasTinker10(cosmo, mdef)
+        return bf.get_halo_bias(cosmo,
+                                halo_mass,
+                                a)
+    else:
+        raise NotImplementedError("No b(M) fitting function implemented for "
```
beckermr: This probably raised a CCLError before.

beckermr:
One more comment. We should copy paste the old unittests instead of making new ones. This will help us make sure the API is the same.

pyccl/__init__.py
damonge: do we want to have two copies of sigmaM? Now it lives in power, but the API is still the same.

pyccl/massfunction.py
damonge: Those bias functions are completely useless: they will return errors unless you use the virial overdensity **exactly**. I don't know anyone who ever managed to use them at all, because floating-point comparison of Delta_vir always failed.
I think it doesn't make sense to waste any cycles enforcing backward-compatibility of those.

pyccl/massfunction.py
damonge: If we're not supporting the sheth&tormen ones anymore, I think we should check here.

pyccl/__init__.py
beckermr: The API includes function locations in the python package, so we have to.

pyccl/massfunction.py
beckermr: We have to for API compatibility. If someone, somewhere did manage to use them, we should support it. It can be as fragile as it was before, it just needs to be there.

pyccl/massfunction.py
beckermr: The errors raised are also part of the API.

damonge:
alright, past tests added!

pyccl/__init__.py
damonge: this doesn't make much sense to me. Users called `sigmaM` as `pyccl.sigmaM`, and they can keep doing so.
If we do this, we need to either change the name of the function in the new location or make it harder for users to call it (e.g. `pyccl.power.sigmaM`).

beckermr: Ahhh sorry. I mean we need a copy of sigmaM inside the massfunction module.

pyccl/massfunction.py
damonge: OK, but it won't be as fragile as it was before. This will now call the new implementation of this function, which will not check for Delta_vir (because it doesn't have to, this bias function was actually derived for FoF masses)

pyccl/massfunction.py
damonge: OK, for the reasons above I'll just change this for a CCLError then.

pyccl/massfunction.py
beckermr: That's fine. Then we have improved things! Improved implementations of APIs are great!

pyccl/__init__.py
damonge: ok, sorry, still confused: this will create two copies, one in power, one in massfunction. Both with the same signature, both imported in __init__. Won't this create problems?

pyccl/massfunction.py
damonge: :+1: 
OK, great.

pyccl/__init__.py
beckermr: I think you can put `from .power import sigmaM` at the top of `massfunction.py` and this will work.

beckermr: you don't need to import into the init twice. I said the wrong thing above.

beckermr: we simply need a copy at `pyccl.massfunction.sigmaM`

damonge: ok, let me check that

pyccl/massfunction.py
damonge: Done

pyccl/massfunction.py
damonge: Done

pyccl/__init__.py
damonge: that seems to work, done

pyccl/massfunction.py
```diff
@@ -0,0 +1,72 @@
...
+@deprecated(hal.MassFunc)
+def massfunc(cosmo, halo_mass, a, overdensity=200):
+    """Halo mass function, dn/dlog10M.
+    Args:
+        cosmo (:obj:`Cosmology`): Cosmological parameters.
+        halo_mass (float or array_like): Halo masses; Msun.
+        a (float): scale factor.
+        overdensity (float): overdensity parameter (default: 200)
+    Returns:
+        float or array_like: Halo mass function; dn/dlog10M.
```
beckermr: Doc string needs spaces between the parts.

pyccl/massfunction.py
```diff
@@ -0,0 +1,72 @@
...
+@deprecated(hal.HaloBias)
+def halo_bias(cosmo, halo_mass, a, overdensity=200):
+    """Tinker et al. (2010) halo bias
+    Args:
+        cosmo (:obj:`Cosmology`): Cosmological parameters.
+        halo_mass (float or array_like): Halo masses; Msun.
+        a (float): Scale factor.
+        overdensity (float): Overdensity parameter (default: 200).
+    Returns:
+        float or array_like: Halo bias.
```
beckermr: Ditto here.

pyccl/massfunction.py
```diff
@@ -0,0 +1,72 @@
...
+
+@deprecated(hal.mass2radius_lagrangian)
+def massfunc_m2r(cosmo, halo_mass):
+    """Converts smoothing halo mass into smoothing halo radius.
+    .. note:: This is R=(3M/(4*pi*rho_m))^(1/3), where rho_m is the mean
+    matter density.
+    Args:
+        cosmo (:obj:`Cosmology`): Cosmological parameters.
+        halo_mass (float or array_like): Halo masses; Msun.
+    Returns:
```
beckermr: Ditto here.

damonge: Done

pyccl/massfunction.py
damonge: Done

pyccl/massfunction.py
damonge: These were the previous docstrings. Sigh. Done!

beckermr: Ugh sorry! Never to late to make things better!

Issue #695, opened 2019-10-30
Update all examples to match v2


elisachisari:
We should verify that all examples work and if not, update them. This is the list of Jupyter networks that need to be revisited: 

- [ ] examples/3x2demo.ipynb
- [ ]  examples/Halo model.ipynb
- [ ]  examples/Correlation.ipynb
- [ ]  examples/RSD_benchmark.ipynb
- [ ]  examples/Correlation_3d.ipynb
- [ ]  examples/Correlation_3dRSD.ipynb
- [ ]  examples/SLAC Feb2018 Demo.ipynb
- [ ]  examples/GeneralizedTracers.ipynb
- [ ]  examples/RSD_benchmark_read.ipynb
- [ ]  examples/Benchmark_Comparisons.ipynb
- [ ]  examples/NFW_profile_3D_2D_demo.ipynb
- [ ]  examples/Power spectrum example.ipynb
- [ ]  examples/MCMC Likelihood Analysis.ipynb
- [ ]  examples/Halo-mass-function-example.ipynb
- [ ]  examples/Distance Calculations Example.ipynb
- [ ]  examples/muSigma_modified_gravity_demo.ipynb
- [ ]  examples/Lensing angular power spectrum.ipynb
- [ ]  examples/Reading-writing-Cosmology-objects.ipynb

elisachisari:
Upon discussion on the Slack channel, we are probably going to get rid of some of these examples. I would propose to re-structure examples as follows:

- [ ]  An example for core functions like H(z), distances, etc.
- [ ]  An example for power spectra functionality, including halo model and MG.
- [ ]  An example for Cells + correlations.
- [ ]  The benchmark comparison notebook.
- [ ]  examples/Reading-writing-Cosmology-objects.ipynb
- [ ]   examples/NFW_profile_3D_2D_demo.ipynb
- [ ]   examples/MCMC Likelihood Analysis.ipynb
- [ ]   examples/GeneralizedTracers.ipynb

We'd go from 18 notebooks to 8. We'd be losing the RSD and the 3D correlation, but they appear in the benchmark notebook. Thoughts?

elisachisari:
So I'm thinking of keeping the benchmark notebook in this repo. This is where all the benchmark testing happens and otherwise we'd have to duplicate all the benchmark files in the CCLX repo. The other examples we would move.

elisachisari:
[...]

PR #705, opened 2019-11-18
Updating benchmark comparison notebook to v2

elisachisari:
Related to issue #695.


elisachisari:
Some progress on this: I was able to bring most of the tests to their original status by matching the tests/test*py set-ups. There are still some issues: power spectra have decreased in accuracy for some splines and for cases with massive neutrinos. Some of this could be linked to my set-up (4 tests failing in this computer, though travis says everything is ok). But not all the failures are due to this. The spline accuracy check has decreased in quality since v2, not only for this computer.

elisachisari:
Problems currently encountered in insets 21,22,26.

elisachisari:
Also 25 (baryons) has decreased in accuracy

elisachisari:
The most recent commit runs the benchmark notebook using a system where the tests (except new fast-pt capability) are passing. 

The changes in the benchmark notebook I was able to identify compared to the previous released version are:
* The BCM test decreased from 1e-12 to 1e-9. However, the tolerance in the python test is set to 1e-4, so it is obviously still passing.
* The spline accuracy checks look very different. I can’t find this anywhere in the repo. I think this test was removed if it was ever run automatically in the test suite.
* C_ell plots have changed in detail, but maintain the same accuracy. It could be a consequence of changes in some integration parameters, or minor differences in the cosmology that I haven’t been able to match.
* xi_3D also changed a bit, but the order of magnitude of the accuracy checks has remained the same.
* All non Limber plots have been removed.

c-d-leonard:
@elisachisari looks like you've added the last plot - any idea what's happening with the one travis test failing? I've not seen that issue before. I'll approve as soon as we resolve that.

elisachisari:
This looks like a travis internal problem. I re-started the jobs there.

elisachisari:
@c-d-leonard Tests pass now, I brought master in and that solved the issue.

c-d-leonard:
Sorry for the delay on this. Looks great! I'm approving - want to do the honours of merging @elisachisari?

PR #713, opened 2019-11-27
Return tracer functions

### LINKED ISSUE #712: ENH add methods to get out effective quantities from tracers
It would be very helpful to have canned methods to get out the effective IA bias as a function of redshift etc from the new tracer objects. 

@damonge for viz

damonge:
Closes #712 
Currently can return radial kernel. If @beckermr approves of this setup I'll implement transfer functions, ell-factors, and bessel derivative orders.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   - "./.travis/install.sh"
   - export PATH=$HOME/miniconda/bin:$PATH
   - source activate test-environment
-  - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then conda install -y llvm-openmp clang=4.0.1; fi
+  - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then conda install -y llvm-openmp; fi
   # this is a hack for conda osx compilers
   - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then export DYLD_FALLBACK_LIBRARY_PATH=${CONDA_PREFIX}/lib; fi
   - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then export CONDA_BUILD_SYSROOT=/; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
   - "./.travis/install.sh"
   - export PATH=$HOME/miniconda/bin:$PATH
   - source activate test-environment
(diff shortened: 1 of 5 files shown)


damonge:
Actually, I implemented everything. Let me know what you think.

damonge:
as I said in the issue: why do you prefer that? If you called the constructor you should have access to those quantities already.

beckermr:
Not necessarily. This is a feature I want so if you don’t want to do it I will.

damonge:
No, I'm fine doing it. Just to be clear: would it be enough with just returning the input arrays?

beckermr:
No need to redo it. I’d rather they use the internal methods to recompile them. This would be a better check.

damonge:
Humm, that's gonna be annoying, I think. For number counts it's as simple as dividing the radial kernel by H(z), but for lensing you'd have to deconvolve the N(z) from the lensing kernel. Is this what you were thinking?

beckermr:
I’m ok with the dndz being separate. I think having access to what is actually in the kernel is important if at all possible.

damonge:
OK, so the IA bias and the galaxy bias are just given by the transfer functions, so you can get those by using the `get_transfer` method I just implemented (I can write a wrapper of it in the WeakLensing tracer for the IA bias and another one for the galaxy bias in the NumberCounts tracer).

The N(z) is almost completely lost in the WeakLensing tracer, since it's convolved into the lensing kernel. The magnification bias s(z) is also lost into the magnification kernel.

beckermr:
Per our discussion, we should add the dndz stuff. Does that go here or another PR?

damonge:
I wasn't sure if that's what you wanted.
That can be done easily at the python level. Is there anything else you wanted to have in there?

beckermr:
I think that is fine for now. Anything that we cannot get back by some other means.

damonge:
Do you want the input array (easy version) or the array interpolated onto whatever z values you pass as input (slightly harder, probably will use `scipy.interpolate`?

beckermr:
i'd say a get_dndz method?

damonge:
Which takes z as an argument and returns an interpolated dndz? Or which just returns the original input arrays?

beckermr:
Interpolated sorry.

damonge:
Done @beckermr !

PR #717, opened 2020-01-01
Halo model P(k)

### LINKED ISSUE #668: Halo theory in CCL

damonge:
Functionality for halo-model-based power spectra

Closes #668


damonge:
This is ready for review (@beckermr @c-d-leonard @elisachisari )
What I've done:

- Written a generic halo model power spectrum calculator.
- Tested that it reproduces the currently implemented halo model power spectrum (which is limited to just the matter power spectrum with very specific choices of halo mass functions, biases, profiles etc.) to the required accuracy.

What I haven't done yet:

- Remove all traces of the previous halo model power spectrum calculator. I've left this for the next PR in order to simplify this one.

damonge:
(@beckermr just checking that this one didn't slip under the radar, since I made two PRs today - perfectly fine if you're too busy)

beckermr:
Yup doing it today. Was at the doctor. All is well. Thanks for poking me because I do forget!

pyccl/halos/pk_hm.py
```diff
@@ -0,0 +1,569 @@
...
+
+        Args:
+            prof (:class:`~pyccl.halos.profiles.HaloProfile`):
+                halo profile for which the second-order moment
+                is desired.
+            cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
+            k (float or array_like): comoving wavenumber in Mpc^-1.
+            M (float or array_like): halo mass in units of M_sun.
+            a (float): scale factor.
+            prof_2 (:class:`~pyccl.halos.profiles.HaloProfile`):
```
beckermr: This underscore is funny.

beckermr: the one in the name.

pyccl/halos/pk_hm.py
```diff
@@ -0,0 +1,569 @@
...
+
+    Args:
+        cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
+        l10M_min (float): logarithmic mass (in units of solar mass)
+            corresponding to the lower bound of the integrals in
+            mass. Default: 8.
+        l10M_max (float): logarithmic mass (in units of solar mass)
+            corresponding to the upper bound of the integrals in
+            mass. Default: 16.
+        nl10M (int): number of samples in log(Mass) to be used in
```
beckermr: these parameter names could be more verbose and clear (e.g., log10M_min)

pyccl/halos/pk_hm.py
```diff
@@ -0,0 +1,569 @@
...
+        integration_method_M (string): integration method to use
+            in the mass integrals. Options: "simpson" and "spline".
+            Default: "simpson".
+        k_min (float): some of the integrals solved by this class
+            will often be normalized by their value on very large
+            scales. This parameter (in units of inverse Mpc)
+            determines what is considered a "very large" scale.
+            Default: 1E-5.
+    """
+    def __init__(self, cosmo, **kwargs):
```
beckermr: The defaults need to be in the function signature. Using bare `kwargs` doesn't do the error checking on misspelled keywords which leads to silent bugs.

damonge:
Thanks a lot @beckermr . Yes, let me make a notebook.

damonge:
Ok, @beckermr have a look here:
https://github.com/LSSTDESC/CCLX/blob/hm_pk/Halo-model-Pk.ipynb

The idea behind the `ProfileCovar` object is that it should tell you how to compute the two-point correlator of two profiles. Usually this is just the product of the mean profiles in most halo models, but this is not always the case. I give an example of this in the notebook above.

pyccl/halos/pk_hm.py
damonge: Do you mean you'd prefer `prof2` to `prof_2`? All done in that case.

pyccl/halos/pk_hm.py
damonge: Done

pyccl/halos/pk_hm.py
damonge: Done

beckermr:
Thanks for the notebook! That helps a lot! But more questions. Is `HMCalculator` supposed to hold any halo model computation we want overtime? Is the plan to add more and methods, if say we want three-point stats or we want to compute a non-trivial 4pt function for some covariance?

I'm trying to figure out how we grow with this API.

damonge:
Very good question.
Yes, my intention was that we would grow this overtime.

So, in all N-point halo model calculations we need to compute integrals of the form:
.

So, for example, `HMCalculator.mean_profile` computes:

`HMCalculator.bias` computes:

`HMCalculator.pk` computes:


If we want to include bispectra later on, we'll need to add the following two integrals:


So, for instance, I could change the names of these functions so that it'd be easier to generalize them later on (e.g. `mean_profile` -> `mean_profile_1pt`. `bias` -> `bias_1pt`. `pk` -> `power_spectrum`.

damonge:
I could also rename the current `ProfileCovar` to `Profile2pt`

beckermr:
Got it. Then I think we should refactor the API a bit. One giant object for all of this stuff might get unwieldy. Instead, what if we did this.

1. We have a halo model integral computer thingy which computes the various `I`'s etc. 
2. We create functions that use the integrator thing and combine them to compute pk, bispectra, etc.

One more question. Why is the profile covar not a part of the profile object itself?

damonge:
OK, sounds good. This is pretty much what it's doing right now (the various internal methods `_u_k_from_arrays` and `_b_k_from_arrays` do this, so I think the only thing we need to do according to your philosophy is to take the `bias`, `mean_profile` and `pk` functions out of the object. Correct?

The answer to your second question: the profile covar must combine different profiles (not just one).

beckermr:
Agreed we are mostly there! I would be ok with an integrator class that had methods named according to the standard `I` notation above. These should all share state. By factoring out the combination of the parts into other code, we can enable more flexible combinations in the future.

damonge:
OK, refactored! I renamed things a bit, and move the covariance stuff to its own separate file

pyccl/halos/halo_model.py
```diff
@@ -0,0 +1,523 @@
...
+    def _check_hbias(self, hbias):
+        if not isinstance(hbias, HaloBias):
+            raise TypeError("hbias must be of type `HaloBias`")
+
+    def _check_prof(self, prof):
+        if not isinstance(prof, HaloProfile):
+            raise TypeError("prof must be of type `HaloProfile`")
+
+
+def halomod_mean_profile_1pt(cosmo, hmc, k, a, massfunc, prof,
```
beckermr: I would make this a method on HMCalculator and call it something like `I_1_0` or maybe `int_1_0` or w/e. Ditto for the other integral.

pyccl/halos/halo_model.py
```diff
@@ -0,0 +1,523 @@
...
+                                            mbf0, uk) * norm
+
+    if np.ndim(a) == 0:
+        out = np.squeeze(out, axis=0)
+    if np.ndim(k) == 0:
+        out = np.squeeze(out, axis=-1)
+    return out
+
+
+def halomod_power_spectrum(cosmo, hmc, k, a, massfunc, hbias, prof,
```
beckermr: Why doesn't this function use the two functions above?

beckermr: Sorry no wrong question. 🤦‍♂ 

We should add methods like `I_2_0` or whatever right?

pyccl/halos/halo_model.py
damonge: These things already exist within `HMCalculator` as private methods. This is just me exposing them outside of it.

damonge: OK, wait, let me see if I can do this in a cleaner way.

damonge:
OK @beckermr , I've compactified things as much as possible. The current structure is set up so that you don't recompute quantities (e.g. mass functions) twice. If I separated everything into individual integrals, this wouldn't be the case anymore (e.g. you would compute mass functions twice for the power spectrum). This way, `HMCalculator` allows you to calculate all of these integrals given mass functions, biases, profiles etc. already evaluated at the correct mass nodes (for which it also provides useful methods).

pyccl/halos/halo_model.py
damonge: Check now

damonge:
@beckermr : just checking that this is still in the radar

beckermr:
yup. I am still very confused by the PR. So the halo model P(k) is 

`I_0_v * I_0_v + I_1_v * I_1_u * P_lin(k)`

for example. 

I would expect the `I_*` methods to live in the calculator and the power spectrum call to be very simple, like literally doing that formula and some minimal setup. 

So why is it calling all of these internal methods on `HMCalculator`? Do we need to implement some sort of caching mechanism in `HMCalculator`?

damonge:
We could think about something like that. The main reason right now is that many of these integrals need similar ingredients, for instance the mass function. So if you compute all of them independently, you end up recalculating things unnecessarily.

damonge:
Note that the `I_*` methods do live in the calculator. It's just that they accept arrays of evaluated mass functions instead of evaluating them themselves.

beckermr:
I think we should cache the ingredients.

beckermr:
I am thinking on this hard because I don't want this API to last and want to figure out how to do that.

damonge:
Why is that better than the current implementation (where each function takes care of computing what it needs). As it is the `HMCalculator` is a way to automatically sample and integrate whatever you need.

beckermr:
We shouldn't be calling private methods on HMCalculator. Right now you are calling a bunch of checks etc each time. So we either need to put the pk methods back on it, or we should fully separate them. I think we should do the second thing.

damonge:
OK, let me try something.

damonge:
OK @beckermr have another look. There will be some docstrings missing, but see if the API makes sense.

pyccl/halos/halo_model.py
```diff
@@ -0,0 +1,557 @@
...
+            value of `k`.
+        """
+        # Compute mass function and halo bias
+        self._get_ingredients(a, cosmo, True)
+        uk = prof.fourier(cosmo, k, self.mass, a,
+                          mass_def=self.mass_def).T
+        i11 = self._integrate_over_mbf(uk)
+        return i11
+
+    def I_0_2(self, cosmo, k, a, prof_2pt, prof1, prof2):
```
beckermr: I think you need None defaults here. I'd put prof1 first, and then default prof_2pt and prof2 to None.

pyccl/halos/concentration.py
```diff
@@ -337,7 +337,7 @@ def _setup(self):
         if self.mdef.Delta == 'vir':
             self.A = 7.85
             self.B = -0.081
-            self.C = -0.47
+            self.C = -0.71
```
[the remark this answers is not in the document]
damonge: Yes, done.

pyccl/halos/hbias.py
```diff
@@ -144,10 +144,19 @@ class HaloBiasSheth99(HaloBias):
             a mass definition object.
             this parametrization accepts FoF masses only.
             If `None`, FoF masses will be used.
+        mass_def_strict (bool): if False, consistency of the mass
```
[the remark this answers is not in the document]
damonge: Done

benchmarks/test_halomod.py
```diff
@@ -24,23 +24,35 @@ def test_halomod(model):
...
+    prf = ccl.halos.HaloProfileNFW(cM)
+    hmc = ccl.halos.HMCalculator(cosmo, hmf, hbf, mass_def)
+
+    z = 0.
     k = data_z0[:, 0] * cosmo['h']
     pk = data_z0[:, -1] / (cosmo['h']**3)
-    pk_ccl = ccl.halomodel_matter_power(cosmo, k, 1)
+    pk_ccl = ccl.halos.halomod_power_spectrum(cosmo, hmc, k, 1./(1+z),
+                                              prf, normprof1=True)
     tol = pk * HALOMOD_TOLERANCE
```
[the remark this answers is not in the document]
damonge: Yes, and it's faster. I was happy to see that! All of these integrands are pretty smooth...

pyccl/halos/hmfunc.py
```diff
@@ -201,10 +201,19 @@ class MassFuncSheth99(MassFunc):
             a mass definition object.
             this parametrization accepts FoF masses only.
             If `None`, FoF masses will be used.
+        mass_def_strict (bool): if False, consistency of the mass
```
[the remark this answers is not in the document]

pyccl/tests/test_pkhm.py
```diff
@@ -0,0 +1,222 @@
...
+        if sk == 0:
+            k = 0.1
+        else:
+            k = np.logspace(-2., 0., sk)
+        if sa == 0:
+            a = 1.
+        else:
+            a = np.linspace(0.5, 1., sa)
+        p = func(k, a)
+        assert np.shape(p) == sh
```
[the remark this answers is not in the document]
damonge: Done

pyccl/halos/halo_model.py
```diff
@@ -0,0 +1,557 @@
...
+            raise NotImplementedError("Only \'simpson\' and 'spline' "
+                                      "supported as integration methods")
+        elif self.precision['integration_method_M'] == 'simpson':
+            from scipy.integrate import simps
+            self.integrator = simps
+        else:
+            self.integrator = self._integ_spline
+
+        self.a_current_mf = -1
+        self.a_current_bf = -1
```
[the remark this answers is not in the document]
damonge: Done

pyccl/halos/halo_model.py
damonge: I'd prefer not to default prof_2pt here (I don't think it's so practical, since one would have to instantiate the default class every time - e.g. at every new value of the scale factor).
Otherwise, I've changed the ordering and defaulted prof2

damonge:
Thanks @beckermr . All done!

damonge:
Ping @beckermr

PR #704, opened 2019-11-14
Halo profiles

damonge:
Addresses #668 
Incidentally, it also
Closes #217

PR #720, opened 2020-01-07
Array resampling

damonge:
this implements a python-level routine to resample arrays through interpolation.

This is part of #704

### DIFF
--- a/pyccl/pyutils.py
+++ b/pyccl/pyutils.py
@@ -7,6 +7,13 @@ import functools
 import warnings
 import numpy as np
 
+extrap_types = {'none': lib.f1d_extrap_0,
+                'constant': lib.f1d_extrap_const,
+                'linx_liny': lib.f1d_extrap_linx_liny,
+                'linx_logy': lib.f1d_extrap_linx_logy,
+                'logx_liny': lib.f1d_extrap_logx_liny,
+                'logx_logy': lib.f1d_extrap_logx_logy}
+
 
 def check(status, cosmo=None):
     """Check the status returned by a ccllib function.
@@ -282,6 +289,28 @@ def _vectorize_fn4(fn, fn_vec, cosmo, x, a, d, returns_status=True):
     # Check result and return
     check(status, cosmo_in)
     return f
+
+
+def resample_array(x_in, y_in, x_out,
+                   extrap_lo='none', extrap_hi='none',
+                   fill_value_lo=0, fill_value_hi=0):
+    if extrap_lo not in extrap_types.keys():
+        raise ValueError("'%s' is not a valid extrapolation type. "
+                         "Available options are: %s"
+                         % (extrap_lo, extrap_types.keys()))
+    if extrap_hi not in extrap_types.keys():
+        raise ValueError("'%s' is not a valid extrapolation type. "
+                         "Available options are: %s"
+                         % (extrap_hi, extrap_types.keys()))
+
+    status = 0
+    y_out, status = lib.array_1d_resample(x_in, y_in, x_out,
+                                          fill_value_lo, fill_value_hi,
+                                          extrap_types[extrap_lo],
+                                          extrap_types[extrap_hi],
+                                          x_out.size, status)
+    check(status)
+    return y_out
 
 
 def deprecated(new_function=None):
--- a/pyccl/pyutils.py
+++ b/pyccl/pyutils.py
@@ -294,6 +294,26 @@ def _vectorize_fn4(fn, fn_vec, cosmo, x, a, d, returns_status=True):
 def resample_array(x_in, y_in, x_out,
                    extrap_lo='none', extrap_hi='none',
                    fill_value_lo=0, fill_value_hi=0):
+    """ Interpolates an input y array onto a set of x values.
+
+    Args:
+        x_in (array_like): input x-values.
+        y_in (array_like): input y-values.
+        x_out (array_like): x-values for output array.
+        extrap_lo (string): type of extrapolation for x-values below the
+            range of `x_in`. 'none' (for no interpolation), 'constant',
+            'linx_liny' (linear in x and y), 'linx_logy', 'logx_liny' and
+            'logx_logy'.
+        extrap_hi (string): type of extrapolation for x-values above the
+            range of `x_in`.
+        fill_value_lo (float): constant value if `extrap_lo` is
+            'constant'.
+        fill_value_hi (float): constant value if `extrap_hi` is
+            'constant'.
+    Returns:
+        array_like: output array.
+    """
+
     if extrap_lo not in extrap_types.keys():
         raise ValueError("'%s' is not a valid extrapolation type. "
                          "Available options are: %s"
--- a/src/ccl_correlation.c
+++ b/src/ccl_correlation.c
@@ -70,7 +70,9 @@ static void ccl_tracer_corr_fftlog(ccl_cosmology *cosmo,
 
   //Interpolate input Cl into array needed for FFTLog
   ccl_f1d_t *cl_spl=ccl_f1d_t_new(n_ell,ell,cls,cls[0],0,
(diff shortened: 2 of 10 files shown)


pyccl/pyutils.py
```diff
@@ -284,6 +291,28 @@ def _vectorize_fn4(fn, fn_vec, cosmo, x, a, d, returns_status=True):
     return f
 
 
+def resample_array(x_in, y_in, x_out,
+                   extrap_lo='none', extrap_hi='none',
+                   fill_value_lo=0, fill_value_hi=0):
```
beckermr: Doc string.

src/ccl_correlation.c
```diff
@@ -69,31 +69,20 @@ static void ccl_tracer_corr_fftlog(ccl_cosmology *cosmo,
...
+  ccl_f1d_t *cl_spl=ccl_f1d_t_new(n_ell,ell,cls,cls[0],0,
+				  ccl_f1d_extrap_const,ccl_f1d_extrap_logx_logy);
   if(cl_spl==NULL) {
     free(l_arr);
     free(cl_arr);
     *status=CCL_ERROR_MEMORY;
-    ccl_cosmology_set_status_message(cosmo, "ccl_correlation.c: ccl_tracer_corr_fftlog ran out of memory\n");
+    ccl_cosmology_set_status_message(cosmo,
+				     "ccl_correlation.c: ccl_tracer_corr_fftlog "
+				     "ran out of memory\n");
```
beckermr: Wrong error string.

src/ccl_correlation.c
```diff
@@ -213,19 +184,14 @@ static void ccl_tracer_corr_bessel(ccl_cosmology *cosmo,
...
-  #pragma omp parallel default(none) \
-                       shared(cosmo, status, wtheta, n_ell, ell, cls, \
-                              corr_type, cl_spl, theta, n_theta) \
-                       private(w, F, result, eresult, local_status, ith, \
-                               gslstatus, cp)
+#pragma omp parallel default(none)				      \
+  shared(cosmo, status, wtheta, n_ell, ell, cls,		      \
+	 corr_type, cl_spl, theta, n_theta)				\
+  private(w, F, result, eresult, local_status, ith,			\
+	  gslstatus, cp)
```
beckermr: Please preserve the visual spacing here.

src/ccl_correlation.c
```diff
@@ -271,7 +224,8 @@ static void ccl_tracer_corr_bessel(ccl_cosmology *cosmo,
         //TODO: Split into intervals between first bessel zeros before integrating
         //This will help both speed and accuracy of the integral.
         gslstatus = gsl_integration_qag(&F, 0, cosmo->spline_params.ELL_MAX_CORR, 0,
-                                        cosmo->gsl_params.INTEGRATION_EPSREL, cosmo->gsl_params.N_ITERATION,
+                                        cosmo->gsl_params.INTEGRATION_EPSREL,
+					cosmo->gsl_params.N_ITERATION,
```
beckermr: spacing is weird.

src/ccl_correlation.c
```diff
@@ -390,10 +333,9 @@ static void ccl_tracer_corr_legendre(ccl_cosmology *cosmo,
   }
 
   int local_status, i_L;
-
 #pragma omp parallel default(none) \
-                     shared(cosmo, theta, cl_arr, wtheta, n_theta, status, corr_type) \
-                     private(Pl_theta, i, i_L, local_status)
+  shared(cosmo, theta, cl_arr, wtheta, n_theta, status, corr_type)	\
+  private(Pl_theta, i, i_L, local_status)
```
beckermr: again spacing

src/ccl_f1d.c
```diff
@@ -30,26 +32,124 @@ ccl_f1d_t *ccl_f1d_t_new(int n,double *x,double *y,double y0,double yf)
...
+    spl->der_lo = (y[1]-y[0])/(x[1]-x[0]);
+  }
+  else if(spl->extrap_lo_type == ccl_f1d_extrap_linx_logy) {
+    spl->der_lo = log(y[1]/y[0])/(x[1]-x[0]);
+  }
+  else if(spl->extrap_lo_type == ccl_f1d_extrap_logx_liny) {
+    spl->der_lo = (y[1]-y[0])/log(x[1]/x[0]);
+  }
+  else if(spl->extrap_lo_type == ccl_f1d_extrap_logx_logy) {
+    spl->der_lo = log(y[1]/y[0])/log(x[1]/x[0]);
```
beckermr: Do we want to let divide by zero errors just go through here, or should we attempt to catch them?

pyccl/pyutils.py
damonge: Done

src/ccl_correlation.c
damonge: This was the automatic spacing emacs gave me. I've gone back to the previous one.

src/ccl_correlation.c
damonge: Done

src/ccl_correlation.c
damonge: Done

src/ccl_f1d.c
damonge: Caught

damonge:
Thanks. Back to you @beckermr !

src/ccl_f1d.c
```diff
@@ -12,44 +12,168 @@
...
+  else if(spl->extrap_hi_type == ccl_f1d_extrap_logx_logy) {
+    if((y[n-1]*y[n-2]<=0) || (x[n-1]*x[n-2]<=0))
+      *status=CCL_ERROR_SPLINE;
+    else
+      spl->der_hi = log(y[n-1]/y[n-2])/log(x[n-1]/x[n-2]);
+  }
+  else // No extrapolation
+    spl->der_hi = 0;
   return spl;
 }
```
beckermr: Do we need to do anything with status above before we return here?

damonge: Actually I've changed things. Now instead of changing the status, the extrapolator is set to "no extrapolation", so that, if extrapolation is ever needed, an error will be thrown.

damonge: This is because often extrapolation won't be possible because of wrong edges, but it's also never needed anyway (it happens in some of the correlation function tests)

beckermr: Yikes this feels a bit too clever. It is not wrong, but I can see this causing an MCMC chain to fail right in the middle. I think maybe this is not the best idea.

damonge: Humm, not sure why this is a problem (this is exactly the same behaviour that `ccl_correlation.c` had before).
1- Code initalizes this through CCL with some extrapolation option
2- CCL realizes that this extrapolation is not possible and defaults to "no extrapolation"
3- At some point the MCMC asks CCL to extrapolate, and CCL throws an error.

Why is this more dangerous than CLASS failing?

beckermr: Well if someone is asking for log interp but the bin limits are <=0, then that is an actual bug/problem. We should fail immediately instead of letting the error go along silently in the hopes we do not hit it. It is a matter if principle.

damonge: These flags only apply to extrapolation, not interpolation. So we will fail only if extrapolation is ever needed. I think this is right, but I can change it to the previous option if you prefer.

beckermr: Yeah I think it should fail ASAP. People should have to think about this.

damonge: OK, done.
To get back to your initial question: I don't think we need to do anything there, the status gets checked after initializing these structures in ccl_tracers and ccl_correlation.

src/ccl_correlation.c
beckermr: please turn off tabs! ;p

damonge:
Great, thanks! I'll wait for the tests to pass.

PR #722, opened 2020-01-07
Better FFTLog

damonge:
This PR:
- Is part of #704 
- Fixes a bug in our implementation of FFTLog (which wasn't affecting our code before)
- Makes the FFTLog interface a bit easier to use for 2D and 3D-like Hankel transforms. It also adds more documentation.
- Makes this directly accessible through python.


pyccl/ccl_fftlog.i
```diff
@@ -0,0 +1,63 @@
+%module ccl_f1d
```
beckermr: This line is wrong

pyccl/tests/test_fftlog.py
```diff
@@ -0,0 +1,37 @@
...
+        g2 = gamma(0.5 * (alpha + mu))
+        den = np.pi**(dim/2.) * 2**alpha
+        return g1 / (g2 * den * r**(dim - alpha))
+
+    nk = 1024
+    k_arr = np.logspace(-4, 4, nk)
+    fk_arr = f(k_arr)
+
+    status = 0
+    result, status = ccl.ccllib.fftlog_transform(1, k_arr, fk_arr,
```
beckermr: Do we want a nicer python wrapper for this?

pyccl/ccl_fftlog.i
damonge: Good catch! Done

pyccl/tests/test_fftlog.py
damonge: Sure, but I don't want to make it too accessible, because people will misuse it. I'll make a `_fftlog_transform` function.

damonge:
OK, done. Back to you @beckermr !

damonge:
It's just flake8's authoritarian attitude. I'll fix it in a bit.

Issue #721, opened 2020-01-07
Support 2.7


damonge:
Should we keep supporting python2.7?

Pros:
- Maybe someone is still using it.

Cons:
- Our tests take twice as long to run on travis.

damonge:
As part of this, we could use the following trick from @rmjarvis (quoting from slack):
```
allow_failures:
        - python: 2.7
fast_finish: true
```
This will run 2.7, but leave it for last and return success when 3.6 passes.
Testing this in #723

elisachisari:
Merged with #723

PR #723, opened 2020-01-10
Travis to allow 2.7 failures for faster testing

damonge:
See #721

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,18 +2,26 @@ language: generic
 
 matrix:
   include:
-  - os: linux
-    sudo: required
-    env: TOXENV=py27
-  - os: linux
-    sudo: required
-    env: TOXENV=py36
-  - os: osx
-    osx_image: xcode9.4
-    env: TOXENV=py27
-  - os: osx
-    osx_image: xcode9.4
-    env: TOXENV=py36
+    - os: linux
+      sudo: required
+      env: TOXENV=py27
+      name: linux (Python2.7)
+    - os: linux
+      sudo: required
+      env: TOXENV=py36
+      name: linux (Python3.6)
+    - os: osx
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,18 +2,27 @@ language: generic
...
+      env: TOXENV=py27
+      name: linux (Python2.7)
+    - os: osx
+      osx_image: xcode9.4
+      env: TOXENV=py27
+      name: OSX (Python2.7)
+
+  allow_failures:
+    - name: linux (Python 2.7)
+    - name: OSX (Python 2.7)
```
rmjarvis: The name string has to match exactly. You have an extra space here (or missing above).

damonge: Geez... Thanks a lot @rmjarvis !

rmjarvis: Yay.  Now it works.  :)  Although weirdly it ran one of the py2.7 ones before the second (OSX) 3.6.  Not sure how to dictate the order it chooses.  For mine I thought it always did the allow_failures ones after the ones that weren't allowed to fail.  But maybe that's not strictly true...

damonge: Yes, I saw that and was confused. I may dig around a bit

damonge: From https://travis-ci.community/t/why-are-jobs-include-not-run-in-the-correct-order/1132

> Indeed, we push jobs that are allowed to fail to the end of the stage, so that required jobs tend to (there is no guarantee) be executed first, so that the result is known earlier rather than later. This happens silently without any visual cues.

rmjarvis: No guarantee.  OK.  :)

damonge: Oh well... 
Can I get a review of this anyway?

damonge:
Thanks a lot!

PR #725, opened 2020-01-14
DEP remove old C/C++ examples

### LINKED ISSUE #714: DEP remove c/c++ examples
We should remove the old examples in c and c++.

beckermr:
Closes #714


damonge:
But tests are failing

examples/CMakeLists.txt
```diff
@@ -1,17 +0,0 @@
-add_executable(ccl_sample_baryons ccl_sample_baryons.c)
```
damonge: I wonder if the problem has to do with having deleted this. Maybe the root CMakeLists.txt should be changed too.

beckermr: I remove the ref to it from the main one. Should work now I think.

beckermr:
OK @damonge this is working now.

PR #726, opened 2020-01-16
Spline integral

damonge:
This branch:

- Defines a spline integrator function at the C level
- Allows the user to switch to it (instead of using the default GSL integrator)

Three reasons for this:

- Spline integrals are potentially faster and not much less accurate than GSL's quad.
- They won't fail, which can be convenient (they may just give you a less accurate answer).
- I need this type of integration for the halo model power spectrum branch.

### DIFF
--- a/include/ccl_cls.h
+++ b/include/ccl_cls.h
@@ -41,7 +41,7 @@ void ccl_angular_cls_nonlimber(ccl_cosmology *cosmo,
          ccl_cl_tracer_collection_t *trc2,
          ccl_f2d_t *psp,
          int nl_out,int *l_out,double *cl_out,
-	 int *status);
+         int *status);
 
 CCL_END_DECLS
 #endif
--- a/include/ccl_cls.h
+++ b/include/ccl_cls.h
@@ -21,7 +21,9 @@ void ccl_angular_cls_limber(ccl_cosmology *cosmo,
        ccl_cl_tracer_collection_t *trc1,
        ccl_cl_tracer_collection_t *trc2,
        ccl_f2d_t *psp,
-       int nl_out, double *l_out, double *cl_out, int *status);
+       int nl_out, double *l_out, double *cl_out,
+       ccl_integration_t integration_method,
+       int *status);
 /**
  * Computes non-Limber power spectrum for two different tracers at a given ell.
  * @param cosmo Cosmological parameters
@@ -39,7 +41,7 @@ void ccl_angular_cls_nonlimber(ccl_cosmology *cosmo,
          ccl_cl_tracer_collection_t *trc2,
          ccl_f2d_t *psp,
          int nl_out,int *l_out,double *cl_out,
-         int *status);
+	 int *status);
 
 CCL_END_DECLS
 #endif
--- a/include/ccl_utils.h
+++ b/include/ccl_utils.h
@@ -57,8 +57,8 @@ double ccl_j_bessel(int l,double x);
  * @return spline integral
  */
 double ccl_integ_spline(int n,double *x,double *y,
-			double a, double b,
-			gsl_interp_type *T, int *status);
+                        double a, double b,
+                        gsl_interp_type *T, int *status);
 
 CCL_END_DECLS
 
--- a/include/ccl_utils.h
+++ b/include/ccl_utils.h
@@ -10,7 +10,7 @@
 CCL_BEGIN_DECLS
 
 typedef enum ccl_integration_t {
-  ccl_integration_quad = 500,  // GSL's quad
+  ccl_integration_qag_quad = 500,  // GSL's quad
   ccl_integration_spline = 501,  // Spline integral
 } ccl_integration_t;
 
@@ -58,7 +58,7 @@ double ccl_j_bessel(int l,double x);
  */
 double ccl_integ_spline(int n,double *x,double *y,
                         double a, double b,
-                        gsl_interp_type *T, int *status);
+                        const gsl_interp_type *T, int *status);
 
 CCL_END_DECLS
 
--- a/include/ccl_utils.h
+++ b/include/ccl_utils.h
@@ -9,6 +9,11 @@
 
 CCL_BEGIN_DECLS
 
+typedef enum ccl_integration_t {
+  ccl_integration_quad = 500,  // GSL's quad
+  ccl_integration_spline = 501,  // Spline integral
+} ccl_integration_t;
+
 /**
  * Compute bin edges of N-1 linearly spaced bins on the interval [xmin,xmax]
  * @param xmin minimum value of spacing
--- a/include/ccl_utils.h
+++ b/include/ccl_utils.h
@@ -42,6 +42,19 @@ double * ccl_log_spacing(double xmin, double xmax, int N);
 double ccl_j_bessel(int l,double x);
 //Spherical Bessel function of order l (adapted from CAMB)
 
+/**
+ * Compute spline integral.
+ * @param n number of elements in input array.
+ * @param x input x-values.
+ * @param y input y-values.
+ * @param a lower end of integration range.
+ * @param b upper end of integration range (use b<a if you just want to integrate over all of y).
+ * @return spline integral
+ */
+double ccl_integ_spline(int n,double *x,double *y,
+			double a, double b,
+			gsl_interp_type *T, int *status);
+
 CCL_END_DECLS
 
 #endif
--- a/pyccl/ccl_cls.i
+++ b/pyccl/ccl_cls.i
@@ -22,7 +22,7 @@ void angular_cl_vec(ccl_cosmology * cosmo,
                     ccl_cl_tracer_collection_t *clt2,
                     ccl_f2d_t *pspec, double l_limber,
                     double* ell, int nell,
-		    int integration_type,
+                    int integration_type,
                     int nout, double* output,
                     int *status) {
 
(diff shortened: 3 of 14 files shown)


src/ccl_cls.c
```diff
@@ -137,22 +137,91 @@ static double cl_integrand(double lk, void *params) {
...
-  int clastatus, lind;
-  integ_cl_par ipar;
-  gsl_function F;
-  double lkmin, lkmax, l;
-  double result, eresult;
+static void integ_cls_limber_spline(ccl_cosmology *cosmo,
+				    integ_cl_par *ipar,
+				    double lkmin, double lkmax,
+				    double *result, int *status)
+{
```
beckermr: put this bracket at the end of the previous line

src/ccl_cls.c
```diff
@@ -217,50 +292,31 @@ void ccl_angular_cls_limber(ccl_cosmology *cosmo,
         // Get integration limits
         get_k_interval(cosmo, trc1, trc2, l, &lkmin, &lkmax);
 
-        // Integrate
-        gslstatus = gsl_integration_qag(&F, lkmin, lkmax, 0,
```
beckermr: You got rid of the qag method. We should keep the old code the same and simply add a new method.

src/ccl_cls.c
```diff
@@ -217,50 +292,31 @@ void ccl_angular_cls_limber(ccl_cosmology *cosmo,
...
       }
     }
 
-    gsl_integration_workspace_free(w);
-    gsl_integration_cquad_workspace_free(w_cquad);
+    if(integration_method == ccl_integration_quad)
+      gsl_integration_workspace_free(w);
 
     if (local_status) {
       #pragma omp atomic write
```
beckermr: I forgot about atomic writes in the previous PR.

src/ccl_cls.c
damonge: this has just been moved into the `integ_cls_limber_quad` function.

beckermr: Ahhhh I missed it! Sorry! I'll soldier on!

rmjarvis:
> We can change C APIs, but not functionality. So we need to put that back and then add the spline one.
I don't think this is true.  Or at least "functionality" here should mean "do the integral".  Not use a specific algorithm to do the integral.  If the spline integral is always better, it's fine to just switch to that being the only method.

beckermr:
Fair point. I generally don't like changes like this because they can cause surprising results to users later.

rmjarvis:
Sure.  Especially if there is a bug.  :)
My point is that API consistency doesn't require that the numerical calculated values are the same.  Just that the syntax and nominal function of the code is preserved.  Doing things (that are thought to be) better is perfectly acceptable.

include/ccl_cls.h
```diff
@@ -39,7 +41,7 @@ void ccl_angular_cls_nonlimber(ccl_cosmology *cosmo,
          ccl_cl_tracer_collection_t *trc2,
          ccl_f2d_t *psp,
          int nl_out,int *l_out,double *cl_out,
-         int *status);
+	 int *status);
```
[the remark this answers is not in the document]
damonge: My emacs's automatic spacing. TBH, the spacing in this file is kind of arbitrary, but whatever. Fixed.

include/ccl_utils.h
```diff
@@ -42,6 +47,19 @@ double * ccl_log_spacing(double xmin, double xmax, int N);
...
+ * @param x input x-values.
+ * @param y input y-values.
+ * @param a lower end of integration range.
+ * @param b upper end of integration range (use b<a if you just want to integrate over all of y).
+ * @return spline integral
+ */
+double ccl_integ_spline(int n,double *x,double *y,
+			double a, double b,
+			gsl_interp_type *T, int *status);
+
```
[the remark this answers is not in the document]
damonge: Same. done

pyccl/ccl_cls.i
```diff
@@ -22,6 +22,7 @@ void angular_cl_vec(ccl_cosmology * cosmo,
                     ccl_cl_tracer_collection_t *clt2,
                     ccl_f2d_t *pspec, double l_limber,
                     double* ell, int nell,
+		    int integration_type,
```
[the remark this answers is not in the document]
damonge: Done

pyccl/ccl_cls.i
```diff
@@ -70,7 +71,8 @@ void angular_cl_vec(ccl_cosmology * cosmo,
     for (int i=index_nonlimber_last+1; i < nell; i++)
       _ell[i - (index_nonlimber_last+1)] = ell[i];
 
-    ccl_angular_cls_limber(cosmo, clt1, clt2, pspec, (nell - (index_nonlimber_last+1)), _ell, _cl_ell, status);
+    ccl_angular_cls_limber(cosmo, clt1, clt2, pspec, (nell - (index_nonlimber_last+1)), _ell, _cl_ell,
+			   integration_type, status);
```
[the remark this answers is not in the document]
damonge: Done

src/ccl_cls.c
damonge: This is not the style I normally use in C but I see it's been made standard in CCL now, so that's fine fine. Done!

src/ccl_cls.c
damonge: I've now used them in ccl_haloprofiles.c

damonge:
Done! Thanks @beckermr

include/ccl_utils.h
```diff
@@ -9,6 +9,11 @@
 
 CCL_BEGIN_DECLS
 
+typedef enum ccl_integration_t {
+  ccl_integration_quad = 500,  // GSL's quad
```
beckermr: So this method is the same as it was previously. Maybe we should call this `ccl_integration_qag_cquad` or something? After all, `c_quad` is the backuo method.

pyccl/cls.py
```diff
@@ -13,7 +13,7 @@
 
 
 def angular_cl(cosmo, cltracer1, cltracer2, ell, p_of_k_a=None,
-               l_limber=-1.):
+               l_limber=-1., limber_integration_method='quad'):
```
beckermr: Ditto here.

pyccl/cls.py
```diff
@@ -25,6 +25,10 @@ def angular_cl(cosmo, cltracer1, cltracer2, ell, p_of_k_a=None,
             the non-linear matter power spectrum will be used.
         l_limber (float) : Angular wavenumber beyond which Limber's
             approximation will be used. Defaults to -1.
+        limber_integration_method (string) : integration method to be used
+            for the Limber integrals. Possibilities: 'quad' (GSL's `quad`
```
beckermr: And here.

pyccl/tests/test_swig_interface.py
```diff
@@ -102,7 +102,7 @@ def test_swig_cls():
         ccllib.angular_cl_vec,
         COSMO,
         None, None, None,
-        1, 0,
+        1, 0, 501,
```
beckermr: Can we use the enum? This magic number is confusing.

src/ccl_cls.c
```diff
@@ -137,22 +137,89 @@ static double cl_integrand(double lk, void *params) {
...
+static void integ_cls_limber_spline(ccl_cosmology *cosmo,
+				    integ_cl_par *ipar,
+				    double lkmin, double lkmax,
+				    double *result, int *status) {
+  int ik;
+  int nk = (int)(fmax((lkmax - lkmin) / cosmo->spline_params.DLOGK_INTEGRATION + 0.5,
+		      1))+1;
+  double *fk_arr = NULL;
+  double *lk_arr = NULL;
+  lk_arr = ccl_linear_spacing(lkmin, lkmax, nk);
```
beckermr: Is linear the best here? I don't actually know. Asking for a friend.

src/ccl_cls.c
```diff
@@ -217,50 +290,31 @@ void ccl_angular_cls_limber(ccl_cosmology *cosmo,
...
+          ccl_raise_gsl_warning(local_status, "ccl_cls.c: ccl_angular_cl_limber():");
           cl_out[lind] = NAN;
           local_status = CCL_ERROR_INTEG;
         }
       }
     }
 
-    gsl_integration_workspace_free(w);
-    gsl_integration_cquad_workspace_free(w_cquad);
+    if(integration_method == ccl_integration_quad)
```
beckermr: I don't think you need this since you init `w` to NULL above.

include/ccl_utils.h
damonge: good point (I named it from memory). Done.

pyccl/cls.py
damonge: Done

pyccl/cls.py
damonge: Done

pyccl/tests/test_swig_interface.py
damonge: done (not that it matters, since all the inputs here are crap)

src/ccl_cls.c
damonge: The original qag integrator is solving the integral in log(k) too. If you look at the integrand, it is a lot smoother in logk than in k (that's honestly all the thinking that has gone into it).

src/ccl_cls.c
damonge: Done

src/ccl_cls.c
beckermr: ohhhhh darn. I read `ccl_linear_spacing` and assumed it was linear, but you are feeding in logs! My bad! Sorry!

PR #727, opened 2020-01-17
Two parallelized functions

damonge:
This PR:

* Paralelizes FFTLog.
* Paralelizes the spline integrator.
* Makes the latter accessible from python.

### DIFF
--- a/src/ccl_fftlog.c
+++ b/src/ccl_fftlog.c
@@ -1,11 +1,11 @@
 #include <stdlib.h>
 #include <math.h>
 
+#include <complex.h>
 #include <fftw3.h>
 
 #include <gsl/gsl_sf_result.h>
 #include <gsl/gsl_sf_gamma.h>
-#include <complex.h>
 #include "ccl.h"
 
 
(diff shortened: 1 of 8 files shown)


src/ccl_fftlog.c
```diff
@@ -140,83 +141,126 @@ static void fht(int npk, int N,
...
+
+        #pragma omp for
+        for(int j = 0; j < npk; j++) {
+          for(int i = 0; i < N; i++)
+            a[i] = prefac_pk[i] * pk[j][i];
+
+          fftw_execute_dft(forward_plan,a,b);
+          for(int m = 0; m < N; m++)
+            b[m] *= u[m] / (double)(N);       // divide by N since FFTW doesn't normalize the inverse FFT
+          fftw_execute_dft(reverse_plan,b,b);
```
beckermr: Is it safe to reuse plans from multiple threads?

damonge: Yes (in fact having plans be initialized within parallel regions is a no-no)
http://www.fftw.org/fftw3_doc/Thread-safety.html

damonge:
Right. So most of the FFTs we have to solve will be 1D and over vectors of O(1000) elements. In my experience, splitting these transforms over several cores only pays off when the vector is very large, when you have multi-dimensional transforms, or maybe when you only have a single transform.
If you have to solve several FFTs of the same size, I think doing each of them in a core in parallel is more efficient.

damonge:
This will happen, for instance, for halo model power spectra (where you have to compute the Fourier profile for a large number of halo masses).

beckermr:
Good enough for me!

damonge:
Awesome!

PR #734, opened 2020-02-07
Nonlin models

damonge:
Implements fastPT

Things left to do:

- [x] Settle on precomputation of workspace.
- [x] Settle on input P_lin
- [x] Settle on input k sampling.
- [x] Get a benchmark (regression test is an option here).


damonge:
This notebook showcases what the intended use of the API would be:
https://github.com/LSSTDESC/CCLX/blob/nonlin_models/NonlinearBiasFASTPT_detailed.ipynb

.travis/install.sh
```diff
@@ -42,3 +42,5 @@ py36)
     cython "camb>=1"
   ;;
 esac;
+
+pip install https://github.com/JoeMcEwen/FAST-PT/archive/v3.0.2.tar.gz --user
```
beckermr: Don't use a user install here.

beckermr: You should add --no-deps as well.

damonge: Thanks!

damonge: @beckermr It still doesn't seem to be able to find fastpt. pip installs successfully, but it still can't find it. I imagine this has to do with miniconda, but I'm not familiar with that at all. Let me know if you have any thoughts.

beckermr: OK. Let me dig into it.

beckermr: Try this.

beckermr: I put in a stupid bug. Should work now.

jablazek:
Some thoughts, which I will come back and edit as I go through this review.

1. We want the PTWorkspace to be cached by CCL to be accessed for future calculations (like the power spectrum). I think this is automatically done already? A bit more complex, can it be cached in such a way that it is accessible to future steps in a chain being run through Firecrown or cobaya?

2. I want the tracers to be specified in a reasonably efficient way, both for elegance and to avoid error (e.g. asking for nonlinear bias in one place but forgetting to ask for it somewhere else). This was the original reason I wanted everything specified in a single Tracer object.

3. Mainly a question for Matt. Are you happy with how this will interact with Firecrown? We can expand this discussion on the related Firecrown issue.

jablazek:
@damonge @beckermr : it looks like all the PT stuff is now in nl_pt. So instead of ccl.[pt thing] it is ccl.nl_pt.[pt thing]. Is this the preferred structure for non core features?

damonge:
@jablazek 
Thanks a lot! OK, here's my thinking:
1. Sure, I can make it return the PTCalculator if required.
2. I'd rather keep PTTracers and standard Tracers separated. One of them is used to get P(k)s, the other one is used to get C_ells (given a P(k)). It's the same logic why we haven't associated halo profiles with tracers: sure, they are descriptions of physical quantities we may want to correlate, but they serve different purposes in the code.

About `nl_p`: I decided to move things there for the same reason we moved all the halo model stuff into `halos`: this is a module that has significant functionality and, more importantly, it may see a lot of growth in the future (e.g. eventually we'll want bispectra, or trispectra for covariances etc.).

jablazek:
@damonge : there a couple of other things I wanted to check. I will try to finish in the next ~24 hours.

jablazek:
Why the change Workspace -> Calculator? I don't have a strong preference, just curious. I have updated a couple mentions of "workspace."

damonge:
Just to make it consistent with the same vocabulary I used for the halo model module

jablazek:
This is a bit confusing syntax:
>def get_pt_pk2d(cosmo, tracer1, ptc=None, tracer2=None, ...

Is there any way to have ptc not split the tracers, if we are making one optional? Alternatively, make both tracers mandatory to avoid user confusion.

damonge:
Humm. Sure, makes sense. I'll just put it after tracer2

jablazek:
Moving this conversation here from the comments. We can probably turn this into an "improvement" issue after the initial PR:
---

TODO (JAB): Restore ability to pass in pk. Could be useful for custom cases where interpolated Plin isn't needed. And could help with speed-up in some cases.

DAM: Can you describe a case when this is useful?

JAB: We could use a low-res array of (k,Pk), perhaps at fixed cosmology, although in that latter case, we could just precompute the PT terms as well. More generally, sometimes you may want to do PT on something that isn't CCL's output for Pk. For instance, some PT use an iterative 1loop calculation to more accurately deal with nonlinearities or scale-dependent growth. In this case, you would want to pass in Pk from the previous iteration rather than Plin.

jablazek:
> Humm. Sure, makes sense. I'll just put it after tracer2

ah, yeah, that's the obvious solution. my brain isn't quite working.

jablazek:
Another feature addition:
TODO: should we use one_loop_dd_bias_b3nl instead of one_loop_dd_bias

jablazek:
This is mentioned in the above list of features, but it may be slightly higher priority. Do we want to allow nonlinear P_mm to be from 1loop PT? This would require an additional param, since right now use_nonlin is a bool.

jablazek:
Currently, we need two calls to get E and B modes for IA auto spectrum.

jablazek:
@damonge and @beckermr : I'm happy with this. There is one high-priority feature we may want to integrate before merging: allowing 1loop PT to be used for P_d1d1. I can try to do this tonight. The other feature additions mentioned above are probably safely left for the future.

I checked the equations for the various power spectra - they look good to me. Do we want to implement tests for things beyond P_gg? Other tests we want to implement at this stage?

damonge:
Thanks @jablazek 

OK, so:
1- On input P(k):
>We could use a low-res array of (k,Pk), perhaps at fixed cosmology, although in that latter case, we could just precompute the PT terms as well. More generally, sometimes you may want to do PT on something that isn't CCL's output for Pk. For instance, some PT use an iterative 1loop calculation to more accurately deal with nonlinearities or scale-dependent growth. In this case, you would want to pass in Pk from the previous iteration rather than Plin.

How about passing an optional input `Pk2D` structure containing the power spectrum you want to use as the input linear power spectrum that we want to do PT on? If `None`, then the internal linear power spectrum (scaled by linear growth) will be used.

2- k spacing
>  I think allowing the user to pass a generic (k,Pk) is a good feature. It isn't needed if everything is using CCL interpolations for Plin, but that may not always be the case. This could all go in the next iteration of feature additions.

I think this would be taken care of by the above.

3- Growth scaling:
> NOTE: we should add an option that does PT at every a value. This isn't hard, it just takes longer.

Are we happy with scaling with linear growth for the moment and implementing this option later? I don't care much either way.

4- IA BB
> Currently, we need two calls to get E and B modes for IA auto spectrum.

good point. I'll fix that so that, if you pass `return_ia_bb=True`, then both EE and BB will be returned.

5- Pd1d1: currently this is either linear or halofit. As we have said, using FASTPT's own 1-loop power spectrum would be another option. Are we happy to just enable this here? Shouldn't take any time. I imagine it'd also be good if we could get that into the CCL cosmology as its internal non-linear matter power spectrum (as an option to halofit), but that could be done in a different PR.

6- Benchmarks: I've created benchmarks directly from fastpt for all possible combinations. They currently pass (and in achieving that I was able to fix a bug in CCL, so yay!). I don't think we need anything else.

So, current to-do list:

- [ ] IA BB in a single call
- [ ] Non-factorisable growth scaling: do we think it's worth doing this now?
- [ ] Option for 1-loop P_mm

I think I only need an answer for the second one @jablazek

jablazek:
@damonge : will reply to your questions below. Also, note that I am currently without slack or a working version of pyccl (so I can't test things myself, although I can still write code).

One other issue I noticed while working on the 1loop SPT option:
Right now, if you don't pass it a PTC object, it initializes one with everything (including IA), which is "slow" if you only want bias. Should probably parse the tracers first to see what is needed.

Similarly, if you pass a PTC object that is missing some required quantities (e.g. for IA), it crashes with the correct message. But we could instead go back into the object and add the needed quantities (with a message saying that this is being done).

The first issue is more important than the second.

jablazek:
> How about passing an optional input `Pk2D` structure containing the power spectrum you want to use as the input linear power spectrum that we want to do PT on? If `None`, then the internal linear power spectrum (scaled by linear growth) will be used.
> 

I think this is good. What do you mean by 2D? k and a? If we are also passing a, we need to generalize how growth is (or isn't) applied, as mentioned below. Also, needs a check that k values are log spaced.

> 3- Growth scaling:
> 
> Are we happy with scaling with linear growth for the moment and implementing this option later? I don't care much either way.

Yes, definitely a niche feature to add later.

> 
> 4- IA BB
> > Currently, we need two calls to get E and B modes for IA auto spectrum.
> 
> good point. I'll fix that so that, if you pass `return_ia_bb=True`, then both EE and BB will be returned.

perfect.

> 
> 5- Pd1d1: currently this is either linear or halofit. As we have said, using FASTPT's own 1-loop power spectrum would be another option. Are we happy to just enable this here? Shouldn't take any time. I imagine it'd also be good if we could get that into the CCL cosmology as its internal non-linear matter power spectrum (as an option to halofit), but that could be done in a different PR.
> 

I'm working on this now. It's not quite as trivial as I would like (at least the way I am doing it). I'll push asap and you can take a look.

jablazek:
@damonge : ok, I think I have resolved the issue of SPT for Pd1d1, which can also be used just to get the matter power spectrum if you want. Also, this should mean that IA and bias are only initialized if needed. I can't test, so there may be bugs.

damonge:
@beckermr @jablazek @c-d-leonard : I think this is ready for review

jablazek:
@damonge : two things:
1) did we want to update p_ii to output E and B modes at the same time if requested? We probably want to be able to also ask for just B modes.
2) why the change from `elif` to `if` in the line: `if self.with_dd:`
This was designed to only call the basic dd if dd_bias wasn't also called (since the latter does the basic calculation as well)

damonge:
@jablazek 
1- Yep, forgot about that one. Will do tomorrow.
2- That was throwing errors all over the place. I'll check it tomorrow too.

jablazek:
> 2- That was throwing errors all over the place. I'll check it tomorrow too.

Weird. Can you post an error? It looks correct to me, but I'm still without my computer and unable to debug.

damonge:
Alright @jablazek, I've fixed both things. Let's review this thing!
Just tagging possible reviewers: @c-d-leonard @beckermr @elisachisari 
As we said above, this notebook describes how the API works: https://github.com/LSSTDESC/CCLX/blob/nonlin_models/PerturbationTheoryPk.ipynb

jablazek:
>Alright @jablazek, I've fixed both things. Let's review this thing!

@damonge : did a quick review of the latest changes - they look good to me, and I understand now why it was throwing an error before. I'm happy. Will wait for others to review, and happy to revisit anything.

c-d-leonard:
> Alright @jablazek, I've fixed both things. Let's review this thing!
> Just tagging possible reviewers: @c-d-leonard @beckermr @elisachisari
> As we said above, this notebook describes how the API works: https://github.com/LSSTDESC/CCLX/blob/nonlin_models/PerturbationTheoryPk.ipynb

A couple of small comments on the example notebook before I do a proper review:

Tiny issue but under heading 'PTTracers' here you say 'Here we create two' when I think you mean three.

Also, if tracer2 is an optional variable in  get_pt_pk2d, can we see an example where it is left to it's default value? I'm not sure what that behaviour would be offhand.

pyccl/pyutils.py
```diff
@@ -463,3 +465,26 @@ def _spline_integrate(x, ys, a, b):
...
+        f3 = NoneArr
+    else:
+        f1 = np.atleast_1d(np.array(f_arg[0], dtype=float))
+        f2 = np.atleast_1d(np.array(f_arg[1], dtype=float))
+        if arr3:
+            f3 = np.atleast_1d(np.array(f_arg[2], dtype=float))
+    if arr3:
+        return f1, f2, f3
+    else:
+        return f1, f2
```
c-d-leonard: How is this change related to fastpt and the current PR?

pyccl/tracers.py
```diff
@@ -549,26 +548,3 @@ def _check_returned_tracer(return_val):
...
-        f3 = NoneArr
-    else:
-        f1 = np.atleast_1d(np.array(f_arg[0], dtype=float))
-        f2 = np.atleast_1d(np.array(f_arg[1], dtype=float))
-        if arr3:
-            f3 = np.atleast_1d(np.array(f_arg[2], dtype=float))
-    if arr3:
-        return f1, f2, f3
-    else:
-        return f1, f2
```
c-d-leonard: Okay, I see the above has been moved from there, but what is the relevance to this PR?

jablazek:
>Also can we get the changelog and the note updated?
I can look at the note. I will leave the changelog for @damonge, unless he objects.

pyccl/pyutils.py
damonge: This function has become more generally useful, so I moved it here so it can be imported from `pyutils.py`, rather than `tracers.py`

pyccl/tracers.py
damonge: See above!

damonge:
@c-d-leonard thanks a lot for this! I've updated the changelog. @jablazek let me know if I can help with the note.
I'll fix the example notebook as you suggest.

elisachisari:
I've had a look at this PR and ran the example notebook
https://github.com/LSSTDESC/CCLX/blob/nonlin_models/PerturbationTheoryPk.ipynb
Could we possibly make a small addition that shows how to extract the IA C_ell B-modes as well?

damonge:
Done @elisachisari

pyccl/nl_pt/power.py
```diff
@@ -0,0 +1,583 @@
...
+            cd (array_like): overdensity bias for the IA tracer
+                being correlated at the same set of input redshifts.
+
+        Returns:
+            array_like: 2D array of shape `(N_k, N_z)`, where `N_k` \
+                is the size of this object's `ks` attribute, and \
+                `N_z` is the size of the input redshift-dependent \
+                biases and growth factor.
+        """
+        warnings.warn(
```
beckermr: This warning is a bit much. I'd say we put this in the docstring and not bother people.

pyccl/nl_pt/power.py
```diff
@@ -0,0 +1,583 @@
...
+            calculator.
+        tracer2 (:class:`~pyccl.nl_pt.tracers.PTTracer`): the second
+            tracer being correlated. If `None`, the auto-correlation
+            of the first tracer will be returned.
+        sub_lowk (bool): if True, the small-scale white noise
+            contribution will be subtracted for number counts
+            auto-correlations.
+        nonlin_pk_type (str): type of 1-loop matter power spectrum
+            to use. 'linear' for linear P(k), 'halofit' for halofit
+            non-linear power spectrum, 'spt' for standard perturbation
```
beckermr: halofit as the option here is not accurate. The non-linear power could be an emulator. The options should be linear, non-linear or spt

benchmarks/test_ptpk.py
```diff
@@ -0,0 +1,61 @@
...
+ptt['i'] = pt.PTIntrinsicAlignmentTracer(c1=1.9, c2=2.1, cdelta=2.3)
+ptt['m'] = pt.PTMatterTracer()
+
+# Calculator
+ptc = pt.PTCalculator(with_NC=True, with_IA=True,
+                      log10k_min=-4, log10k_max=2,
+                      nk_per_decade=20, pad_factor=0.5)
+
+# Read data
+dirdat = os.path.join(os.path.dirname(__file__), 'data')
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/tracers.py
```diff
@@ -0,0 +1,156 @@
...
+    If a number is provided, a constant bias is assumed.
+    If `None`, a bias of zero is assumed.
+
+    Args:
+        c1 (float or tuple of arrays): a single number or a
+            tuple of arrays (z, c1(z)) giving the first-order
+            alignment bias.
+        c2 (float or tuple of arrays): as above for the
+            second-order alignment bias.
+        cdelta (float or tuple of arrays): as above for the
```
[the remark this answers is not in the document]
damonge: Hmm, I'd rather not remove functionality that is actually useful. Could we just open an issue for this one and potentially extend the standard tracer functionality?

pyccl/tests/test_pt_power.py
```diff
@@ -0,0 +1,174 @@
+import numpy as np
```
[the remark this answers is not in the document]
damonge: I've done something minimal. I can't think of anything else other than repeating the benchmarks (which are already testing the values of all of these functions).

pyccl/nl_pt/power.py
damonge: Done

pyccl/nl_pt/power.py
damonge: Done

damonge:
Thanks a lot @beckermr ! Back to you.

PR #736, opened 2020-02-12
Halo model cleanup

### LINKED ISSUE #733: Clean outdated halo model power spectrum
Perform a general cleanup after #717

damonge:
Closes #733


damonge:
@beckermr can I get a review?
Three main things have happened here:
- I've removed all traces of the previous halo model, which has cleaned up a lot of stuff in C. The functionality is still the same, but trying to use the previous halo model will trigger deprecation warnings.
- We now have a method to populate the non-linear power spectrum from an external 2D spline - this could be useful in the future. Right now we use the generalized halo model to get one of these and then copy it to the internal non-linear matter power spectrum.
- In order to fully reproduce the previous functionality I had to enhance our current implementation of the Tinker mass functions and bias. Before, we could only use them for Delta_matter-like mass definitions, but now we can use them for any spherical overdensity definition.

pyccl/halos/hbias.py
```diff
@@ -303,31 +304,44 @@ def __init__(self, cosmo, mass_def=None, mass_def_strict=True):
...
     def _check_mdef_strict(self, mdef):
-        if isinstance(mdef.Delta, str):
-            return True
-        elif (mdef.Delta < 200.) or (mdef.Delta > 3200.) or \
-             (mdef.rho_type != 'matter'):
+        if mdef.Delta == 'fof':
             return True
         return False
 
+    def _get_Delta_m(self, cosmo, a):
```
beckermr: Should this method live on the parent class? Also, it should be able to handle vir, right?

pyccl/halos/hmfunc.py
```diff
@@ -312,6 +312,18 @@ def __init__(self, cosmo, mass_def=None, mass_def_strict=True):
     def _default_mdef(self):
         self.mdef = MassDef200m()
 
+    def _pd(self, ld):
+        return 10.**(-(0.75/(ld - 1.8750612633))**1.2)
+
+    def _get_Delta_m(self, cosmo, a):
```
beckermr: Same comment here.

pyccl/halos/hmfunc.py
```diff
@@ -430,6 +438,15 @@ def __init__(self, cosmo, mass_def=None, mass_def_strict=True):
     def _default_mdef(self):
         self.mdef = MassDef200m()
 
+    def _get_Delta_m(self, cosmo, a):
```
beckermr: Ditto here

pyccl/tests/test_halomodel.py
```diff
@@ -41,6 +42,35 @@ def test_halo_concentration(m):
...
+    mf, c = mf_c
+    cosmo = ccl.Cosmology(
+        Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
+        transfer_function='bbks', matter_power_spectrum='linear',
+        mass_function=mf, halo_concentration=c)
+    a = 0.8
+    k = np.geomspace(1E-2, 1, 10)
+    p = ccl.twohalo_matter_power(cosmo, k, a)
+
+    assert np.all(np.isfinite(p))
```
beckermr: Let's check the values against a separate comp using the new halo model classes.

pyccl/tests/test_halomodel.py
```diff
@@ -41,6 +42,35 @@ def test_halo_concentration(m):
...
+def test_halomodel_choices_raises():
+    cosmo = ccl.Cosmology(
+        Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
+        transfer_function='bbks', matter_power_spectrum='linear',
+        mass_function='tinker')
+    a = 0.8
+    k = np.geomspace(1E-2, 1, 10)
+
+    with pytest.raises(ValueError):
+        ccl.twohalo_matter_power(cosmo, k, a)
```
beckermr: Did this raise before? I want to make sure.

pyccl/tests/test_power.py
```diff
@@ -8,6 +8,23 @@
...
+    1,
+    1.0,
+    [0.3, 0.5, 10],
+    np.array([0.3, 0.5, 10])
+])
+def test_nonlin_power_halomod(k):
+    a = 0.8
+    pk = ccl.nonlin_matter_power(COSMO_HM, k, a)
+    assert np.all(np.isfinite(pk))
+    assert np.shape(pk) == np.shape(k)
```
beckermr: Again it'd be good to test against a computation using the new classes in the test.

src/ccl_f2d.c
```diff
@@ -8,6 +8,88 @@
 
 #include "ccl.h"
 
+ccl_f2d_t *ccl_f2d_t_copy(ccl_f2d_t *f2d_o, int *status)
```
beckermr: I know we don't have C tests so much, but this function seems like it needs one. Can we expose this interface in python and test it there maybe?

src/ccl_power.c
```diff
@@ -475,12 +472,18 @@ void ccl_cosmology_compute_linear_power(ccl_cosmology* cosmo, ccl_f2d_t *psp, in
     cosmo->computed_linear_power = true;
 }
 
+void ccl_cosmology_compute_nonlin_power_from_f2d(ccl_cosmology *cosmo,
```
beckermr: Do we actually need this function, or can we do the copy right in the function below?

pyccl/halos/hbias.py
damonge: This is very specific of the Tinker mass functions, but I guess it doesn't hurt to put it in the parent class if I document it. On it.

Yes, it handles vir (and we test for that)

pyccl/halos/hmfunc.py
damonge: as above

pyccl/halos/hmfunc.py
damonge: as above

pyccl/tests/test_halomodel.py
damonge: It did, but it did so at the C level and we didn't test for it.

src/ccl_f2d.c
damonge: Humm, how about exposing a python function that updates the non-linear power spectrum and then testing that the non-linear spectrum matches the input one? Exposing this function would be more useful in the future than an abstract function that copies python-level Pk2Ds, I think.

src/ccl_power.c
damonge: this is what I'd want to expose in python

pyccl/halos/hbias.py
damonge: Done

pyccl/halos/hmfunc.py

pyccl/halos/hmfunc.py

src/ccl_f2d.c
beckermr: Sure! As long as we test it somewhere.

pyccl/tests/test_power.py
damonge: Done

pyccl/tests/test_halomodel.py
damonge: Done

src/ccl_f2d.c
damonge: OK, but wait, we are already testing this when we check that the updated non-linear power spectrum matches the halo model implementation, right? I.e. what I just implemented here: https://github.com/LSSTDESC/CCL/blob/b9f85eb9fd7e478935edd6d3e03dadcea5618ca8/pyccl/tests/test_power.py#L40

damonge:
Thanks! Back to you @beckermr

src/ccl_f2d.c
beckermr: Did we check that all of the internals are copied correctly?

src/ccl_power.c
damonge: No, good point. Let me do that quickly.

src/ccl_f2d.c
damonge: No, good point. Let me do that now.

damonge: Done here: https://github.com/LSSTDESC/CCL/blob/e457cfdd2892d1ea7b670d98276a0a592487b573/pyccl/tests/test_power.py#L17

src/ccl_power.c
damonge: I'd rather not do this here, though. We'll be forced to do it when we try to get things directly from CAMB, and that seems like an isolated issue to me.

beckermr:
Ack changelog too.

damonge:
Awesome, thanks! I'm gonna ignore the 0.1% decrease in test coverage though...

beckermr:
Yup. I give very little about test coverage. If the code is over 80%, great.

PR #739, opened 2020-03-12
First step to input distance

chrgeorgiou:
Adding keyword arguments, conditional expressions and initializing a C function.

### DIFF
--- a/include/ccl_background.h
+++ b/include/ccl_background.h
@@ -331,14 +331,29 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
 /**
  * //TODO: Fill
  * @param cosmo Cosmological parameters
- * @param na integer indicating size of arrays a and chi_a
- * @param a scale factor at locations where the input chi_a has been pre-computed
+ * @param na integer indicating size of array a
+ * @param a scale factor at locations where the input arrays are pre-computed
  * @param chi_a comoving distance computed at values of a
+ * @param E_a Hubble parameter dividied by H_0 as a function of a
  * @param status Status flag. 0 if there are no errors, nonzero otherwise.
  * For specific cases see documentation for ccl_error.c
  * @return void
  */
 void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], double E_a[], int *status);
+
+/**
+ * //TODO: Fill
+ * @param cosmo Cosmological parameters
+ * @param na integer indicating size of array a
+ * @param a scale factor at locations where the input arrays are pre-computed
+ * @param growth_arr Growth factor array, defined as D(a)=P(k,a)/P(k,a=1), assuming no scale dependence.
+ * It is assumed that D(a<<1)~a so that D(1.0) will be used for normalization.
+ * @param fgrowth_arr Derivative of the growth array.
+ * @param status Status flag. 0 if there are no errors, nonzero otherwise.
+ * For specific cases see documentation for ccl_error.c
+ * @return void
+ */
+void ccl_cosmology_growth_from_input(ccl_cosmology* cosmo, int na, double a[], double growth_arr[], double fgrowth_arr[], int* status);
 
 /**
  * Compute the growth function and a spline to be stored
--- a/include/ccl_background.h
+++ b/include/ccl_background.h
@@ -329,6 +329,18 @@ double ccl_mu_MG(ccl_cosmology * cosmo, double a, int *status);
 void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
 
 /**
+ * //TODO: Fill
+ * @param cosmo Cosmological parameters
+ * @param na integer indicating size of arrays a and chi_a
+ * @param a scale factor at locations where the input chi_a has been pre-computed
+ * @param chi_a comoving distance computed at values of a
+ * @param status Status flag. 0 if there are no errors, nonzero otherwise.
+ * For specific cases see documentation for ccl_error.c
+ * @return void
+ */
+void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], int *status);
+
+/**
  * Compute the growth function and a spline to be stored
  * in the cosmology structure.
  * @param cosmo Cosmological parameters
--- a/include/ccl_background.h
+++ b/include/ccl_background.h
@@ -329,7 +329,7 @@ double ccl_mu_MG(ccl_cosmology * cosmo, double a, int *status);
 void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
 
 /**
- * //TODO: Fill
+ * Store user input arrays in splines for the comoving radial distance chi(a), hubble parameter E(a) as well as a(chi).
  * @param cosmo Cosmological parameters
  * @param na integer indicating size of array a
  * @param a scale factor at locations where the input arrays are pre-computed
@@ -342,7 +342,7 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
 void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], double E_a[], int *status);
 
 /**
- * //TODO: Fill
+ * Store user input arrays in splines for the growth factor, growth rate and normalization.
  * @param cosmo Cosmological parameters
  * @param na integer indicating size of array a
  * @param a scale factor at locations where the input arrays are pre-computed
--- a/include/ccl_background.h
+++ b/include/ccl_background.h
@@ -338,7 +338,7 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
  * For specific cases see documentation for ccl_error.c
  * @return void
  */
-void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], int *status);
+void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], double E_a[], int *status);
 
 /**
  * Compute the growth function and a spline to be stored
--- a/pyccl/ccl_background.i
+++ b/pyccl/ccl_background.i
@@ -134,7 +134,6 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 /* The directive gets carried between files, so we reset it at the end. */
 %feature("pythonprepend") %{ %}
 
-// FIXME: I cannot place this inside the other pythonprepend features.
 %inline %{
 
 void cosmology_distances_from_input(ccl_cosmology * cosmo,
--- a/pyccl/ccl_background.i
+++ b/pyccl/ccl_background.i
@@ -10,6 +10,8 @@
 %apply (double* IN_ARRAY1, int DIM1) {(double* a2, int na2)};
 %apply (double* IN_ARRAY1, int DIM1) {(double* chi, int nchi)};
 %apply (double* IN_ARRAY1, int DIM1) {(double* hoh0, int nhoh0)};
+%apply (double* IN_ARRAY1, int DIM1) {(double* growth, int ngrowth)};
+%apply (double* IN_ARRAY1, int DIM1) {(double* fgrowth, int nfgrowth)};
 %apply (int DIM1, double* ARGOUT_ARRAY1) {(int nout, double* output)};
 
 %include "../include/ccl_background.h"
@@ -138,6 +140,12 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 void cosmology_distances_from_input(ccl_cosmology * cosmo,
         double* a, int na, double* chi, int nchi, double* hoh0, int nhoh0, int *status) {
     ccl_cosmology_distances_from_input(cosmo, na, a, chi, hoh0, status);
+}
+
+void cosmology_growth_from_input(ccl_cosmology * cosmo,
+        double* a, int na, double* growth, int ngrowth, double* fgrowth, int nfgrowth,
+        int * status) {
+    ccl_cosmology_growth_from_input(cosmo, na, a, growth, fgrowth, status);
 }
 
 %}
--- a/pyccl/ccl_background.i
+++ b/pyccl/ccl_background.i
@@ -111,14 +111,12 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 
 %}
 
-
 %feature("pythonprepend") %{
     if numpy.shape(a1) != (nout,):
         raise CCLError("Input shape for `a1` must match `(nout,)`!")
     if numpy.shape(a2) != (nout,):
         raise CCLError("Input shape for `a2` must match `(nout,)`!")
 %}
-
 
 %inline %{
 
@@ -132,3 +130,13 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 
 /* The directive gets carried between files, so we reset it at the end. */
 %feature("pythonprepend") %{ %}
+
+// FIXME: I cannot place this inside the other pythonprepend features.
+%inline %{
+
+void cosmology_distances_from_input(ccl_cosmology * cosmo,
+        double* a, int na, double* chi, int nchi, int *status) {
+    ccl_cosmology_distances_from_input(cosmo, na, a, chi, status);
+}
+
+%}
--- a/pyccl/ccl_background.i
+++ b/pyccl/ccl_background.i
@@ -9,6 +9,7 @@
 %apply (double* IN_ARRAY1, int DIM1) {(double* a1, int na1)};
 %apply (double* IN_ARRAY1, int DIM1) {(double* a2, int na2)};
 %apply (double* IN_ARRAY1, int DIM1) {(double* chi, int nchi)};
+%apply (double* IN_ARRAY1, int DIM1) {(double* hoh0, int nhoh0)};
 %apply (int DIM1, double* ARGOUT_ARRAY1) {(int nout, double* output)};
 
 %include "../include/ccl_background.h"
@@ -135,8 +136,8 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 %inline %{
 
 void cosmology_distances_from_input(ccl_cosmology * cosmo,
-        double* a, int na, double* chi, int nchi, int *status) {
-    ccl_cosmology_distances_from_input(cosmo, na, a, chi, status);
+        double* a, int na, double* chi, int nchi, double* hoh0, int nhoh0, int *status) {
+    ccl_cosmology_distances_from_input(cosmo, na, a, chi, hoh0, status);
 }
 
 %}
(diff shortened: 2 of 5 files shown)


src/ccl_background.c
```diff
@@ -621,6 +621,122 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
+    gsl_spline_free(E); // Note: you are allowed to call gsl_free() on NULL
+    gsl_spline_free(chi);
+    gsl_spline_free(a);
+    E = NULL;
+    chi = NULL;
+    a = NULL;
+  }
+
+  /*
+  // FIXME: Is this necessary?
```
damonge: No! Do not modify any of the memory in the input arrays.

src/ccl_background.c
```diff
@@ -621,6 +621,122 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
+  if (!*status){
+    if (gsl_spline_init(chi, a, chi_a, na)){//in Mpc
+      *status = CCL_ERROR_SPLINE;
+      ccl_cosmology_set_status_message(
+        cosmo, "ccl_background.c: ccl_cosmology_distances_from_input(): Error creating  chi(a) spline\n");
+    }
+  }
+
+  // Reverse the order of chi(a) so that we can initialize the a(chi) spline, which needs monotonically decreasing x-array.
+  reverseArray(chi_a, 0, na-1);
```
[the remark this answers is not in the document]
damonge: Actually, now that I think about it, I see that you've gone through quite some trouble to reverse the arrays in place (i.e. avoid allocating new arrays). However, this is kind of dangerous: you're modifying the input arrays, which you may want to use for something else afterwards. So here, instead, what I would do is create new arrays (e.g. `chi_a_reverse` and `a_reverse`) and fill them up with the `chi_a` and `a` in reverse order. That way you don't touch the input arrays.

pyccl/core.py
```diff
@@ -148,6 +148,14 @@ class Cosmology(object):
             m_nu_type = 'equal', and 'equalize', which will redistribute
             masses to be equal right before calling the emualtor but results in
             internal inconsistencies. Defaults to 'strict'.
+        background_on_input (:obj:`bool`, optional): If True, the comoving distance
+            will be read as input from the chi_z array. #TODO Doc
+        chi_z (array_like, optional): Comoving distance computed at points indicated
+            by the z_array. If background_on_input is `False`, this is ignored. #TODO Doc
+        hoh0 (array_like, optional): Hubble parameter over the value of H0.
+            If background_on_input is `False`, this is ignored. #TODO Doc
+        z_array (array_like, optional): Redshift array with values on which \chi(z)
```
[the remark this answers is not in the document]
chrgeorgiou: Done!

pyccl/core.py
```diff
@@ -634,9 +649,19 @@ def compute_distances(self):
...
-        check(status, self)
+        if not self.background_on_input:
+            status = 0
+            status = lib.cosmology_compute_distances(self.cosmo, status)
+            check(status, self)
+        else:
+            # TODO: Check that chi_z and z_array are arrays of the same size.
+            # TODO: Check that z_array has the correct order.
+            status = 0
+            a = 1./(1+self.z_array)
```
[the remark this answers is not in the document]
chrgeorgiou: Yes indeed, I created z decreasing. I suppose it would be good to check if the input a_array is indeed monotonically increasing at some point in the code, perhaps here.

src/ccl_background.c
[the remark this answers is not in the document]
chrgeorgiou: True, I thought about that as well. I just noticed that the python arrays "self.chi_z" (in the core.py) where not actually reversed, and only the C-level arrays where changed, inside the function, so I thought it wasn't necessary to define new arrays (mostly to not take up too much memory). But I suppose the safe thing to do is to define new, reversed arrays.

src/ccl_background.c
```diff
@@ -621,6 +621,122 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
+      *status = CCL_ERROR_SPLINE;
+      ccl_cosmology_set_status_message(
+        cosmo, "ccl_background.c: ccl_cosmology_distances_from_input(): Error creating  chi(a) spline\n");
+    }
+  }
+
+  if (*status){ //If there was an error, free the GSL splines and return
+    gsl_spline_free(E); // Note: you are allowed to call gsl_free() on NULL
+    gsl_spline_free(chi);
+    gsl_spline_free(a);
```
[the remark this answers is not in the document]
chrgeorgiou: Yes, thanks!

elisachisari:
Looks like travis tests are failing due to flake8. That should be an easy fix. Please see here: https://flake8.pycqa.org/en/latest/user/index.html for flake8 docs.

src/ccl_background.c
```diff
@@ -621,6 +621,122 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
+  //Check for too little memory
+  if (a == NULL || E_a == NULL || chi_a == NULL || E == NULL || chi == NULL || achi == NULL) {
+    *status=CCL_ERROR_MEMORY;
+    ccl_cosmology_set_status_message(
+      cosmo, "ccl_background.c: ccl_cosmology_distances_from_input(): ran out of memory\n");
+  }
+
+  // FIXME: This is raised and I don't know why.
+  /*
+  //Check for messed up scale factor conditions
```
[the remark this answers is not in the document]
chrgeorgiou: Within `ccl_background.c` these two are used in the functions to compute distances and growth, to initialize the scale factor array at which points the various distances and functions will be computed. Since now these are going to be from input, perhaps it is only necessary to check for a few basic things (e.g. the scale factor is not larger than 1 or smaller than some pre-determined value)?

src/ccl_background.c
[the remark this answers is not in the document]
damonge: OK, thanks a lot. Let's do that for safety.

pyccl/core.py
```diff
@@ -161,7 +178,9 @@ def __init__(
             baryons_power_spectrum='nobaryons',
             mass_function='tinker10',
             halo_concentration='duffy2008',
-            emulator_neutrinos='strict'):
+            emulator_neutrinos='strict',
+            background_on_input=False, chi_a=None, hoh0=None, a_array=None,
```
damonge: Let's think of a homogeneous way of calling these variables. I'd suggest appending `_array` to all of them. Also, since the assumption is that all the arrays (`chi_a`, `hoh0`, `growth_array` and `fgrowth_array`) share the same `a_array`, I would put `a_array` before all the other ones in the function signature

pyccl/core.py
```diff
@@ -634,9 +661,29 @@ def compute_distances(self):
...
-        status = 0
-        status = lib.cosmology_compute_distances(self.cosmo, status)
-        check(status, self)
+        if not self.background_on_input:
+            status = 0
+            status = lib.cosmology_compute_distances(self.cosmo, status)
+            check(status, self)
+        else:
+            # Check that input arrays have the same size.
+            if not (self.a_array.shape == self.chi_a.shape == self.hoh0.shape):
```
damonge: We should create a few quick unit tests for all of these ifs (i.e. tests that check that the corresponding errors are raised). I can help do that, ping me on slack.

src/ccl_background.c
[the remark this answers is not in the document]
damonge: OK, I think this is fine, since you're checking for this at the python level.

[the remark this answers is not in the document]
damonge: Actually, related to this: ` cosmo->spline_params.A_SPLINE_MIN` is annoyingly used [here](https://github.com/LSSTDESC/CCL/blob/1f08a44733cafee5492d3d1914ecaf4ca7f09640/src/ccl_tracers.c#L458)

Could you update this variable within this function call to reflect the actual lower bound of the scale factor in the distance arrays?

src/ccl_background.c
```diff
@@ -621,6 +621,150 @@ void ccl_cosmology_compute_distances(ccl_cosmology * cosmo, int *status)
...
+  if (!*status){
+    if (gsl_spline_init(chi, a, chi_a, na)){//in Mpc
+      *status = CCL_ERROR_SPLINE;
+      ccl_cosmology_set_status_message(
+        cosmo, "ccl_background.c: ccl_cosmology_distances_from_input(): Error creating  chi(a) spline\n");
+    }
+  }
+
+  // Reverse the order of chi(a) so that we can initialize the a(chi) spline, which needs monotonically decreasing x-array.
+  double chi_a_reversed[na], a_reversed[na];
```
damonge: This may get you into trouble if `na` is too large, so unfortunately we should allocate these dynamically.

damonge: Also, this blocked should be wrapped inside an `if (!*status)` block.

chrgeorgiou: Would that be something like this? 

```suggestion
  chi_a_reversed = malloc(sizeof(double)*na);
  a_reversed = malloc(sizeof(double)*na);
```

Also, to update the A_SPLINE_MIN, you mean to update this value to the lowest value of the input scale factor by the user, (i.e. cosmo->spline_params.A_SPLINE_MIN = a[0])?

damonge: Yes, and yes! You'll have to check that malloc has worked (i.e. check that the arrays are not NULL), and then rememer to free them up afterwards.

include/ccl_background.h
```diff
@@ -328,6 +328,33 @@ double ccl_mu_MG(ccl_cosmology * cosmo, double a, int *status);
  */
 void ccl_cosmology_compute_distances(ccl_cosmology * cosmo,int *status);
 
+/**
+ * //TODO: Fill
```
damonge: Remove this?

include/ccl_background.h
```diff
@@ -328,6 +328,33 @@ double ccl_mu_MG(ccl_cosmology * cosmo, double a, int *status);
...
+ * @param chi_a comoving distance computed at values of a
+ * @param E_a Hubble parameter dividied by H_0 as a function of a
+ * @param status Status flag. 0 if there are no errors, nonzero otherwise.
+ * For specific cases see documentation for ccl_error.c
+ * @return void
+ */
+void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[], double chi_a[], double E_a[], int *status);
+
+/**
+ * //TODO: Fill
```
damonge: Remove this?

pyccl/ccl_background.i
```diff
@@ -132,3 +131,13 @@ void scale_factor_of_chi_vec(ccl_cosmology * cosmo, double* chi, int nchi,
 
 /* The directive gets carried between files, so we reset it at the end. */
 %feature("pythonprepend") %{ %}
+
+// FIXME: I cannot place this inside the other pythonprepend features.
```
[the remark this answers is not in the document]
damonge: This is fine actually. Just remove the comment.

pyccl/core.py
```diff
@@ -148,6 +148,24 @@ class Cosmology(object):
...
+            this is ignored. #TODO Doc
+        hoh0_array (array_like, optional): Hubble parameter divided by the
+            value of H0. If background_on_input is `False` this is ignored.
+             #TODO Doc
+        growth_array (array_like, optional): Growth factor array, defined as
+            D(a)=P(k,a)/P(k,a=1), assuming no scale dependence. It is assumed
+            that D(a<<1)~a so that D(1.0) will be used for normalization.
+            If background_on_input is `False` this is ignored. #TODO Doc
+        fgrowth_array (array_like, optional): Growth rate array.
+            If background_on_input is `False` this is ignored. #TODO Doc
```
damonge: I think the docstrings are fine. Remove TODOs

pyccl/core.py
```diff
@@ -182,6 +202,15 @@ def __init__(
 
         self._build_cosmo()
 
+        # TODO: include these in the self._params_init_kwargs?
```
damonge: Let's not do this yet.

pyccl/core.py
```diff
@@ -182,6 +202,15 @@ def __init__(
...
+        # User input arrays:
```
damonge: Might be a good place to check that if `background_on_input == True` then, all the input arrays are not `None` (and not `None`s).

pyccl/tests/test_background.py
```diff
@@ -111,3 +119,73 @@ def test_background_rho_x(a, kind, is_comoving):
...
+                          A_s=2e-9)
+    # Where the input quantities are calculated.
+    a_arr = np.linspace(0.1, 1, 100)
+    chi_from_ccl = ccl.background.comoving_radial_distance(cosmo, a_arr)
+    hoh0_from_ccl = ccl.background.h_over_h0(cosmo, a_arr)
+
+    cosmo_input = ccl.Cosmology(Omega_c=0.27, Omega_b=0.05, h=0.7, n_s=0.965,
+                                A_s=2e-9, background_on_input=True,
+                                a_array=a_arr, chi_array=chi_from_ccl,
+                                hoh0_array=hoh0_from_ccl)
```
damonge: this should fail if you don't pass growth arrays

pyccl/tests/test_background.py
```diff
@@ -111,3 +119,73 @@ def test_background_rho_x(a, kind, is_comoving):
...
+                          A_s=2e-9)
+    # Where the input quantities are calculated.
+    a_arr = np.linspace(0.1, 1, 100)
+    growth_from_ccl = ccl.background.growth_factor(cosmo, a_arr)
+    fgrowth_from_ccl = ccl.background.growth_rate(cosmo, a_arr)
+
+    cosmo_input = ccl.Cosmology(Omega_c=0.27, Omega_b=0.05, h=0.7, n_s=0.965,
+                                A_s=2e-9, background_on_input=True,
+                                a_array=a_arr, growth_array=growth_from_ccl,
+                                fgrowth_array=fgrowth_from_ccl)
```
damonge: As above for the chi and H/H0 arrays.

damonge:
For context, [here](https://github.com/damonge/md_share/blob/master/extending_CCL.md)'s what we're trying to do

beckermr:
I am not happy with the proliferation of input arguments here as we discussed but I am used to being shouted down.

damonge:
Thanks a lot @beckermr . I was thinking about this. The problem is that, if we want to create a cosmology with input distances, we still need the cosmological parameters as well (we need them for e.g. lensing kernels, power spectra etc.).

So, in order to avoid having too many parameters in the constructor, how about this:
1. We remove those parameters from the constructor.
2. We create a new method called `set_background_from_arrays` that passes these arrays. If the background for this cosmology has already been initialized, then this call will raise an error.
3. If this method has been called, then the cosmology background quantities will be initialized from these arrays (otherwise, we compute our own distances/growth).

What do you think?

beckermr:
That works for me. What other quantities do we plan on inputting besides the background? We should make sure to name the methods in such a way as they make sense down the road.

damonge:
OK, great! @chrgeorgiou , could you implement something along these lines?

@beckermr this is the plan.
1. distances, E(z), linear growth factor and linear growth rate.
2. linear matter power spectrum
3. Non-linear matter power spectrum
4. Eventually extend things so we actually read in **a set of** core power spectra (corresponding to all the auto- and cross-correlations of the matter overdensity, bardeen potential, and possibly CDM+baryons -- I may be missing some others here.)

Clearly the 4th one would change a lot of the internal parts of CCL, and should be done for v3 but I think it's vital if we ever want to do useful things with non-LCDM theories.

beckermr:
I see. I am ok with `set_XYZ` functions for all of this. Others should chime in.

beckermr:
We need to think about how ppl would interpret the matter power spectrum parameters if we are inputting other stuff.

chrgeorgiou:
>     2. We create a new method called `set_background_from_arrays` that passes these arrays. If the background for this cosmology has already been initialized, then this call will raise an error.

Can you explain what you mean by "cosmology has been initialized"? I was under the impression that, right now, one cannot call a pyccl.Cosmology() object without specifying certain parameters (e.g. sigma8 or As, https://github.com/LSSTDESC/CCL/blob/1f08a44733cafee5492d3d1914ecaf4ca7f09640/pyccl/core.py#L353). So, would the idea be to create a cosmo = pyccl.Cosmology() object with the desired cosmological parameters but then also call a function cosmo.set_background_from_arrays() that would store the splines we want?

damonge:
Right, what I mean is that `set_background` would start with:
```
if self.has_distances or self.has_growth:
    raise ValueError("Background has already been initialized and cannot be reset")
```
or something like that.

> So, would the idea be to create a cosmo = pyccl.Cosmology() object with the desired cosmological parameters but then also call a function cosmo.set_background_from_arrays() that would store the splines we want?

Yes, that's right. It think it should be just a matter of you moving things around a bit a the python level.

beckermr:
I know we discussed something along the lines of adding a new object that encapsulates the cosmology computations in addition to the cosmology. Where did we land on that @damonge?

damonge:
Yeah, I'm not sure how to do that in a staged way (i.e. without working on a 1000-line PR). I think I'd rather wait on this until at least stages 1, 2 and 3 are there.

beckermr:
I would think the following would work. 

1. PR to reconfigure the internals of the current Cosmology object to separate the steps of making the data to be interpolated and inputting that data into the proper spots in the C layer.
2. PR to separate the code from 1 into two objects w/ defaulting 

There are some interface issues to be worked out around parameters etc. Probs best to make an issue.

damonge:
Sorry for the delay @beckermr . I'm afraid we may not be talking about the same thing here, but I think at least point 1 of what you said is essentially what @chrgeorgiou has done for distances and growth.

Now as for the second object you mention, I'm not sure what you were thinking that would be. Is it a subclass of Cosmology (or even a superclass) that takes arrays as inputs rather than cosmological parameters?

damonge:
@beckermr what do you think about the above?

beckermr:
Right now we are adding a new public API that we would eventually want to deprecate. My step 1 above is an internal refactor that would be there to then support the addition of a new public API in terms of two classes later.

damonge:
I'm not sure I understand. Isn't that what is currently done by `compute_distances` [here](https://github.com/chrgeorgiou/CCL/blob/7214e78498e67d3816a6b63d98db078fec89993e/pyccl/core.py#L637) at the python level, or by `ccl_compute_distances` [here](https://github.com/chrgeorgiou/CCL/blob/7214e78498e67d3816a6b63d98db078fec89993e/src/ccl_background.c#L444) and `ccl_compute_distances_from_input` [here](https://github.com/chrgeorgiou/CCL/blob/7214e78498e67d3816a6b63d98db078fec89993e/src/ccl_background.c#L628) at the C level?

beckermr:
Yes, but there is a new public API on the Cosmology object. This API would very likely not be present in a later version when we moved to two things.

damonge:
would you rather this not be a public method for now?

beckermr:
IMHO yes. I'd prefer an internal refactor to separate the computation of the spline data and the splines themselves. Then once these things are separate, we can then decide on a new set of public APIs that are backwards compatible and also support cobaya etc.

damonge:
OK, sounds good. @chrgeorgiou, can you change `set_background_from_arrays` into `_set_background_from_arrays`?

chrgeorgiou:
Done, no problem!

damonge:
Ack, sorry! And just to be consistent, can we do the same for `background_on_input` ?

chrgeorgiou:
Done! @damonge @beckermr

damonge:
@beckermr I've reviewed this and would like to merge it. Any thoughts?

damonge:
Awesome, thanks!

Issue #354, opened 2018-03-23
malloc error in ClTracerLensing when setting A_SPLINE_MAX < 1


tilmantroester:
When creating a ClTracerLensing object with A_SPLINE_MAX set to less than 1, Python gives a malloc error. 
```
python(26540,0x7fffb026f3c0) malloc: *** mach_vm_map(size=18446744056529686528) failed (error code=3)
*** error: can't allocate region
*** set a breakpoint in malloc_error_break to debug
ERROR: Could not allocate memory for linear-spaced array (N=-2147483647)
```
The exception is just a type error
```
TypeErrorTraceback (most recent call last)
<ipython-input-2-eabf94f2e53d> in <module>()
     18 
     19 lens1 = ccl.ClTracerLensing(ccl_cosmology, has_intrinsic_alignment=False,
---> 20                             n=[z_nofz, nofz],)

/Users/yooken/miniconda/envs/python3/lib/python3.5/site-packages/pyccl/cls.py in __init__(self, cosmo, has_intrinsic_alignment, n, z, bias_ia, f_red)
    193                  has_intrinsic_alignment=has_intrinsic_alignment,
    194                  z=z, n=n, bias=None, mag_bias=None,
--> 195                  bias_ia=bias_ia, f_red=f_red)
    196 
    197 

/Users/yooken/miniconda/envs/python3/lib/python3.5/site-packages/pyccl/cls.py in __init__(self, cosmo, tracer_type, has_rsd, has_magnification, has_intrinsic_alignment, z, n, bias, mag_bias, bias_ia, f_red, z_source)
     98                             self.z_rf, self.rf,
     99                             self.z_source,
--> 100                             status )
    101 
    102     def __del__(self):

TypeError: 'int' object is not iterable
```

Here is  the code to cause this error, with A_SPLINE_MAX=0.98:
```
import os
import numpy as np
import pyccl as ccl

os.environ["CCL_PARAM_FILE"] = "./ccl_params_modified.ini"

ccl_cosmology = ccl.Cosmology(Omega_c=0.3,
                              Omega_b=0.05,
                              Omega_k=0.0,
                              h=0.7,
                              n_s=0.96,
                              sigma8=0.8)

z_nofz = np.linspace(0.1, 4, 1000)
nofz = np.exp(-0.5*(z_nofz-1.0)**2/0.2**2)

lens1 = ccl.ClTracerLensing(ccl_cosmology, has_intrinsic_alignment=False,
[...]

PR #748, opened 2020-03-30
Force A_SPLINE_MAX to be 1.0.

tilmantroester:
Currently the `A_SPLINE_MAX` spline parameter can in principle be set to something `<= 1.0`. This however breaks things (#354). This PR forces `A_SPLINE_MAX` to 1.0 by throwing an error if set to anything else. 

It would probably be good to check the other spline parameters has well at some point.

### DIFF
--- a/pyccl/tests/test_spline_params.py
+++ b/pyccl/tests/test_spline_params.py
@@ -1,0 +1,17 @@
+import pytest
+import pyccl as ccl
+
+def test_spline_params():
+    cosmo = ccl.Cosmology(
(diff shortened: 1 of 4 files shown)


tilmantroester:
It just occurred to me that I can define A_SPLINE_MAX as const. Currently this gives a cryptic error message when trying to set it in python. I'll see if I can overwrite the setter to catch this. This would be a much cleaner solution I think.

tilmantroester:
Closed in favour of #751 .

PR #751, opened 2020-04-02
Make A_SPLINE_MAX functionally a constant

tilmantroester:
This is a cleaner implementation of PR #748 . On the C-level, it makes `A_SPLINE_MAX` a const and checks its value in `ccl_cosmology_create`. On the Python-level it overwrites the setter to throw a `RuntimeError` with an informative error message.

I think the only way to bypass these checks is to create a new `ccl_spline_params` struct with  A_SPLINE_MAX != 1 and replace the `spline_params` in an existing `ccl_cosmology` struct. If a user goes that far, they can probably break other things just as easily.

### DIFF
--- a/pyccl/ccl_core.i
+++ b/pyccl/ccl_core.i
@@ -4,7 +4,8 @@
 /* put additional #include here */
 %}
 
-// SWIG black magic
+// SWIG black magic. Change the behaviour of setting A_SPLINE_MAX to thowing an
+// error.
 %typemap(memberin) double A_SPLINE_MAX {
     if($input) {
         PyErr_SetString(PyExc_RuntimeError, "A_SPLINE_MAX is fixed to 1.0 and is not mutable.");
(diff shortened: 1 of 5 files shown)


pyccl/ccl_core.i
```diff
@@ -4,6 +4,18 @@
 /* put additional #include here */
 %}
 
+// SWIG black magic
```
damonge, line 7: OK, I'm not familiar with this at all. Can you explain or point me to a reference?

tilmantroester: Well. There's http://www.swig.org/Doc3.0/SWIGDocumentation.html#Typemaps but that's not particularly useful. My very limited understanding is that typemaps hook into the C<-> python interface to deal with type conversions etc. The `memberin` typemap hooks into actions with structure members. In this case it thows an exception when there's and input.

I tried all afternoon to get things like %extend or %immutable to work, or overriding the setter functions but none of that seemed to do anything. I was sceptical about SWIG before but now I feel like converting everything to something else like pybind might not be the worst idea.

damonge:
OK, things seem to be failing on Travis. If you don't want to spend too much time on this, I'm happy for you to resurrect the previous PR and just wait on this for v3 (at which point we could make A_SPLINE_MAX a C-level macro and retire it from the API).

c-d-leonard:
It does look like we are plassing on travis now - @tilmantroester @damonge ? Can this be merged?

PR #753, opened 2020-04-08
Linear power spectrum from input

chrgeorgiou:
Extension of the Cosmology class adding the option for the user to parse his own linear matter power spectrum, which CCL then uses to create the splines for the power spectrum.


pyccl/core.py
```diff
@@ -762,6 +774,29 @@ def compute_linear_power(self):
         status = lib.cosmology_compute_linear_power(self.cosmo, psp, status)
         check(status, self)
 
+    def _compute_linear_power_from_arrays(self):
+        if not self._linear_power_on_input:
+            raise ValueError("Cannot compute linear power spectrum from"
+                             "input without input arrays initialized.")
+        from .pk2d import Pk2D  # FIXME: Is it okay to call this here?
```
damonge, line 781: I think it'd be best to put this at the beginning of the file

pyccl/core.py
```diff
@@ -953,3 +988,42 @@ def _set_background_from_arrays(self, a_array=None, chi_array=None,
                     or (hoh0_array is None) or (growth_array is None)
                     or (fgrowth_array is None)):
                 raise ValueError("Input arrays not parsed.")
+
+    def _set_linear_power_from_arrays(self, a_array=None, k_array=None,
+                                      pk_array=None):
+        """
+        # TODO: Docstring.
```
damonge, line 995: Can you write this while we're at it?

pyccl/tests/test_background.py
```diff
@@ -192,3 +192,34 @@ def test_input_arrays_raises():
         cosmo_input = ccl.Cosmology(Omega_c=0.27, Omega_b=0.05, h=0.7,
                                     n_s=0.965, A_s=2e-9)
         cosmo_input._set_background_from_arrays()
+
+
+def test_input_lin_power_spectrum():
```
damonge, line 197: Can you move this test to `test_power.py`?

src/ccl_power.c
```diff
@@ -459,9 +459,12 @@ void ccl_cosmology_compute_linear_power(ccl_cosmology* cosmo, ccl_f2d_t *psp, in
         break;
 
       case ccl_boltzmann_isitgr:
-	rescaled_mg_flag = 0; 
+      	rescaled_mg_flag = 0;
```
damonge, line 462: Ah, my bad. Can you remove this extra space?

pyccl/core.py
```diff
@@ -762,6 +780,28 @@ def compute_linear_power(self):
...
+                      a_arr=self.a_array,
+                      lk_arr=np.log(self.k_array),
+                      pk_arr=self.pk_array,
+                      is_logp=False,
+                      extrap_order_lok=1,
+                      extrap_order_hik=2,
+                      cosmo=None)
+
+        # needed to init some models
+        self.compute_growth()
```
damonge, line 797: Do we really need this with the new scheme?

pyccl/tests/test_background.py
```diff
@@ -192,3 +192,21 @@ def test_input_arrays_raises():
         cosmo_input = ccl.Cosmology(Omega_c=0.27, Omega_b=0.05, h=0.7,
                                     n_s=0.965, A_s=2e-9)
         cosmo_input._set_background_from_arrays()
+
+
+def test_input_linpower_raises():
```
damonge, line 197: Can we also move this to `test_power.py`?

pyccl/core.py
chrgeorgiou: You're right, I think it's no longer necessary, I will remove it.

pyccl/core.py
```diff
@@ -19,7 +20,8 @@
     'bbks':             lib.bbks,
     'boltzmann_class':  lib.boltzmann_class,
     'boltzmann_camb':   lib.boltzmann_camb,
-    'boltzmann_isitgr':   lib.boltzmann_isitgr,
+    'pklin_from_input': lib.pklin_from_input,
```
beckermr, line 23: This is part of the public API.

chrgeorgiou: Ah, what are the necessary changes so this is not the case?

beckermr: nothing new appears in the python api that we wouldn't take out later.

damonge: OK @chrgeorgiou let's do the following.
1. Remove this line.
2. (more instructions below)

pyccl/core.py
```diff
@@ -953,3 +990,45 @@ def _set_background_from_arrays(self, a_array=None, chi_array=None,
...
+            sampled (i.e. the resolution of `a_array` and `k_array` is high
+            enough to sample the main features in the power spectrum).
+            For reference, CCL will use bicubic interpolation to evaluate
+            the power spectrum at any intermediate point in k and a.
+        """
+        if self.has_linear_power:
+            raise ValueError("Linear power spectrum has been initialized"
+                             "and cannot be reset.")
+        else:
+            if (self._config_init_kwargs['transfer_function']
```
damonge, line 1022: 2. Insert two lines just above this one that read:
```
self.cosmo.config.transfer_function_method = lib.pklin_from_input
self._config_init_kwargs['transfer_function'] = 'pklin_from_input'
```

pyccl/core.py
```diff
@@ -953,3 +984,41 @@ def _set_background_from_arrays(self, a_array=None, chi_array=None,
...
+            For reference, CCL will use bicubic interpolation to evaluate
+            the power spectrum at any intermediate point in k and a.
+        """
+        if self.has_linear_power:
+            raise ValueError("Linear power spectrum has been initialized"
+                             "and cannot be reset.")
+        else:
+            if ((a_array is None) or (k_array is None)
+                    or (pk_array is None)):
+                raise ValueError("Input arrays not parsed.")
```
c-d-leonard, line 1018: This error messages strikes me as potentially being a bit unclear to some users, can we have something slightly more descriptive like "One or more of the input array for a, k, or pk is not parsed"?

pyccl/tests/test_power.py
```diff
@@ -170,3 +170,46 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+    cosmo_input = ccl.Cosmology(Omega_c=0.27, Omega_b=0.05, h=0.7, n_s=0.965,
+                                A_s=2e-9)
+    cosmo_input._set_background_from_arrays(a_array=a_arr,
+                                            chi_array=chi_from_ccl,
+                                            hoh0_array=hoh0_from_ccl,
+                                            growth_array=growth_from_ccl,
+                                            fgrowth_array=fgrowth_from_ccl)
+    cosmo_input._set_linear_power_from_arrays(a_arr, k_arr, pk_arr)
+    cosmo_input.compute_linear_power()
+    pk_CCL_input = ccl.power.linear_matter_power(cosmo_input, k_arr, 0.5)
```
c-d-leonard, line 198: I think I see what is happening here, but having to call cosmo_input.compute_linear_power() and then also call ccl.power.linear_matter_power() seems like a clunky thing to need to do which will be confusing for some users ... do we envision this as the standard way to do this is this something which is temporary while we are keeping this under the hood?

include/ccl_config.h
```diff
@@ -20,7 +20,8 @@ typedef enum transfer_function_t
   ccl_bbks              = 2,
   ccl_boltzmann_class   = 3,
   ccl_boltzmann_camb    = 4,
-  ccl_boltzmann_isitgr  = 5,
+  ccl_pklin_from_input  = 5,
+  ccl_boltzmann_isitgr  = 6,
```
c-d-leonard, line 24: This isn't a big deal but is there a reason not to just add ccl_pklin_from_input on as number 6?

pyccl/core.py
chrgeorgiou: Yes, thanks for pointing that out!

pyccl/tests/test_power.py
chrgeorgiou: It is not necessary to call "compute_linear_power()" for this to work, I think it was left there from my debugging. I have now removed it.

include/ccl_config.h
chrgeorgiou: I think when I resolved a merge conflict I had already the pklin_from_input set to 5. I have now changed it to 6.

PR #760, opened 2020-04-24
k_nl integral

billwright93:
Implementation of the below integral which computes the scale for a non-linear cut. Currently, the basic code changes have been made, but unit tests and benchmarks are missing, and not all documentation has been updated.
image


billwright93:
As far as I'm aware the only remaining issues are: 1) not being able to re-build the 0000-ccl_note in my local environment (issue with mkauthlist); 2) a similar issue with building readthedocs; 3) flake8 really hating ccllib.py.

billwright93:
Is it necessary to describe what the new benchmark test consists of? If so, is the wiki the best place for this?

pyccl/power.py
```diff
@@ -108,3 +108,17 @@ def sigma8(cosmo):
...
+
+def kNL(cosmo, a):
+    """Scale for the non-linear cut.
+
+    Args:
+        cosmo (:class:`~pyccl.core.Cosmology`): Cosmological parameters.
+        a (float or array_like): Scale factor(s), normalized to 1 today.
+
+    Returns:
+        float or array-like: Scale of non-linear cut-off; Mpc^-1.
```
slosar, line 121:
```suggestion
        float or array-like: Scale of non-linear cut-off; Mpc^-1.
    Note:
        k_NL is calculated based on Lagrangian perturbation theory as the 
        inverse of variance of the displacement field, i.e 
           k_NL = 1/Sigma = [1/(6 pi^2)  int P_L(k) dk]^{-1/2}     
```

src/ccl_power.c
```diff
@@ -768,3 +774,60 @@ smoothed with a tophat filter of comoving size 8 Mpc/h
...
+                                          0.0, cosmo->gsl_params.INTEGRATION_KNL_EPSREL,
+                                          workspace,&k_NL,NULL,NULL);
+    if(gslstatus != GSL_SUCCESS) {
+      ccl_raise_gsl_warning(gslstatus, "ccl_power.c: ccl_kNL():");
+      *status |= gslstatus;
+    }
+  }
+  gsl_integration_cquad_workspace_free(workspace);
+
+  return pow(sqrt(k_NL/(6*M_PI*M_PI))*ccl_growth_factor(cosmo, a, status), -1); //remove M_LN10 -- what is it? //is ccl_growth_factor there to set the power spectrum to the correct redshift? Why not do this in the integrand?
```
slosar, line 832: Let's split this into two lines for readbility. Something like

```
double sigma2 = 1/(6...) * ...
return pow (sigma2,-0.5);
```

src/ccl_power.c
```diff
@@ -768,3 +774,60 @@ smoothed with a tophat filter of comoving size 8 Mpc/h
...
+
+  KNL_pars par;
+  par.status = status;
+
+  par.cosmo=cosmo;
+  gsl_integration_cquad_workspace *workspace =  NULL;
+  gsl_function F;
+  F.function=&kNL_integrand;
+  F.params=&par;
+  double k_NL;
```
slosar, line 815: call this variable PL_integral (or something else clear) for readability, because it only becomes k_NL at the end.

billwright93: Good point, will change.

src/ccl_power.c
billwright93: Yep this would be better, will change.

pyccl/power.py
billwright93: Just to check -- is this "Note:" type tag compatible with ReadTheDocs, as I can't see it being used anywhere else in the CCL/pyccl directory? What I see intead is ".. note::" being used. Can you advise @slosar ?

billwright93:
@slosar , to answer your question "Where is EPSREL_KNL defined?", it is defined in src/ccl_core.c -- see line 53 (#define GSL_EPSREL_KNL 1E-5). The value of 10^-5 doesn't seem to be too stringent -- it allows the benchmark to pass on my machine at least. Let me know if you want any of this changed.

pyccl/power.py
billwright93: I've implemented the above with the ".. note::" syntax.

billwright93:
@beckermr The two instances of the flake8 issue which I think are causing the Travis-CI checks to fail are:
pyccl/correlation.py:141:43: E741 ambiguous variable name 'l'
pyccl/tests/test_correlation.py:102:50: E741 ambiguous variable name 'l'

I haven't touched either of these files in this PR, so I guess this would be a problem on the master branch too.

damonge:
I fixed that here: https://github.com/LSSTDESC/CCL/pull/783
However, that required changing quite a lot of the python files, since flake8 has become more of a brat lately. If people are happy with that PR (@slosar?) we could merge that, and this should be fixed if you then merge master into this PR.

damonge:
@billwright93 if you merge master into this branch the flake8 problems will go away!

billwright93:
> 
> 
> @billwright93 if you merge master into this branch the flake8 problems will go away!

Great thanks, will give it a go!

PR #761, opened 2020-04-24
Explicitly import isitgr

elisachisari:
Make sure isitgr is imported every time a new test is run. If this is not done, then some of the tests will fail. Specifically, "test_power_mg2.py". It is likely that this is coming from memory/object handling issues within isitgr which should be fixed separately.

### DIFF
--- a/benchmarks/test_correlation_MG.py
+++ b/benchmarks/test_correlation_MG.py
@@ -4,6 +4,7 @@ import pyccl as ccl
 from scipy.interpolate import interp1d
 import pytest
 
+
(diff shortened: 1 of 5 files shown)


mishakb:
@elisachisari should we just not here that this was fixed by adding an import isitgr at the top of boltzmann.py and it is not due to memory/object handling issues within isitgr.  :)

elisachisari:
We further checked this in the `debug_benchmark` branch and continue to find a failure of the tests related to `test_power_mg2.py`. @mishakb has kindly opened an issue in isitgr to fix this, though we have a workaround for the time being.

PR #763, opened 2020-04-27
added ptc output option

jablazek:
This is a minor feature addition to allow `get_pt_pk2d()` to return the `ptc` it initialized.

@damonge : do you remember why this was its own function which takes a `ptc` rather than a method of the `ptc` itself? Was it so that the user never needs to worry about the `ptc` if they don't want to?


jablazek:
@damonge : any thoughts on this?

pyccl/nl_pt/power.py
```diff
@@ -380,7 +380,8 @@ def get_pmm(self, Pd1d1_lin, g4):
 def get_pt_pk2d(cosmo, tracer1, tracer2=None, ptc=None,
                 sub_lowk=False, nonlin_pk_type='nonlinear',
                 a_arr=None, extrap_order_lok=1, extrap_order_hik=2,
-                return_ia_bb=False, return_ia_ee_and_bb=False):
+                return_ia_bb=False, return_ia_ee_and_bb=False,
+                return_ptc=False):
```
damonge, line 384: This new parameter needs to be documented

pyccl/nl_pt/power.py
```diff
@@ -580,10 +581,16 @@ def get_pt_pk2d(cosmo, tracer1, tracer2=None, ptc=None,
                         lk_arr=np.log(ptc.ks),
                         pk_arr=p_pt[1].T,
                         is_logp=False)
-        return pt_pk_ee, pt_pk_bb
+        if return_ptc:
```
damonge, line 584: Could we add a quick unit test for this

pyccl/nl_pt/power.py
```diff
@@ -580,10 +581,16 @@ def get_pt_pk2d(cosmo, tracer1, tracer2=None, ptc=None,
...
+            return pt_pk_ee, pt_pk_bb, ptc
+        else:
+            return pt_pk_ee, pt_pk_bb
     else:
         pt_pk = Pk2D(a_arr=a_arr,
                      lk_arr=np.log(ptc.ks),
                      pk_arr=p_pt.T,
                      is_logp=False)
-        return pt_pk
+        if return_ptc:
```
damonge, line 593: ... and this?

c-d-leonard:
@jablazek any update on this?

jablazek:
@c-d-leonard, not yet. but thanks for the poke.

pyccl/nl_pt/power.py
jablazek: Should these go in `pyccl/tests/`? or in `benchmarks/`?
@damonge @c-d-leonard

jablazek:
@damonge @c-d-leonard : this should be ready for the next round of review. I think we are waiting for the swap to github action to test the build and coverage?

pyccl/nl_pt/power.py
damonge: `pyccl/tests/`

damonge:
yes, but this shouldn't affect how we write the tests

damonge:
@jablazek I meant to say, if you can check that the new tests run, then I'd say we can still merge even if travis doesn't work.

pyccl/tests/test_pt_power.py
```diff
@@ -233,3 +234,26 @@ def test_translate_IA_norm_raises():
     with pytest.raises(ValueError):
         c_1, c_d, c_2 = ccl.nl_pt.translate_IA_norm(COSMO, ZZ2, a1=a_1_v,
                                                     Om_m2_for_c2=False)
+                                                    
+def test_return_ptc():
```
damonge, line 238: If this passes and flake8 doesn't complain, we can still merge this even if travis complains

damonge:
@jablazek it seems there are still flake8 issues

jablazek:
thanks. should be fixed.

jablazek:
@damonge , we good?

damonge (approved):
Looks good!

PR #765, opened 2020-04-27
Remove warnings for v2.1

rmjarvis:
While helping Elisa and Mustapha try to find their problems running the test suite on master, I noticed that you currently have a lot of warnings show up both when compiling and when running the tests.  IMO, it's a good idea to try to remove all these warnings.  So I went ahead and did so for you.

At least 3 of the compiler warnings were actual bugs (related to runtime failures, so didn't affect much in normal operations).  

The python warnings were mostly CCL deprecations, so for these I added a test using assert_warns to check that the deprecation warning was raised, and to catch it from being output to the terminal.

Finally, there were a couple other warnings related to numpy and collections that aren't a problem yet, but might become a problem in the future.  So worth fixing now before they become an error.

### DIFF
--- a/benchmarks/test_power_nu.py
+++ b/benchmarks/test_power_nu.py
@@ -31,7 +31,6 @@ def test_power_nu(model):
 
     a = 1
 
-
     data_lin = np.loadtxt("./benchmarks/data/model%d_pk_nu.txt" % (model+1))
     k_lin = data_lin[:, 0] * cosmo['h']
     pk_lin = data_lin[:, 1] / (cosmo['h']**3)
--- a/benchmarks/test_power_nu.py
+++ b/benchmarks/test_power_nu.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pytest
+import warnings
 
 import pyccl as ccl
 
@@ -30,10 +31,19 @@ def test_power_nu(model):
 
     a = 1
 
+
     data_lin = np.loadtxt("./benchmarks/data/model%d_pk_nu.txt" % (model+1))
     k_lin = data_lin[:, 0] * cosmo['h']
     pk_lin = data_lin[:, 1] / (cosmo['h']**3)
-    pk_lin_ccl = ccl.linear_matter_power(cosmo, k_lin, a)
+
+    with warnings.catch_warnings():
+        # Linear power with massive neutrinos raises a warning.
+        # Ignore it.
+        # XXX: Do you really want to be raising a warning for this?
+        #      This seems spurious to me.  (MJ)
+        warnings.simplefilter("ignore")
+        pk_lin_ccl = ccl.linear_matter_power(cosmo, k_lin, a)
+
     err = np.abs(pk_lin_ccl/pk_lin - 1)
     assert np.allclose(err, 0, rtol=0, atol=POWER_NU_TOL)
 
--- a/pyccl/tests/test_cls.py
+++ b/pyccl/tests/test_cls.py
@@ -3,6 +3,8 @@ import pytest
 
 import pyccl as ccl
 
+from numpy.testing import assert_raises
+
 COSMO = ccl.Cosmology(
     Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
     transfer_function='bbks', matter_power_spectrum='linear')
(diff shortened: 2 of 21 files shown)


rmjarvis:
Also, there were a few places where the deprecated syntax was still being used in the tests, where it didn't seem like it was just to check that the old version still worked (e.g. some old == new kinds of tests).  These should really be converted to the new syntax.

I did so in a couple of cases where it was obvious to me what to do, but I left TODOs where I wasn't as clear what the new syntax was.  In particular, it looked like `halomodel_matter_power`, `onehalo_matter_power` and `twohalo_matter_power` would require some significant boilerplate to convert to the new API.  Maybe I'm wrong, and it's actually really simple.  But if so, it might imply that these should be retained as convenience functions just with documentation about what they are equivalent to.

c-d-leonard:
@rmjarvis Thanks a lot for this! I just checked out this branch and ran the tests (`pytest -vv benchmarks`) and I still see quite a few deprecation warnings, am I missing something?

@damonge the halo model upgrade is your baby, do you want to fix this syntax now or leave it (given we still support the old API at this point)?

rmjarvis:
@c-d-leonard Thanks.  I'd meant to do benchmarks too, but forgot.  Working on that now.  Almost done.

damonge:
Re benchmarks and deprecated halo model code: it was recommended by my reviewer back then that we left the benchmarks using the old code because that way we could test that the code still worked as intended. I get no joy from arguing about these things, so that's why that code is still there. I'm happy for it to go.

Happy to review this when you're done!

rmjarvis:
The best way to deal with deprecations, imo, is to have a new test that checks the equivalence between old and new.  There are a few of these around, which I think are good.
And then change all other tests to use the new API.

rmjarvis:
I took the liberty of changing the benchmark code to use the new API.  But there are still places in the regular tests that I think might want some attention.

rmjarvis:
@c-d-leonard @damonge I think I'm done fiddling now.  Benchmarks are all green now too, and passing flake8 tests.  (Still waiting on the OSX run on travis, but I don't expect a problem there.)  So, ready for a review.

benchmarks/test_power_nu.py
```diff
@@ -33,7 +34,15 @@ def test_power_nu(model):
     data_lin = np.loadtxt("./benchmarks/data/model%d_pk_nu.txt" % (model+1))
     k_lin = data_lin[:, 0] * cosmo['h']
     pk_lin = data_lin[:, 1] / (cosmo['h']**3)
-    pk_lin_ccl = ccl.linear_matter_power(cosmo, k_lin, a)
+
+    with warnings.catch_warnings():
+        # Linear power with massive neutrinos raises a warning.
+        # Ignore it.
+        # XXX: Do you really want to be raising a warning for this?
+        #      This seems spurious to me.  (MJ)
```
damonge, line 42: I tend to agree with these (and have done so for a while), so I'll open an issue.

pyccl/tests/test_cls.py
```diff
@@ -46,6 +48,24 @@ def test_cls_smoke(p_of_k_a):
                     COSMO, tracers[j], tracers[i], ell, p_of_k_a=p_of_k_a)
                 assert np.allclose(corr, corr_rev)
 
+    # Check invalid dndz
```
damonge, line 51: Great

pyccl/tracers.py
```diff
@@ -404,7 +393,10 @@ def add_tracer(self, cosmo, kernel=None,
         self._trc.append(_check_returned_tracer(ret))
 
     def __del__(self):
-        if hasattr(self, '_trc'):
+        # Sometimes lib is freed before some Tracers, in which case, this
+        # doesn't work.
+        # So just check that lib.cl_tracer_t_free is still a real function.
+        if hasattr(self, '_trc') and lib.cl_tracer_t_free is not None:
```
damonge, line 399: Nice, I got bitten in the ass by this one before and couldn't understand what was happening. I'll put this in namaster too.

benchmarks/test_power_nu.py
c-d-leonard: Yeah, we do probably overwarn on this, we probably need to go through and reconsider how we handle this. Thanks for raising the issue.

rmjarvis:
> I still get one warning when running the tests and also the benchmarks - about imp being deprecated in favour of importlib. Did you intentionally leave this one in?

I don't see the warning.  Nor do I see any use of `imp` in CCL.  Maybe this is coming from one of the dependencies on your system?

c-d-leonard:
Okay, you're probably right, forget I said anything.

Thanks a lot for your work on this @rmjarvis, definitely a big improvement. I'm going to merge.

PR #767, opened 2020-04-28
Add Jarvis to authors of note

rmjarvis:
Add my name to note authors.  :)


rmjarvis:
I think easier to update the pdf at the end when you're ready to tag or publish or whatever.  PDFs don't diff well in git.

rmjarvis:
I usually do things like that directly on master just prior to a tag.  Then there are fewer versions of binary files being stored in the git repo, but you still get the correct version at the times you need it.

c-d-leonard:
I agree with you in principle - it's messy and it creates merge conflicts. I think for now let's merge this and I'm going to open an issue on changing our process on this going forward.

PR #769, opened 2020-04-29
update entry in authors.csv

mishakb:
Updated entry in authors.cvs as requested before Tagging new version. 
@c-d-leonard: if all looks good just merge it since you will get to this before I do in the morning and before the telecon  :).  I did not push the pdfs as discussed but they look good locally.


c-d-leonard:
Closed and re-opened in an attempt to re-trigger travis

PR #776, opened 2020-05-04
DOC added dev install docs

### LINKED ISSUE #772: "Lightweight" developer install for python updates?
I have had persistent problems with developer installation using `cmake`. I suspect that some of these issues are related to mac OS X, and especially the increase of built-in restrictions, although I'm not sure. Installation of released CCL with `conda` is easy.

Since most new developer features are pure python, can there be a developer install for python-only features that doesn't touch `cmake`? Or does something like this exist already?

beckermr:
This PR adds docs for a simple dev env.

Closes #772

cc @jablazek @c-d-leonard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 ## Python library
--
+- Added more developer documentation (#776).
 
 ## C library
 -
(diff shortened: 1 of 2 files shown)


beckermr:
New docs look good! https://ccl.readthedocs.io/en/dev-docs/source/developer_installation.html#boostrapping-a-ccl-development-environment-with-conda

beckermr:
I tested all of the commands in a new env locally so they def work.

PR #783, opened 2020-05-21
Quick LCDM cosmology

### LINKED ISSUE #782: Convenience cosmologies
We should have a few preset cosmologies one can use without worrying too much about passing parameters. This is to avoid having to remember what the names of the parameters are every time you need to declare a cosmology just to run a quick test.

damonge:
This creates a small subclass of `Cosmology` that just uses the best-fit Planck-2018 LCDM parameters. I'm gonna tag the following people:
- @slosar since he wanted this
- @beckermr since he didn't want this
- @erminiacalabrese since she may be able to vet the parameter values I've chosen

Closes #782


pyccl/core.py
```diff
@@ -1023,3 +1023,32 @@ def _set_linear_power_from_arrays(self, a_array=None, k_array=None,
...
+    """A cosmology roughly corresponding to the best-fit parameters from
+    the 2018 Planck release (arXiv:1807.06209).
+
+    Args:
+        **kwargs (dict): a dictionary of parameters passed as arguments
+            to the `Cosmology` constructor. It should not contain any of
+            the LambdaCDM parameters (`"Omega_c"`, `"Omega_b"`, `"n_s"`,
+            `"sigma8"`, `"A_s"`, `"h"`), since these are fixed.
+    """
+    def __init__(self, **kwargs):
```
beckermr, line 1038: This use of `**kwargs` and the docstring is great for developers since it is very easy code to write, but horrible for users since now they cannot find any of the information they might need about how to use this object.

damonge: I thought this was relatively standard (see e.g. https://matplotlib.org/devdocs/api/_as_gen/matplotlib.pyplot.imshow.html). Note that the documentation should show a link to the default cosmology object, where the full set of parameters is listed.

beckermr: Yeah the matplotlib docs are awful to use.

damonge: It feels silly to repeat the full documentation and set of parameters here, right?

beckermr: shrug - up to you

You can also access the __doc__ magic of the other class and paste it in.

pyccl/core.py
```diff
@@ -1023,3 +1023,32 @@ def _set_linear_power_from_arrays(self, a_array=None, k_array=None,
...
+        h = 0.6732
+        ob = 0.022383
+        oc = 0.12011
+        ns = 0.96605
+        s8 = 0.8120
+        p = {'h': h,
```
damonge, line 1044: These are the values @erminiacalabrese should check

erminiacalabrese: these are correct if you want the official Planck 2018 best-fit parameters from the nominal combo

damonge: Awesome, thanks!

jcolinhill: Planck also assumes a non-zero neutrino mass in their definition of the "base LCDM" model.  It is fixed to M_nu = 0.06 eV.  There is a file distributed with CLASS that contains all the default assumptions:
https://github.com/lesgourg/class_public/blob/master/base_2018_plikHM_TTTEEE_lowl_lowE_lensing.ini
(although the way this is implemented yields N_eff = 3.044 rather than 3.046)

damonge: Ah good point @jcolinhill 
So, not sure what to do here. On the one hand, we could ignore neutrinos (and maybe remove the `Planck18` from the class name to avoid confusion), since the main point of this class is just to provide a quick way to get a LCDM-like cosmology.
On the other, we can put in neutrinos (this may require more work, to make sure N_eff is right and all that), and that way we get Planck18 proper. However, certain things that we don't support for cosmologies with neutrinos won't work, which will be confusing.
I lean towards the first option. Thoughts @slosar ?

slosar: I think Planck should have mnu = 0.06 and N_eff  3.046, because 0.002 REALLY does not matter. And for things that are bad we issue warnings anyway.
I also think we should have a VanillaCosmology, which is just h=0.68, Om=0.3, mnu=0 cosmology. Here everything is rounded to make it clear that it is just an approximate quick cosmology and mnu=0 will make most of boltzmann calculations faster.

damonge: OK, I think we should only have one of those. I'm happy to go for "CosmologyVanilla".

damonge: (because of the title of this PR)

damonge:
Ping @slosar 
I've just made it into a VanillaLCDM cosmology.

PR #784, opened 2020-05-21
SZ sugar

damonge:
Adds capabilities to compute tSZ-related quantities.

Things left to do:

- [x] Fourier transform FFTLog.
- [x] GNFW instead of just Planck2013.
- [x] Check documentation.
- [x] Finish unit test with class_sz.


tilmantroester:
If you need benchmark code, I got some for the tSZ kernel.

damonge:
Ah great. You were already on my list of suspects for this!

damonge:
OK @tilmantroester what kind of benchmarks could you provide?
Ideally something like C_ell^yy and C_ell^ym would be great... Note that I've only implemented the Arnaud GNFW profile here for now

tilmantroester:
What I could provide easily is a the tSZ redshift kernel. Cls would require a bit more work. I have 
 C_ell^ym in principle from CosmoSIS but those would be based on a non-UPP halo model (Alex Mead's HMx). The python interface is not yet stable though. But in principle that could be plugged into the generalised tracers in CCL to get C_ell^ym to compare.

damonge:
@tilmantroester would you mind reviewing this?

I've benchmarked the full power spectrum calculation against [class_sz](https://github.com/borisbolliet/class_sz), after iterating with @borisbolliet (who has been most helpful!).

@beckermr I've introduced functionality to update the internal parameters of halo profiles through [this kind of stuff](https://github.com/LSSTDESC/CCL/blob/bfcf3aa14ba0f30d180500a98f288a2faad41951/pyccl/halos/profiles.py#L922). Having will be very useful, since many of these parameters are "fast" and shouldn't trigger a recalculation of slow things like Fourier transforms (this will be also true when we implement HODs and other profiles). I know you don't like kwargs, so I'd like to hear your thoughts on this particular point (no need to look at the full PR).

beckermr:
@damonge Why do you need the kwargs? Just list each one and set it to None. Then you can test and if it is non-None, then reset it.

damonge:
yep, that would work too. I just wondered if kwargs would be useful for likelihood design (e.g. if all the likelihood parameters are in a dictionary, you can just pass the full dictionary through, without having to pick the params in it that this particular profile depends on)

beckermr:
This sounds convenient but ultimately it is a recipe for errors. By making things explicit, we can help catch cases where ppl pass things that they think do one thing but don't actually do anything at all.

damonge:
Alright, let me do this.

damonge:
Alright, done. @tilmantroester , let me know if you think you'll be able to take a look.

pyccl/tests/test_profiles.py
```diff
@@ -70,6 +70,21 @@ def test_empirical_smoke(prof_class):
...
+    smoke_assert_prof_real(p)
+    p.update_parameters(mass_bias=0.7,
+                        alpha=1.24)
+    assert p.alpha == 1.24
+    assert p.mass_bias == 0.7
+    assert p.beta == beta_old
+    for n in ['P0', 'P0_hexp', 'alpha',
+              'beta', 'gamma', 'alpha_P',
+              'c500', 'mass_bias']:
+        p.update_parameters(**{n: 1.})
```
tilmantroester, line 85: Does this just test if `update_parameters` accepts `n` as a kwarg?

pyccl/tracers.py
```diff
@@ -542,6 +543,38 @@ def __init__(self, cosmo, z_source, n_samples=100):
...
+        n_chi (float): number of intervals in the radial comoving
+            distance on which we sample the kernel.
+    """
+    def __init__(self, cosmo, z_max=6., n_chi=1024):
+        self.chi_max = comoving_radial_distance(cosmo, 1./(1+z_max))
+        chi_arr = np.linspace(0, self.chi_max, n_chi)
+        a_arr = scale_factor_of_chi(cosmo, chi_arr)
+        # This is \sigma_T / (m_e * c^2)
+        prefac = 4.01710079e-06
+        w_arr = prefac * a_arr
```
tilmantroester, line 572: The kernel depends on whether the pressure is in physical or comoving coordinates. The docstrings don't specify, is that documented elsewhere?
This has tripped my up before when trying to match factors of `a` to hydrosim quantities.

pyccl/tests/test_profiles.py
damonge: It did, but that's not great, so now it also tests that the parameter actually gets updated (which in turn has allowed me to fix a minibug).

pyccl/tracers.py
damonge: Right, this should be physical, I believe (as in the one that goes in Eq. 2 of https://arxiv.org/pdf/1909.09102.pdf). I've documented this.

damonge:
Thanks @tilmantroester ! Back to you.

pyccl/halos/profiles.py
```diff
@@ -843,3 +844,214 @@ def _real(self, cosmo, r, M, a, mass_def):
...
+    mass :math:`M_{500}` via the mass bias parameter :math:`(1-b)`
+    as :math:`\\tilde{M}_{500}=(1-b)M_{500}`. :math:`\\tilde{r}_{500}`
+    is the overdensity halo radius associated with :math:`\\tilde{M}_{500}`
+    (note the intentional tilde!), and the profile is defined for
+    a halo overdensity :math:`\\Delta=500` with respect to the
+    critical density.
+
+    The default arguments (other than `mass_bias`), correspond to the
+    profile parameters used in the Planck 2013 (V) paper.
+
```
tilmantroester, line 879: Could you add the units of the pressure profile here as well for completeness?

damonge: Good point. Done!

damonge:
Awesome, thanks!

PR #787, opened 2020-06-18
ENH number counts integrals

beckermr:
This PR has number counts integrals


pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+        dvda = dvdz * abs_dzda
+
+        # this is used to convert quantities per dm to per dlog10m (e.g.
+        # the selection function )
+        dmdlog10m = self._mass * np.log(10)
+
+        # now do m intergrals in a loop
+        mint = np.zeros_like(a)
+        sint = np.zeros_like(a)
+        for i, _a in enumerate(a):
```
damonge, line 188: If this is too slow we may want to write a quick 2D integrator that openMPs over a. This wouldn't be too difficult, since we already have a spline integrator at the C level. I had to do something similar for the fftlog wrapper.

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+        and scale factor.
+
+        Note that the selection function is normalized to integrate to unity and
+        assumed to represent the selection probaility per unit scale factor and
+        per unit mass.
+
+        Args:
+            cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
+            sel (callable): function of mass and scale factor that returns the
+                possibly unnormalized selection function
```
damonge, line 160: There should be some information on what the expected dimensions of the inputs and outputs of this function are

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+        mint = np.zeros_like(a)
+        sint = np.zeros_like(a)
+        for i, _a in enumerate(a):
+            self._get_ingredients(_a, cosmo, False)
+            _selm = np.atleast_2d(sel(self._mass, _a)).T
+            mint[i] = self._integrator(
+                dvda[i] * self.mf[..., :] * _selm[..., :] * dmdlog10m,
+                self._lmass
+            )
+            sint[i] = self._integrator(_selm[..., :] * dmdlog10m, self._lmass)
```
damonge, line 195: I am not convinced that the selection function is something that should be normalized, or something that has units of M^-1 (what I said above about it being a probability distribution). You can't have two things that are per unit mass, integrate over mass once and then get something that does not have units of M^-1 (i.e. counts).

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+
+        # now do scale factor integrals and normalize
+        mtot = self._integrator(mint, a)
+        stot = self._integrator(sint, a)
```
damonge, line 199: See above

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+
+        if np.allclose(stot, 0):
+            return 0
+        else:
+            return mtot / stot
```
damonge, line 204: And see above

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+        # get a values for integral
+        a = np.linspace(cosmo.cosmo.spline_params.A_SPLINE_MIN, 1, na)
+        z = 1.0 / a - 1
+
+        # compute the volume element
+        abs_dzda = 1 / a / a
+        da = background.angular_diameter_distance(cosmo, a)
+        ez = background.h_over_h0(cosmo, a)
+        dh = physical_constants.CLIGHT_HMPC / cosmo['h']
+        dvdz = dh * (1.0 + z)**2 * da**2 / ez
```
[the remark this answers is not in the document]
beckermr: fixed

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,69 @@ def profile_norm(self, cosmo, a, prof):
...
+        dvda = dvdz * abs_dzda
+
+        # this is used to convert quantities per dm to per dlog10m (e.g.
+        # the selection function )
+        dmdlog10m = self._mass * np.log(10)
```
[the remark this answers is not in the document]
beckermr: fixed

beckermr:
Thanks @damonge! I fixed a bunch of things and addressed your comments. The code is still too slow and I can only match the numcosmo benchmarks to a percent or so.

pyccl/halos/halo_model.py
beckermr: this seems fast enough for now

beckermr:
OK @damonge. This PR is ready for another look. I'd say overall the integrals have an error around a percent or so. The benchmarks agree to a big better than this, but comparing to an adaptive integrator shows a bigger difference.

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,67 @@ def profile_norm(self, cosmo, a, prof):
...
+        :math:`sel(M,a)` is the selection function as a function of halo mass
+        and scale factor.
+
+        Note that the selection function is normalized to integrate to unity and
+        assumed to represent the selection probaility per unit scale factor and
+        per unit mass.
+
+        Args:
+            cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
+            sel (callable): function of mass and scale factor that returns the
```
damonge, line 159: Is there any situation where the selection function would be cosmology-dependent?

pyccl/halos/halo_model.py
```diff
@@ -137,6 +140,67 @@ def profile_norm(self, cosmo, a, prof):
...
+        ez = background.h_over_h0(cosmo, a)
+        dh = physical_constants.CLIGHT_HMPC / cosmo['h']
+        dvdz = dh * dc**2 / ez
+        dvda = dvdz * abs_dzda
+
+        # now do m intergrals in a loop
+        mint = np.zeros_like(a)
+        for i, _a in enumerate(a):
+            self._get_ingredients(_a, cosmo, False)
+            _selm = np.atleast_2d(sel(self._mass, _a)).T
```
damonge, line 193: Since you're transposing things here anyway, would it make sense to just ask that `sel` return arrays of shape ``(len(a), len(m))``?

damonge:
To clarify: does the adaptive integrator agree with numcosmo better than CCL?
Note that all the halo model integrals before matched the previous implementation of the halo model in CCL, which did use slow adaptive integrators. The mass function integrands are usually smooth enough that you don't need anything too fancy...

beckermr:
> To clarify: does the adaptive integrator agree with numcosmo better than CCL?
Note that all the halo model integrals before matched the previous implementation of the halo model in CCL, which did use slow adaptive integrators. The mass function integrands are usually smooth enough that you don't need anything too fancy...

No. Our spline integrator in the HMCalculator class agrees the best.

pyccl/halos/halo_model.py
beckermr: It could be, but this entire integral is at fixed cosmology so we can assume that dependence is fixed.

pyccl/halos/halo_model.py
beckermr: I was trying to match the dimensions of other quantities in the code under the assumption that this was useful. Happy to change.

beckermr:
@damonge I left some comments above. If you are ok with those, then let's merge.

pyccl/halos/halo_model.py
damonge: Ah wait, so are we doing this stupidly in other places too?

damonge: (and by we I mean me...)

beckermr: yes

damonge: Oh, actually, I remember there was a reason for this, to do with what's best to integrate profiles (which is a limiting factor often). OK, nevermind. Let's approve this.

PR #793, opened 2020-06-25
HOD-HOD-HOD

damonge:
HOD

- [x] Generalize and receive N_central and N_star from input
- [x] Unit test
- [x] Documented
- [x] Benchmarked (I suggest pinging the authors of https://arxiv.org/abs/1706.05422)


damonge:
@anicola you're absolutely right! Just fixed those two. Back to you!

Issue #770, opened 2020-04-29
CCL git repository is insanely large


EiffL:
A fresh clone of CCL this 4/29/2020 weighs in at a whopping 230MB \o/  The current files on master account for 63 MB, and the rest is hidden in the git history.

So, I'm suspecting some of the benchmark data is to blame. I haven't kept a close eye on that, but I would say we probably should purge the `data`  folder from the git history, and instead use a git-lfs folder for all data or massive files, to keep the repo light.

Someone with more experience with current CCL should chime in though :-)

EiffL:
Hum.... also, looks like the pdfs are committed every time... that's about 500Kb everytime, quickly add up

c-d-leonard:
So @EiffL do you have a suggestion about what to do instead of committing the pdfs? I agree it's dumb (@rmjarvis just pointed this out too and I opened an issue #768 ) - he said that his policy is to just make the pdf himself in master pre-releases, however we have always historically had a total embargo on anybody pushing anything to master. We could make an exception for this, but I'm curious if you have another good workaround.

damonge:
I would argue that no one other than ourselves is looking at the note, and that we could just remove its pdf from the repo (and ask users to compile it themselves)

EiffL:
We "could" setup some travis magic that execute a "make" in the doc folders upon for instance a new release. not sure what's the best trigger for this

tilmantroester:
Was the paper branch squash merged? Else there might be a lot of figures etc in the git history there.

rmjarvis:
I think there is not much utility in having the pdf in the git repo.  Perhaps you could periodically put a current version on the wiki.  Say every time you release a new version, update the note on the wiki.

EiffL:
(just need to make sure the paper compiles, through travis I guesss)

rmjarvis:
You could certainly add that as a "unit" test.  Run make in the paper directories and ensure that the pdf was generated.

c-d-leonard:
I like the idea of having it on the wiki. @elisachisari do you have thoughts on this from your experience with the paper and note?

rmjarvis:
FYI about the wiki, it has its own git repo, which in my experience is often easier to deal with than the web interface.  cf. https://help.github.com/en/github/building-a-strong-community/adding-or-editing-wiki-pages#adding-or-editing-wiki-pages-locally

EiffL:
[...]

Issue #768, opened 2020-04-28
Keeping note pdf in git?


c-d-leonard:
It has been pointed out by @rmjarvis that committing the note pdf to the repo each time it is changed creates issues with diffing. I also have found it to create unnecessary merge conflicts. Do we really want to keep the pdf of the note in the repo on each branch, or do we want to do something like just make it on master every so often? Tagging @elisachisari @damonge for comment.

c-d-leonard:
This has been resolved by PR #796, the note and paper pdf, as well as other produced-upon-compilation files like the authors.tex and contributions.tex files are no longer kept in the repo. I added a page on the wiki which links to the note; this should be updated at minimum at each release.

PR #796, opened 2020-07-02
Remove pdfs etc from repo

c-d-leonard:
This PR (partially) addresses Issues #770 and #768, which are to do with the fact that the git repo is getting larger and larger in part due to tracking files we don't really need to track.

So far, I have removed the note and paper PDFs, as well as the authors.tex and contributions.tex files which are generated upon compilation in each case from authors.csv (which stays). Proposal is to make a new page on the wiki where we can upload the current note at least up to the most recent release version. (I think making people upload it every time there is a new PR merged is probably overkill but we can discuss.)

It was also mentioned that we could remove some of the data files from the repo for the same reason, but I'm not really sure how we would do this since the test suite depends on them. Thoughts @rmjarvis @EiffL @beckermr ?


beckermr:
Just to be clear, this PR will remove them going forward, but they are still in the git history. We should merge to stop the repo growth though.

rmjarvis:
Agreed.  The tex files will diff just fine in git, so they won't bloat the repo.  The problem is only the binary files that git can't diff, so has to store a fresh copy of every time.

rmjarvis:
(And I posted my comments about the data files on #770.)

rmjarvis:
I'd also add to travis the command to build the pdf in the script section.  So one of the "unit tests" would be to make sure the two pdfs build properly.

c-d-leonard:
Thanks both!

@beckermr I included the tex files that I did because those particular ones are produced upon compilation (from authors.csv) which we wouldn't expect someone modifying the note to do on a regular basis under this new set-up. I don't mind keeping them but it seems weird to keep some of the compilation products (these) and not others (the pdf), thoughts?

@rmjarvis  I see, will have a look and see if I can get something like this working here

beckermr:
> @beckermr I included the tex files that I did because those particular ones are produced upon compilation (from authors.csv) which we wouldn't expect someone modifying the note to do on a regular basis under this new set-up. I don't mind keeping them but it seems weird to keep some of the compilation products (these) and not others (the pdf), thoughts?

Ohh sorry! I thought they were sources. Let's delete them!

.travis.yml
```diff
@@ -34,6 +34,8 @@ install:
 script:
 - flake8 pyccl
 - flake8 --exclude=data benchmarks
+- cd doc/0000-ccl_note && make && cd ..
+- cd doc/ccl_paper && make apj && cd ..
```
rmjarvis, line 38: I'm guessing this won't work because it's two layers deep.
I'd suggest
```
pushd doc/000-ccl_note && make && popd
pushd doc/ccl_paper && make && popd
```

c-d-leonard: True, let's see if it passes now.

c-d-leonard:
I would like to get some wider buy-in before moving benchmark data to the wiki because it represents a relatively substantial shift in structure. I'd like to go ahead and merge this and then do that as a different PR. I'm assuming this is okay because @beckermr has already approved.

rmjarvis:
Sorry, it looks like getting LaTex working on travis isn't as easy as I would have guessed.  This might help: https://malramsay.com/post/compiling_latex_on_travis/

beckermr:
I'd do the latex on linux only and use yum/apt-get to install the needed packages.

.travis/install.sh
```diff
@@ -33,17 +33,18 @@ py27)
...
     compilers pkg-config setuptools_scm pytest pandas pytest-cov \
-    cython "camb>=1" isitgr traitlets mkauthlist
+    cython "camb>=1" isitgr traitlets
   ;;
 esac;
 
 # we have to activate the cond env before we install this
 source activate test-environment
 pip install https://github.com/JoeMcEwen/FAST-PT/archive/v3.0.2.tar.gz --no-deps
+pip install mkauthlist
```
beckermr, line 49: you want --no-deps here and then to make sure any deps are in the list of conda installs above

c-d-leonard:
Many tweaks later - the note successfully makes on travis in the linux test (I haven't asked it to make in the osx test as we discussed). @beckermr and / or @rmjarvis would you mind having another quick look at what i've done to achieve this in the travis folder to see if there are any issues there?

PR #797, opened 2020-07-06
Double free bug

joezuntz:
This deals with a crash when computing correlation functions from C_ell that are all zero.

It tries to both deal with a lower level issue in the C code as well as short-cutting to avoid attempting the integral altogether.

I've added a test also.


joezuntz:
I have fixed flake8 errors in my new additions to the code - the remaining one is from code that was there before.

damonge:
Flake8 should be ignoring that error (and it was before): https://github.com/LSSTDESC/CCL/blob/c1464c6dd3470b0baa0133d51a68e3bdfcfb441f/setup.cfg#L2

@beckermr ?

joezuntz:
I've just merged master into this, so if that was the problem we'll find out shortly.

damonge:
Humm, it shouldn't (unless you branched off a while ago)

joezuntz:
It was a week or two ago, or so I thought.  But looks like it is fixed now: https://travis-ci.org/github/LSSTDESC/CCL/jobs/705352217

joezuntz:
Yep, looks like that's fixed it.

src/ccl_correlation.c
```diff
@@ -71,6 +71,16 @@ static void ccl_tracer_corr_fftlog(ccl_cosmology *cosmo,
 				  ccl_f1d_extrap_const,
 				  ccl_f1d_extrap_logx_logy,
 				  status);
+  if (*status) {
```
damonge, line 74: Can we add a test for this? I think this should get triggered if you pass an array of C_ells that is 0 everywhere and then 1 somewhere in the middle (to avoid the shortcut)

joezuntz: Good idea - added.

joezuntz:
Tests pass and I've added the new test @damonge mentioned.  Is there anything else people would like me to look at here?

PR #800, opened 2020-07-08
Sigma(M) in 2D

### LINKED ISSUE #609: ccl_sigmaR does not use the 2D P(k,z) spline to compute sigmaR
Comparing CCL and NumCosmo, which already implements the 2D (k, z) spline for P(k, z), we noticed that the 5e-5 required precision fails: 
- sigma at z > 0.8 and is attained between 0 <= z < 0.8 and  0.1 < R < 100;
- dlnsigma_dlog10M at z > 0.8 and is attained between 0 <= z < 0.8 and  0.3 < R < 100, if the interval for R starts at 0.1 then it already fails at z = 0.

The lower bound for R is controlled by K_MAX_SPLINE. But even if we increase its value, we still have problems at high z (z ~1).

_Originally posted by @pennalima in https://github.com/LSSTDESC/CCL/issues/79#issuecomment-472574231_

I just realized that CCL does not use the P(k,z) to compute sigma_R at redshifts z>0. It always compute at z=0 and then multiply the result by the growth function. This leads to lower precisions at high redshifts.

damonge:
We currently extrapolate sigma(M,z) in z using the linear growth factor. This fixes that.

Closes #609


damonge:
Currently works but it's rather slow.

damonge:
Closes #609

damonge:
Now addressed in #843

PR #804, opened 2020-07-22
Added a check for the value of Omega_k.

t-ferreira:
Values of Omega_k < -1.0135 (approximately), causes the kernel to crash, so I added an error message.


.python-version
```diff
@@ -0,0 +1 @@
+3.6.0
```
elisachisari, line 1: This looks like a user-specific file. Should it be unstaged?

t-ferreira: My apologies! Yes, it should.

damonge:
Just checking, is there anything particular about -1.0135?

damonge:
If the value is approximate, it might be best to leave a small safety buffer and cut at -1 instead.

Issue #780, opened 2020-05-13
pyccl.neutrinos.Omeganuh2 dodginess 


slosar:
It is very dodgy.
First, the documentation says:
```Docstring:
Calculate Omega_nu * h^2 at a given scale factor given the sum of the
neutrino masses.

.. note:: for all practical purposes, Neff is simply N_nu_mass.

Args:
    a (float or array-like): Scale factor, normalized to 1 today.
    m_nu (float or array-like): Neutrino mass (in eV)
    T_CMB (float, optional): Temperature of the CMB (K). Default: 2.725.

Returns:
    float or array_like: corresponding to a given neutrino mass.
File:      ~/anaconda3/lib/python3.6/site-packages/pyccl/neutrinos.py
Type:      function
```

but it wouldn't take a float as a second argument:
```
In [37]: pyccl.neutrinos.Omeganuh2(1/1150,0.3)                                                                         
---------------------------------------------------------------------------
CCLError                                  Traceback (most recent call last)
<ipython-input-37-94034cba1530> in <module>
----> 1 pyccl.neutrinos.Omeganuh2(1/1150,0.3)

~/anaconda3/lib/python3.6/site-packages/pyccl/neutrinos.py in Omeganuh2(a, m_nu, T_CMB)
     41     # Call function
     42     OmNuh2, status = lib.Omeganuh2_vec(N_nu_mass, T_CMB,
---> 43                                        a, m_nu, a.size, status)
     44 
     45     # Check status and return

~/anaconda3/lib/python3.6/site-packages/pyccl/ccllib.py in Omeganuh2_vec(N_nu_mass, T_CMB, a, mnu, nout, status)
   1313 
   1314     if numpy.shape(mnu) != (3,):
-> 1315         raise CCLError("Input shape for `mnu` must match `(3,)`!")
   1316 
   1317 

CCLError: Input shape for `mnu` must match `(3,)`!
```

More importantly, it is not symmetric, e.g
```
In [38]: pyccl.neutrinos.Omeganuh2(0.5,(0.3,0.,0.))                                                                      
Out[38]: 0.025933626497893495

In [39]: pyccl.neutrinos.Omeganuh2(0.5,(0.,0.,0.3))                                                                    
Out[39]: 0.00026937577197676493
```

slosar:
I guess it is returning the first omnuh2 instead of all three.

c-d-leonard:
Sorry to be so slow getting to this but yeah, this is not good. I'll have a look into this.

PR #808, opened 2020-07-23
Fix omeganuh2

c-d-leonard:
Fix the issues described in Issue #780 where ccl.Omeganuh2 has unexpected behaviour.

### DIFF
--- a/src/ccl_neutrinos.c
+++ b/src/ccl_neutrinos.c
@@ -170,14 +170,7 @@ double ccl_Omeganuh2 (double a, int N_nu_mass, double* mnu, double T_CMB, int* s
 
   Tnu = T_CMB*pow(4./11.,1./3.);
   a4 = a*a*a*a;
-  // Check if mnu=0. We assume that in the massless case mnu is a
-  // pointer to a single element and that element is 0. This should in principle never be called.
-  if (mnu[0] < 0.00017) {  // Limit taken from Lesgourges et al. 2012
-    prefix_massless = NU_CONST  * Tnu * Tnu * Tnu * Tnu;
-    return N_nu_mass*prefix_massless*7./8./a4;
-  }
 
-  // And the remaining massive case.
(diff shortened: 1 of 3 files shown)


c-d-leonard:
@slosar could you take a look at this? it fixes the Omeganuh2 dodginess you pointed out.

In particular, the C-level function was doing a thing where if the first quantity in the mass vector was ~0, it was using the massless case and only returning that. This was a buggy holdover from old code. The Omeganuh2 function is intended for the massive neutrino case, and in the case someone would call it with a zero or very low mass neutrino (e.g. by accident) you still get the right answer if you compute the full integral as for the massive case, so I've just had it do that.

I also fixed the issue where it wasn't accepting a float; that was a weird hard-coding thing in the ccl_neutrions.i file, should be fine now.

src/ccl_neutrinos.c
```diff
@@ -170,14 +170,7 @@ double ccl_Omeganuh2 (double a, int N_nu_mass, double* mnu, double T_CMB, int* s
 
   Tnu = T_CMB*pow(4./11.,1./3.);
   a4 = a*a*a*a;
-  // Check if mnu=0. We assume that in the massless case mnu is a
-  // pointer to a single element and that element is 0. This should in principle never be called.
-  if (mnu[0] < 0.00017) {  // Limit taken from Lesgourges et al. 2012
```
slosar, line 175: But did you actually have to remove this line? It does seem to be sensible to revert to the default calculation ofr low enough mass.

c-d-leonard: Fair point. I was being a bit lazy by removing it as it almost never should get hit (only for the case where we have to post-hoc fudge equal neutrino masses for cosmic emu, I think). But I've reintroduced that case now (in the below loop so we make sure to check them all and not just the first as was causing some of the original weird behaviour).

c-d-leonard: Thoughts @slosar? If you're happy could you approve?

PR #834, opened 2020-12-04
updated correlation.py to correlations.py to avoid import clash

### LINKED ISSUE #744: Update correlation documentation
The readthedocs documentation for correlation functions (including correlation_3d) is incorrect. It may be just that the function is incorrectly specified (with an extra `.correlation` in the call), but there may be other issues. We should also update documentation once the added functionality in Issue #743 is included.

jablazek:
This is pretty basic. Just trying to avoid a namespace clash. I don't think this will break the API, which was already broken by this clash.

Should resolve #744


jablazek:
Although, it WILL break code that explicitly imports the desired function with `import pyccl.correlation.correlation_3d` or similar.

I have updated the name of the file in tests/ to reflect the new correlations.py name. Within that test, it was already using the full import `pyccl as ccl` option. The benchmarks appear to test the actual `correlation` function in its own file, so no name change is needed there.

beckermr (approved):

jablazek:
How does coveralls work? It seems not to recognize that correlation.py has been replaced. Is this because the build is failing?

beckermr:
Fix the build and then worry about it.

jablazek:
OK, I think that builds will all fail until the change to github actions. So we can wait.

jablazek:
@beckermr, I think we're good now with the build and coveralls.
@c-d-leonard, any comments before we merge?

damonge:
Just checking: are we waiting on anything here @jablazek @c-d-leonard ?

beckermr:
Let's merge.

PR #835, opened 2020-12-04
Tk3D

damonge:
This will enable the Limber integrals needed for connected trispectrum covariances


damonge:
@anicola this is ready for review. A summary of what the modifications are (in order of importance):

Actual new content:
- At the C level, a structure and functions that holds the splines necessary to store evaluate functions of 3 variables: f(k1,k2,a). This is done by using bicubic interpolation in (k1,k2) for a fixed value of a, and linear interpolation in between values of a. The main purpose of this is to support calculations involving the connected trispectra used for power spectrum covariances.
- At the python level, a class that is able to save the same purpose (essentially translating between numpy arrays and C pointers. As an option, we allow for these functions to be factorizable as f(k1,k2,a)=f(k1,a)*f(k2,a), which can come in handy for SSC covariances.

Other bits:
- I deprecated the `ccl_f2d_customgrowth` option for CCL 2D splines, since we never ever used it.
- Fixed bits of old documentation here and there.

c-d-leonard:
Status here? @damonge @anicola ?

anicola:
This looks great to me @damonge. The only comment I have is if it might be worth adding a test somewhere that the arrays `a_arr`, `lk_arr`,  passed to `ccl_f3d_t_new` are strictly increasing.

anicola (commented):
This looks great to me @damonge. The only comment I have is if it might be worth adding a test somewhere that the arrays a_arr, lk_arr, passed to ccl_f3d_t_new are strictly increasing.

damonge:
@anicola Thanks a lot. Good point, I've added this check!

anicola (approved):
Fantastic! Thanks a lot @damonge!

PR #836, opened 2020-12-04
Github actions

damonge:
Testing GA here. If it works, this PR will switch us from travis to github actions.


damonge:
OK, @rmjarvis , @vitenti this branch is ready for review.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+
+        - name: Set up Python ${{ matrix.py }}
+          uses: actions/setup-python@v2
+          with:
+            python-version: ${{ matrix.py }}
+
+        - name: Install GFortran macOS
+          if: matrix.os == 'macos-latest'
+          run: |
+            brew cask install gfortran || true
```
rmjarvis, line 28: Could leave a comment here about the || true bit.  Since you got it from me, I can say that this is because brew returns a failure exit code if the thing is already installed and current (which would often be the case if you used a cache, which you aren't doing yet).  So || true means this statement always is considered to succeed.  If there was a real failure in brew installation, then something later will fail instead.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
...
+
+        - name: Benchmarks
+          run: OMP_NUM_THREADS=2 pytest -vv benchmarks --cov=pyccl --cov-append
+
+        - name: Coveralls
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          run: |
+            pip install -U coveralls
+            coveralls
```
rmjarvis, line 96: Maybe consider adding a cache?  I don't have any experience yet how important it is for GHA.  But it sped things up a lot in Travis to cache pip and brew directories.  Here is what that would look like I think:
```
           - name: Cache pip, brew
              uses: actions/cache@v2
              with:
                path: |
                  ~/.cache/pip
                  $HOME/Library/Caches/Homebrew
                  /usr/local/Cellar
                key: ${{ runner.os }}-${{ matrix.py }}-cache
                restore-keys: |
                    ${{ runner.os }}-${{ matrix.py }}-cache
                    ${{ runner.os }}-
```

rmjarvis (approved):
Looks good.  Just a few suggestions, which you can take or leave.

vitenti:
For whoever is going to merge, it can be a good idea to squash this during merge to avoid polluting master with test commits.

damonge:
@vitenti yep, that's the standard procedure we use
@rmjarvis thanks a lot! I've implemented these and will merge as soon as things pass.

.github/workflows/ci.yml
damonge: Done

.github/workflows/ci.yml
```diff
@@ -0,0 +1,112 @@
+name: continuous-integration
+on: [push]
```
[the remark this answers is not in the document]
damonge: Ok, sounds good. Let me give it a try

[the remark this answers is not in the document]
damonge: Question @rmjarvis : is this really supposed to be `main` or `master`? It seems the actions are not being triggered after including this.

.github/workflows/ci.yml
damonge: OK, caching didn't seem to have a huge impact, but I've left it in just in case (doesn't hurt to have it).

rmjarvis:
For ccl it would be master. I switched my repos to all use main for the main branch.

Issue #819, opened 2020-09-02
Access non-linear power spectrum predictions from CAMB/CLASS


tilmantroester:
Currently we only make use of the linear power spectrum from CAMB and CLASS. Both codes also implement non-linear power spectrum methods, e.g., various versions of HALOFIT and HMCode.

Especially the latter would be useful to access. An immediate application would be the txpipe-reanalysis project that requires an HMCode implementation for the KV450 analysis (see https://github.com/LSSTDESC/txpipe-reanalysis/issues/26).

This could be implemented through some extra piping within CCL, or through the Pk2D interface.

PR #837, opened 2020-12-04
[WIP] Expose nonlinear power spectra from Boltzmann codes

tilmantroester:
Currently CCL only uses the linear power spectrum prediction from Boltzmann codes, with limited options for nonlinear power spectra. Boltzmann codes provide a range of fast nonlinear models that can be used. For example, CAMB provides an extensive list of codes, among them HMCode.

This PR uses the HMCode implementation in CAMB to provide nonlinear predictions for the matter power spectrum, including modelling of baryons, for CCL. It addresses issue #819.  

To make the baryon modelling accessible, the HMCode parameters A and eta (for the 2016 model), and the AGN feedback strength logT (for the 2020 model) are exposed. To get this to work consistently and for the yaml output, these are also added to the C layer.
This kinda changes the Python API (by adding new but optional parameters to `Cosmology.__init__`) and clearly changes the C API (by changing `ccl_parameters_create`). 
I'd rather have avoided that but right now everything needs to go through the C layer. For example, the `write_yaml` method uses the C-layer dumper function. NB, `read_yaml` doesn't use the C-layer reader. 
While there is a `read_yaml` function in the C-layer, it is not being used. I've added the new parameters to it but this breaks loading old yaml files due to the super simple implementation of the C `read_yaml` function.

If not for the yaml functionality, the HMCode parameters could probably live in the Python layer in a structure adjacent to `_params`.

## To discuss
Should the HMCode parameters get added to the `ccl_parameter` struct or better live exclusively in the Python layer? This would require some refactor of the yaml functionality. Which might be a good thing. Currently it is quite painful to add any new functionality since everything is still tightly bound to the C layer.

## Todo
- [ ] Add tests
- [ ] Fix existing tests
- [ ] Expose nonlinear power spectra in CLASS

### DIFF
--- a/benchmarks/ccl_test_angpow.c
+++ b/benchmarks/ccl_test_angpow.c
@@ -63,7 +63,7 @@ static void test_angpow_precision(struct angpow_data * data)
   ccl_parameters ccl_params = ccl_parameters_create(data->Omega_c, data->Omega_b, data->Omega_k,
 						    data->Neff, data->mnu, data->mnu_type,data->w_0,
 						    data->w_a, data->h, data->A_s, data->n_s,
-						    -1,-1,-1,data->mu_0, data->sigma_0,-1,NULL,NULL, &status);
+						    -1,-1,-1,-1,-1,-1,data->mu_0, data->sigma_0,-1,NULL,NULL, &status);
   ccl_params.Omega_g=0.;
   ccl_params.Omega_l=data->Omega_v;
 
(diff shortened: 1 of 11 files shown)


damonge:
OK, do we really want the `write_yaml` at all? (@joezuntz ?)
And if we do, could we just put it in python?

joezuntz:
Yes, fine to drop this and replace it with pure python. In practice what we use in TXPipe just needs to initialize a cosmology, not fully serialize e.g. calculations that have already been run.  So yes, please go ahead.

tilmantroester:
Replaced by #854

PR #843, opened 2020-12-22
Calculator mode

### LINKED ISSUE #609: ccl_sigmaR does not use the 2D P(k,z) spline to compute sigmaR

### LINKED ISSUE #815: Correlation functions using built-in Pk->corr
Add functionality to allow an arbitrary Pk to be transformed into 3d or projected correlation functions. (This is a stub issue. @slosar @damonge can add more details/requirements)

### LINKED ISSUE #840: Observable calculator mode
## The issue
It is currently hard for CCL to support arbitrary non-standard cosmologies. E.g. imagine we're told by someone in TJP that they want to explore an early Dark Energy model with neutrinos, warm dark matter and Vainshtein screening, and accounting for baryons using HMCODE. Including each of those consistently in CCL would take a lot more work than we can support. These things, however, are implemented in different versions of CAMB/CLASS etc.

## The proposed solution
We enable a "calculator mode" for CCL so it can take in basic observables (distances, power spectra) from these and use them consistently to carry out calculations of more complex quantities (shear/galaxy power spectra, correlation functions, halo model stuff). This would also make CCL easier to integrate into likelihood frameworks like cobaya or cosmosis. When we've discussed this in the past, it was well received, so I'm hoping this is not too controversial.

## The specifics
Here's a concrete proposal of what the new API would look like. There's a chance that none of the previous API would have to be deprecated, but we'll see.

### 1. Declaring a `Cosmology`
Although ideally we would like the CCL `Cosmology` object to not care about cosmological parameters when in "calculator mode", in practice many of the higher-level calculations we do (e.g. halo model) do need to know e.g. what Omega_M is. So, instead of subclassing the `Cosmology` class into a calculator-mode object, I propose that we just add 3 more arguments to its `__init__` method, which will trigger the calculator mode:

- `background`: a dictionary with three items: "a", an array of scale factor values, "h_over_h0" an array with H(a)/H(a=1), and "chi", an array containing the comoving radial distance to scale factor a. If `None`, CCL will compute these on its own.
- `pk_lin`: a dictionary in the form `{'a': a_arr, 'k':k_arr, 'q1_x_q2': pkq1q2}`, where `a_arr` is an array of scale factors, `k_arr` is an array of comoving wave numbers, and `pkq1q2` is a 2D array containing the linear power spectrum between quantities `q1` and `q2` sampled at the values of `a_arr` and `k_arr`. The intention is that `q1` and `q2` could be any of e.g. [these quantities](https://camb.readthedocs.io/en/latest/transfer_variables.html#transfer-variables). There can be an arbitrary number of `q1_x_q2` combinations, with any names you want, but at least one of them should be `delta_matter_x_delta_matter`, designating the linear matter power spectrum. CCL will create `Pk2D` objects out of all of these and will store them at the python level. The C-level linear power spectrum structure attached to a C-level cosmology structure will disappear, and all interaction with the C-level functions will happen via the `f2d` C structures attached to these `Pk2D` objects. If `None`, CCL will just compute `delta_matter_x_delta_matter` on its own as it does right now.
- `pk_nl`: same as `pk_lin` for the non-linear power spectrum with the following caveat: if any of the `q1_x_q2` combinations listed in `pk_lin` is **not** listed in `pk_nl`, CCL will compute the corresponding non-linear power spectrum via its halofit implementation. This will happen for all combinations if `pk_nl==None`.

The default value for all these arguments will be `None` (i.e. the current standard behavior in CCL).

### 2. Impact on the rest of the API
All CCL functions that currently depend on a P(k) will take an optional `p_of_k_a` argument (the same way that `angular_cl` currently does), specifying the power spectrum that should be used for that calculation. This argument can then take 3 possible values:

- `None`, in which case the `delta_matter_x_delta_matter` power spectrum is used. This will be the default value.
- A string of the form `q1_x_q2`, specifying which of the stored power spectra should be used. If not present, an exception will be thrown.
- A `Pk2D` object containing the power spectrum that should be used.

A quick scan of the current API tells me these functions would be affected: [angular_cl](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/cls.py#L15), [correlation_3d](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/correlations.py#L113), [correlation_multipole](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/correlations.py#L145), [correlation_3dRsd](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/correlations.py#L180), [correlation_3dRsd_avgmu](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/correlations.py#L220), [correlation_pi_sigma](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/correlations.py#L255), [sigmaR](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L64), [sigmaV](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L80), [sigma8](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L97), [kNL](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L113), [halomod_power_spectrum](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/halos/halo_model.py#L406), [halomod_Pk2D](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/halos/halo_model.py#L540).

[linear_matter_power](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L7) and [nonlin_matter_power](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L23) will continue to return the matter power spectrum (`delta_matter_x_delta_matter`).

[sigmaM](https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/power.py#L39) will always use the linear matter power spectrum, since this is what is used for halo model quantities.

Thoughts? @c-d-leonard @slosar @beckermr @elisachisari @mishakb and anyone else.

damonge:
Closes #840 
Closes #815 
Closes #609


damonge:
OK, this is ready for review. This is a big PR, and I apologize in advance to whoever reviews it, but I don't think this could have been done efficiently via mini-PRs, since the work involved cutting and reconnecting a lot of the C-level wires of CCL. Because of this, I discovered a number of small issues that I decided to fix while I was making the modifications specific to this PR.

Below is a summary of the changes implemented. I'm tagging a few people whose eyes on this would be very welcome if they have the time: @beckermr @slosar @c-d-leonard.

Two disclaimers:
- Although there are substantial changes, this PR does not alter the existing API.
- The execution time of typical operations (Limber integrals etc.) does not change at all, and #844 is there to prove it (in fact it may be better to merge that one first). This is mostly a rewiring of the CCL internals.
- Although many lines have changed, in many cases these are small things (indentations, outdated stuff, removed old code etc.), so I actually think reviewing this should be a lot easier than it looks.


## Changes in this PR

### Main change
Implements the "calculator mode" discussed in #840 as a subclass of the `Cosmology` class [here](https://github.com/LSSTDESC/CCL/blob/ded8ec754f5c2105498b0c66380aa3c3010faf8b/pyccl/core.py#L938). Please have a look at the docstring, which specifies how to use this functionality, and let me know if anything doesn't make sense. I can create a notebook showing how the life cycle of these things would work, but I'd rather get some comments first.

### Associated changes in python:

**New functionality**
- General methods (`linear_power`, `nonlin_power`) to compute linear and non-linear power spectra attached to a cosmology object.
- `sigmaR`, `sigmaV`, `sigma8`, `sigmaM`, and `kNL` no longer assume scale-independent growth (see below about sigma_M).
- `bcm_correct_pk2d` applies the BCM correction to arbitrary `Pk2D` objects.
- The following functions now take an optional argument to specify which power spectrum should be used with them: `sigmaR`, `sigmaV`, `sigma8`, `kNL`, `angular_cl`, `correlation_3d`, `correlation_multipole`. `correlation_3dRsd`, `correlation_3dRsd_avgmu`, `correlation_pi_sigma`.
- `Pk2D` objects can be created as `empty` (i.e. without a C-level `ccl_f2d_t` structure attached), to be filled out later. This was convenient for the following new Pk2D methods:
- `Pk2D` class method `pk_from_model` that creates a new `Pk2D` object from a numerical model (BBKS, E&H or cosmic-emu).
- `Pk2D` class method `halofit_it` that creates a new `Pk2D` object by applying halofit on another `Pk2D` object.

**Removed hidden python methods**
- The temporary code written to populate the `Cosmology` object from arrays (`_set_linear_power_from_arrays` etc.). We agreed to do this once the proper API was in place.
- `_vectorize_fn2`, which was only used by functions that have now been modified.

**New tests**
- For all new functionality.
- Some untested previous functionality (the `bcm` methods, and some odds and ends in `core.py`).

**Documentation**
- All new functionality is documented.
- The `correlations` module wasn't being compiled by rtd after it changed names.

### Associated changes in C:

**New C-level functionality**
- Code to apply the BCM correction to general `ccl_f2d_t` power spectra.
- Code to apply HALOFIT to generic `ccl_f2d_t` power spectra.
- Code to create Pk2D objects from the BBKS, Eisenstein & Hu and cosmic-emu models.
- Code to rescale generic `ccl_f2d_t` power spectra by sigma8 and mu/sigma. `ccl_cosmology_spline_linpower_musigma` is now called `ccl_rescale_musigma_s8`. Instead of creating a new `ccl_f2d_t` power spectrum with fixed sampling rate in k and a, this new function just applies the rescaling by mu/sigma and sigma8 to an input `ccl_f2d_t` structure. This is both faster and more memory-efficient.
- `ccl_sigmaR`, `ccl_sigma8`, `ccl_sigmaV` and `ccl_kNL` no longer assume scale-independent growth.

**Removed C code**
- linear and non-linear power spectra attached to the C-level cosmology structure. This has changed the signature of some C-functions which used these internal power spectra. They now depend on input `ccl_f2d_t` structures.
- C-level functions to evaluate the C-level power spectrum splines attached to the C-level cosmology structure.

**Sigma_M**
The splines holding `sigma_M` and its derivative have been modified so that:
- They don't assume a scale independent growth anymore.
- We calculate `dlogsigmaM/dlogM` from the spline derivatives, instead of via finite differences + additional spline storage.
This was necessary in order to make the calculator mode truly work as a general observable calculator (and it also closes one of our most annoying outstanding issues). I tested that this does not change the initialization time significantly (<0.05s on my laptop - compare with ~0.2s for applying halofit). This was achieved by downsampling the sigmaM spline (we were over-sampling it by a factor ~8). 

**General cleanup**
This included (although more is needed): unused variables and functions, uninitialized variables, indentations, outdated comments, etc. I also removed some unused C-level functions (`ccl_parameters_create_flat_lcdm`). These were already not used before these code modifications.

### Bugfixes
- The previous functions that loaded growth from arrays didn't normalize it properly.
- Some of the C-level error messages were wrong.
- `write_yaml` wasn't catching the C-level errors properly.

include/ccl_musigma.h
```diff
@@ -10,7 +10,8 @@ CCL_BEGIN_DECLS
  ^ @param psp The linear power spectrum to spline.
  * @param status, integer indicating the status
  */
-void ccl_cosmology_spline_linpower_musigma(ccl_cosmology* cosmo, ccl_f2d_t *psp, int isitgr_flag, int* status);
+void ccl_rescale_musigma_s8(ccl_cosmology* cosmo, ccl_f2d_t *psp,
+                            int mg_rescale, int* status);
```
beckermr, line 14: Note for self: figure out why isitgr_flag was renamed.

include/ccl_neutrinos.h
```diff
@@ -31,15 +31,6 @@ typedef enum ccl_neutrino_mass_splits{
...
 } ccl_neutrino_mass_splits;
 
-/**
- * Spline for the phasespace integral required for getting the fractional energy density of massive neutrinos.
- * Returns a gsl spline for the phase space integral needed for massive neutrinos.
- * @param status Status flag. 0 if there are no errors, nonzero otherwise.
- * For specific cases see documentation for ccl_error.c
- * @return spl, the gsl spline for the phasespace integral required for massive neutrino calculations.
- */
-gsl_spline* calculate_nu_phasespace_spline(int *status);
```
beckermr, line 41: Note for self: Why do we not need this anymore?

beckermr (commented):
I got about 1/3 of the way through but need a break. One comment on the docs. I think if you run the `rtd/apidoc.sh` script, it will make the api docs again. I am guessing you may need to do this? I don't actually recall how well we hooked all of this up.

pyccl/ccl_power.i
```diff
@@ -12,30 +12,6 @@
...
 
-/* The python code here will be executed before all of the functions that
-   follow this directive. */
-%feature("pythonprepend") %{
-    if numpy.shape(k) != (nout,):
-        raise CCLError("Input shape for `k` must match `(nout,)`!")
-%}
-
-%inline %{
-void linear_matter_power_vec(ccl_cosmology * cosmo, double a, double* k, int nk,
```
beckermr, line 23: Note to self: figure out what replaces this.

pyccl/pk2d.py
```diff
@@ -104,6 +108,62 @@ def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
...
+            status = ret
+        else:
+            pk2d.psp, status = ret
+
+        check(status, cosmo)
+        pk2d.has_psp = True
+        return pk2d
+
+    @classmethod
+    def halofit_it(Pk2D, cosmo, pk_linear):
```
beckermr, line 146: This function name is clever but a little hard to parse. Maybe `apply_halofit` or `compute_halofit`?

pyccl/pk2d.py
```diff
@@ -149,3 +209,43 @@ def __del__(self):
...
+        else:
+            raise ValueError("p_of_k_a must be a pyccl.Pk2D object, "
+                             "a string, or None")
+
+        if is_linear:
+            cosmo.compute_linear_power()
+            pk = cosmo._pk_lin
+        else:
+            cosmo.compute_nonlin_power()
+            pk = cosmo._pk_nl
```
beckermr, line 246: I suspect we need an additional interface on the cosmology object that looks like `get_linear_power` and `get_nonlin_power` that pulls these attributes from the dicts and calls the compute functions if needed.

pyccl/tests/test_background.py
```diff
@@ -24,7 +24,7 @@
 input_a_array = np.linspace(0.1, 1, 100)
 input_chi = ccl.background.comoving_radial_distance(COSMO, input_a_array)
 input_hoh0 = ccl.background.h_over_h0(COSMO, input_a_array)
-input_growth = ccl.background.growth_factor(COSMO, input_a_array)
+input_growth = ccl.background.growth_factor_unnorm(COSMO, input_a_array)
```
beckermr, line 27: Why did this switch to unnormalized?

pyccl/tests/test_background.py
```diff
@@ -128,17 +128,19 @@ def test_input_arrays():
     a_arr = np.linspace(0.1, 1, 100)
     chi_from_ccl = ccl.background.comoving_radial_distance(cosmo, a_arr)
     hoh0_from_ccl = ccl.background.h_over_h0(cosmo, a_arr)
-    growth_from_ccl = ccl.background.growth_factor(cosmo, a_arr)
+    growth_from_ccl = ccl.background.growth_factor_unnorm(cosmo, a_arr)
```
beckermr, line 131: Same question here about normalization.

pyccl/tests/test_pk2d.py
```diff
@@ -65,6 +66,60 @@ def test_pk2d_smoke():
     assert_(not np.isnan(psp.eval(1E-2, 0.5, cosmo)))
 
 
+@pytest.mark.parametrize('model', ['bbks', 'eisenstein_hu'])
+def test_pk2d_from_model(model):
+    cosmo = ccl.Cosmology(
+        Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
+        transfer_function=model)
+    pk = ccl.Pk2D.pk_from_model(cosmo, model=model)
```
beckermr, line 74: So if I make a Cosmology object with BBKS but then call pk_from_model with model='eisenstein_hu', will the code fail? If it does, why do I need to pass the model string at all? This API confuses me a bit.

pyccl/tests/test_power.py
```diff
@@ -32,7 +32,7 @@ def test_halomod_f2d_copy():
         ccl.CCLWarning,
         ccl.nonlin_matter_power, COSMO_HM, 1., 0.8)
     pk_new = pk2d.eval(1., 0.8, COSMO_HM)
-    psp_old = COSMO_HM.cosmo.data.p_nl
+    psp_old = COSMO_HM._pk_nl['delta_matter_x_delta_matter'].psp
```
beckermr, line 35: Private attribute access is a bad sign. I think the `get_nonlin_power` method I suggested above might be a good idea.

pyccl/tests/test_power.py
```diff
@@ -187,89 +205,278 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+    # Test again passing only linear Pk, but letting HALOFIT do its thing
+    kl_arr = np.logspace(-4, 1, 1000)
+    pkl_arr = np.array([ccl.power.linear_matter_power(cosmo, kl_arr, a)
+                        for a in a_arr])
+    cosmo_input = ccl.CosmologyCalculator(
+        Omega_c=0.27, Omega_b=0.05, h=0.7,
+        n_s=0.965, A_s=2e-9,
+        pk_linear={'a': a_arr, 'k': kl_arr,
+                   'delta_matter_x_delta_matter': pkl_arr},
+        use_halofit=True)
```
beckermr, line 382: Niiiiiice.

pyccl/tests/test_power.py
```diff
@@ -187,89 +205,278 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+                        for a in a_arr])
+    cosmo_input = ccl.CosmologyCalculator(
+        Omega_c=0.27, Omega_b=0.05, h=0.7,
+        n_s=0.965, A_s=2e-9,
+        pk_linear={'a': a_arr, 'k': kl_arr,
+                   'delta_matter_x_delta_matter': pkl_arr,
+                   'a_x_b': pkl_arr},
+        pk_nonlin={'a': a_arr, 'k': k_arr,
+                   'delta_matter_x_delta_matter': -pk_arr},
+        use_halofit=True)
```
beckermr, line 400: This is going to trip people up and create weird bugs where people think they are getting one thing but not another. I wonder if we should not allow this calling signature and make people be more explicit about what they want? We could force use_halofit to be a dict in this case mapping the linear power name to True or False.

pyccl/tests/test_power.py
```diff
@@ -187,89 +205,278 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+
+    # Non-parsable power spectrum
     with pytest.raises(ValueError):
-        cosmo_input._set_nonlin_power_from_arrays()
+        ccl.CosmologyCalculator(
+            Omega_c=0.27, Omega_b=0.05, h=0.7,
+            n_s=0.965, sigma8=0.8,
+            pk_nonlin={'a': a_arr, 'k': k_arr,
+                       'delta_matter_x_delta_matter': pk_arr,
+                       'a_y_b': pk_arr})
```
beckermr, line 453: A test with `x_x_y` is probably a good idea.

pyccl/tests/test_power.py
```diff
@@ -187,89 +205,278 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+
+    # k not present
+    with pytest.raises(ValueError):
+        ccl.CosmologyCalculator(
+            Omega_c=0.27, Omega_b=0.05, h=0.7,
+            n_s=0.965, sigma8=0.8,
+            pk_nonlin={'a': a_arr, 'kk': k_arr,
+                       'delta_matter_y_delta_matter': pk_arr})
+
+    # delta_matter_x_delta_matter not present
```
beckermr, line 438: Does the calculator always have to have the matter power spectrum?

src/ccl_f1d.c
```diff
@@ -150,7 +150,8 @@ double ccl_f1d_t_eval(ccl_f1d_t *spl,double x)
       return spl->y_ini * pow(x/spl->x_ini, spl->der_lo);
     }
     else {
-      ccl_raise_gsl_warning(CCL_ERROR_SPLINE_EV, "ccl_utils.c: ccl_splin_eval(): "
+      ccl_raise_gsl_warning(CCL_ERROR_SPLINE_EV,
+                            "ccl_f1d.c: ccl_f1d_t_eval(): "
 			    "x-value below range.");
```
beckermr, line 155: Indentation here.

src/ccl_f1d.c
```diff
@@ -171,7 +172,8 @@ double ccl_f1d_t_eval(ccl_f1d_t *spl,double x)
       return spl->y_end * pow(x/spl->x_end, spl->der_hi);
     }
     else {
-      ccl_raise_gsl_warning(CCL_ERROR_SPLINE_EV, "ccl_utils.c: ccl_splin_eval(): "
+      ccl_raise_gsl_warning(CCL_ERROR_SPLINE_EV,
+                            "ccl_f1d.c: ccl_f1d_t_eval(): "
 			    "x-value above range.");
```
beckermr, line 177: Ditto

src/ccl_f1d.c
```diff
@@ -180,7 +182,7 @@ double ccl_f1d_t_eval(ccl_f1d_t *spl,double x)
     double y;
     int stat=gsl_spline_eval_e(spl->spline,x,NULL,&y);
     if (stat!=GSL_SUCCESS) {
-      ccl_raise_gsl_warning(stat, "ccl_utils.c: ccl_splin_eval(): "
+      ccl_raise_gsl_warning(stat, "ccl_f1d.c: ccl_f1d_t_eval(): "
 			    "x-value outside range.");
```
beckermr, line 186: ditto

src/ccl_mass_conversion.c
```diff
@@ -118,12 +118,12 @@ void ccl_convert_concentration(ccl_cosmology *cosmo,
...
   double d_factor = delta_old/delta_new;
   for(ii=0;ii<nc;ii++) {
-    double c_new_h;
     st=convert_concentration_single(d_factor, c_old[ii], &(c_new[ii]), c_old[ii]);
     if(st!=GSL_SUCCESS) {
       *status=CCL_ERROR_ROOT;
       ccl_cosmology_set_status_message(cosmo,
-				       "ccl_mass_conversion.c: NR solver failed to find a root\n");
+        "ccl_mass_conversion.c: ccl_convert_concentration(): "
+        "NR solver failed to find a root\n");
```
beckermr, line 126: I don't think NR solver is the right thing here?

src/ccl_massfunc.c
```diff
@@ -48,127 +48,117 @@ static double sigmaM_m2r(ccl_cosmology *cosmo, double halomass, int *status)
...
+  m = ccl_linear_spacing(cosmo->spline_params.LOGM_SPLINE_MIN,
+                         cosmo->spline_params.LOGM_SPLINE_MAX, nm);
   if (m == NULL ||
       (fabs(m[0]-cosmo->spline_params.LOGM_SPLINE_MIN)>1e-5) ||
       (fabs(m[nm-1]-cosmo->spline_params.LOGM_SPLINE_MAX)>1e-5) ||
       (m[nm-1]>10E17)) {
     *status = CCL_ERROR_MEMORY;
-    ccl_cosmology_set_status_message(cosmo,"ccl_cosmology_compute_sigmas(): Error creating linear spacing in m\n");
+    ccl_cosmology_set_status_message(cosmo,
+                                     "ccl_massfunc.c: ccl_cosmology_compute_sigmas(): "
```
beckermr, line 78: `ccl_cosmology_compute_sigma():`

src/ccl_massfunc.c
```diff
@@ -48,127 +48,117 @@ static double sigmaM_m2r(ccl_cosmology *cosmo, double halomass, int *status)
...
 
+  gsl_spline2d *lsM = NULL;
   if (*status == 0) {
-    dlnsigma_dlogm = gsl_spline_alloc(cosmo->spline_params.M_SPLINE_TYPE, nm);
-    if (dlnsigma_dlogm == NULL) {
+    lsM = gsl_spline2d_alloc(gsl_interp2d_bicubic, nm, na);
+    if (lsM == NULL) {
       *status = CCL_ERROR_MEMORY;
+      ccl_cosmology_set_status_message(cosmo,
+                                       "ccl_massfunc.c: ccl_cosmology_compute_sigmas(): "
```
beckermr, line 139: `ccl_cosmology_compute_sigma()`

src/ccl_massfunc.c
```diff
@@ -48,127 +48,117 @@ static double sigmaM_m2r(ccl_cosmology *cosmo, double halomass, int *status)
...
+                                       "memory allocation\n");
```
beckermr, line 140: Make this more specific?

src/ccl_utils.c
```diff
@@ -282,7 +285,8 @@ void ccl_integ_spline(int ny, int nx,double *x,double **y,
 
   if((b>x[nx-1]) || (a<x[0])) {
     ccl_raise_warning(CCL_ERROR_SPLINE,
-		      "ERROR: integration limits beyond interpolated range\n");
+		      "ccl_utils.c: ccl_integ_spline(): "
+                      "integration limits beyond interpolated range\n");
```
beckermr, line 289: spacing.

src/ccl_background.c
```diff
@@ -714,23 +714,39 @@ void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[
...
+                                     double growth_arr[], double fgrowth_arr[], int* status)
 {
   int chistatus;
   if (cosmo->computed_growth)
     return;
 
+  double *growth_normed = NULL;
   gsl_spline * growth = gsl_spline_alloc(cosmo->spline_params.A_SPLINE_TYPE, na);
   gsl_spline * fgrowth = gsl_spline_alloc(cosmo->spline_params.A_SPLINE_TYPE, na);
+  double growth0 = growth_arr[na-1];
```
beckermr, line 727: We need a check here or in the python code that ensure this element is at or close to `a=1`.

src/ccl_background.c
```diff
@@ -714,23 +714,39 @@ void ccl_cosmology_distances_from_input(ccl_cosmology * cosmo, int na, double a[
...
+      ccl_cosmology_set_status_message(
+           cosmo, "ccl_background.c: ccl_cosmology_growth_from_input(): ran out of memory\n");
+    }
+  }
+
+  if (*status == 0) {
+    for(int ia=0; ia<na; ia++)
+      growth_normed[ia] = growth_arr[ia]/growth0;
+
+    chistatus = gsl_spline_init(growth, a, growth_normed, na);
```
beckermr, line 749: We are now storing the normalized growth instead of the unnormalized growth in the same spot. Do we have a test that we are not applying the normalization twice in other places in the code?

pyccl/core.py
```diff
@@ -760,51 +704,37 @@ def _compute_linear_power_internal(self):
...
+        if trf == 'boltzmann_class':
+            pk = get_class_pk_lin(self)
+        elif trf == 'boltzmann_isitgr':
+            rescale_mg = False
+            pk = get_isitgr_pk_lin(self)
+        elif trf == 'boltzmann_camb':
+            pk = get_camb_pk_lin(self)
+        elif trf in ['bbks', 'eisenstein_hu']:
+            rescale_s8 = False
+            rescale_mg = False
```
beckermr, line 721: Why can't we use mu-Sigma with EH98 or BBKS?

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
                              "%s " % list(p.keys()))
         kwargs.update(p)
         super(CosmologyVanillaLCDM, self).__init__(**kwargs)
+
+
+class CosmologyCalculator(Cosmology):
```
beckermr, line 938: Why are we not allowing mu-Sigma MG for the calculator?

pyccl/core.py
beckermr: NVM. This matches what we did before.

src/ccl_correlation.c
```diff
@@ -518,28 +525,20 @@ INPUT:  cosmology, scale factor a, beta (= growth rate / bias),
...
-      "ccl_correlation.c: ccl_correlation_multipole(): non-linear power spctrum has not been computed!");
-    return;
-  }
-
   N_ARR = (int)(cosmo->spline_params.N_K_3DCOR * log10(cosmo->spline_params.K_MAX / cosmo->spline_params.K_MIN));
 
   k_arr = ccl_log_spacing(cosmo->spline_params.K_MIN, cosmo->spline_params.K_MAX, N_ARR);
   if (k_arr == NULL) {
     *status = CCL_ERROR_MEMORY;
     strcpy(cosmo->status_message,
```
beckermr, line 540: These strcpy calls should use the set status message function.

src/ccl_musigma.c
```diff
@@ -184,85 +128,187 @@ void ccl_cosmology_spline_linpower_musigma(ccl_cosmology* cosmo, ccl_f2d_t *psp,
...
+      sigma8 = ccl_sigma8(cosmo, psp, status);
 
-    if (*status == 0) {
-      // Calculate normalization factor using computed value of sigma8, then
-      // recompute P(k, a) using this normalization
-      log_sigma8 = 2*(log(cosmo->params.sigma8) - log(sigma8));
-      for(int j = 0; j<na*nk; j++)
-        lpk_ln[j] += log_sigma8;
-    }
+      if (*status == 0) {
```
beckermr, line 229: There is a lot of somewhat duplicated spline resetting code here. It might be good to add an extra function in the spline class that does all of this there to avoid duplication and make long-term changes to the spline stuff easier.

src/ccl_background.c
beckermr: I think this check is done in python. Can you add a comment here to that effect assuming I understood the code correctly?

pyccl/core.py
```diff
@@ -760,51 +704,37 @@ def _compute_linear_power_internal(self):
...
-                      cosmo=None)
-
-        psp = pk_lin.psp
+        # Populate power spectrum splines
+        trf = self._config_init_kwargs['transfer_function']
+        pk = None
+        rescale_s8 = True
+        rescale_mg = True
+        if trf == 'boltzmann_class':
+            pk = get_class_pk_lin(self)
```
beckermr, line 713: IIRC you should not need to have rescale_s8 = True for any of the boltzmann code options. The python wrappers all do this already. We should double-check before removing this bit of code.

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
...
+            transfer_function=transfer_function,
+            matter_power_spectrum=matter_power_spectrum)
+
+        # Parse arrays
+        has_bg = background is not None
+        has_dz = growth is not None
+        has_pklin = pk_linear is not None
+        has_pknl = pk_nonlin is not None
+
+        # Background
```
beckermr, line 1065: This init function is giant. Let's break out each of the background, growth, lin and nonlin parts into separate methods.

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
...
+            a = background['a']
+            chi = background['chi']
+            hoh0 = background['h_over_h0']
+            # Check that input arrays have the same size.
+            if not (a.shape == chi.shape == hoh0.shape):
+                raise ValueError("Input arrays must have the same size.")
+            # Check that `a` is a monotonically increasing array.
+            if not np.array_equal(a, np.sort(a)):
+                raise ValueError("Input scale factor array is not "
+                                 "monotonically increasing.")
```
beckermr, line 1082: Why not sort and reorder the other arrays?

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
...
+                # Initialize and store
+                pk = Pk2D(pkfunc=None,
+                          a_arr=pk_linear['a'], lk_arr=lk, pk_arr=pk,
+                          is_logp=use_log, extrap_order_lok=1,
+                          extrap_order_hik=2, cosmo=None)
+                self._pk_lin[n] = pk
+            # Set linear power spectrum as initialized
+            self._has_pk_lin = True
+
+        # Linear power spectrum
```
beckermr, line 1160: Comment is wrong.

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
...
+                qs = n.split('_x_')
+                if len(qs) != 2:
+                    raise ValueError("Power spectrum label %s could " % n +
+                                     "not be parsed. Label must be of the " +
+                                     "form 'q1_x_q2'")
+                pk = pk_linear[n]
+                if pk.shape != (na, nk):
+                    raise ValueError("Power spectrum %s has shape " % n +
+                                     str(pk.shape) + " but shape " +
+                                     "(%d, %d) was expected." % (na, nk))
```
beckermr, line 1146: Did we want to check that a and k are properly ordered here?

pyccl/core.py
```diff
@@ -21,13 +22,15 @@
...
+    'calculator': lib.pklin_from_input
 }
 
 matter_power_spectrum_types = {
     'halo_model': lib.halo_model,
     'halofit': lib.halofit,
     'linear': lib.linear,
-    'emu': lib.emu
+    'emu': lib.emu,
+    'calculator': lib.pklin_from_input
```
beckermr, line 33: This line is wrong. It should be `pknl_from_input`.

pyccl/core.py
```diff
@@ -1148,3 +933,278 @@ def __init__(self, **kwargs):
...
+            T_CMB=None, background=None, growth=None,
+            pk_linear=None, pk_nonlin=None, use_halofit=False):
+        if pk_linear:
+            transfer_function = 'calculator'
+        else:
+            transfer_function = 'boltzmann_camb'
+        if pk_nonlin or use_halofit:
+            matter_power_spectrum = 'calculator'
+        else:
+            matter_power_spectrum = 'halofit'
```
beckermr, line 1047: The API in this if else block is confusing to say the least. If I only pass pk_linear, then use_halofit being set or not won't matter because the base cosmology object will always apply halofit?

beckermr (changes requested):
Nice job!

OK. I've done my best. Besides some small questions and a little bit of cruft in the C code that needs fixing, my main question is how the new API works with respect to the various halofit + linear power spectrum options. I left comments on the code in various spots on this, but some additional general explanation and examples would be great!

damonge:
Thanks a lot @beckermr, I really appreciate this (especially given your other commitments and the size of this PR). I'll address all this ASAP.

pyccl/tests/test_background.py
damonge: It should have been unnormalized all along. Otherwise, there is no way to make the `growth_factor_unnorm` function could work with normalized input arrays (since the normalization is lost).

pyccl/tests/test_background.py
damonge: As above

pyccl/tests/test_pk2d.py
damonge: No, this works for any model, regardless of what was chosen for the cosmology (i.e. you could use `transfer_function="boltzmann_class"` and then compute the BBKS power spectrum with this method. I just make the `transfer_function` flag coincide here so I can compare both power spectra further down. I can make this explicit if needed.

include/ccl_neutrinos.h
damonge: This function is not used anywhere else, so I made it static and removed it from the `.h` file.

pyccl/ccl_power.i
damonge: By this: https://github.com/LSSTDESC/CCL/blob/6dcd05f7c3990dd7706f58d6996b1c86e3965639/pyccl/power.py#L62

pyccl/pk2d.py
damonge: Humm, I wrote evaluators [here](https://github.com/LSSTDESC/CCL/blob/6dcd05f7c3990dd7706f58d6996b1c86e3965639/pyccl/power.py#L8) and [here](https://github.com/LSSTDESC/CCL/blob/6dcd05f7c3990dd7706f58d6996b1c86e3965639/pyccl/power.py#L28). Do you think we need something that returns the Pk2D object itself? Not sure if that'd be needed too often.

damonge: On second thought, I see what you mean. Let me write those.

pyccl/core.py
damonge: Is that right? I thought not, based on [this comment](https://github.com/LSSTDESC/CCL/blob/6dcd05f7c3990dd7706f58d6996b1c86e3965639/pyccl/boltzmann.py#L59).

beckermr: ACK. you are right! NVM!

src/ccl_mass_conversion.c
damonge: Humm, `convert_concentration_single` uses an NR solver to translate between mass definitions, and this checks if the function has failed (which only happens if the NR solver fails).

beckermr: ahh ok. nvm then!

pyccl/core.py
damonge: Fixed

pyccl/core.py
damonge: Fixed

src/ccl_correlation.c
damonge: Done. I caught a few other of these in other files while I was at it.

src/ccl_background.c
damonge: Yep, that's right. Added.

src/ccl_utils.c
damonge: Done

src/ccl_massfunc.c
damonge: Done

src/ccl_massfunc.c
damonge: Done

src/ccl_massfunc.c
damonge: Done

src/ccl_f1d.c
damonge: Done

src/ccl_f1d.c
damonge: Done. I think at some point I changed my emacs from tabs to spaces, so when I modify old code things don't match anymore (although it looks right on the editor).

src/ccl_f1d.c
damonge: Done

pyccl/pk2d.py
damonge: Bummer :-P
Yep, changed to `apply_halofit`

damonge:
OK, while I'm addressing all the other ones, let me explain what I envisaged the API to do:
- If `background is None`, then the distance-redshift relation is computed by CCL itself. Otherwise it's loaded from the corresponding arrays.
- Same thing with `growth`.
- If `pk_linear is None`, then this `Cosmology` does not have a linear power spectrum (and an exception will be thrown if you try to calculate it). Otherwise, it's loaded from the corresponding arrays, one of which must be `delta_matter_x_delta_matter`.
- If `pk_nonlin is None` **and** `use_halofit == False`, then this `Cosmology` does not have a non-linear power spectrum (and an exception will be thrown if you try to calculate it). If `pk_nonlin is not None`, then the non-linear power spectra in that dictionary will be loaded into this `Cosmology`. If `use_halofit == False`, then one of those power spectra must be `delta_matter_x_delta_matter` (since that non-linear power spectrum must always exist).
- If `use_halofit == True`, then all the power spectra contained in `pk_linear` but not contained in `pk_nonlin` will be halofit-transformed and loaded into the `Cosmology` as non-linear power spectra. If `pk_nonlin` was `None`, that means all of the linear power spectra will be transformed (which will include `delta_matter_x_delta_matter`).

Question number 1 is: is this too complicated? If so, we could consider scrapping `use_halofit` altogether: if you use the calculator mode, you're responsible for providing any linear and non-linear power spectra you want to use with CCL.

Question number 2 is a bit more involved: do we envisage CCL eventually superseding CAMB/CLASS in terms of non-linear power spectrum prescriptions? This is already the case with `emu`, for instance, but CAMB/CLASS win in terms of hmcode and other methods. In that case, we may want to replace `use_halofit` with `nonlin_model`, where `nonlin_model` is a dictionary with items of the form `{'a_x_b': 'model'}`, with `'model'` specifying which of the CCL-supported non-linear models should be used with the `'a_x_b'` power spectrum if it's not present in `pk_nonlin`.

pyccl/tests/test_power.py
damonge: See my comment/questions below.

pyccl/tests/test_power.py
damonge: Humm, what do you mean? This should not fail. Do you mean testing that `x_x_y` in particular does not fail?

pyccl/tests/test_power.py
damonge: See my comment/questions about the API below. The idea is that this needs to eventually be present, so if `use_halofit = False`, `pk_nonlin` should contain `delta_matter_x_delta_matter`.

pyccl/core.py
damonge: Do you mean do that ourselves rather than force users to pass ordered arrays? I'd rather do the latter...

pyccl/core.py
damonge: do you mean check that the arrays are sorted? Or check if `pk.shape == (nk, na)` and, if so, flip it?

pyccl/core.py
damonge: See comments below

pyccl/core.py
damonge: Done

pyccl/tests/test_power.py
damonge: Fixed

pyccl/tests/test_pk2d.py
damonge: I've made this explicitly clear.

pyccl/pk2d.py
damonge: Ok, I've created them [here](https://github.com/LSSTDESC/CCL/blob/e8883f969a67809c38381fffed47391f54280024/pyccl/core.py#L874) and [here](https://github.com/LSSTDESC/CCL/blob/e8883f969a67809c38381fffed47391f54280024/pyccl/core.py#L893).

pyccl/core.py
damonge: My idea was that the calculator mode is intended for you to construct your own fancy non-standard cosmology from arrays. The main aim of the cosmological parameters needed to initialize it is that certain things (mostly the halo model) require knowledge of what Omega_M and h are. If you want to use CCL's mu-Sigma, isitgr etc., just use the standard CCL cosmology.

src/ccl_musigma.c
damonge: I've streamlined this function so the resplining only happens once. The previous version was still relying too much on the old code. May be worth looking at the file again rather than looking at the diff... :-(

src/ccl_background.c
damonge: Tested [here](https://github.com/LSSTDESC/CCL/blob/30138e2bc6c13d01f72c3bb6176d1968c62ade68/pyccl/tests/test_background.py#L157) now

pyccl/core.py
```diff
@@ -871,6 +871,44 @@ def compute_sigma(self):
...
+            name (:obj:`str` or `None`): name of the power spectrum to
+                return. If `None`, `'delta_matter_x_delta_matter'` will
+                be used.
+
+        Returns:
+            :class:`~pyccl.pk2d.Pk2D` object containing the linear
+            power spectrum with name `name`.
+        """
+        if name is None:
+            name = 'delta_matter_x_delta_matter'
```
beckermr, line 888: You should make this the default in the function signature.

pyccl/core.py
```diff
@@ -871,6 +871,44 @@ def compute_sigma(self):
...
+            name (:obj:`str` or `None`): name of the power spectrum to
+                return. If `None`, `'delta_matter_x_delta_matter'` will
+                be used.
+
+        Returns:
+            :class:`~pyccl.pk2d.Pk2D` object containing the non-linear
+            power spectrum with name `name`.
+        """
+        if name is None:
+            name = 'delta_matter_x_delta_matter'
```
beckermr, line 907: You should make this the default in the function signature.

pyccl/core.py
damonge: meaning `name='delta_matter_x_delta_matter'` as default value?

pyccl/core.py
beckermr: You will need to keep this if statement as well, which is fine too. Same for above.

pyccl/core.py
beckermr: yup.

damonge: ok, done

pyccl/core.py
damonge: Ok done

beckermr:
> Question number 1 is: is this too complicated? If so, we could consider scrapping use_halofit altogether: if you use the calculator mode, you're responsible for providing any linear and non-linear power spectra you want to use with CCL.

I think API above is fine, except for my comments on it below around `use_halofit`.

> Question number 2 is a bit more involved: do we envisage CCL eventually superseding CAMB/CLASS in terms of non-linear power spectrum prescriptions? This is already the case with emu, for instance, but CAMB/CLASS win in terms of hmcode and other methods. In that case, we may want to replace use_halofit with nonlin_model, where nonlin_model is a dictionary with items of the form {'a_x_b': 'model'}, with 'model' specifying which of the CCL-supported non-linear models should be used with the 'a_x_b' power spectrum if it's not present in pk_nonlin.

So I think you are right that we should leave things a bit more general. Here is my proposal. Replace the `use_halofit` keyword with something like `nonlin_perscription` (but a better name) which can take the following options

- `None`: do not compute nonlinear power for anything (default)
- `'halofit'`: applies halofit to anything without a nonlin power spectrum
- dict mapping linear power name (e.g. `a_x_c`) to a string (e.g. `'halofit'`) or `None` (with the string or None being interpreted per power spectrum as in the last two items)

Then we can add `'hmcode'` or w/e later and things just work as expected. This is also more customizable which I think might be useful later.

pyccl/core.py
beckermr: Both maybe?

pyccl/core.py
beckermr: I did mean do it outselves but I am happy to go with your proposal too.

pyccl/core.py
beckermr: nvm per discussion above.

src/ccl_musigma.c
beckermr: looks good!

pyccl/tests/test_power.py
beckermr: Yup. But I saw later you split the string on `_x_`. Have you put `blah_x_x_y` through the code?

beckermr: Or maybe `blah_x_x_x`?

pyccl/core.py
beckermr: Fair enough. FWIW, most of the prescriptions for LSS quantities that need these other parameters are calibrated assuming LCDM. Thus I think the main use for these params is in things like computing 2pt functions etc. This is fine OFC, just something for users to be mindful of.

pyccl/tests/test_power.py
damonge: Ok, good point. I wonder if we should make the separator `:` instead of `_x_`. These issues should not happen then, since it's a single character that is unlikely to appear on observable names.

beckermr: yep that would work

damonge:
OK, sounds good, done. Have a look at the latest commit. `use_halofit` is now `nonlinear_model`. Lots of `if` statements, but I think it does what we want it to.

pyccl/tests/test_power.py
damonge: Done!

pyccl/tests/test_power.py
```diff
@@ -141,14 +141,14 @@ def test_sigmaR_smoke(r):
     1.0,
     [0.3, 0.5, 10],
     np.array([0.3, 0.5, 10])])
-def test_sigmaV_smoke(r):
+def atest_sigmaV_smoke(r):
```
beckermr: why the change from test to atest?

pyccl/tests/test_power.py
```diff
@@ -187,89 +211,364 @@ def test_power_sigma8norm_norms_consistent(tf):
...
+
+    # Non-parsable power spectrum
     with pytest.raises(ValueError):
-        cosmo_input._set_nonlin_power_from_arrays()
+        ccl.CosmologyCalculator(
+            Omega_c=0.27, Omega_b=0.05, h=0.7,
+            n_s=0.965, sigma8=0.8,
+            pk_nonlin={'a': a_arr, 'k': k_arr,
+                       'delta_matter:delta_matter': pk_arr,
+                       'a_y_b': pk_arr})
```
beckermr, line 545: should this be `:`?

beckermr:
Besides the failing tests, this basically looks close to merging. One more comment above.

pyccl/tests/test_power.py
damonge: We want to test that it fails. Changed it to `;`

beckermr: ahhhh great!

beckermr (approved):

beckermr:
LGTM! Really nice work on this one!

damonge:
Thanks a lot! I'll wait a couple of days in case others want to comment.

slosar:
This looks really great. While I stuff myself with unhealthy stuff, damonge codes away! :)
I think it would be great to have some tutorial notebook to explain how this is all used in practice. Ideally written by someone else based on the discussion above so that we see if it actually makes sense for an average user. (but this is independent of this PR).

c-d-leonard (commented):
Looks great @damonge, thank you so much! I've left a couple of small comments which are more questions for my own understanding than issues.

When I ran the benchmark tests locally I found that several of the benchmarks/test_emu.py tests throw a warning about how we don't compute the linear growth rate in models with massive neutrinos. This doesn't happen on master. It would be good to figure out if this is deliberate and if so catch the warning so it doesn't who up when the tests are run.

c-d-leonard:
> This looks really great. While I stuff myself with unhealthy stuff, damonge codes away! :)
> I think it would be great to have some tutorial notebook to explain how this is all used in practice. Ideally written by someone else based on the discussion above so that we see if it actually makes sense for an average user. (but this is independent of this PR).

Also I 100% agree on this, a tutorial nobody would be great and having it not written by @damonge would probably be valuable. I'll see if we can get a volunteer.

src/ccl_fftlog.c
```diff
@@ -44,11 +44,6 @@ static double complex lngamma_fftlog(double complex z)
   return lnr.val + I*phi.val;
 }
 
-static double complex gamma_fftlog(double complex z)
-{
-  return cexp(lngamma_fftlog(z));
-}
-
```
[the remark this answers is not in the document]
damonge: we weren't using it for anything

src/ccl_power.c
```diff
@@ -12,19 +12,6 @@
...
-  for (int ii=0; ii < na; ii++) {
-    double a = a_arr[ii];
-    for(int jj=0; jj < nk; jj++) {
-      double k = exp(lk_arr[jj]);
-      double fbcm = ccl_bcm_model_fka(cosmo, k, a, status);
-      pk2d[jj+nk*ii] += log(fbcm);
-    }
-  }
-}
-
```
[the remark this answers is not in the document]
damonge: moved to a proper function in `ccl_bcm.c`

pyccl/power.py
```diff
@@ -1,7 +1,47 @@
...
+            compute (which should be stored in `cosmo`). Defaults to
+            the non-linear matter power spectrum.
+
+    Returns:
+        float or array_like: Non-linear power spectrum.
+    """
+    cosmo.compute_nonlin_power()
+    if p_of_k_a not in cosmo._pk_nl:
+        raise KeyError("Power spectrum %s unknown" % p_of_k_a)
+    return cosmo._pk_nl[p_of_k_a].eval(k, a, cosmo)
```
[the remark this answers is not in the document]
damonge: because these functions are not meant to return the **matter** power spectrum (i.e. they don't have the word `matter` in the name).

damonge:
@c-d-leonard Thanks! The warning is gone.

pyccl/power.py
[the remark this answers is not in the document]
c-d-leonard: ah okay right thanks.

src/ccl_fftlog.c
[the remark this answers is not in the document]
c-d-leonard: good reason :)

src/ccl_power.c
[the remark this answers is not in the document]
c-d-leonard: ah right thanks

c-d-leonard:
Awesome - ready to merge then from my perspective!

beckermr:
Merge it!!!!

damonge:
So much pressure...
Thanks guys!

PR #844, opened 2020-12-24
Timing benchmark

damonge:
This PR adds a benchmark test that ensures the time to compute the SRD data vector does not exceed a maximum threshold.
Useful to make sure new code additions don't make the code much slower.


damonge:
This was actually merged with #843

Issue #817, opened 2020-08-07
read_yaml 


empEvil:
The current yaml reader for cosmology information, is not optimal. An example is that if you don't define `sigma8: 'nan'` then the reader will crash, if you are using `A_s` instead. It also can't seem to just load in the default values for a bunch of other parameters, `bcm_log10Mc`, `bcm_etab` e.g.

joezuntz:
The yaml functions were things I added as a quick fix for things I wanted to do in TXPipe.  They're not at all flexible or generic, and are due for a revamp.

damonge:
Closed by #852

Issue #842, opened 2020-12-22
Deprecate `z_mg` and `df_mg`


damonge:
Has anyone ever used these parameters? Otherwise I think they should go.
https://github.com/LSSTDESC/CCL/blob/593ed60ce7d9763590856fd253e6683ca18269f5/pyccl/core.py#L131

mishakb:
I think no-one ever used these. I agree they should probably go.

c-d-leonard:
I never use them and I don't know of anyone who has. This was our first attempt at including modified gravity as I recall - we may have outgrown it. @TessaBaker @gvalogia any thoughts? (These parameters allow a general non-standard growth factor to be passed as an array - but I think this functionality is preserved in generality in @damonge's new 'Calculator Mode' which should be merged shortly, right David?)

TessaBaker:
I haven't used these either. I think having a modified growth factor option is still desirable, but sounds like David has it covered?

damonge:
Yes, I think keeping these now is superfluous. The calculator mode allows a lot more freedom, and losing these parameters would simplify the API in the next version significantly.

c-d-leonard:
Cool, let's get rid of them!

gvalogia:
Yes, I also don't think this feature is necessary. I can see the logic behind this, I haven't seen anyone using this parametrization recently.

PR #852, opened 2021-02-15
YAML writing in python

tilmantroester:
This PR removes the C-level yaml reader and writer routines. The reader wasn't being used and the writer has now been replaced by a python level reader. This makes it easier to add new parameters in the future. 
The C-level writer wasn't consistent anymore in any case (e.g., C-level wrote `z_mgrowth`, while Python-level read `z_mg`).

The PR also allows file pointers and streams to be used instead of file names. This enables passing around information without need to write to disk.

Removed `df_mg` and `z_mg` from the yaml reader, in anticipation of their eventual removal (#842).


pyccl/core.py
```diff
@@ -229,27 +229,53 @@ def write_yaml(self, filename):
...
-            params = yaml.load(fp, Loader=yaml.Loader)
+        if isinstance(filename, str):
+            with open(filename, 'r') as fp:
+                params = yaml.load(fp, Loader=yaml.Loader)
+        else:
+            params = yaml.load(filename, Loader=yaml.Loader)
+
+        # Check for missing values in yaml files created with the C-level
+        # yaml dumper
+        if any([p not in params for p in ["m_nu", "m_nu_type"]]):
```
damonge, line 272: Is this the only case that needs to be checked?

damonge (commented):
So nice to see unneeded code go. thanks! Just one comment, and I'll approve as soon as the tests pass

pyccl/core.py
tilmantroester: Technically `z_mg` and `df_mg` as well but I don't think that had ever worked (C-level calls them `z_mgrowth`). The new MG parameters (`c1_mg`, etc) don't have a fall-back either, so the readers already fail on yaml files created with old versions of CCL. 
So I'm tempted to replace that check with just a check for a `KeyError` and give a warning if any parameters are missing.

tilmantroester: The current version should handle this more cleanly and robustly.

tilmantroester:
This should now handle missing values much better and address #817.

damonge (approved):
LGTM, but would be good to check if this messes up any of what @joezuntz has been doing

joezuntz:
This is fine from my end - we only use the python interface.  I might need to regenerate one file but that's all.

damonge:
Closes #817

PR #854, opened 2021-02-15
Expose nonlinear CAMB power spectra

tilmantroester:
Redo of PR #837. Much simpler thanks to David's calculator mode and yaml handling in python.

Currently CCL only uses the linear power spectrum prediction from Boltzmann codes, with limited options for nonlinear power spectra. Boltzmann codes provide a range of fast nonlinear models that can be used. For example, CAMB provides an extensive list of codes, among them HMCode.

This PR uses the HMCode implementations in CAMB to provide nonlinear predictions for the matter power spectrum, including modelling of baryons, for CCL. It addresses issue #819.  

The nonlinear power spectra are accessed by setting `matter_power_spectrum="camb"`. To specify parameters, I added a generic `extra_parameters` kwarg to `Cosmology.__init__`. This is meant to take a dictionary and act as a catch-all for more specialised parameters, in order to avoid further bloating of the call signature of `Cosmology.__init__`.

Serialisation seems to work out of the box. 

I added a test comparing an HMCode power spectrum from CAMB to that of CCL. I get agreement at the 2e-3 level, which might be due to interpolation effects and subtleties in how CCL calls CAMB (there's a lot of fiddling with the neutrinos going on in `get_camb_pk_lin`).

Example:
```python
ccl_cosmo = ccl.Cosmology(
    Omega_c=Omega_c, Omega_b=Omega_b, h=h, m_nu=0.0,
    A_s=A_s, n_s=n_s,
    transfer_function="boltzmann_camb",
    matter_power_spectrum="camb",
    extra_parameters={"camb":
         {"halofit_version": "mead2020",
          "HMCode_logT_AGN": logT_AGN, # Feedback strength
          "lmax" : 2600 # Also allow access to some other CAMB parameters for good measure
         }})

pk_nonlin_ccl = ccl.nonlin_matter_power(ccl_cosmo, k, 1.0)
```

### DIFF
--- a/pyccl/boltzmann.py
+++ b/pyccl/boltzmann.py
@@ -35,9 +35,9 @@ def get_camb_pk_lin(cosmo, nonlin=False):
             non-linear power spectrum as well.
 
     Returns:
-        :class:`~pyccl.pk2d.Pk2D`: Power spectrum \
-            object. The linear power spectrum. If `nonlin=True`, returns a
-            tuple (pk_lin, pk_nonlin).
+        :class:`~pyccl.pk2d.Pk2D`: Power spectrum object. The linear power \
+            spectrum. If ``nonlin=True``, returns a tuple \
+            ``(pk_lin, pk_nonlin)``.
     """
 
     # Comment from Jarvis: TODO clean up this and other assert
--- a/pyccl/boltzmann.py
+++ b/pyccl/boltzmann.py
@@ -49,8 +49,7 @@ def get_camb_pk_lin(cosmo, nonlin=False):
     # Get extra CAMB parameters that were specified
     extra_camb_params = {}
     try:
-        extra_camb_params = \
-            cosmo._params_init_kwargs["extra_parameters"]["camb"]
+        extra_camb_params = cosmo["extra_parameters"]["camb"]
     except (KeyError, TypeError):
         pass
 
--- a/pyccl/boltzmann.py
+++ b/pyccl/boltzmann.py
@@ -24,7 +24,7 @@ from .pk2d import Pk2D
 from .errors import CCLError
 
 
-def get_camb_pk_lin(cosmo):
+def get_camb_pk_lin(cosmo, nonlin=False):
     """Run CAMB and return the linear power spectrum.
 
     Args:
@@ -42,6 +42,14 @@ def get_camb_pk_lin(cosmo):
     assert HAVE_CAMB, (
         "You must have the `camb` python package "
         "installed to run CCL with CAMB!")
+
+    # Get extra CAMB parameters that were specified
+    extra_camb_params = {}
+    try:
+        extra_camb_params = \
+            cosmo._params_init_kwargs["extra_parameters"]["camb"]
+    except (KeyError, TypeError):
+        pass
 
     # z sampling from CCL parameters
     na = lib.get_pk_spline_na(cosmo.cosmo)
@@ -136,14 +144,25 @@ def get_camb_pk_lin(cosmo):
         w=cosmo['w0'],
         wa=cosmo['wa']
     )
-    # cp.set_cosmology()
+
+    if nonlin:
+        cp.NonLinearModel = camb.nonlinear.Halofit()
+        halofit_version = extra_camb_params.get("halofit_version", "mead")
+        options = {k: extra_camb_params[k] for k in
+                   ["HMCode_A_baryon",
+                    "HMCode_eta_baryon",
+                    "HMCode_logT_AGN"] if k in extra_camb_params}
+        cp.NonLinearModel.set_params(halofit_version=halofit_version,
+                                     **options)
+
     cp.set_matter_power(
         redshifts=[_z for _z in zs],
-        kmax=10,
-        nonlinear=False)
-    assert cp.NonLinear == camb.model.NonLinear_none
+        kmax=extra_camb_params.get("kmax", 10.0),
+        nonlinear=nonlin)
+    if not nonlin:
+        assert cp.NonLinear == camb.model.NonLinear_none
 
-    cp.set_for_lmax(5000)
+    cp.set_for_lmax(extra_camb_params.get("lmax", 5000))
     cp.InitPower.set_params(
         As=A_s_fid,
         ns=cosmo['n_s'])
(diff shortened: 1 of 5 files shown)


damonge:
I haven't looked at the code yet but I'm in general in favour of the idea. However I know @beckermr has strong feelings about kwargs, so might be good to hear his thoughts ;-)

beckermr:
This API seems fine.

tilmantroester:
Would it make sense to have separate `extra_parameters` and `extra_config`? The latter could be used for things like internal accuracy parameters.

I've now made it so that the configuration parameters are also serialised. I assume the fact that the config options (e.g., `transfer_function`) were not being serialised was due to limitations of the C-level dumper. But do let me know if there were other reasons for that design choice (@joezuntz, @beckermr).

From the user perspective, there's now little distinction between the parameters and configuration, so a single `extra_parameters` might be sufficient.

beckermr:
Yes let's just have one random bag of extra parameters, not two. 

Thanks for fixing the serialization.

pyccl/tests/test_nonlin_camb_power.py
```diff
@@ -0,0 +1,53 @@
...
+    # import matplotlib.pyplot as plt
+
+    # plt.loglog(k, pk_nonlin_camb[0])
+    # plt.loglog(k, pk_nonlin_ccl)
+    # plt.semilogx(k, pk_nonlin_ccl/pk_nonlin_camb[0]-1)
+    # plt.show()
+    assert np.allclose(pk_nonlin_camb, pk_nonlin_ccl, rtol=3e-3)
+
+
+# if __name__ == "__main__":
```
damonge, line 52: remove comments

pyccl/tests/test_nonlin_camb_power.py
```diff
@@ -0,0 +1,53 @@
...
+    ccl_cosmo = ccl.Cosmology(Omega_c=Omega_c, Omega_b=Omega_b, h=h, m_nu=0.0,
+                              A_s=A_s, n_s=n_s,
+                              transfer_function="boltzmann_camb",
+                              matter_power_spectrum="camb",
+                              extra_parameters={"camb":
+                                                {"halofit_version": "mead2020",
+                                                 "HMCode_logT_AGN": logT_AGN}})
+    pk_nonlin_ccl = ccl.nonlin_matter_power(ccl_cosmo, k, 1.0)
+
+    # import matplotlib.pyplot as plt
```
damonge, line 43: remove comments

pyccl/core.py
```diff
@@ -747,7 +758,31 @@ def compute_linear_power(self):
...
+                pk_nl_from_camb = True
+            pk = get_camb_pk_lin(self, nonlin=pk_nl_from_camb)
+            if pk_nl_from_camb:
+                pk, pk_nl = pk
+                self._pk_nl['delta_matter:delta_matter'] = pk_nl
+                self._has_pk_nl = True
+                rescale_mg = False
+                rescale_s8 = False
+                if abs(self["mu_0"]) > 1e-14:
+                    warnings.warn("You want to compute the non-linear power "
```
damonge, line 772: Humm, is this right? I think we didn't have this warning before, right?

pyccl/core.py
```diff
@@ -164,13 +167,15 @@ class Cosmology(object):
...
-        emulator_neutrinos: `str`, optional): If using the emulator for
+        emulator_neutrinos (:obj: `str`, optional): If using the emulator for
             the power spectrum, specified treatment of unequal neutrinos.
             Options are 'strict', which will raise an error and quit if the
             user fails to pass either a set of three equal masses or a sum with
             m_nu_type = 'equal', and 'equalize', which will redistribute
             masses to be equal right before calling the emulator but results in
             internal inconsistencies. Defaults to 'strict'.
+        extra_parameters (:obj:`dict`, optional): Dictionary holding extra
+            parameters. Defaults to None.
```
damonge, line 178: Maybe explain that these will be understood by CAMB (and we may update to also CLASS later on)

pyccl/core.py
```diff
@@ -747,7 +758,31 @@ def compute_linear_power(self):
...
+                rescale_s8 = False
+                if abs(self["mu_0"]) > 1e-14:
+                    warnings.warn("You want to compute the non-linear power "
+                                  "spectrum using CAMB. This cannot be "
+                                  "consistently done with mu_0 > 0.",
+                                  category=CCLWarning)
+                if np.isfinite(self["sigma8"]) \
+                        and not np.isfinite(self["A_s"]):
+                    if abs(self["sigma8"] - sigma8(self, pk)) > 1e-6:
+                        warnings.warn("You want to compute the non-linear "
```
damonge, line 779: OK, so with this, essentially it's not really practical to use `sigma8` as an input parameter if you're going to calculate both linear and non-linear power spectra with CAMB, right?
In that case, should we throw an exception instead of a warning?

pyccl/tests/test_nonlin_camb_power.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # plt.loglog(k, pk_nonlin_camb[0])
+    # plt.loglog(k, pk_nonlin_ccl)
+    # plt.semilogx(k, pk_nonlin_ccl/pk_nonlin_camb[0]-1)
+    # plt.show()
+    assert np.allclose(pk_nonlin_camb, pk_nonlin_ccl, rtol=3e-3)
```
damonge, line 49: Can we check at two different redshifts? I think 3E-3 is OK, considering that we're computing our own halofit in one case.

damonge (changes requested):
Thanks @tilmantroester . A few comments below

pyccl/core.py
tilmantroester: This is usually checked in `compute_nonlin_power` (https://github.com/LSSTDESC/CCL/blob/master/pyccl/core.py#L817) but since `has_nonlin_power` is set to `True`, that check needs to be done earlier.

damonge: Sorry, you're right. I read too quickly and thought this was about neutrinos.

pyccl/core.py
tilmantroester: Yeah, that might be better. I thought the pre-camb rescaling might work but I just checked and it does not.

pyccl/boltzmann.py
```diff
@@ -43,6 +46,14 @@ def get_camb_pk_lin(cosmo):
         "You must have the `camb` python package "
         "installed to run CCL with CAMB!")
 
+    # Get extra CAMB parameters that were specified
+    extra_camb_params = {}
+    try:
+        extra_camb_params = \
+            cosmo._params_init_kwargs["extra_parameters"]["camb"]
```
beckermr, line 53: Anything with an underscore (`_`) in front of it is considered private. So in this case, the Cosmology class needs to add a public API that exposes this information. We should not be calling things that are considered private, even from our own code.

pyccl/core.py
beckermr: Yeah you'd have to call camb twice.

pyccl/boltzmann.py
tilmantroester: Now exposed through `__getitem__`, like the other parameters.

pyccl/tests/test_nonlin_camb_power.py
tilmantroester: Done.

pyccl/tests/test_nonlin_camb_power.py
tilmantroester: Done.

pyccl/tests/test_nonlin_camb_power.py
tilmantroester: Done.

pyccl/core.py
tilmantroester: Now changed to throwing an exception.

pyccl/core.py
tilmantroester: What markup format is being used for the docstrings? I'd to list the currently supported parameters but to so with sane formatting.

damonge: I'd say that's what the CAMB documentation is there for. Here I'd just say "these can currently include any parameters that can be parsed by CAMB, but will be extended in the future" or something like that.

pyccl/boltzmann.py
damonge: Looks good to me!

damonge (approved):
@tilmantroester this looks good to me once you add the extra sentence in the documentation. Once you do that, feel free to merge if the tests pass (which they will!)

beckermr (changes requested):
Please add a changelog entry.

beckermr (approved):

tilmantroester:
Thanks for reviewing @damonge @beckermr !

Issue #779, opened 2020-05-13
ERROR STOP Fluid dark energy model does not allow w crossing -1


Andromedanita:
I am using CCL installed on NERSC which is version 2.1.0 and when I try to calculate P(k, z) I get this error: 
> ERROR STOP Fluid dark energy model does not allow w crossing -1

which is due to CAMB using Fluid DE model instead of the PPF.

This is the cosmology I used:
cosmo =  ccl.Cosmology(Omega_c= 0.2139426, Omega_b=0.0441039, h=0.719, n_s=0.95, 
          sigma8=0.8, Omega_k=0.0, Omega_g=None, 
         w0=-1., wa=-0.01, transfer_function='boltzmann_camb', 
         matter_power_spectrum='halofit')

Is there an easy way to switch CAMB to use PPF model in CCL?

c-d-leonard:
@Andromedanita just wondering how you're getting on with this or if you need any support?

Andromedanita:
I was just wondering if it would be possible to get someone who knows C better than me on this?

damonge:
@Andromedanita don't know enough about CAMB. How would you normally ask CAMB to use PPF? It should be possible to pass that through python without touching the C code.

damonge:
And this would be the function to look at/modify: https://github.com/LSSTDESC/CCL/blob/641ba87226565f1f8c894f8d4570127c53fdec76/pyccl/boltzmann.py#L27

Andromedanita:
@damonge Thanks for your response. In CAMB itself there is a parameter in the param.ini file that we can set to choose which DE model to use. So I basically just set "dark_energy_model = PPF" instead of fluid which is the default option.
Would I need to add "cp.dark_energy_model = PPF" similar to line 77 of the code you mentioned?

Also, this might sound stupid but I am currently using CCL that is already installed on NERSC and I have had issues installing it locally before so I am not sure how I can test and use it.

damonge:
I think you'd need to modify [this line](https://github.com/LSSTDESC/CCL/blob/641ba87226565f1f8c894f8d4570127c53fdec76/pyccl/boltzmann.py#L133) specifically, and change it into `dark_energy_model=camb.dark_energy.DarkEnergyPPF`

Unfortunately you need to modify ccl for this, so you will need to install a new version yourself...

We can probably help with installation issues.

Andromedanita:
Okay thank you. I will try installing it again and ask for help.
[...]

PR #857, opened 2021-02-18
Camb dark energy model

t-ferreira:
I included an option to choose which type of dark energy you want when using camb. It defaults to "fluid", now one can easily set it to "ppf".

I simply set "self._params.DE_model_camb = DE_model_camb" because I was having a difficult time with the C files.

### DIFF
--- a/pyccl/tests/test_power.py
+++ b/pyccl/tests/test_power.py
@@ -581,3 +581,12 @@ def test_camb_de_model():
         extra_parameters={"camb": {"dark_energy_model": "pf"}})
     with pytest.raises(ValueError):
         ccl.linear_matter_power(cosmo, 1, 1)
+
+
+def test_camb_correct_de_model():
+    """Check that w is not less than -1, if the chosen dark energy model for
+    CAMB is fluid."""
+    cosmo = ccl.CosmologyVanillaLCDM(
+        transfer_function='boltzmann_camb', w0=-1, wa=-1)
+    with pytest.raises(ValueError):
+        ccl.linear_matter_power(cosmo, 1, 1)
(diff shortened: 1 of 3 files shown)


damonge:
@tilmantroester would your PR (#854) be indirectly addressing this? We want to avoid having too many different ways of doing the same thing

tilmantroester:
Yes, in the sense that the dark energy model could be specified in `extra_parameters["camb"]`.

t-ferreira:
@tilmantroester I could do that, if you wouldn't mind.

t-ferreira:
@tilmantroester I added a test to check for a w crossing -1, if ppf is not selected as the dark energy model.

pyccl/tests/test_power.py
```diff
@@ -572,3 +572,21 @@ def test_input_nonlin_raises():
...
+        ccl.linear_matter_power(cosmo, 1, 1)
+
+
+def test_camb_correct_de_model():
+    """Check that w is not less than -1, if the chosen dark energy model for
+    CAMB is fluid."""
+    cosmo = ccl.CosmologyVanillaLCDM(
+        transfer_function='boltzmann_camb', w0=-1, wa=-1)
+    with pytest.raises(ValueError):
+        ccl.linear_matter_power(cosmo, 1, 1)
```
tilmantroester, line 592: Apologies for the delay getting back to this. Maybe just add a
```python
cosmo = ccl.CosmologyVanillaLCDM(
        transfer_function='boltzmann_camb', w0=-1, wa=-1, extra_parameters={"camb": {"dark_energy_model": "ppf"}}
)
ccl.linear_matter_power(cosmo, 1, 1)
```
to check that the PPF version is running? That would directly check for the case of the issue that prompted this (#779).

c-d-leonard:
@tilmantroester could you have a look? looks like Tassia may have added the most recent check you asked about

tilmantroester:
Looks good to me!

tilmantroester (approved):

PR #861, opened 2021-02-24
ISW

damonge:
Implements an ISW tracer.

### DIFF
--- a/pyccl/__init__.py
+++ b/pyccl/__init__.py
@@ -42,7 +42,7 @@ from .massfunction import massfunc, halo_bias, massfunc_m2r
 
 # Cl's and tracers
 from .tracers import Tracer, NumberCountsTracer, WeakLensingTracer, CMBLensingTracer, \
-    tSZTracer, get_density_kernel, get_kappa_kernel, get_lensing_kernel
+    tSZTracer, ISWTracer, get_density_kernel, get_kappa_kernel, get_lensing_kernel
 from .cls import angular_cl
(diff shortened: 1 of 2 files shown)


c-d-leonard:
@damonge is this ready for review? missing benchmarks?

damonge:
Ready for review @c-d-leonard

c-d-leonard:
I will either review this or find someone to review it tomorrow.

damonge:
There is no rush. FYI, I couldn't find a proper implementation of this in CLASS (it'd be mixed up with a bunch of other low-redshift effects), so I implemented my own alternative version in python for a benchmark... Open to criticism on that front

c-d-leonard (commented):
This looks good to me. My only hesitation is the benchmark like you said. I think if we can it would be good to at least have someone else code it up just in case you made the same silly error twice. I'm happy to do it - it looks like it should be pretty straightforward? Or we can see if anyone else is keen or has their own code lying around. I'll check on the channel.

c-d-leonard:
Also when I run the tests locally I'm getting a failure on test_nonlin_camb_power to do with the keyword 'mead2020'. I updated my camb but still there - are you seeing that?

damonge:
Nope, but it seems to run fine on gha

c-d-leonard:
@pennalima Do I remember right that you said numcosmo can calculate the ISW power spectrum for a benchmark here? What we're looking for is equation 6 in 1710.03238.

Issue #863, opened 2021-03-11
CCL_ERROR_SPLINE in CalculatorCosmology when passing decreasing a


tilmantroester:
Passing the `a` array in `pk_nonlin` in decreasing order causes a CCL_ERROR_SPLINE:

```python
decreasing_a = np.linspace(0.1, 1.0, 100)[::-1]
k = np.logspace(-3, 1, 50)
dummy_Pk = np.ones((len(increasing_a), len(k)))

ccl_cosmo = ccl.CosmologyCalculator(Omega_c=0.25,
                                    Omega_b=0.05,
                                    A_s=2.1e-9,
                                    n_s=0.97,
                                    h=0.7,
                                    pk_nonlin={"a" : decreasing_a, 
                                               "k" : k, 
                                               "delta_matter:delta_matter": dummy_Pk})
```

Error:
```
---------------------------------------------------------------------------
CCLError                                  Traceback (most recent call last)
<ipython-input-51-c6c6bf0dac6e> in <module>
     10                                     pk_nonlin={"a" : decreasing_a, 
     11                                                "k" : k,
---> 12                                                "delta_matter:delta_matter": dummy_Pk})

~/Codes/miniconda/envs/analysis/lib/python3.7/site-packages/pyccl/core.py in __init__(self, Omega_c, Omega_b, h, n_s, sigma8, A_s, Omega_k, Omega_g, Neff, m_nu, m_nu_type, w0, wa, T_CMB, background, growth, pk_linear, pk_nonlin, nonlinear_model)
   1152             self._init_pklin(pk_linear)
   1153         if has_pknl:
-> 1154             self._init_pknl(pk_nonlin, has_nonlin_model)
   1155         self._apply_nonlinear_model(nonlinear_model)
   1156 

~/Codes/miniconda/envs/analysis/lib/python3.7/site-packages/pyccl/core.py in _init_pknl(self, pk_nonlin, has_nonlin_model)
   1285                       a_arr=pk_nonlin['a'], lk_arr=lk, pk_arr=pk,
   1286                       is_logp=use_log, extrap_order_lok=1,
-> 1287                       extrap_order_hik=2, cosmo=None)
   1288             self._pk_nl[n] = pk
   1289         # Set non-linear power spectrum as initialized

[...]

PR #864, opened 2021-03-11
Give useful error message when trying to pass decreasing a array to CalculatorCosmology

tilmantroester:
Addresses issue #863.

### DIFF
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -1276,6 +1276,11 @@ class CosmologyCalculator(Cosmology):
                              "and 'delta_matter:delta_matter' "
                              "(at least)")
 
+        # Check that `a` is a monotonically increasing array.
+        if not np.array_equal(pk_linear['a'], np.sort(pk_linear['a'])):
+            raise ValueError("Input scale factor array in `pk_linear` is not "
+                             "monotonically increasing.")
+
         # needed for high-z extrapolation
         self.compute_growth()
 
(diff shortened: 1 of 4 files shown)


tilmantroester:
I ran into another unhelpful error message when the tracer and power spectra have different a_max. I'll add a check for that as well.

damonge:
@tilmantroester did you want to add more changes? Otherwise this looks good to me.

tilmantroester:
@damonge I've now added a check in Pk2D as well.

I found that if the redshift ranges of the tracers and P(k) don't agree then this sometimes causes integration errors. Should that be left to the user to figure out or should this cause a warning?

damonge:
Humm, I think it should be up to the users to provide Pk arrays that cover the appropriate range of k/a for their purposes. But is the error message particularly obscure in this case?

tilmantroester:
It's a C-level integration error in `angular_cl` and should only occur when using custom P(k) arrays. I guess at that point users can be expected to check that their inputs are consistent.

damonge:
Maybe we could add a check at the C-level comparing the limits in k and a of the tracers and power spectra. I'm happy for that to be addressed somewhere else.

damonge (approved):
LGTM!

tilmantroester:
Thanks for the review!
I have a python-level check that I can polish if that issue comes up again.

Issue #865, opened 2021-03-17
Low accuracy on WeakLensingTracer at low redshifts


tilmantroester:
The heuristic that determines the support of the lensing kernel is overly aggressive. 
For the case of n(z) of the first tomographic bin of KiDS-1000, CCL determines a chi_min of ~200 Mpc. This causes >5% deviations in the Cls:

![lensing_tracer_chi_min](https://user-images.githubusercontent.com/26582252/111535444-03c8c980-8761-11eb-9667-5f43bfe6d149.png)

Forcing chi_min to a lower value by setting `WL_tracer_fix._trc[0].chi_min = 5.0` after the creation of the tracer recovers the result from cosmosis. I originally suspected cosmosis to be the culprit but the cosmosis result was robust against a number of tests and various changes in accuracy parameters. The behaviour of CCL when manually changing chi_min behaves as expected, so I am inclined to believe that CCL is the issue here.

If CCL is indeed the culprit, we might need to look into the benchmarks for the Cls, since the deviation is much larger than the advertised accuracy of CCL. This was likely introduced sometime in the last ~2 years, as this was not an issue when I ran similar tests back then. 

I will look into exposing `CCL_FRAC_RELEVANT`, which controls the support estimation, in `ccl_cl_tracer_t_new` and `WeakLensingTracer`.

damonge:
That's weird. By default chi_min should go to zero for anything involving weak lensing. Let me look into it. What does the N(z) look like, and do you have a code snippet?

damonge:
OK, I think I know what's happening. I bet the N(z) is poorly sampled (I guess on intervals Dz~0.05 or 0.1), right?
There are two solutions:
a) in your particular case, oversample the N(z) a bit.
b) In general, we can make sure CCL enforces chi_min = 0 for anything lensing related.

tilmantroester:
The n(z) is indeed sampled at dz=0.05, which is a common case though. 

When searching for the kernel support, currently the comoving distance `chi_min` is set to where the kernel has a value greater than some fraction of the maximum. For discontinuous or sparsely sampled kernels, there is no guarantee that the volume of the kernel between `chi_min` and `chi_min - Delta_chi` is small. 
I think an easy and general fix would be to take `chi_min` as the comoving distance _before_ the kernel gets larger than the threshold value:

```c
      // Find minimum
      for (ichi=0; ichi < n_w-1; ichi++) {
        if (fabs(w_w[ichi+1]) >= w_max) {
[...]

PR #866, opened 2021-03-18
Fix estimation of kernel support.

tilmantroester:
A more conservative estimate of the kernel support. Addresses #865.


damonge (approved):
LGTM!

damonge:
@tilmantroester did you check that this fixes the KiDS-specific error you noticed?

tilmantroester:
Yes, Cls look good now. The chi_min gets set to 0 now. Before it would get set to the first sampling point, which turned out to be z~0.05 ~ 200 Mpc:
lensing_tracer_chi_min

damonge:
Great, let's merge! Thanks a lot @tilmantroester

PR #868, opened 2021-04-07
Fix AttributeError in Tk3D.__del__

tilmantroester:
`Tk3D.__del__` calls `lib.f3d_t_free`, which however isn't exposed. In other, words a memory leak. Since this happens in `__del__` the `AttributeError` exception that is raised is usually ignored. It's not entirely clear to me why this isn't at least reported as a warning when pytest runs though. Maybe something funny about when `__del__` is called when going out of scope.

This should be fixed by including the relevant header file in `ccl_tk3d.i`.


damonge (approved):
great job tracking this Tilman!

PR #869, opened 2021-04-12
Draft implementation of Cov NG terms

carlosggarcia:
I have started the implementation of the missing terms in the Cov NG. So far I have implemented the `T^2h_13` term and checked that runs but haven't checked if the result makes sense. 

I'm following Eq. 3.25 of https://arxiv.org/pdf/1912.08209.pdf


pyccl/halos/profiles_3pt.py
```diff
@@ -0,0 +1,68 @@
+from .profiles import HaloProfile
+
+
+class Profile3pt(object):
```
damonge, line 4: So, although it is great to have a proper implementation of this, the annoying thing is that this is a very, very specific implementation, which does not correspond to a general 3-point function, but to the specific case <u(ku) v1(kv) v2(kv)>. For this reason I'm tempted to say that, for now, the 3,1 part of the connected trispectrum should just involve products of profiles rather than a general 3-point cumulant...

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
         out = np.squeeze(out, axis=-1)
     return out
 
+def halomod_trispectrum_2h_22(cosmo, hmc, k, a, prof1, p12_of_k_a, prof2=None,
```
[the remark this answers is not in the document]
carlosggarcia: Done in next commit

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+                              prof12_2pt=None, prof3=None, prof4=None,
```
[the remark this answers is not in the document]
carlosggarcia: Reordered

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+            norm3 = get_norm(normprof3, prof3, aa)
+        if prof4 is None:
+            norm4 = norm3
+        else:
+            norm4 = get_norm(normprof4, prof4, aa)
+
+        norm = norm1 * norm2 * norm3 * norm4
+
+        # Compute trispectrum at this redshift
+        p12 = get_pk(p12_of_k_a)
```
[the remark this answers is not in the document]
carlosggarcia: Yeah. Fixed this bit, however I haven't checked if this T^2_22 returns something sensible. I'll do once we're happy with T^2_13, as you said

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+        out[ia, :, :] = tk_2h_22 * norm
+
+    if np.ndim(a) == 0:
+        out = np.squeeze(out, axis=0)
+    if np.ndim(k) == 0:
+        out = np.squeeze(out, axis=-1)
+        out = np.squeeze(out, axis=-1)
+    return out
+
+def halomod_trispectrum_2h_13(cosmo, hmc, k, a, prof1, p1_of_k_a,
```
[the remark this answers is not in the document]
carlosggarcia: Done

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+                              normprof3=False, normprof4=False,
+                              p2_of_k_a=None, p3_of_k_a=None, p4_of_k_a=None):
+    """ Computes the halo model 2-halo trispectrum for four different
+    quantities defined by their respective halo profiles. The 2-halo
+    trispectrum for four profiles :math:`u_{1,2}`, :math:`v_{1,2}` is
+    calculated as:
+
+    .. math::
+        T^{2h}_{13}_{u_1,u_2,v_1,v_2}(k_u,k_v,a) =
+        P_lin(k_u) * I^1_1(k_{u_1}|u_1) * I^1_3(k_{u_1, v}|u_1, v_{1,2}) + 3 perm
```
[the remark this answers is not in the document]
carlosggarcia: Changed in next commit

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+        i234 = hmc.I_1_3(cosmo, k_use, aa, prof2, prof234_3pt, prof2=prof3,
+                         prof3=prof4)
+        # Permutation 1
+        # p2 = p1  # (because k_a = k_b)
+        i2 = hmc.I_1_1(cosmo, k_use, aa, prof2)[:, None]
+        i134 = hmc.I_1_3(cosmo, k_use, aa, prof1, prof134_3pt, prof2=prof3,
+                         prof3=prof4)
+        ### Attention to axis order change!
+        # Permutation 2
+        p3 = get_pk(p3_of_k_a)(aa)[None, :]
```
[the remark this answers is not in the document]
carlosggarcia: Done in next commit

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+        i214 = hmc.I_1_3(cosmo, k_use, aa, prof4, prof214_3pt, prof2=prof1,
+                         prof3=prof2).T
+        # Permutation 4
+        # p4 = p3  # (because k_c = k_d)
+        i4 = hmc.I_1_1(cosmo, k_use, aa, prof3)[None, :]
+        i231 = hmc.I_1_3(cosmo, k_use, aa, prof3, prof231_3pt, prof2=prof2,
+                         prof3=prof1).T
+        ####
+
+        tk_2h_22 = p1 * (i1 * i234 + i2 * i134) + p3 * (i3 * i214 + i4 * i231)
```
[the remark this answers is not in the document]
carlosggarcia: Done in next commit. Just realized that `tk_2h_22` there should be `tk_2h_13`

pyccl/halos/profiles_3pt.py
carlosggarcia: hmm, should I change it already?

pyccl/halos/halo_model.py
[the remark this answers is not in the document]
carlosggarcia: Fixed the name in next commit.

pyccl/halos/halo_model.py
```diff
@@ -803,6 +870,515 @@ def get_norm(normprof, prof, sf):
...
+                              normprof4=False, p13_of_k_a=None,
+                              p14_of_k_a=None):
+    """ Computes the halo model 1-halo trispectrum for four different
+    quantities defined by their respective halo profiles. The 2-halo
+    trispectrum for four profiles :math:`u_{1,2}`, :math:`v_{1,2}` is
+    calculated as:
+
+    .. math::
+        T^{2h}_{22}_{u_1,u_2;v_1,v_2}(k_u,k_v,a) =
+        P_lin(k_u) * I^1_2(k_u|u_{1,2}) * I^1_2(k_v|u_{1,2}) + 2 perm
```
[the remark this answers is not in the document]
carlosggarcia: Upated in next commit

pyccl/halos/profiles_3pt.py
c-d-leonard: @damonge thoughts on this?

c-d-leonard:
FYI @carlosggarcia this is failing the tests due to some flake8 formatting issues. You can see them if you climb on the Details above for the continuous integration tests. Let me know if you need any help with fixing these.

c-d-leonard:
Hi @carlosggarcia just checking whether the update is still there's no update? no worries if so :) just doing an inventory before next telecon

carlosggarcia:
@c-d-leonard I'm afraid it is. Sorry! I'll try get back to this in a few weeks.

c-d-leonard:
Looks like some progress here :) @carlosggarcia any support you need on this at this point?

carlosggarcia:
@c-d-leonard, thanks for asking! @damonge is already helping me with the maths. I think that will be enough for now.

carlosggarcia:
@c-d-leonard same as two months ago. I'll try to give it a push in the next two weeks.

pyccl/halos/profiles_3pt.py
carlosggarcia: Done.`I_1_3` computes now `I_1_1(k) * I_1_2(k')`

pyccl/halos/halo_model.py
```diff
@@ -1744,7 +1743,7 @@ def get_kr_f2_f2T_X():
     if np.ndim(k) == 0:
         out = np.squeeze(out, axis=-1)
         out = np.squeeze(out, axis=-1)
-    return out
+    return t1113 + t1122 # out
```
carlosggarcia, line 1746: shouldn't this be outputting `out`?  You need to multiply by the `i1`, `i2`, etc. (see Eq 3.25 of https://arxiv.org/pdf/1912.08209.pdf) to the trispectrum. Do you know why you can reproduce the external code's results without these? Furthermore, you need to make sure the output has the right dimensions given the input `a` and `k`.

carlosggarcia:
Finally, this is ready for review! @damonge

pyccl/halos/halo_model.py
```diff
@@ -247,6 +247,53 @@ def I_1_1(self, cosmo, k, a, prof):
...
+    def I_1_3(self, cosmo, k, a, prof, *, prof2=None, prof_2pt, prof3=None):
+        """ Solves the integral:
+
+        .. math::
+            I^1_3(k,a|u_2, v_1, _v2) = \\int dM\\,n(M,a)\\,b(M,a)\\,
+            \\langle u_2(k,a|M) v_1(k',a|M) v_2(k',a|M)\\rangle,
+
+        approximated to
+        .. math::
+            I^1_3(k,a|u_2, v_1, _v2) = I^1_1(k,a|u_2) I^1_2(k',a|v_1, v_2)
```
damonge, line 259: Humm, I don't think this equation is correct. What you're approximating is <uvv> ~ <u><vv>, right? where the angle brackets are an average over profile realisations, not over mass

pyccl/halos/halo_model.py
```diff
@@ -309,18 +356,27 @@ def I_1_2(self, cosmo, k, a, prof, *, prof2=None, prof_2pt):
...
             prof2 = prof
-
         self._check_mass_def(prof, prof2)
         self._get_ingredients(cosmo, a, get_bf=True)
-        uk = prof_2pt.fourier_2pt(cosmo, k, self._mass, a, prof, prof2=prof2).T
-        return self._integrate_over_mbf(uk)
+        uk = prof_2pt.fourier_2pt(cosmo, k, self._mass, a, prof,
+                                  prof2=prof2, diag=diag)
+        # uk = prof_2pt.fourier_2pt(cosmo, k, self._mass, a, prof, prof2=prof2,
+        #                           diag=diag).T
```
damonge, line 373: remove commented lines

pyccl/halos/halo_model.py
```diff
@@ -309,18 +356,27 @@ def I_1_2(self, cosmo, k, a, prof, *, prof2=None, prof_2pt):
                 a profile covariance object
                 returning the the two-point moment of the two profiles
                 being correlated.
+            diag (bool): If True, both halo profiles depend on the same k. If
+            False, they will depend on k and k', respectively. Default True.
```
damonge, line 360: Can you check if this reads well in the doc pages?
Also, is the idea that, if diag = False, then k is a 2D array, not a 1D array?

pyccl/halos/halo_model.py
```diff
@@ -384,4 +440,5 @@ def I_0_22(self, cosmo, k, a, prof, *,
             uk34 = prof34_2pt.fourier_2pt(
                 cosmo, k, self._mass, a, prof3, prof2=prof4).T
 
+        # TODO: Is this k ordering correct?
```
damonge, line 443: is it?

pyccl/halos/pk_4pt.py
```diff
@@ -495,6 +500,81 @@ def _allocate_profiles(prof, prof2, prof3, prof4, prof12_2pt, prof34_2pt):
     return prof, prof2, prof3, prof4, prof12_2pt, prof34_2pt
 
 
+def _allocate_profiles1pt(prof, prof2, prof3, prof4):
+    if not isinstance(prof, HaloProfile):
```
damonge, line 504: I think we don't need to check the type (we stopped doing this), since it'll fail anyway with an equally useful python-level message

damonge: How is this function different to the already-existing `_allocate_profiles` above it?

pyccl/halos/pk_4pt.py
```diff
@@ -495,6 +500,81 @@ def _allocate_profiles(prof, prof2, prof3, prof4, prof12_2pt, prof34_2pt):
...
+    if prof4 is None:
+        prof4 = prof3
+    elif not isinstance(prof4, HaloProfile):
+        raise TypeError("prof4 must be of type `HaloProfile` or `None`")
+
+    return prof, prof2, prof3, prof4
+
+
+def _allocate_profiles2(prof, prof2, prof3, prof4, prof13_2pt, prof14_2pt,
+                        prof24_2pt, prof32_2pt):
```
damonge, line 523: As above, no need to check for types.

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+            kk = k.flatten()
+            if isinstance(p_of_k_a, Pk2D):
+                pk = p_of_k_a(kk, aa, cosmo)
+            elif (p_of_k_a is None) or (str(p_of_k_a) == 'linear'):
+                pk = cosmo.get_linear_power()(kk, aa)
+            elif str(p_of_k_a) == 'nonlinear':
+                pk = cosmo.get_nonlin_power()(kk, aa)
+            else:
+                raise TypeError("p_of_k_a must be `None`, \'linear\', "
+                                "\'nonlinear\' or a `Pk2D` object")
```
damonge, line 681: could we set up the Pk2D outside of the integrand (so it only happens once)? Should help  with speed and would be a bit neater.

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+        norm = norm1 * norm2 * norm3 * norm4
+        p = get_isotropized_pkr(p_of_k_a, aa)
+
+        # Compute trispectrum at this redshift
+        # P(k1 - k1 = 0) = 0
+        # p12 = get_isotropized_pk(p_of_k_a, 0 * kkth, aa)
+        # i12 = hmc.I_1_2(cosmo, k_use, aa, prof, prof12_2pt,
+        #                 prof2=prof2)[:, None]
+        # i34 = hmc.I_1_2(cosmo, k_use, aa, prof3, prof34_2pt,
+        #                 prof2=prof4)[None, :]
```
damonge, line 701: remove commented lines

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+        elif (p_of_k_a is None) or (str(p_of_k_a) == 'linear'):
+            def pkf(sf):
+                return cosmo.get_linear_power()(k_use, sf)
+        elif str(p_of_k_a) == 'nonlinear':
+            def pkf(sf):
+                return cosmo.get_nonlin_power()(k_use, sf)
+        else:
+            raise TypeError("p_of_k_a must be `None`, \'linear\', "
+                            "\'nonlinear\' or a `Pk2D` object")
+        return pkf
```
damonge, line 807: We could call this just once outside the loop, right? (just setting the function, I mean)

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+            pk = p_of_k_a(k, a, cosmo)
+        elif (p_of_k_a is None) or (str(p_of_k_a) == 'linear'):
+            pk = cosmo.get_linear_power()(k, a)
+        elif str(p_of_k_a) == 'nonlinear':
+            pk = cosmo.get_nonlin_power()(k, a)
+        else:
+            raise TypeError("p_of_k_a must be `None`, \'linear\', "
+                            "\'nonlinear\' or a `Pk2D` object")
+
+        return pk
```
damonge, line 943: As before, should we set up the Pk2D just once?

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
```
damonge, line 1109: as above

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+        # Compute profile normalizations
+        norm1, norm2, norm3, norm4 = _get_norms(prof, prof2, prof3, prof4,
+                                                cosmo, aa, hmc)
+        norm = norm1 * norm2 * norm3 * norm4
+
+        # Permutation 0
+        # Bpt_1_2_34 = 0
+        # i1 = hmc.I_1_1(cosmo, k_use, aa, prof)[:, None]
+        # i2 = hmc.I_1_1(cosmo, k_use, aa, prof2)[:, None]
+        # i34 = hmc.I_1_2(cosmo, k_use, aa, prof3, prof34_2pt, prof2=prof4)
```
damonge, line 993: remove commented lines

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+                        prof_2pt=prof14_2pt, diag=False)
+        # Permutation 4: 1 <-> 4
+        i31 = hmc.I_1_2(cosmo, k_use, aa, prof3, prof2=prof,
+                        prof_2pt=prof13_2pt, diag=False)
+
+        # Permutation 5: 12 <-> 34
+        # Bpt_3_4_12 = 0
+        # i3 = hmc.I_1_1(cosmo, k_use, aa, prof3)[None, :]
+        # i4 = hmc.I_1_1(cosmo, k_use, aa, prof4)[None, :]
+        # i12 = hmc.I_1_2(cosmo, k_use, aa, prof, prof12_2pt, prof2=prof2)
```
damonge, line 1017: remove commented lines

pyccl/halos/pk_4pt.py
```diff
@@ -505,3 +585,976 @@ def _logged_output(*arrs, log):
...
+    tkk_2h_22 = halomod_trispectrum_2h_22(cosmo, hmc, np.exp(lk_arr), a_arr,
+                                          prof, prof2=prof2,
+                                          prof3=prof3, prof4=prof4,
+                                          prof13_2pt=prof13_2pt,
+                                          prof14_2pt=prof14_2pt,
+                                          prof24_2pt=prof24_2pt,
+                                          prof32_2pt=prof32_2pt,
+                                          p_of_k_a=p_of_k_a)
+
+    tkk_2h_13 = halomod_trispectrum_2h_13(cosmo, hmc, np.exp(lk_arr), a_arr,
```
damonge, line 1272: I would suggest that, if the `_22` and `_13` components of this are not meant to be called separately by users, we just make them private methods.

pyccl/halos/profiles_2pt.py
```diff
@@ -71,14 +73,25 @@ def fourier_2pt(self, cosmo, k, M, a, prof, *, prof2=None):
         if prof2 is None:
             prof2 = prof
 
+        HP = HaloProfile
+        if not (isinstance(prof, HP) and isinstance(prof2, HP)):
+            raise TypeError("prof and prof2 must be HaloProfile")
```
damonge, line 78: Do we need this?

pyccl/halos/profiles_2pt.py
```diff
@@ -111,6 +124,9 @@ def fourier_2pt(self, cosmo, k, M, a, prof, *, prof2=None):
                 is desired. If ``None``, the assumption is that you want
                 an auto-correlation. Note that only auto-correlations
                 are allowed in this case.
+            diag (bool): If True, both halo profiles depend on the same k. If
+            False, they will depend on k and k', respectively and we will
+            approximate <uk uk'> to <uk><uk'>. Default True.
```
damonge, line 129: Meh, we should actually modify `fourier_variance` in the HOD profile to do this properly, rather than approximating.

pyccl/halos/profiles_2pt.py
```diff
@@ -60,6 +60,8 @@ def fourier_2pt(self, cosmo, k, M, a, prof, *, prof2=None):
                 second halo profile for which the second-order moment
                 is desired. If ``None``, the assumption is that you want
                 an auto-correlation, and `prof` will be used as `prof2`.
+            diag (bool): If True, both halo profiles depend on the same k. If
+            False, they will depend on k and k', respectively. Default True.
```
damonge, line 64: Can you specify that in this case the output will be 2D (in k and k')?

pyccl/halos/profiles_2pt.py
damonge: Also, isn't there a similar thing for the CIB?

damonge (commented):
Thanks a lot @carlosggarcia , this is really great.

See comments below. One overall comment: the new functions in `pk_4pt` could be optimized by comparing profiles, and avoiding recomputing some integrals if they've been already computed (you can check out what's being done in the 1h trispectrum). If you can see an easy way of doing that now, go for it, but I'm also happy to just leave a comment and optimise this later on (since this is already a huge PR!).

pyccl/halos/halo_model.py
carlosggarcia: It depends on what's the default for CCL. When building my 2D arrays, I am doing the following:
```
k = k_use[:, None]
kp = k_use[None, :]
```
so the second index correspond to k'. This is not what is happening here. However, this works if the profiles of `uk12` and `uk34` are the same. What should I change? This line or the others lines?

pyccl/halos/pk_4pt.py
carlosggarcia: For the 4h case, we only need the profiles, without the 2pt ones.
There is another difference. in the already existing `_allocate_profiles` we have
```
if prof4 is None:
    prof4 = prof2
```
whereas in the `_allocate_profiles1pt` has `prof4 = prof3`, which I think makes more sense. However, changing this in `_allocate_profiles` breaks a tkssc test. Which one should I leave?

pyccl/halos/profiles_2pt.py
carlosggarcia: I don't remember why I added it. I think I added it because I was having some errors while testing my merge with master, probably because the order of the arguments changed and I was passing the wrong object. I can remove it if you think it will not be helpful.

pyccl/halos/profiles_2pt.py
carlosggarcia: Should I modify this in this PR? Regarding CIB, I didn't touch it. Should I add the diag argument too?

pyccl/halos/halo_model.py
carlosggarcia: I think now is correct. I wasn't sure what you meant about the average.

pyccl/halos/halo_model.py
carlosggarcia: Done

pyccl/halos/halo_model.py
carlosggarcia: Checked and fixed. Yes `diag=False` returns a 2D array (k, k')

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/profiles_2pt.py
carlosggarcia: Done

carlosggarcia:
I have addressed the easy comments. I'll look now how to optimize it.

carlosggarcia:
back to you @damonge

pyccl/halos/halo_model.py
```diff
@@ -247,6 +247,54 @@ def I_1_1(self, cosmo, k, a, prof):
...
+        """ Solves the integral:
+
+        .. math::
+            I^1_3(k,a|u_2, v_1, v_2) = \\int dM\\,n(M,a)\\,b(M,a)\\,
+            \\langle u_2(k,a|M) v_1(k',a|M) v_2(k',a|M)\\rangle,
+
+        approximated to
+
+        .. math::
+            I^1_3(k,a|u_2, v_1, v_2) \\sim I^1_1(k,a|u_2) I^1_2(k',a|v_1, v_2),
```
damonge, line 260:
```suggestion
        where we approximate

        .. math::
            \\langle u_2(k,a|M) v_1(k',a|M) v_2(k', a|M)\\rangle \\sim \\langle u_2(k,a|M)\\rangle \\langle v_1(k',a|M) v_2(k', a|M)\\rangle,
```

pyccl/halos/halo_model.py
damonge: Don't think so. Let me suggest something

pyccl/halos/halo_model.py
damonge: It's really up to you. If you prefer it this way, just remove the comment (or change it to specify what's going on)

pyccl/halos/pk_4pt.py
damonge: We need to choose one default behaviour and stick to it in all these functions. I actually think the behaviour in `_allocate_profiles` makes a tiny bit more sense. Since this is also what's used by the 1h and SSC terms, which are already part of the API, I suggest we keep that (otherwise we break the API anyway).

damonge: and going back to my original point, to avoid rewriting code, in that case, simply call `_allocate_profiles1pt` inside `_allocate_profiles`

pyccl/halos/profiles_2pt.py
damonge: meh, let's leave it.

pyccl/halos/profiles_2pt.py
damonge: Humm, let's leave it as it is. Can you add a TODO comment somewhere?

pyccl/halos/halo_model.py
carlosggarcia: Suggestion accepted. Now I see what you meant. Thanks!

pyccl/halos/profiles_2pt.py
damonge: But you should do a similar thing for the CIB profile here: https://github.com/LSSTDESC/CCL/blob/a427bddaabdacb3d43188d7c3b4b4f1fcd1f0b78/pyccl/halos/profiles_2pt.py#L134

damonge (commented):
A few more comments (plus remaining previous conversations)

pyccl/halos/halo_model.py
carlosggarcia: Done. Left as it was for Tkk_1h to avoid changing the default ordering (i.e. (k', k)).

pyccl/halos/pk_4pt.py
carlosggarcia: Done

pyccl/halos/profiles_2pt.py
carlosggarcia: Done

carlosggarcia:
I think all it's done now. Back to you @damonge. Thanks for reviewing!

pyccl/tests/test_hmcalculator_integrals.py
```diff
@@ -0,0 +1,99 @@
...
+    prof1 = P1
+
+    I = hmc.I_1_1(cosmo, k_use, aa, prof1)
+    I01 = hmc.I_0_1(cosmo, k_use, aa, prof1)
+
+    # Test correct shape
+    assert I.shape == (nk, )
+
+    # Check that you're accounting for the bias
+    assert np.max(np.abs(I / I01 - 1)) > 0.1
```
damonge, line 48: Nice new tests

pyccl/tests/test_profiles.py
```diff
@@ -165,6 +165,34 @@ def test_cib_smoke():
...
+                                         mass_def='200c')
+    p2pt = ccl.halos.Profile2ptCIB()
+
+    # Test diag=False
+    # TODO: I comment this out because the CIB_2pt returns np.array(number)
+    # which is not a float. Is this intended? This is not the case for
+    # Profile2pt.fourier_2pt
+    # F = p2pt.fourier_2pt(COSMO, 1., 1E13, 1., p1, prof2=p2,
+    #                    diag=False)
+    # assert isinstance(F, float)
```
damonge, line 182: This is a weird numpy behaviour. The reason is that, in this case, `fourier_2pt` returns a squeezed array, which, even when it's a scalar, still has a `ndarray` type. This doesn't really matter, because it will actually behave as a scalar. In fact, when you take one of these squeezed arrays and multiply them by any other scalar (even another squeezed array), the result becomes a float.

So what's going on here, is that the standard `Profile2pt` gets the two squeezed profiles and multiplies them together (the result being a float). The CIB and HOD versions, however, return the squeezed array that their `fourier_variance` spits out, which is still a squeezed array.

As I said, none of this matters, since any of these squeezed arrays will become floats as soon as you do anything with them. So, instead, how about testing `assert np.ndim(F) == 0`?

pyccl/tests/test_profiles.py
```diff
@@ -322,6 +350,24 @@ def test_hod_2pt():
     with pytest.raises(ValueError):
         p2.fourier_2pt(COSMO, 1., 1e13, 1., pgood, prof2=pbad)
 
+    # Test diag = False
+    # TODO: I comment this out because the HOD_2pt returns np.array(number)
+    # which is not a float. Is this intended? This is not the case for
+    # Profile2pt.fourier_2pt
+    # F = p2.fourier_2pt(COSMO, 1., 1E13, 1., pgood, prof2=pgood,
+    #                    diag=False)
+    # assert isinstance(F, float)
```
damonge, line 359: see above

pyccl/tests/test_profiles.py
```diff
@@ -165,6 +165,34 @@ def test_cib_smoke():
...
+    F = p2pt.fourier_2pt(COSMO, [1., 2], 1E13, 1., p1, prof2=p2, diag=False)
+    assert F.shape == (2, 2)
+    F = p2pt.fourier_2pt(COSMO, [1., 2], [1e12, 5e12, 1e13], 1., p1, prof2=p2,
+                         diag=False)
+    assert F.shape == (3, 2, 2)
+    F2 = ccl.halos.Profile2pt().fourier_2pt(COSMO, [1., 2],
+                                            [1e12, 5e12, 1e13], 1., p1,
+                                            prof2=p2, diag=False)
+    print(F)
+    print(F2)
```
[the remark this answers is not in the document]
carlosggarcia: Done

pyccl/tests/test_profiles.py
carlosggarcia: Done

pyccl/tests/test_profiles.py
carlosggarcia: Done

PR #874, opened 2021-05-12
implemented r_corr in 1h 2-pt correlator

nikfilippas:
I sub-classed `Profile2pt` into a new class `Profile2ptR`, where the parameter `r_corr` is implemented.

- The motivation behind this is that 1-halo terms of different profiles may not necessarily be fully correlated. We can set the degree of correlation between the 1-halo terms by scaling `uk1*uk2` by `1+r_corr`.
- For the trivial case where `r_corr=0` it simply returns the 2-point correlator of its parent class, `Profile2pt`.
- I've added tests for that, and for the case where the 1-halo terms are completely uncorrelated (`1+r_corr=0`).


nikfilippas:
I think the unit test is written correctly now, so this is ready to be merged once the checks have been completed.

pyccl/halos/profiles_2pt.py
```diff
@@ -106,3 +106,72 @@ def fourier_2pt(self, prof, cosmo, k, M, a,
...
+        self.r_corr = r_corr
+
+    def update_parameters(self, r_corr=None):
+        """ Update any of the parameters associated with this 1-halo
+        2-point correlator. Any parameter set to `None` won't be updated.
+        """
+        if r_corr is not None:
+            self.r_corr = r_corr
+
+    def fourier_2pt(self, prof, cosmo, k, M, a, r_corr=None,
```
damonge, line 129: if `r_corr` is an attribute of the object, why pass it here as a parameter?
I would choose either one or the other

pyccl/halos/profiles_2pt.py
```diff
@@ -106,3 +106,72 @@ def fourier_2pt(self, prof, cosmo, k, M, a,
...
+
+
+class Profile2ptR(Profile2pt):
+    """ This class inherits from the 1-halo 2-point correlator between two
+    halo profiles. Sometimes we need not assume that the 1-halo terms are
+    fully correlated. We introduce a term `r_corr` which scales the product
+    of the fourier halo profiles by `(1+r_corr)`. In the trivial case where
+    `r_corr=0` this returns the 2-point correlator of the parent class.
+    Example usecases can be found in arXiv:1909.09102 and arXiv:2102.07701.
+    """
```
damonge, line 118: If you're gonna make `r_corr` an input parameter, it needs to be defined in the docstring above.

damonge (changes requested):
Thanks a lot @nikfilippas . I think we just need to clarify whether `r_corr` is an attribute or a function parameter. Your call as far as I'm concerned.

pyccl/halos/profiles_2pt.py
nikfilippas: Yes, I agree. That's why I personally wouldn't make a separate class for `Profile2ptR`; I would just add `r_corr` as an argument of `Profile2pt.fourier_2pt`.
I don't see why `Profile2pt` prohibits this by:

>     More complicated cases should be implemented by subclassing
>     this class and overloading the :meth:`~Profile2pt.fourier_2pt`
>     method.

Adding `r_corr` in `Profile2ptR` would simplify things.

damonge: I'm fine with you modifying `Profile2pt` if you prefer!

damonge: (changing the docstring accordingly)

nikfilippas:
Closing and re-opening because of a `git` issue, and because the implementation changes.

PR #875, opened 2021-05-15
HOD profile: relax requirement to only form satellites when centrals are present

nikfilippas:
- In the current implementation of the HOD profile, satellites can only form when centrals are present. Although mostly true, this assumption breaks in the case of (recent) mergers or smaller clusters.
- Following `arXiv:1909.09102` I've introduced the boolean parameter `ns_independent` in `HaloProfileHOD` which relaxes this assumption.
- In essence, when `ns_independent==True` the satellite contribution is not multiplied by `N_c` (which is a smoothed step function).

The result of this, in reference to the cluster mass `M` is:
- When `M < Mmin_0` the real and fourier -space profiles are not zero anymore. (After all, there is some mass somewhere in there!)
- When `M > Mmin_0` the real and fourier -space profiles of an HOD instantiated with `ns_independent=True` are equal to an HOD instantiated with `ns_independent=False`.
- When `M = Mmin_0` due to how the heaviside function is defined in `numpy`, the independent case has 2x the power of the dependent case (minus 0.5).


nikfilippas:
This, too, should now be ready for merging once the checks are completed.

nikfilippas:
@damonge COVERALLS is correctly picking up that I am not testing the `fourier_variance` method of HOD for `ns_independent=True`. Could you confirm that the implementation is right? I'll add it to the unit test later.

pyccl/tests/test_profiles.py
```diff
@@ -120,6 +120,42 @@ def test_hod_smoke():
...
+
+    c = ccl.halos.ConcentrationDuffy08(M200)
+    hmd = c.mdef
+    p1 = ccl.halos.HaloProfileHOD(c_M_relation=c,
+                                  lMmin_0=12.,
+                                  ns_independent=False)
+    p2 = ccl.halos.HaloProfileHOD(c_M_relation=c,
+                                  lMmin_0=12.,
+                                  ns_independent=True)
+    # M < Mmin
```
damonge, line 136: Very nice tests!

damonge (changes requested):
Thanks a lot @nikfilippas. The implementation looks good, we just need a test of the profile variance.

pyccl/tests/test_profiles.py
nikfilippas: I've already done the test (for M < Mmin). Do you think it would make sense to include the other 2 cases (as with the `_real` and `_fourier` methods)?

damonge: Hadn't seen those. OK, this looks good to me!

damonge (changes requested):
LGTM

damonge (approved):
LGTM (sorry, misclicked)

Issue #820, opened 2020-09-11
Halo model enhancements


damonge:
There are a couple of nice things in the latest [hmcode paper](https://arxiv.org/abs/2009.01858) that would be nice to add to our halo model calculator functions:
- 1-halo large-scale damping
- smoothing out the transition region

There are many other interesting things in the paper, but these two would be trivial to implement.
And this would be a separate issue from that of allowing CCL to use the latest hmcode to compute the non-linear Pk.

c-d-leonard:
@damonge is this the same as what is being worked on in PR #877 ?

damonge:
Partly, yes (although this doesn't address the halo exclusion thing at low ell for 1-halo)

nikfilippas:
I've implemented everything in #877 after Danielle's suggestion, and I've linked this to my PR.

PR #877, opened 2021-05-17
Enhancements of the Halo Model

nikfilippas:
Addresses #820 (after @c-d-leonard pointed out it was almost the same thing)

Inspired by the HMCODE-2020 paper (`arXiv:2009.01858`) I have added the following two extensions to the halo model:

1. Large scale damping of the 1-halo term, via `supress_1h` in `halomod_power_spectrum`. This has to be a function of scale factor (eq. 17).
2. Smoothing of the 1-halo/2-halo transition region, via `smooth_transition` in `halomod_power_spectrum`. This also has to be a function of scale factor (eq. 23).

--------------------------------------------------------
**_OLD TEXT_**:

It is known that the halo model is not accurate at the quasi-linear regime (the transition between the 2h and 1h terms). This can be accounted for in the models in two primary ways:
- ell-level correction: smooth-out the transition between the 2h- and 1h- terms. This is irrelevant to CCL as the correction can be done fully outside of the package, when the user already has computed `angular_cl`.
- k-level correction: as the halo model concerns haloes, this is a more physically motivated approach. The power spectrum can be corrected via a time- and scale-dependent correction function of the form `f(k, a, cosmo)`.

CCL currently does not support this. I've implemented the argument `f_ka` in `halomod_power_spectrum` which takes care of this correction, by multiplying the resulting power spectrum with the result of the input function.


nikfilippas:
This is also properly unit tested now, waiting for coveralls to report.

pyccl/halos/halo_model.py
```diff
@@ -509,6 +510,18 @@ def halomod_power_spectrum(cosmo, hmc, k, a, prof,
             term in the first equation above) won't be computed.
         get_2h (bool): if `False`, the 2-halo term (i.e. the second
             term in the first equation above) won't be computed.
+        smooth_transition (function or None):
+            Modify the halo model 1-halo/2-halo transition region
+            via a time-dependent function :math:`\\alpha(a)`,
+            defined as in HMCODE-2020: :math:`P(k,a)=
```
damonge, line 516: Can you add a reference to the paper (either arXiv reference or Mead et al. 2020)

pyccl/halos/halo_model.py
```diff
@@ -509,6 +510,18 @@ def halomod_power_spectrum(cosmo, hmc, k, a, prof,
...
+            (P_{1h}^{\\alpha(a)}(k)+P_{2h}^{\\alpha(a)}(k))^{1/\\alpha}`.
+            If `None` the extra factor is just 1.
+        supress_1h (function or None):
+            Supress the 1-halo large scale contribution by a
+            time- and scale-dependent function :math:`k_*(a)`,
+            defined as in HMCODE-2020:
+            :math:`\\frac{(k/k_*(a))^4}{1+(k/k_*(a))^4}`.
```
damonge, line 523: Same here about the reference

pyccl/halos/halo_model.py
```diff
@@ -579,13 +608,28 @@ def pkf(sf):
...
             pk_1h = 0.
 
-        # Total power spectrum
-        out[ia, :] = (pk_1h + pk_2h) * norm
+        # Transition region
+        if smooth_transition is None:
+            out[ia, :] = (pk_1h + pk_2h) * norm
+        else:
+            alpha = smooth_transition(aa)
+            if alpha < 0.4 and not flag_alpha:
```
damonge, line 623: Why 0.4 in particular?

pyccl/halos/halo_model.py
```diff
@@ -579,13 +608,28 @@ def pkf(sf):
...
+                flag_alpha = True
+            out[ia, :] = (pk_1h**alpha + pk_2h**alpha)**(1/alpha) * norm
+
+    if flag_alpha:
```
damonge, line 627: I am not sure we need this warning. What's the rationale?

pyccl/tests/test_pkhm.py
```diff
@@ -138,6 +138,10 @@ def f(k, a):
                            'h2': True, 'itg': 'spline',
                            'p2': None},
                           {'cv': None, 'norm': True,
+                           'pk': 'linear', 'h1': True,
```
damonge, line 141: Why are you adding one extra test here? It seems what you added is just a repetition of the same test, right?

pyccl/tests/test_pkhm.py
```diff
@@ -182,6 +186,44 @@ def test_pkhm_pk2d():
...
+                                           normprof1=True, normprof2=True,
+                                           supress_1h=None, get_2h=False)
+    pk1 = ccl.halos.halomod_power_spectrum(COSMO, hmc, k_arr, a_arr, P1,
+                                           normprof1=True, normprof2=True,
+                                           supress_1h=ks0, get_2h=False)
+    assert np.allclose(pk0, pk1, rtol=0)
+    pk2 = ccl.halos.halomod_power_spectrum(COSMO, hmc, k_arr, a_arr, P1,
+                                           normprof1=True, normprof2=True,
+                                           supress_1h=ks1, get_2h=False)
+    assert np.allclose(pk2, 0, rtol=0)
```
damonge, line 225: Can you add one more that is completely quantitative? I.e. make k_star a sensible value (0.05 or whatever), and test that the resulting pk is equal to pk0 multiplied by the corresponding k-dependent factor.

damonge (changes requested):
Looks great, thanks a lot @nikfilippas. A few minor points below.

pyccl/tests/test_pkhm.py
nikfilippas: This addition is irrelevant to my PR. I added one extra test to cover for an extra line of code that wasn't covered before (when there's a `prof2` and `normprof2==False`).

damonge: Oh, I see. Didn't realise `norm` wasn't the same in both cases. OK, good.

pyccl/halos/halo_model.py
nikfilippas: Here in the figure, I'm plotting P(k, a=1) for the autocorrelation of the NFW profile, for the smoothing parameter ranging in 0.2 < α < 2.
For α > 1 the sharpening does not behave in a weird manner, but smoothing (for α < 1) only makes sense up until α ~ 0.4. After that it doesn't really smooth the transition, rather, it affects the entire P(k). I could have been stricter and imposed a warning for α < 0.5, but I figured there might be other profiles where α being so low might make sense.
alphas

pyccl/halos/halo_model.py
nikfilippas: I think I explain the rationale of the warning, along with the value I've chosen, in my reply below.

pyccl/halos/halo_model.py
nikfilippas: The keyword is `smooth_transition` anyway. The point is that users shouldn't expect that only the transition changes for any α < 0.4.

nikfilippas: And here it the equivalent plot for P(k, a=0.3).
alphas_2

nikfilippas: And here's the same thing for the HOD profile's autocorrelation (using the default HOD params) for scale factor 0.3.
HOD

nikfilippas: And lastly, for the GNFW profile (using the default params) and scale factor 0.3.
GNFW

pyccl/tests/test_pkhm.py
nikfilippas: Done!

pyccl/halos/halo_model.py
nikfilippas: Done.

pyccl/halos/halo_model.py
nikfilippas: Done.

pyccl/halos/halo_model.py
damonge: Ok, thanks.
So alpha<0.4 may not seem sensible to you (or to me), but it's still a valid operation someone might want to perform. For instance, one may want to marginalize over alpha in an MCMC, which will often cross this boundary.
I'd much rather not burden the code with extra lines and checks in this case, so I'd remove this check altogether.

damonge (changes requested):
OK, thanks! Final point

pyccl/halos/halo_model.py
nikfilippas: Yep, that makes sense. I'll remove it. Thanks!

nikfilippas: Done.

damonge (approved):
LGTM!

PR #883, opened 2021-06-07
Tuning knob param for 1-halo 2-point correlations

nikfilippas:
Cont. work from #874 . Basically, a parameter `r_corr` which controls the sign of the correlation of two profiles, which do not necessarily have to be fully correlated.

In the original implementation I subclassed `Profile2pt` into `Profile2ptR`, because there was an explicit statement to only subclass `Profile2pt` for more complicated 2-point moments, but in hindsight, there is no reason for this prohibition to exist. So I am implementing it directly into the base class.


nikfilippas:
@damonge I think this is ready for review now.

damonge (approved):
LGTM!

PR #887, opened 2021-06-09
Methods of Cosmology

nikfilippas:
I have added few lines of code to `core.py` to make (almost) all of the `pyccl` functions which accept `cosmo` as their first argument, methods of the `Cosmology` class.

### New methods, organized by module

1. **pyccl.background**
  - growth_factor
  - growth_factor_unnorm
  - growth_rate
  - comoving_radial_distance
  - comoving_angular_distance
  - angular_diameter_distance
  - luminosity_distance
  - h_over_h0
  - distance_modulus
  - scale_factor_of_chi
  - omega_x
  - rho_x

2. **pyccl.bcm**
  - bcm_model_fka
  - bcm_correct_pk2d

3. **pyccl.boltzmann**
 - get_camb_pk_lin
 - get_isitgr_pk_lin
 - get_class_pk_lin

4. **pyccl.cls**
  - angular_cl

5. **pyccl.correlations**
  - correlation
  - correlation_3d
  - correlation_multipole
  - correlation_3dRsd
  - correlation_3dRsd_avgmu
  - correlation_pi_sigma

6. **pyccl.covariances**
  - angular_cl_cov_cNG

7. **pyccl.pk2d**
  - parse_pk2d

8. **pyccl.power**
  - linear_power
  - nonlin_power
  - linear_matter_power
  - nonlin_matter_power
  - sigmaM
  - sigmaR
  - sigmaV
  - sigma8
  - kNL

9. **pyccl.tracers**
  - get_density_kernel
  - get_lensing_kernel
  - get_kappa_kernel

10. **pyccl.halos.halo_model**
  - halomod_mean_profile_1pt
  - halomod_bias_1pt
  - halomod_power_spectrum
  - halomod_Pk2D
  - halomod_trispectrum_1h
  - halomod_Tk3D_1h

11. **pyccl.halos.massdef**
  - mass2radius_lagrangian
  - convert_concentration   ==> I added this one in `__init__.py` because it's useful

12. **pyccl.nl_pt.power**
  - get_pt_pk2d

13. **pyccl.nl_pt.tracers**
  - translate_IA_norm

### These functions take `cosmo` as their first argument but were not added (because they are not in `__init__.py`)
- _get_concentration
- _get_mf_hb
- _Sig_MG

### These functions take `cosmo` as their first argument but were not added (because they are deprecated)
  - einasto_profile_3d
  - halo_bias
  - halo_concentration
  - halomodel_matter_power
  - hernquist_profile_3d
  - massfunc
  - massfunc_m2r
  - nfw_profile_2d
  - nfw_profile_3d
  - onehalo_matter_power
  - twohalo_matter_power

### Advantages of this method
1. New functions do not have to be manually added in `Cosmology`
2. It is backwards-compatible! (I think)

### DIFF
--- a/pyccl/__init__.py
+++ b/pyccl/__init__.py
@@ -24,6 +24,9 @@ from .background import growth_factor, growth_factor_unnorm, \
     h_over_h0, luminosity_distance, distance_modulus, scale_factor_of_chi, \
     omega_x, rho_x
 
+# Boltzmann solvers
+from .boltzmann import get_camb_pk_lin, get_isitgr_pk_lin, get_class_pk_lin
+
 # Generalized power spectra
 from .pk2d import Pk2D, parse_pk2d
 
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -5,8 +5,9 @@ can compute a set of theoretical predictions.
 import warnings
 import numpy as np
 import yaml
-import inspect
+from inspect import getmembers, isfunction, signature
 
+import pyccl
 from . import ccllib as lib
 from .errors import CCLError, CCLWarning
 from ._types import error_types
@@ -194,6 +195,17 @@ class Cosmology(object):
                                      "HMCode_logT_AGN": 7.8}}
 
     """
+    # Go through all functions in the main package and the subpackages
+    # and make every function that takes `cosmo` as its first argument
+    # an attribute of this class.
+    subs = [pyccl, pyccl.halos, pyccl.nl_pt]
+    funcs = [getmembers(sub, isfunction) for sub in subs]
+    funcs = [func for sub in funcs for func in sub]
+    for name, func in funcs:
+        pars = signature(func).parameters
+        if list(pars)[0] == "cosmo":
+            vars()[name] = func
+
     def __init__(
             self, Omega_c=None, Omega_b=None, h=None, n_s=None,
             sigma8=None, A_s=None,
@@ -301,7 +313,7 @@ class Cosmology(object):
 
         # Get the call signature of Cosmology (i.e., the names of
         # all arguments)
-        init_param_names = inspect.signature(cls).parameters.keys()
+        init_param_names = signature(cls).parameters.keys()
 
         # Read the values we need from the loaded yaml dictionary. Missing
         # values take their default values from Cosmology.__init__
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -197,10 +197,10 @@ class Cosmology(object):
     # Go through all functions in the main package and the subpackages
     # and make every function that takes `cosmo` as its first argument
     # an attribute of this class.
-    from . import background, bcm, \
+    from . import background, boltzmann, bcm, \
         cls, correlations, covariances, \
         pk2d, power, tracers, halos, nl_pt
-    subs = [background, bcm, cls, correlations, covariances,
+    subs = [background, boltzmann, bcm, cls, correlations, covariances,
             pk2d, power, tracers, halos, nl_pt]
     funcs = [getmembers(sub, isfunction) for sub in subs]
     funcs = [func for sub in funcs for func in sub]
@@ -208,7 +208,7 @@ class Cosmology(object):
         pars = signature(func).parameters
         if list(pars)[0] == "cosmo":
             vars()[name] = func
-    del background, bcm, cls, correlations, covariances, \
+    del background, boltzmann, bcm, cls, correlations, covariances, \
         pk2d, power, tracers, halos, nl_pt
 
     def __init__(
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -7,7 +7,6 @@ import numpy as np
 import yaml
 from inspect import getmembers, isfunction, signature
 
-import pyccl
 from . import ccllib as lib
 from .errors import CCLError, CCLWarning
 from ._types import error_types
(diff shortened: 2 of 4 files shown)


nikfilippas:
Does it make sense to have this unit tested? I am not sure because it's kind of tautological.

c-d-leonard:
Re: unit tests - I think just need to have some tests that ensure if we do try to call these functions via method to cosmology we don't get errors. Benchmark tests I would say are not necessary here if it's just changing the piping of identical code to make them methods.

I'm a bit confused about why those functions aren't in __init__.py, I want to check if that's just an oversight for some of them ..

nikfilippas:
I've unit tested this now. The test checks that all functions in pyccl (that are exposed in `__init__`) have been added as attributes of the `Cosmology` object.

Regarding why some of them are not in `__init__`, maybe that was done to discourage direct calls to the Boltzmann solvers as this is handled separately in CCL. I don't mind adding them in this PR if people agree.

nikfilippas:
I think this is ready for review now. Tagging both of you because you were interested in this PR.

nikfilippas:
I also added the boltzmann solvers in `__init__.py` now, so it's easy to get the `Pk2D` objects directly.

pyccl/core.py
```diff
@@ -194,6 +194,23 @@ class Cosmology(object):
...
+            pk2d, power, tracers, halos, nl_pt]
+    funcs = [getmembers(sub, isfunction) for sub in subs]
+    funcs = [func for sub in funcs for func in sub]
+    for name, func in funcs:
+        pars = signature(func).parameters
+        if list(pars)[0] == "cosmo":
+            vars()[name] = func
+    del background, boltzmann, bcm, cls, correlations, covariances, \
+        pk2d, power, tracers, halos, nl_pt
+
```
damonge, line 213: Can we check how much extra time all these new lines add to the creation of a standard cosmology?
I'd like to check this both on a laptop and on a server (@nikfilippas you can use glamdring).

pyccl/__init__.py
```diff
@@ -24,6 +24,9 @@
     h_over_h0, luminosity_distance, distance_modulus, scale_factor_of_chi, \
     omega_x, rho_x
 
+# Boltzmann solvers
+from .boltzmann import get_camb_pk_lin, get_isitgr_pk_lin, get_class_pk_lin
```
damonge, line 28: why did you need this?

pyccl/core.py
```diff
@@ -194,6 +194,23 @@ class Cosmology(object):
...
+        cls, correlations, covariances, \
+        pk2d, power, tracers, halos, nl_pt
+    subs = [background, boltzmann, bcm, cls, correlations, covariances,
+            pk2d, power, tracers, halos, nl_pt]
+    funcs = [getmembers(sub, isfunction) for sub in subs]
+    funcs = [func for sub in funcs for func in sub]
+    for name, func in funcs:
+        pars = signature(func).parameters
+        if list(pars)[0] == "cosmo":
+            vars()[name] = func
```
damonge, line 210: can you explain what this line is doing?

pyccl/__init__.py
nikfilippas: Why not? It might be useful for anyone who wants to get the camb or class power spectra in a `Pk2D` object in a single line of code. Plus, these functions are not prepended with an underscore, so I don't see why they shouldn't be exposed.

pyccl/core.py
nikfilippas: For every function in the functions exposed in the `__init__.py` files, check the function signature. If the first argument of the funciton is "cosmo", then add it to the local (class) variables as an attribute with the same name.

nikfilippas: If it's not clear in the source code, I can always add comments! Let me know.

pyccl/core.py
nikfilippas: I've checked it with Spyder's built-in `%timeit` functionality, and, within the standard deviation, there is no difference at all.
That was with a loop of 70,000 instantiations.

nikfilippas: - with: 42.3 µs ± 135 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
- without: 40.1 µs ± 198 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

damonge: ok, good

pyccl/core.py
damonge: Yeah, can you add a comment? Took me a bit to figure out what `vars()` (with no argument) did

pyccl/core.py
damonge: @nikfilippas did you test this on glamdring?

pyccl/tests/test_cosmology.py
```diff
@@ -10,6 +10,26 @@
...
+    subs = [background, bcm, cls, correlations, covariances,
+            pk2d, power, tracers, halos, nl_pt]
+    funcs = [getmembers(sub, isfunction) for sub in subs]
+    funcs = [func for sub in funcs for func in sub]
+    for name, func in funcs:
+        pars = signature(func).parameters
+        if list(pars)[0] == "cosmo":
+            _ = getattr(cosmo, name)
+
+
```
damonge, line 32: Can you add a test that runs some of the functions (e.g. pick 5 of them with different complexities) as a `Cosmology` method and as a standalone function, and checks that the result is the same?

damonge (changes requested):

pyccl/core.py
nikfilippas: Timings on glamdring:
- with: `37.0 μs per loop (10000 loops)`
- without: `37.3 μs per loop (10000 loops)`

No statistically significant difference then.

pyccl/core.py
nikfilippas: I actually just realised there is already a comment above the loop, explaining what I do in the lines below. Is this enough explanation?

pyccl/tests/test_cosmology.py
nikfilippas: Done.

damonge (approved):
Ok, looks good to me!

pyccl/tests/test_cosmology.py
```diff
@@ -10,6 +10,40 @@
 import pyccl as ccl
 
 
+def test_cosmo_methods():
+    """ Check that all pyccl functions that take cosmo
+    as their first argument are methods of the Cosmology object.
+    """
+    from inspect import getmembers, isfunction, signature
+    from pyccl import background, bcm, \
```
nikfilippas, line 18: I am not testing for all the modules here. Commit incoming...

nikfilippas: Done. I am now testing the `boltzmann` module as well.
Other than that, I added `neutrinos` and `tk3d` even though they contain no functions with `cosmo`, just in case there are any future additions in there (so that this part of the code won't need maintenance).
@damonge do you still approve?

damonge: looks good to me, yes.

damonge (approved):

Issue #884, opened 2021-06-08
Inconsistent use of cosmo in HMCalculator, MassFunc, HaloBias


nikfilippas:
A `Cosmology` object is needed to initialize an instance of the `HMCalculator`. This is used to calculate `rho0` and store it as a class attribute. Later on, `cosmo` is required when calling any of the `HMCalculator` methods.

If users initialize `hmc` with one cosmology, and then call it with a different cosmology CCL will yield inconsistent results without any errors.

A solution is to remove `cosmo` from a required argument of `HMCalculator.__init__`, and allow it to be instantiated without setting `rho0` as an attribute. This quantity can be calculated on the fly whenever `_get_ingredients` is called (which is the only case where `rho0` is used).

**ref code**
```
class HMCalculator(object):
    ...

    def __init__(self, cosmo, ...):
        self._rho0 = rho_x(cosmo, 1., 'matter', is_comoving=True)  # only time cosmo is used in init
        ...

    def _get_ingredients(self, a, cosmo, get_bf):
        # rho0 is only used in here
        self.mf0 = (self._rho0 - ... ... ...)
        self.mbf0 = (self._rho0 - ... ... ...)
```

nikfilippas:
**UPDATE**
Having a closer look at `HMCalculator`.

Suppose we completely get rid of `cosmo` during initialization of the class instance. This still does not solve the problem, because we need to pass **instances** of `MassFunc` and `HaloBias`, which will have been defined for a specific cosmology!

Then, if we call `_get_ingredients` for a different `cosmo` than the one we used to instantiate our mass function and halo bias, we will once again get inconsistent results.

There are two solutions to this issue:
1. Either leave `cosmo` as it is, in which case, for functions that take both `cosmo` and `hmc` this will be like we pass `cosmo` twice (since only one `cosmo` is valid for a given `hmc`);
2. Or pass the mass function and halo bias **classes** (instead of instances) into `HMCalculator.__init__`, in which case, the mass function and the halo bias will be re-computed every time.

Both of these point to putting `HMCalculator` under `Cosmology`, since with (1) there is a one-to-one correspondence of `cosmo` instance to `hmc` instance, and with (2) we do not end up saving time computing the mass function and halo bias, even if we avoid passing cosmology in the `__init__` method of `HMCalculator`.

Any thoughts are welcome. (tagging @damonge because you were reluctant to put `hmc` under `cosmo`)

nikfilippas:
[...]

PR #888, opened 2021-06-10
Fix cosmo inconsistency in Halo Model

nikfilippas:
This breaks API so it must only be merged to the v3 branch (when we make one).

Addresses #884 .

In short, all of `HMCalculator`, `MassFunc` (and subclasses), and `HaloBias` (and subclasses) could be instantiated with one cosmology and then called with another cosmology producing inconsistent results.

I have made the following changes:
- `MassFunc` & `HaloBias`: Allow instantiation with `cosmo` still, but make sure that every time the getter methods are called, the cosmology passed is consistent with the one in instantiation. ~Fail otherwise.~ (upd: not anymore, see comments below). If a cosmology is not passed during instantiation, then users can use any cosmology when calling the getter method. This is (kind of) backwards compatible, in the sense that previously written codes will only break if the cosmologies are not consistent with one another.
- `HMCalculator`: Completely remove `cosmo` from its `__init__` method. No need to have it; use it to calculate `rho0` on the fly. Apart from `rho0`, `cosmo` was also passed to `MassFunc` and `HaloBias`, but these have changed now.
- tests, benchmarks: adhere to the new signature of the halo model calculator.
- `Cosmology`: added a `__eq__` method to check if this cosmology is equivalent to another cosmology.

*Note:* I decided to leave `cosmo` as an optional argument in `MassFunc` and `HaloBias` because it could be used via `_setup` in case any implementations depend on cosmology. Currently, none of the subclasses do, but this saves us from extra work in the future.


nikfilippas:
**Update** (I use shorthand: mf=mass function, hb=halo bias)
- Cosmology only ought to be used in instantiating a `MassFunc` or `HaloBias` object when the mf/hb form factor (`_setup` method) is cosmology-dependent.
- To address this, I've introduced the `_check_cosmo_dependent` method, which can be overridden for mf/hb whose `_setup` depends on cosmology.
- This new method sets the `_cosmo_dependent` attribute, which is a boolean to check whether each mf/hb depends on cosmolgoy.
- This allows users to actually call the `get_mass_function` and `get_halo_bias` methods using a different cosmology to the one used during instantiation. The only difference is that, if the new cosmology is not equivalent to the old one **and** the mf/hb setup is cosmology-dependent, then the setup is re-run.

a.k.a. problem solved with fewer CCL complaints :)

nikfilippas:
**Update**
Since the functions `mass_function_from_name` and `halo_bias_from_name` are there, why not also define `mass_def_from_name` and use all 3 to simplify the `HMCalculator`!

All commits up to this point had to do with making sure we use `cosmo` consistently, but since this breaks API for the `HMCalculator`, I thought I'd do a slight revamp, so it can now be called directly via:
```
hmc = ccl.halos.HMCalculator(massfunc="Tinker08", hbias="Tinker10", mass_def="200m")
```
It will then internally instantiate the halo mass function and halo bias function to-be-used. No `cosmo` needed after all the commits above. So it's still optimized.

If there is a consensus for this usage in v3 I'll keep this change, otherwise I'll just revert it back to the final commit above this comment.

nikfilippas:
@c-d-leonard , @damonge I think it's ready for review now. This PR contains lots of useful stuff that we could also implement in v2 if this was backwards-compatible.
The difference that breaks API is the removal of `cosmo` as an argument of `HMCalculator`. Since with this PR `cosmo` is not going to be used at all in the `__init__` method, I could make a separate PR, this time **with** `cosmo` as a first argument, and make all of this backwards-compatible to be merged to the v2 master separately.
Let me know what you think.

damonge (commented):
Thanks @nikfilippas . I'd actually advocate for removing the cosmology dependence from the mass function and halo bias instantiations altogether. In the future, these classes should be understood as "prescriptions" rather than cosmology-dependent quantities. I don't think any of our mass functions need to be instantiated with a cosmology, and none of the ones we want to include in the future do (e.g. emulators) should either (i.e. they'll just load the emulator, but not evaluate it at a given cosmology).

This would simplify the current API significantly, and it would also simplify this PR quite a lot.

nikfilippas:
The point of keeping `cosmo` as an optional argument in instantiation is **not** to actually compute anything, but merely to set-up the prescription. Are we sure there will never be a mass function or halo bias prescription whose `_setup` method will be a function of, say, `sigma8`? If the answer is yes, I'll very happily remove it. Simplifying is always better.

nikfilippas:
This will be breaking API. Do you reckon it's a good idea to try and preserve it by amending the decorators I've built for #889 (and this merging the two PRs)? For me, it's just a matter of adding an extra block that checks if `cosmo` is passed, and if it is, it just ignores it and warns the user.

I thought we could preserve API in the next release tag, in order to give users enough time to adapt (otherwise all the warnings are not needed), and then actually do the API change in the next major release.

nikfilippas:
I mean, I think that's a good idea, but it's a matter of how much time it will take you and another reviewer to go through all the changes.

damonge:
OK, since we're at it, why don't you do that? I.e. merge this PR and #889 and remove `cosmo` from the initialization of mass functions and biases.

pyccl/halos/massdef.py
```diff
@@ -265,7 +272,43 @@ class MassDefVir(MassDef):
...
+
+    Args:
+        name (string): a mass function name
+
+    Returns:
+        MassDef subclass corresponding to the input name.
+    """
+    if name == "vir":
+        c_m_relation = "Klypin11"
+        return MassDefVir
```
damonge, line 294: you don't seem to be using `c_m_relation` for anything

pyccl/halos/massdef.py
```diff
@@ -265,7 +272,43 @@ class MassDefVir(MassDef):
...
+
+    c_m_relation = "Duffy08"
```
damonge, line 296: I am not sure I understand the logic of the lines below this one. I mean, I understand what you're doing, but it won't work in general, since `Duffy08` is only defined for `Delta=200`. I am not sure it is worth the extra code to have this function.

pyccl/tests/test_pkhm.py
```diff
@@ -19,6 +19,24 @@
 PK2D = ccl.Pk2D(cosmo=COSMO, pkfunc=lambda k, a: a / k)
 
 
+def test_hmc_from_string_smoke():
```
damonge, line 22: can you add a quick test that these calculators return the same value for one of their associated methods/functions if initialized from objects or strings?

damonge (changes requested):
Few comments to be addressed here or when you merge with the other PR

pyccl/halos/massdef.py
nikfilippas: OK

pyccl/halos/massdef.py
nikfilippas: OK. Indeed, I've now simplified the function. Simple is always better.

pyccl/tests/test_pkhm.py
nikfilippas: OK

nikfilippas:
@damonge alright, let's resolve all the issues/requests here first, before merging the two PRs together.

nikfilippas:
Following are the commits where I get rid of `cosmo` in `MassFunc` and `HaloBias`.

nikfilippas:
@damonge all done, let me know if you have any more comments so I can merge the two PRs together.

pyccl/tests/test_pkhm.py
damonge: I couldn't see these tests @nikfilippas

damonge:
@nikfilippas I think this looks good, but we don't want to merge into master. Can you fix the small comment I made about the tests, change the base of this PR to the `star` branch, and ping me again? I'll take one final look before merging.

damonge:
Also, there seem to be conflicts with `halos/__init__.py`. Take a look.

pyccl/tests/test_pkhm.py
nikfilippas: @damonge issue with git, they now show in the line below.

nikfilippas: They're not qualitative with the string meaning of the word, but I do check now, that the type (!) returned is identical, so they all point to the same class.

nikfilippas:
@damonge I think your comment s addressed now. Let me know if this is fine, and I'll rebase this PR to the `star` branch.

nikfilippas:
I've now rebased this PR into branch `star` so I'm closing it.

Issue #881, opened 2021-06-04
Inconsistencies in function arguments


nikfilippas:
I quickly went over the Python code and noticed inconsistencies in some of the function arguments:

1. Some of the CCL functions/methods have different names for arguments/quantities which are essentially the same. For example, `mdef`, `mass_def`, `m_def` are all used in the code. We should adopt uniform naming.
2. The ordering of some function arguments is not the same everywhere. For example, sometimes, `cosmo` is the first argument. Sometimes it is the last. Sometimes, `prof_2pt` goes before `prof2`. Sometimes it goes after. Like grammar rules, we should make a list of all arguments and decide on a universal ordering, and only allow for very specific deviations from this rule.

damonge:
I agree with this. This will break the API (and a lot of existing code) though, so we should make it a milestone for v3.
@nikfilippas do you have a complete list of the specific inconsistencies you have seen (e.g. in which functions does cosmo not feature first)?

nikfilippas:
I started making a list but it is difficult to do because I couldn't remember all the arguments I had already gone through. I think making a complete list of *all* arguments first would make things easier.

Issue #880, opened 2021-06-04
Star operator for keyword-only arguments


nikfilippas:
Previous discussion on this was in #878 . The main point is that `ccl.Cosmology(0.25, 0.05, 0.65, 0.96, 0.81)` does not fail which opens two issues: readability, and the code being foolproof and futureproof.

On a closer look, these are two sides of the same coin. Obviously, `ccl.Cosmology` *should* require *all* arguments to be keyword-only. But what about functions like `rho_x(cosmo, a, species, is_comoving=False)`? This is not obvious at all.

### The `*` operator in increasing levels of strictness

1. *Essential*: changes that have to do with preventing CCL from silently failing (foolproof), and with making user code compatible with possible changes to the CCL function argument structure (futureproof). _Example_: `ccl.Cosmology(0.25, 0.05, 0.67, 0.81, 0.96)` will silently fail if we decide to swap the order of `sigma8` and `n_s` in the future, giving the wrong cosmology. _Solution_: Go through the entirety of Python code and, for all functions and public methods, add the `*` operator where two (or more) consecutive arguments can be swapped (e.g. accidentally) without the user realizing or CCL complaining. This applies to `Cosmology`, `update_parameters` methods, etc.
2. *Strict*: For example, `rho_x(cosmo, a, species, is_comoving=False)` can be called with `rho_x(cosmo, 1., "matter", True)`. It is not clear what `True` at the end refers to, so we might want to change the definition to `rho_x(cosmo, a, species, *, is_comoving=False)` in order to force users to call it like `rho_x(cosmo, 1. "matter", is_comoving=True)`.
3. *Stricter*: Although "matter" is somewhat self-explanatory, one can argue that `"matter"` is not clear either, and that users should know what the first and second arguments must be because they are used widely in CCL (`cosmo`, `a`). So we could change the definition to `rho_x(cosmo, a, *, species, is_comoving=False)` in order to force users to call it like `rho_x(cosmo, 1., species="matter", is_comoving=False)`.
4. *Strictest*: Apart from the obvious first argument, everything else *must* be keyword-only. Change the definition to `rho_x(cosmo, *, a, species, is_comoving=False)` in order to force users to call it like `rho_x(cosmo, a=1., species="matter", is_comoving=False)`. We should note, however, that although "species" and "is_comoving" are informative words, "a" is not, so is the code really more readable if we use this solution?

### My opinion (in bold)
[...]

Issue #886, opened 2021-06-09
Rename cls.py to cells.py; Rename bcm.py to baryons.py.


nikfilippas:
`cls` is a Python-reserved keyword for class methods ([PEP8](https://www.python.org/dev/peps/pep-0008/)).
I don't think it's good to have a module named `cls.py` just like we wouldn't have a module `self.py`.

nikfilippas:
It even gets recognized by my editing programme and gets a different color.
![Screenshot from 2021-06-09 17-26-28](https://user-images.githubusercontent.com/22447385/121373673-f24e2280-c947-11eb-9cb3-f8f58a1c6b66.png)

nikfilippas:
Renaming `bcm.py` to `baryons.py` will enable future addition of baryonic corrections to the power spectrum, other than the one currently implemented (Schneider & Teysser). Some interest in this has been indicated in #236 .
Also, "baryons" is a better name for a module that deals with baryons, just like there is a module named "neutrinos".

c-d-leonard:
I agree that in general baryons is a better name and not opposed to changing to cells. Obviously need to trade this off against API breaking inconvenience for users, so this would be v3 earliest.

nikfilippas:
No need to break API for any of those. I've implemented that in [this](https://github.com/LSSTDESC/CCL/pull/889/commits/49ce08964f4d28c02d54eb09e729df7e52351fe6) commit of PR #889 .

c-d-leonard:
I was thinking in terms of modules paths but you're right I guess that doesn't technically affect the API.

Issue #878, opened 2021-05-21
foolproof+futureproof cosmology instantiation


nikfilippas:
Currently, CCL cosmology objects can be instantiated without passing any keyword-arguments, like so:
`ccl.Cosmology(0.25, 0.05, 0.65, 0.96, 0.81)`. If for some reason the ordering of arguments in CCL changes in the future, then codes built without explicitly declaring the parameters passed will silently fail, giving the wrong cosmology.

We can make all arguments passed in cosmology to be keyword-only. This way, the user must explicitly state which parameter is which, and this is also futureproof in case the ordering changes and/or more parameters are added.

This is easily done by adding the `*` operator after the `self` argument in every cosmological `__init__` method.
Then, if the user tries to pass in just numbers without keywords, the code will exit with the following message:
```
TypeError: Cosmology() takes 1 positional argument but 6 were given
```

This thinking can be extended to any class method that takes in a number of parameters to update (for example halo profiles), but I think it's more important to do it in `ccl.Cosmology_functions()` first.

rmjarvis:
+1 on this.  This is a really nice feature they added in python 3.  I haven't used it much yet in my own code, since I've still been supporting 2.7, but that's done now.  So I'll be adding this to most of my functions pretty soon.

I think a rule of thumb that will usually work is if the first argument is strongly implied by the name of the function, then let that be keyword-less.  E.g. `readFile(file_name)`.  But then anything after that should be keyword-only.  (Indicated by adding the `*` as Nick mentioned).  For functions that don't have an obvious-from-context first argument (like `Cosmology`) make everything keyword-only.

This makes for much more readable user code, so you're really helping your users out by adding this to help them make their own code better.  `ccl.Cosmology(0.25, 0.05, 0.65, 0.96, 0.81)` is pretty cryptic compared to `ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.65, ns=0.96, sigma_8=0.81)`.  You're doing your users a favor forcing them to use the latter.  :)

damonge:
@rmjarvis would this count as a change to the API and thus trigger a new release?

rmjarvis:
It's arguable.  But personally I'd be happy to let it go as a minor release, not a major one.  It's definitely not a bugfix though.  :)

nikfilippas:
[...]

PR #889, opened 2021-06-12
CCL improvements towards v3 (star, args...)

nikfilippas:
**Note**: Contents from this PR have been branched out to PRs 923, 925, and 933. This PR will then rely on the other three, plus 891 and 906. This is an effort to separate the changes to the API from the new features implemented, and to make review significantly easier.

------------------

This PR addresses a multitude of issues and improves on the CCL API, making things consistent and more straightforward on the journey towards CCLv3. API is preserved and users are notified about things that have changed through a newly implemented `CCLDeprecationWarning`. In this way, transitioning to v3 will be as straightforward as making a copy of this branch and addressing all warnings of that type.

Following is a list of the most notable changes implemented:
- Remove cosmology dependence from halo mass functions, biases, and HMCalculator (brought in from branch `fix_cosmo`).
- Star operator throughout `pyccl` to enhance readability, explicitness, and prevent unpredictable behavior.
- Consistency in the order they appear, as well as the name, of function arguments.
- Rename `bcm.py` to `baryons.py` in light of other baryonic models.
- Address all CCL-generated warnings in unit testing.
- Clean up superfluous arguments from `Cosmology`. Specific models moved to `extra_parameters`.
- Several bug fixes.

-----------------------

**OLD TEXT**:

This breaks API almost completely, so **do not merge** until we make a v3 branch.
**UPDATE:** It doesn't break API anymore!

This PR addresses a multitude of issues (?) and (hopefully) improves on the CCL python code. I thought, since people generally agreed with some of the issues I opened, and since these are inter-related (i.e. it's better to make one big PR instead of many small ones because they address changes on the same lines of code), I make the start by creating this branch, and then we can discuss the technicalities and disagreements (which I foresee, but that's good!).

Checklists of things that need to change:
- [X] main `pyccl`
- [X] `pyccl/halos`
- [x] `pyccl/nl_pt`
- [X] unit tests
- [X] benchmarks
- [X] decorator to preserve API (big task)

Very roughly, this PR includes the following:
- `*` operator implemented. (Issue #880 ). Generally, the changes are somewhere between the 1st and 2nd level of 'strictness', as I have laid it out in the issue thread. Arguments `cosmo`, `hmc`, `k`, `M`, `a`, `r`, which generally are not very informative and are used everywhere in CCL can be declared by position. All others become keyword-only. For example, `rho_x(cosmo, a, species, is_comoving=False)` becomes `rho_x(cosmo, a, species, *, is_comoving=False)`.
- Changed the order of some function/method arguments. (Issue #881 ). I aimed for CCL's 'central' args to appear first (`cosmo,` `hmc`, `k`, `M`, `a`, `r`) and then organized the rest according to their relevance to the function they are in. For example, in the halo model functions the order used to be (`cosmo`, `hmc`, `prof`, `prof_2pt`, `prof2`, `p_of_k_a`, `normprof1`, `normprof2`). I think this order is slightly confusing, so I changed everything to (`cosmo`, `hmc`, all_profiles, all_2pts, all_norms, etc.). Note, that since the star is now implemented the order changing (in most functions) doesn't matter as much; just thought I'd make this change now that this PR breaks API anyway.
- Renamed some of the arguments to make them more self-explanatory, now that the star forces users to declare the keywords. This will improve both writing and reading of the codes using CCL. For example, in `angular_cl`, `cltracer1` and `cltracer2` become `tracer1` and `tracer2` (the `cl` part is inferred by the function name).
- Renamed some arguments/attributes for consistency throughout CCL (also issue #881 ). For example, you won't see `mass_def`, `_mdef` and `m_def` anymore. Just `mass_def`. Or, you won't see `mfunc`, `massfunc`, `hmfunc`. Just `mass_function`, which I think is much more clear.

*Note:* A very neat thing you can do with the star, is that you can have function definitions like:
```
def func(a, *, b=1, c):
    return
```
If `c` is not passed the code will exit with a useful message (that it needs kw-only arg `c` to be passed). We couldn't do this without the star. This allowed me to, for example, remove the default `normprof=False` from the halo model power spectra (while keeping it keyword-only), in order to draw users' attention to this argument.


nikfilippas:
core.Cosmology
    Made all cosmological parameters keyword-only.

core.CosmologyCalculator
    Made all cosmological parameters keyword-only.

cls.angular_cl
    Everything apart from the tracers is keyword-only.
    Aim to call it like `cosmo.angular_cl(tr1, tr2, ell=my_ells)`.
    Rename cltracer --> tracer.

background.rho_x
    `is_comoving` is keyword-only
    Aim to call it like `cosmo.rho_x(0.7, "matter", is_comoving=True)`.

boltzmann.get_camb_pk_lin
    `nonlin` is keyword-only
    Aim to call it like `cosmo.get_camb_pk_lin(nonlin=True)`.

correlations.correlation
    Everything apart from `cosmo` is keyword-only.
    This is to avoid confusion with the ordering of (ell, C_ell).

correlations.correlation_3d
    Everything apart from `cosmo` and `a` is keyword-only.
    This is to avoid confusion with the ordering of (a, r).
    Renamed `r` to `dist` to make it more informative.

correlations.correlation_multipole
    Everything apart from `cosmo` and `a` is keyword-only.
    This is to avoid confusion with the other numerical parameters.
    Swapped (beta, l, s) --> (l, s, beta) because of the function name.
    Renamed `s` to `dist` to make it more informative (and be consistent).
    Renamed `l` to `ell` to make it more informative (and be consistent).

correlations.correlation_3dRsd
    Everything apart from `cosmo` and `a` is keyword-only.
    Args `s`, and `mu` are directly related to the function so left the ordering
    as is (cosmo, a, *, s, mu, beta).
    Renamed `s` to `dist` to make it more informative (and be consistent).

correlations.correlation_3dRsd_avgmu
    Everything apart from `cosmo` and `a` is keyword-only.
    Renamed `s` to `dist` to make it more informative (and be consistent).

correlations.correlation_pi_sigma
    Everything apart from `cosmo` and `a` is keyword-only.
    Swapped (beta, pi, sig) --> (pi, sig, beta) because of the function name.
    Swapped (use_spline, p_of_k_a) --> (p_of_k_a, use_spline) because the
    first is more fundamental and the second has to do with func settings.
    Added `use_spline` in docstring.
    Renamed `sig` to `sigma` to make it more informative.

covariances.angular_cl_cov_cNG
    Everything apart from `cosmo`, `cltracer1` and `cltracer2` is keyword-only.
    Renamed `cltracer` to `tracer`.
    Fixed typo in docstring.
    Renamed `tkka` to `t_of_kk_a` similar to `p_of_k_a` used throughout pyccl.
    Swapped arg ordering: all tracers first (this is what the func is about),
    then `t_of_kk_a`, then the `ell` used, and then the rest of the func settings.

neutrinos.Omeganuh2
    Eveyrhintg apart from `a` is keyword-only.
    Renamed function to `Omega_nu_h2` (readability), and consistency with e.g.
    rho_x (we don't say `rhox`).

neutrinos.nu_masses
    All args are keyword-only.
    Renamed `OmNuh2` to `Om_nu_h2` (readability), since it now becomes kw-only.

pk2d.Pk2D
    All args are keyword-only. The arg names are very helpful when read.
    Swapped arg ordering: (a_arr, lk_arr, pk_arr) go together in this order
    (because pk.shape == (na, nk)).
    Even though (a, lk, pk) are very commonly used, they *must* be declared
    because their order does not match the rest of pyccl (k, a, pk).
    Then (pkfunc, cosmo) go together. 
    Then `is_logp` because it refers to pk (or pkfunc) directly.
    Then, other settings.

pk2d.Pk2D.pk_from_model
    Leave as is because the function name points to arg `model`.

pk2d.Pk2D.apply_halofit
    Arg `pk_linear` is keyword-only.
    Aim to call it like `Pk2D.apply_halofit(cosmo, pk_linear=pkl)`.

pk2d.Pk2D.eval
    Swapped (k, a, cosmo) --> (cosmo, k, a) for consistency with all other funcs.

pk2d.parse_pk2d
    Args `cosmo` is always there.
    Arg`p_of_k_a` are expected from the func name.
    Arg `is_linear` becomes keyword-only.
    Aim to call it like `cosmo.parse_pk2d(my_pk, is_linear=True)`.

power.linear_power & power.nonlin_power
    `p_of_k_a` becomes keyword-only.

power.sigmaR & power.sigmaV & power.sigma8 & power.kNL
    `p_of_k_a` becomes keyword-only.

tk3d.Tk3D
    All args are keyword-only. The arg names are very helpful when read.
    Even though (a, lk, pk) are very commonly used, they *must* be declared
    because their order does not match the rest of pyccl (k, a, pk).
    
tracers._Sig_MG
    Changed `k=None` to just `k`, because `k` is required, or it throws a CCLError.
    It needs distances.

tracers.get_density_kernel
    Arg `dndz` becomes keyword-only.

tracers.get_lensing_kernel
    Args `dndz` and `mag_bias` are keyword-only.

tracers.get_kappa_kernel
    Args `z_source` and `nsamples` are keyword-only.
    Added some very common default values (z=1100, n=100).

tracers.Tracer.get_kernel
    Arg `chi` is keyword-only, because chi is not an obvious argument, and
    pyccl's primary variable for distance is scale factor.

tracers.Tracer.get_transfer
    Changed `lk` to `k`, in accordance to other CCL funcs which take k, a.
    These aren't used to make models out of them (like in Pk2D and Tk3D) so
    it makes sense that users input k instead of lk.
    Amended the docs and the code to reflect this change.

tracers.Tracer.add_tracer
    Everything apart from `cosmo` is keyword-only.
    Swapped `is_logt` to right after the transfer functions are given,
    because it is directly related to them. All other args are settings.

tracers.NumberCountsTracer
    Everything apart from `cosmo` is keyword-only.
    Swapped (has_rsd, dndz, bias, mag_bias) --> (dndz, bias, mag_bias, has_rsd)
    because `has_rsd` is more like a boolean setting, while all others are (z, f(z)).

tracers.WeakLensingTracer
    Everything apart from `cosmo` is keyword-only.
    
tracers.CMBLensingTracer
    Everything apart from `cosmo` is keyword-only.
    Added some very common default values (z=1100).

tracers.tSZTracer
    Everything apart from `cosmo` is keyword-only.

halos.halo_model.HMCalculator
    Arg `cosmo` was removed in PR #888.
    All other args are keyword-only.
    Renamed args (massfunc, hbias) --> (mass_function, halo_bias).
    Renamed attributes (_massfunc, _hbias, _mdef) --> (mass_function, halo_bias, mass_def)
    to simplify acess (and make them public, e.g. might want to check what mass_def is used).
    Renamed args (log10M_min, log10M_max, nlog10M) --> (lM_min, lM_max, nlM)
    to improve readability and make them shorter while they still are self-explanatory.
    Renamed `k_min` --> `k_norm` to differentiate from possible integration bounds
    mix-up.
    Added check for mass_def type.

halos.halo_model.HMCalculator._get_ingredients
    Swapped (a, cosmo) --> (cosmo, a) for consistency.

halos.halo_model.HMCalculator.profile_norm
    (cosmo, a, prof) remains unchanged because `prof` relates directly to func name.

halos.halo_model.HMCalculator.number_counts
    Everything apart from `cosmo` is keyword-only.
    Renamed (amin, amax) --> (a_min, a_max) for consistency.
    Swapped (na, amin, amax) --> (amin, amax, na) for consistency.

halos.halo_model.HMCalculator.I_0_2
    Renamed prof1 --> prof.
    Re-ordered (prof, prof_2pt, prof2) --> (prof, prof2, prof_2pt) to simplify things.
    `prof2` and `prof_2pt` are now keyword-only.

halos.halo_model.HMCalculator.I_0_22
    Renamed prof1 --> prof.
    Re-ordered: (cosmo, k, a, <all profiles>, <all 2pts>)

halos.halo_model.halomod_mean_profile_1pt & halomod_bias_1pt
    `normprof` becomes keyword-only.

halos.halo_model.halomod_power_spectrum
    Everything apart (cosmo, hmc, k, a, prof) becomes keyword-only.
    Re-ordered: (cosmo, hmc, k, a, <profiles>, <2pts>, <norms>, <P_lin>, ...)
    Renamed `normprof1` --> `normprof`.
    Removed default value from `normprof` in order to encourage users to pay attention.

halos.halo_model.halomod_Pk2D
    Everything apart (cosmo, hmc, k, a, prof) becomes keyword-only.
    Re-ordered: (cosmo, hmc, k, a, <profiles>, <2pts>, <norms>, <P_lin>, ...)
    Renamed `normprof1` --> `normprof`.
    Removed default value from `normprof` in order to encourage users to pay attention.
    Removed unecessary arg `cosmo` from Pk2D in the code (because a pk_arr is passed).

halos.halo_model.halomod_trispectrum_1h
    Everything apart (cosmo, hmc, k, a, prof) becomes keyword-only.
    Re-ordered: (cosmo, hmc, k, a, <profiles>, <2pts>, <norms>, 
    Renamed `prof1` --> `prof`
    Renamed `normprof1` --> `normprof`.
    Removed default value from `normprof` in order to encourage users to pay attention.

halos.halo_model.halomod_Tk3D_1h
    Everything apart (cosmo, hmc, k, a, prof) becomes keyword-only.
    Re-ordered: (cosmo, hmc, k, a, <profiles>, <2pts>, <norms>, ...)
    Renamed `prof1` --> `prof`
    Renamed `normprof1` --> `normprof`.
    Removed default value from `normprof` in order to encourage users to pay attention.

halos.concentration.Concentration (all subclasses!)
    Arg `mass_def` becomes keyword-only (for readability, because a class instance
    `MassDef` is expected).
    Renamed `self.mdef` --> `self.mass_def`.
    Renamed args `mdef` --> mass_def everywhere.
    Renamed funcs mdef --> mass_def

halos.concentration.get_concentration
    Arg `mass_def_other` becomes keyword-only.

halos.hbias.HaloBias (all subclasses!)
    Arg `mass_def` becomes keyword-only (for readability, because a class instance
    `MassDef` is expected).
    Renamed `self.mdef` --> `self.mass_def`.
    Renamed args `mdef` --> mass_def everywhere.
    Renamed funcs mdef --> mass_def

halos.hbias.get_halo_bias
    Arg `mass_def_other` becomes keyword-only.

halos.hmfunc.MassFunc (all subclasses!)
    Arg `mass_def` becomes keyword-only (for readability, because a class instance
    `MassDef` is expected).
    Renamed `self.mdef` --> `self.mass_def`.
    Renamed args `mdef` --> mass_def everywhere.
    Renamed funcs mdef --> mass_def
    Renamed c_m --> c_m_relation

halos.hmfunc.get_mass_function
    Arg `mass_def_other` becomes keyword-only.

halos.massdef.convert_concentration
    All args except `cosmo` become keyword-only.

halos.massdef.MassDef
    `c_m_relation` becomes keyword-only (Delta & rho_type are expected).
    
halos.massdef.MassDef.get_mass
    (cosmo, R, a) remain unchanged.

halos.massdef.MassDef.get_radius
    (cosmo, M, a) remain unchanged.

halos.massdef.MassDef.translate_mass
    `m_def_other` becomes keyword-only.

halos.profiles_2pt.Profile2pt
    `r_corr` becomes keyword-only.
    
halos.profiles_2pt.Profile2pt.update_parameters
    `r_corr` becomes keyword-only.

halos.profiles_2pt: Profile2pt.fourier_2pt, Profile2ptHOD.fourier_2pt
    Re-ordered (prof, cosmo, k, M, a, prof2, ...) --> (cosmo, k, M, a, prof, prof2, ...)
    Everything after `prof` is keyword-only.

halos.profiles.HaloProfile: real, fourier, projected, cumul2d
    `mass_def` becomes keyword-only.
    Renamed `r_t` --> `r`.

halos.profiles: HaloProfileGaussian, HaloProfilePowerLaw
    All parameters become keyword-only.
    Renamed attribute `r_s` --> `r_scale` for consistency with the params.

halos.profiles: HaloProfileNFW, HaloProfileEinasto, HaloProfileHernquist
    All parameters become keyword-only.
    Renamed `c_M_relation` --> `c_m_relation` for consistency.
    Renamed attribute `cM` --> `c_m_relation` for consistency with the params.

halos.profiles.HaloProfilePressureGNFW, update_parameters
    All parameters become keyword-only.

halos.profiles.HaloProfileHOD, update_parameters
    All parameters become keyword-only.
    Renamed attribute `cM` --> `c_m_relation` for consistency with the params.




#### UNCHANGED
1. pyccl.background
    growth_factor (`cosmo.growth_factor(0.7)`)
    growth_factor_unnorm (`cosmo.growth_factor_unnorm(0.7)`)
    growth_rate (`cosmo.growth_rate(0.7)`)
    comoving_radial_distance (`cosmo.comoving_radial_distance(0.7)`)
    comoving_angular_distance (`cosmo.comoving_angular_distance(0.7)`)
    angular_diameter_distance (`cosmo.angular_diameter_distance(0.7, 0.3)`)
    h_over_h0 (`cosmo.h_over_h0(0.7)`)
    luminosity_distance (`cosmo.luminosity_distance(0.7)`)
    distance_modulus (`cosmo.distance_modulus(0.7)`)
    scale_factor_of_chi (`cosmo.scale_factor_of_chi(10.)`)
    omega_x (`cosmo.omega_x(0.7, "matter")`)

2. pyccl.bcm
    bcm_model_fka (`cosmo.bcm_model_fka(k_arr, 1.)`)
    bcm_correct_pk2d (`cosmo.bcm_correct_pk2d(my_pk)`)

3. pyccl.power
    linear_matter_power (`cosmo.linear_matter_power(k_arr, 1.)`)
    nonlin_matter_power (`cosmo.nonlin_matter_power(k_arr, 1.)`)
    sigmaM (`cosmo.sigmaM(M_arr, 1.)`)

4. pyccl.pyutils
    assert_warns



**Note:** All private methods (or functions starting with "_") remain unchanged.
**Note:** All functions and methods that depend only on `cosmo` remain unchanged.
**Note:** All the deprecated functions remain unchanged.

nikfilippas:
**UPDATE**

Now, all the changes preserve API! :smile: 

I have written two decorators (in `pyutils.py`) which take care of all changes to function names, method names, and attribute names. I assumed that private functions ought not to be called (strictly speaking nothing in Python is private), so I haven't added the decorator to any private functions that have been changed. Only the pubic ones.

The API-preserving wrapper only adds a few milliseconds to each function call, and I've tried to make it optimal. Suggestions are welcome.

nikfilippas:
Okay, now that the coverage is net-positive, the next step (and ultimate test) is to check if the old tests and benchmarks run fine without errors. This is what the following commits are about: I am bringing back the old unit tests and benchmarks from master, but once they run fine I will replace them with the ones in this PR.

nikfilippas:
Okay, now the old tests and old benchmarks work fine! You can see they all ran successfully by clicking the red 'X' mark on the last commit above this comment. Never mind the coverage, as we're only interested in the tests.

The only changes I made were:
1. added `normprof1` as an argument in 4 different places in the tests (`normprof` is a required argument because I removed the default False value, in order to get users to notice it by explicitly declaring it);
2. fixed a bug in `test_profiles.py` where `pytest` effectively wasn't testing because the wrong argument was passed;
3. changed error of type to be `TypeError` instead of `ValueError` in `covariances.py` and its tests.

Only the first of those could be considered as an API breakage (and we could address that by setting a `normprof` default again), but I think it's best if past codes that don't include that **do** break in order to avoid mighty errors!

I am now bringing back the new benchmarks and tests, keeping the slight modifications I made to the actual code.

nikfilippas:
And (finally), I have relaxed the requirement to explicitly state if the halo profile needs to be normalized, by setting it `normprof=None` by default, and then simply throwing a warning if the user has not passed it in the function call.
This concludes the series of "preserving the API" changes, and now this PR could (in theory) be merged to master without a version bump, as it does not break API at all anymore. :smiley:

pyccl/halos/massdef.py
```diff
@@ -225,47 +229,52 @@ def translate_mass(self, cosmo, M, a, m_def_other):
...
 class MassDef200m(MassDef):
     """`MassDef` class for the mass definition with Delta=200 times the matter
     density.
 
     Args:
-        c_m (string): concentration-mass relation.
+        c_m_relation (string): concentration-mass relation.
     """
-    def __init__(self, c_m='Duffy08'):
+    @warn_api()
```
damonge, line 248: should you rename `c_m`->`c_m_reation` in `warn_api`?

pyccl/halos/massdef.py
```diff
@@ -225,47 +229,52 @@ def translate_mass(self, cosmo, M, a, m_def_other):
...
 class MassDefVir(MassDef):
     """`MassDef` class for the mass definition with Delta=Delta_vir times the
     critical density.
 
     Args:
-        c_m (string): concentration-mass relation.
+        c_m_relation (string): concentration-mass relation.
     """
-    def __init__(self, c_m='Klypin11'):
+    @warn_api()
```
damonge, line 276: as above

pyccl/halos/halo_model.py
```diff
@@ -30,46 +31,47 @@ class HMCalculator(object):
...
-        if not isinstance(hbias, HaloBias):
-            raise TypeError("hbias must be of type `HaloBias`")
-        self._hbias = hbias
-        self._mdef = mass_def
-        self._prec = {'log10M_min': log10M_min,
-                      'log10M_max': log10M_max,
-                      'nlog10M': nlog10M,
+        if not isinstance(mass_function, MassFunc):
+            raise TypeError("mass_function must be of type `MassFunc`")
+        self.mass_function = mass_function
```
damonge, line 63: we had intended for these attributes to be private (i.e. not meant to be accessed by the user), so I'd keep the `_`

pyccl/halos/halo_model.py
```diff
@@ -89,6 +91,12 @@ def __init__(self, cosmo, massfunc, hbias, mass_def,
         self._a_current_mf = -1
         self._a_current_bf = -1
 
+    @deprecate_attr(pairs=[("mass_function", "_massfunc"),
+                           ("halo_bias", "_hbias"),
+                           ("mass_def", "_mdef")])
```
damonge, line 96: these ones, I meant

pyccl/halos/halo_model.py
```diff
@@ -318,44 +336,44 @@ def I_0_22(self, cosmo, k, a,
...
+        uk34 = prof34_2pt.fourier_2pt(cosmo, k, self._mass, a, prof3,
+                                      prof2=prof4, mass_def=self.mass_def).T
         i04 = self._integrate_over_mf(uk12[None, :, :] * uk34[:, None, :])
         return i04
 
 
-def halomod_mean_profile_1pt(cosmo, hmc, k, a, prof,
-                             normprof=False):
+@warn_api()
+def halomod_mean_profile_1pt(cosmo, hmc, k, a, prof, *, normprof=None):
```
damonge, line 376: Why `None` instead of `False`?

pyccl/halos/halo_model.py
```diff
@@ -406,7 +424,8 @@ def halomod_mean_profile_1pt(cosmo, hmc, k, a, prof,
     return out
 
 
-def halomod_bias_1pt(cosmo, hmc, k, a, prof, normprof=False):
+@warn_api()
+def halomod_bias_1pt(cosmo, hmc, k, a, prof, *, normprof=None):
```
damonge, line 428: same here

pyccl/halos/halo_model.py
```diff
@@ -460,9 +479,12 @@ def halomod_bias_1pt(cosmo, hmc, k, a, prof, normprof=False):
...
 
 
-def halomod_power_spectrum(cosmo, hmc, k, a, prof,
-                           prof_2pt=None, prof2=None, p_of_k_a=None,
-                           normprof1=False, normprof2=False,
+@warn_api(pairs=[("normprof", "normprof1")],
+          order=["prof_2pt", "prof2", "p_of_k_a", "normprof", "normprof2"])
+def halomod_power_spectrum(cosmo, hmc, k, a, prof, *,
+                           prof2=None, prof_2pt=None,
+                           normprof=None, normprof2=False,
```
damonge, line 486: again with the `None`

pyccl/halos/halo_model.py
```diff
@@ -530,7 +552,7 @@ def halomod_power_spectrum(cosmo, hmc, k, a, prof,
         `k` and `a` respectively. If `k` or `a` are scalars, the
         corresponding dimension will be squeezed out on output.
     """
-    a_use = np.atleast_1d(a)
+    a_use = np.atleast_1d(a).astype(float)
```
damonge, line 555: Why astype(float) here but not for `k_use`?

pyccl/halos/halo_model.py
```diff
@@ -714,22 +740,24 @@ def halomod_Pk2D(cosmo, hmc, prof,
...
-def halomod_trispectrum_1h(cosmo, hmc, k, a,
-                           prof1, prof2=None, prof12_2pt=None,
-                           prof3=None, prof4=None, prof34_2pt=None,
-                           normprof1=False, normprof2=False,
+@warn_api(pairs=[("prof", "prof1"), ("normprof", "normprof1")],
+          order=["prof12_2pt", "prof3", "prof4"])
+def halomod_trispectrum_1h(cosmo, hmc, k, a, prof, *,
+                           prof2=None, prof3=None, prof4=None,
+                           prof12_2pt=None, prof34_2pt=None,
+                           normprof=None, normprof2=False,
```
damonge, line 760: `None`

pyccl/halos/halo_model.py
```diff
@@ -853,10 +882,12 @@ def get_norm(normprof, prof, sf):
...
-def halomod_Tk3D_1h(cosmo, hmc,
-                    prof1, prof2=None, prof12_2pt=None,
-                    prof3=None, prof4=None, prof34_2pt=None,
-                    normprof1=False, normprof2=False,
+@warn_api(pairs=[("prof", "prof1"), ("normprof", "normprof1")],
+          order=["prof12_2pt", "prof3", "prof4"])
+def halomod_Tk3D_1h(cosmo, hmc, prof, *,
+                    prof2=None, prof3=None, prof4=None,
+                    prof12_2pt=None, prof34_2pt=None,
+                    normprof=None, normprof2=False,
```
damonge, line 890: `None`

pyccl/halos/hmfunc.py
```diff
@@ -464,21 +485,22 @@ def _setup(self, cosmo):
                         -0.301, -0.301, -0.319, -0.336])
 
         ldelta = np.log10(delta)
-        self.pA0 = interp1d(ldelta, alpha)
-        self.pa0 = interp1d(ldelta, eta)
-        self.pb0 = interp1d(ldelta, beta)
-        self.pc0 = interp1d(ldelta, gamma)
-        self.pd0 = interp1d(ldelta, phi)
+        extrap_kw = {"bounds_error": False, "fill_value": "extrapolate"}
```
damonge, line 488: We want this to fail if a user passes values for `ldelta` beyond the Tinker interpolation ranges, since things are not tested in that regime.

pyccl/halos/hmfunc.py
```diff
@@ -596,15 +620,18 @@ def _setup(self, cosmo):
...
-        elif int(mdef.Delta) == 200:
-            if (mdef.rho_type != 'matter') and \
-               (mdef.rho_type != 'critical'):
-                return True
-        elif int(mdef.Delta) == 500:
-            if mdef.rho_type != 'critical':
+        # NOTE: The following lines have been commented out because
+        # currently rho_type can only be 'matter' or 'critical' and
+        # so this case cannot be covered.
+        elif int(mass_def.Delta) == 200:
```
damonge, line 629: but you're not checking anywhere if `rho_type` is something else, right?

pyccl/halos/profiles.py
```diff
@@ -147,7 +151,8 @@ def _get_plaw_projected(self, cosmo, a):
         """
         return self.precision_fftlog['plaw_projected']
 
-    def real(self, cosmo, r, M, a, mass_def=None):
+    @warn_api(pairs=[("r", "rt")])
```
damonge, line 154: hold on, what's happening here? I don't see `rt` anywhere

pyccl/halos/profiles.py
```diff
@@ -212,7 +218,8 @@ def fourier(self, cosmo, k, M, a, mass_def=None):
                                       " _fourier method.")
         return f_k
 
-    def projected(self, cosmo, r_t, M, a, mass_def=None):
+    @warn_api(pairs=[("r", "rt")])
```
damonge, line 221: same here. Where is `rt`?

damonge: Note that it makes sense for this variable to be `r_t` and not `r`, since this is a transverse distance.

pyccl/halos/profiles.py
```diff
@@ -236,14 +243,15 @@ def projected(self, cosmo, r_t, M, a, mass_def=None):
             squeezed out on output.
         """
         if getattr(self, '_projected', None):
-            s_r_t = self._projected(cosmo, r_t, M, a, mass_def)
+            s_r_t = self._projected(cosmo, r, M, a, mass_def)
```
damonge, line 246: `r_t` is better.

pyccl/halos/profiles.py
```diff
@@ -236,14 +243,15 @@ def projected(self, cosmo, r_t, M, a, mass_def=None):
...
         else:
-            s_r_t = self._projected_fftlog_wrap(cosmo, r_t, M,
+            s_r_t = self._projected_fftlog_wrap(cosmo, r, M,
```
damonge, line 248: same here

pyccl/halos/profiles.py
```diff
@@ -236,14 +243,15 @@ def projected(self, cosmo, r_t, M, a, mass_def=None):
...
                                                 a, mass_def,
                                                 is_cumul2d=False)
         return s_r_t
 
-    def cumul2d(self, cosmo, r_t, M, a, mass_def=None):
+    @warn_api(pairs=[("r", "rt")])
+    def cumul2d(self, cosmo, r, M, a, *, mass_def=None):
```
damonge, line 254: same here about `r_t`

pyccl/halos/profiles.py
```diff
@@ -268,9 +276,9 @@ def cumul2d(self, cosmo, r_t, M, a, mass_def=None):
             squeezed out on output.
         """
         if getattr(self, '_cumul2d', None):
-            s_r_t = self._cumul2d(cosmo, r_t, M, a, mass_def)
+            s_r_t = self._cumul2d(cosmo, r, M, a, mass_def)
```
damonge, line 279: `r_t`

pyccl/halos/profiles.py
```diff
@@ -268,9 +276,9 @@ def cumul2d(self, cosmo, r_t, M, a, mass_def=None):
...
         else:
-            s_r_t = self._projected_fftlog_wrap(cosmo, r_t, M,
+            s_r_t = self._projected_fftlog_wrap(cosmo, r, M,
```
damonge, line 281: `r_t`

pyccl/halos/profiles.py
```diff
@@ -409,18 +417,19 @@ class HaloProfileGaussian(HaloProfile):
...
-            units of M_sun, `a` is the scale factor and `mdef`
+            units of M_sun, `a` is the scale factor and `mass_def`
             is a :class:`~pyccl.halos.massdef.MassDef` object.
         rho0 (:obj:`function`): the amplitude of the profile.
             It should have the same signature as `r_scale`.
     """
     name = 'Gaussian'
 
-    def __init__(self, r_scale, rho0):
+    @warn_api(pairs=[("r_scale", "r_s")])
```
damonge, line 429: why are you doing this? It seems you're not really renaming it into `r_s`, right?

pyccl/halos/profiles.py
```diff
@@ -454,17 +463,18 @@ class HaloProfilePowerLaw(HaloProfile):
...
-            `mdef` is a :class:`~pyccl.halos.massdef.MassDef` object.
+            `mass_def` is a :class:`~pyccl.halos.massdef.MassDef` object.
         tilt (:obj:`function`): the power law index of the
             profile. The signature of this function should
             be `f(cosmo, a)`.
     """
     name = 'PowerLaw'
 
-    def __init__(self, r_scale, tilt):
+    @warn_api(pairs=[("r_scale", "r_s")])
```
damonge, line 476: same here, I'm confused

damonge (commented):
OK, huge effort and a great ordeal to review, but here we are.
My main concern is that there's a lot of extra code needed just to maintain the old API. I think this is often worth it when it's not as systemic as this, but in this case I might actually be more inclined to be ruthless and just break the API for v3 rather than have all this dangling code around. However, this seems like a large enough decision to be discussed with others, so let's bring it to slack. In the meantime, I've added lots of comments on naming choices and potential bugs. Take a look.

If we do end up keeping all the new api-preserving code I'll ask you to write a few more unit tests just to make sure it does what it's supposed to (right now I'm worried that there seem to be some bugs in the actual calls to `warn_api` and yet you're not seeing anything break in the unit tests).

nikfilippas:
@damonge okay, let's move this to Slack. In the meantime, I'm replying to the individual comments.

pyccl/__init__.py
```diff
@@ -62,7 +62,7 @@
 from .haloprofile import nfw_profile_3d, einasto_profile_3d, hernquist_profile_3d, nfw_profile_2d
 
 # Specific to massive neutrinos
-from .neutrinos import Omeganuh2, nu_masses
+from .neutrinos import Omega_nu_h2, Omeganuh2, nu_masses
```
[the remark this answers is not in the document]
nikfilippas: If you're asking why am I renaming this, it's because I find "Omeganuh2" extremely hard to read. Python universally adopts `snake_case` while C-like languages adopt `SwapCase`. Similar to how we have `halomod_power_spectrum` rather than `HalomodPowerSpectrum`.

pyccl/pk2d.py
```diff
@@ -169,12 +175,10 @@ def apply_halofit(Pk2D, cosmo, pk_linear):
         pk2d.has_psp = True
         return pk2d
 
-    def eval(self, k, a, cosmo):
+    def eval(self, cosmo, k, a):
```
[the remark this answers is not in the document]
nikfilippas: This alone, yes, but there's a patch in all 3 class methods where this happens. It is in the lines below. Copying it here too:
```
        # patch to check if new or old API is used
        from .core import Cosmology
        if not isinstance(cosmo, Cosmology):
            warnings.warn("Official API for Pk2D.eval has changed. "
                          "Argument order (k, a, cosmo) has been replaced "
                          "by (cosmo, k, a).", CCLWarning)
            k, a, cosmo = cosmo, k, a  # old to new API
            assert isinstance(cosmo, Cosmology)
```

pyccl/tests/test_star.py
```diff
@@ -0,0 +1,127 @@
...
+HMC = ccl.halos.HMCalculator(COSMO, mass_function=HMF,
+                             halo_bias=HBF, mass_def=M200)
+PROF = ccl.halos.HaloProfileHOD(c_m_relation=CON)
+COV = ccl.halos.Profile2pt()
+COVh = ccl.halos.Profile2ptHOD()
+PK2D = ccl.boltzmann.get_camb_pk_lin(COSMO)
+
+
+def test_API_preserve_warnings():
+    # 0. no warnings for the following examplary functions
```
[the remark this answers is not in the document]
nikfilippas: OK

pyccl/pyutils.py
```diff
@@ -518,3 +519,216 @@ def assert_warns(wtype, f, *args, **kwargs):
...
+                    news, olds = np.asarray(rename).T.tolist()
+                    news = news[0] if len(rename) == 1 else news
+                    olds = olds[0] if len(rename) == 1 else olds
+                    warnings.warn(
+                        f"Use of argument{s} {olds} is deprecated "
+                        f"in {func.__name__}. Pass the new name{s} of the "
+                        f"argument{s} {news}, respectively.",
+                        CCLWarning)
+
+            # raise normprof warning? (1 of 2)
```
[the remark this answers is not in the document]
nikfilippas: My thinking is "warn about imminent change before making said change". If we don't warn, users won't be aware, and once we totally get rid of the `normprof` default, their codes will inevitably break.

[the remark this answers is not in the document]
nikfilippas: Extra lines of code are just extra lines of code. It's well-commented, and no one really looks at `pyutils`. What matters is timing, and adding these two extra checks takes almost no time.

pyccl/pyutils.py
```diff
@@ -518,3 +519,216 @@ def assert_warns(wtype, f, *args, **kwargs):
...
+                for key in kwargs:
+                    if key in swap:
+                        swap.remove(key)
+
+            # include new positionals to kwarg dict
+            for name, value in zip(pos_names, args):
+                kwargs[name] = value
+            args = args[npos:]
+
+            # deal with the remaining arguments
```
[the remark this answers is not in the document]
nikfilippas: I know right, it's really the best I could do. I added many comments and made all variables almost self-explanatory aiming to read the code like natural language. I can explain it line-by-line in the next meeting, verbally.

pyccl/pyutils.py
```diff
@@ -518,3 +519,216 @@ def assert_warns(wtype, f, *args, **kwargs):
...
+            # warn about API change
+            no_kw = names[npos : npos + len(args)]  # noqa
+            s = "" if len(no_kw) == 1 else "s"
+            no_kw = f"`{no_kw[0]}`" if len(no_kw) == 1 else no_kw
+            warnings.warn(
+                f"Use of argument{s} {no_kw} as positional is deprecated "
+                f"in {func.__name__}. Pass the name{s} of the "
+                f"keyword-only argument{s} explicitly.", CCLWarning)
+
+            # raise normprof warning? (2 of 2)
```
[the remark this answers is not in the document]
nikfilippas: I replied to this above.

pyccl/correlations.py
```diff
@@ -111,14 +113,15 @@ def correlation(cosmo, ell, C_ell, theta, type='NN', corr_type=None,
     return wth
 
 
-def correlation_3d(cosmo, a, r, p_of_k_a=None):
+@warn_api(pairs=[("dist", "r")])
```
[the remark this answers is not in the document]
nikfilippas: DISCUSSION POINT 1

pyccl/correlations.py
```diff
@@ -182,28 +186,30 @@ def correlation_multipole(cosmo, a, beta, l, s, p_of_k_a=None):
...
+    xis, status = lib.correlation_multipole_vec(cosmo, psp, a, beta, ell,
+                                                dist, len(dist), status)
     check(status, cosmo_in)
     if scalar:
         return xis[0]
     return xis
 
 
-def correlation_3dRsd(cosmo, a, s, mu, beta, use_spline=True, p_of_k_a=None):
+@warn_api(pairs=[("dist", "s")])
```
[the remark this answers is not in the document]
nikfilippas: DISCUSSION POINT 3

pyccl/correlations.py
```diff
@@ -230,28 +236,30 @@ def correlation_3dRsd(cosmo, a, s, mu, beta, use_spline=True, p_of_k_a=None):
...
+                                            len(dist), int(use_spline),
+                                            status)
     check(status, cosmo_in)
     if scalar:
         return xis[0]
     return xis
 
 
-def correlation_3dRsd_avgmu(cosmo, a, s, beta, p_of_k_a=None):
+@warn_api(pairs=[("dist", "s")])
```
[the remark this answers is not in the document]
nikfilippas: DISCUSSION POINT 4

pyccl/halos/massdef.py
nikfilippas: OK

pyccl/halos/massdef.py
```diff
@@ -225,47 +229,52 @@ def translate_mass(self, cosmo, M, a, m_def_other):
...
 class MassDef200c(MassDef):
     """`MassDef` class for the mass definition with Delta=200 times the critical
     density.
 
     Args:
-        c_m (string): concentration-mass relation.
+        c_m_relation (string): concentration-mass relation.
     """
-    def __init__(self, c_m='Duffy08'):
+    @warn_api()
```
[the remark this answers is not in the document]
nikfilippas: OK

pyccl/halos/massdef.py
nikfilippas: OK

pyccl/halos/profiles_2pt.py
```diff
@@ -23,28 +26,27 @@ class Profile2pt(object):
...
+    def update_parameters(self, *, r_corr=None):
         """ Update any of the parameters associated with this 1-halo
         2-point correlator. Any parameter set to `None` won't be updated.
         """
         if r_corr is not None:
             self.r_corr = r_corr
 
-    def fourier_2pt(self, prof, cosmo, k, M, a,
-                    prof2=None, mass_def=None):
+    def fourier_2pt(self, cosmo, k, M, a, prof, *,
```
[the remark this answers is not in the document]
nikfilippas: Replied above.

pyccl/halos/profiles_2pt.py
```diff
@@ -67,6 +72,16 @@ def fourier_2pt(self, prof, cosmo, k, M, a,
...
+        # patch to check if new or old API is used
+        from ..core import Cosmology
+        if not isinstance(cosmo, Cosmology):
+            warnings.warn("Official API for Profile2pt.fourier_2pt "
+                          "has changed. Argument order "
+                          "(prof, cosmo, k, M, a) has been replaced by "
+                          "(cosmo, k, M, a, prof).", CCLWarning)
+            prof, cosmo, k, M, a = cosmo, k, M, a, prof  # old to new API
+            assert isinstance(cosmo, Cosmology)
+
```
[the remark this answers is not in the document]
nikfilippas: It's only a matter of `grep -r CCLWarning`. If it makes it any better, I could even write-up a specific `CCLDeprecationWarning`.

Overall if we run `grep -r warn_api` and `grep -r CCLWarning` we can get rid of everything in under 10 minutes (don't take the time too literally).

pyccl/neutrinos.py
```diff
@@ -11,7 +12,8 @@
 }
 
 
-def Omeganuh2(a, m_nu, T_CMB=None):
+@warn_api()
+def Omega_nu_h2(a, *, m_nu, T_CMB=None):
```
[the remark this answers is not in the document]
nikfilippas: Replied above:
If you're asking why am I renaming this, it's because I find `Omeganuh2` extremely hard to read. Python universally adopts `snake_case` while C-like languages adopt `SwapCase`. Similar to how we have `halomod_power_spectrum` rather than `HalomodPowerSpectrum`.

pyccl/neutrinos.py
```diff
@@ -49,12 +51,19 @@ def Omeganuh2(a, m_nu, T_CMB=None):
     return OmNuh2
 
 
-def nu_masses(OmNuh2, mass_split, T_CMB=None):
+@deprecated(Omega_nu_h2)
+@warn_api()
+def Omeganuh2(a, *, m_nu, T_CMB=None):
+    return Omega_nu_h2(a, m_nu=m_nu, T_CMB=T_CMB)
```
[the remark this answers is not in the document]
nikfilippas: My brain subconsciously reads it as "omegana-two" instead of "omega-nu-eitch-squared".

pyccl/neutrinos.py
```diff
@@ -49,12 +51,19 @@ def Omeganuh2(a, m_nu, T_CMB=None):
...
+
+
+@warn_api(pairs=[("Om_nu_h2", "OmNuh2")])
+def nu_masses(*, Om_nu_h2, mass_split, T_CMB=None):
```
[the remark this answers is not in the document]
nikfilippas: DISCUSSION POINT 5

pyccl/pk2d.py
```diff
@@ -183,10 +187,21 @@ def eval(self, k, a, cosmo):
...
         """
+        # patch to check if new or old API is used
+        from .core import Cosmology
+        if not isinstance(cosmo, Cosmology):
+            warnings.warn("Official API for Pk2D.eval has changed. "
+                          "Argument order (k, a, cosmo) has been replaced "
+                          "by (cosmo, k, a).", CCLWarning)
+            k, a, cosmo = cosmo, k, a  # old to new API
+            assert isinstance(cosmo, Cosmology)
+
```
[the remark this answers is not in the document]
nikfilippas: Yes, replied above.

pyccl/halos/halo_model.py
```diff
@@ -30,46 +31,47 @@ class HMCalculator(object):
...
+        lM_min, lM_max (float): lower and upper integration bounds
+            of logarithmic (base-10) mass (in units of solar mass).
+            Default range: 8, 16.
+        nlM (int): number of uniformly-spaced samples in log(Mass)
+            to be used in the mass integrals. Default: 128.
         integration_method_M (string): integration method to use
             in the mass integrals. Options: "simpson" and "spline".
             Default: "simpson".
-        k_min (float): some of the integrals solved by this class
+        k_norm (float): some of the integrals solved by this class
```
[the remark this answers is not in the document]
nikfilippas: That's a good point, but I find `k_long` a bit awkward. How about `k_large_scale`?

pyccl/halos/halo_model.py
```diff
@@ -30,46 +31,47 @@ class HMCalculator(object):
...
             scales. This parameter (in units of inverse Mpc)
             determines what is considered a "very large" scale.
             Default: 1E-5.
     """
-    def __init__(self, cosmo, massfunc, hbias, mass_def,
-                 log10M_min=8., log10M_max=16.,
-                 nlog10M=128, integration_method_M='simpson',
-                 k_min=1E-5):
+    @warn_api(pairs=[("mass_function", "massfunc"), ("halo_bias", "hbias"),
+                     ("lM_min", "log10M_min"), ("lM_max", "log10M_max"),
```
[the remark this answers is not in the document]
nikfilippas: I'm fine with this, but then it won't be consistent with `HaloProfileHOD`'s `lM1`, `lMmin` etc. Any suggestions?

pyccl/halos/halo_model.py
```diff
@@ -30,46 +31,47 @@ class HMCalculator(object):
...
+                     ("nlM", "nlog10M"), ("k_norm", "k_min")])
```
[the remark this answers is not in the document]
nikfilippas: I think `nM` does the job equally well here, what do you think?

pyccl/halos/halo_model.py
nikfilippas: OK. Why though? Someone might want to do something like `isinstance(hmc.mass_function, MassFuncTinker08)`.

pyccl/halos/halo_model.py
nikfilippas: OK this is addressed above, resolving it.

pyccl/halos/halo_model.py
nikfilippas: Because `False` is a valid normalization.

I wanted to use something that is not `True` or `False` (because these are the only choices), as a placeholder in order not to break API. The decorator checks whether `normprof is None`, and if it is, it warns the user.
If we leave it empty, we break API. The default is intended to go completely in the next major version.

pyccl/halos/halo_model.py
nikfilippas: Replied above.

pyccl/halos/halo_model.py
nikfilippas: Replied above.

pyccl/halos/halo_model.py
nikfilippas: If you pass an integer scale factor (i.e. 1) the code breaks, so it was a quick patch. Don't think too much of it, I've taken care of this kind of error in #891 .

pyccl/halos/halo_model.py
```diff
@@ -628,9 +651,12 @@ def pkf(sf):
...
 
 
-def halomod_Pk2D(cosmo, hmc, prof,
-                 prof_2pt=None, prof2=None, p_of_k_a=None,
-                 normprof1=False, normprof2=False,
+@warn_api(pairs=[("normprof", "normprof1")],
+          order=["prof_2pt", "prof2", "p_of_k_a", "normprof", "normprof2"])
+def halomod_Pk2D(cosmo, hmc, prof, *,
+                 prof2=None, prof_2pt=None,
+                 normprof=None, normprof2=False,
```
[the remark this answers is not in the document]
nikfilippas: Replied above.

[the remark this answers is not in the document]
nikfilippas: I see. This one doesn't really matter. I could change it for consistency. Shall I change it?

pyccl/halos/halo_model.py
nikfilippas: Comment above.

pyccl/halos/halo_model.py
nikfilippas: Comment above.

pyccl/halos/hmfunc.py
nikfilippas: The tests are failing for some reason though. There is one test that uses a value slightly outside of the range. And these are tests that were already there before I make any changes. I'll get back to you on that.

pyccl/halos/hmfunc.py
nikfilippas: I commented it out because it wasn't covered in the unit tests. This PR is a catch-all kind of PR, so there are some instances where I've fixed bugs, coverage etc.
I could bring those lines back, but they will stay uncovered.

pyccl/halos/profiles.py
nikfilippas: OK. Thanks, good catch.

pyccl/halos/profiles.py
nikfilippas: How about `r_transverse`?

nikfilippas: By the way, the `master` unit tests not failing when I tested them means that a big chunk of `pyccl` code is still not covered.

pyccl/halos/profiles.py
nikfilippas: Maybe `r_transverse`?

pyccl/halos/profiles.py
nikfilippas: Suggestion above.

pyccl/halos/profiles.py
nikfilippas: Suggestion above.

pyccl/halos/profiles.py
nikfilippas: Suggestion above.

pyccl/halos/profiles.py
nikfilippas: Suggestion above.

pyccl/halos/profiles.py
nikfilippas: OK. Good catch.

EiffL:
Thanks so much Nick for taking on this huge task! I think the changes you propose make sense. But I'm very puzzled by this use of the * operator, I have to say I've never seen it used that way in other (professional) python libraries. Is it reaaaally needed?

I agree that functionally if the user uses the API correctly it's transparent for them, but this still seems overkill to me.... Do you have an example in the wild of a library using that convention? 

Also ;-) if you want to go down this road of stricter argument checks, there is also this new trend of using python typing for inputs and output arguments.

nikfilippas:
Thanks Francois!

I think there is a sound reason why we should use the star. Motivation and previous discussion is in #880 and #878 , but to put it briefly here, currently `ccl.Cosmology(0.25, 0.05, 0.65, 0.96, 0.81)` is a valid call, and this begs for various problems.

Major Python libraries have long implemented the star. [Here](https://numpy.org/doc/stable/reference/generated/numpy.array.html) is an example in Numpy, [here](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html) is an example in Matplotlib, and [here](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.NearestNeighbors.html) is an example in scikit-learn.

EiffL:
Annnnnd I now stand corrected! Thanks for educating me ^^ the fact that I didn't realize this star was there probably means it's innocuous I practice ;-)

Star away!

nikfilippas:
What do people think of #886 ? Implement it in this PR, or close it?

pyccl/halos/halo_model.py
nikfilippas: Do **not** resolve this. I will revert it back to False, since it is not passed in the `kwargs` dictionary if the user doesn't explicitly declare it anyway. This will simplify things.

pyccl/correlations.py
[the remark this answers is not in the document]
nikfilippas: I agree with the ordering. In `pyccl.halos` we use `k, M, a` everywhere. So distance goes before scale factor. Also, like you said, `r` is the obvious function argument.
Would you say `correlation3d(cosmo, r, a, *, p_of_k_a=None)` is better?

pyccl/neutrinos.py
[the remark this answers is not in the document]
nikfilippas: Because the first argument is `Neutrino energy density at z=0 times h^2` which I think is not at all obvious. So the first argument has to be kw-only. Then, `mass_split` is easier to understand, but it's after `Om_nu_h2`, so it can't be positional.

pyccl/pk2d.py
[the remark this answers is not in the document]
nikfilippas: I think we've resolved this now, right?

pyccl/halos/profiles.py
nikfilippas: BTW the unit tests from master didn't fail on that bug because all `r`, `rt` and `r_t` are treated as positional everywhere.

nikfilippas: Also, the fact that it is a transverse distance should be reflected in the docstring. Currently, the `r_t` has the same explanation as the `r` in `HaloProfile.real`.

nikfilippas:
**UPDATE**
Work towards this PR is now complete. Once the discussion points and the reviewers' comments have been resolved, I will duplicate this PR with the API-breaking, faster version of it.

pyccl/halos/halo_model.py
```diff
@@ -1005,7 +1006,10 @@ def halomod_Tk3D_1h(cosmo, hmc, prof, *,
                                  normprof=normprof, normprof2=normprof2,
                                  normprof3=normprof3, normprof4=normprof4)
     if use_log:
-        if np.any(tkk <= 0):
+        # avoid zeros (this is system-dependent)
```
nikfilippas, line 1009:
```
pyccl/tests/test_tkk1h.py::test_tkk1h_tk3d
  /home/runner/work/CCL/CCL/pyccl/halos/halo_model.py:143: RuntimeWarning: divide by zero encountered in double_scalars
    norm = 1. / self._integrate_over_mf(uk0)

pyccl/tests/test_tkk1h.py::test_tkk1h_tk3d
  /home/runner/work/CCL/CCL/pyccl/halos/halo_model.py:887: RuntimeWarning: invalid value encountered in multiply
    out[ia, :, :] = tk_1h * norm
```

pyccl/halos/halo_model.py
nikfilippas: Right, so I reverted the default back to `False` now, instead of `None`. The decorator will still throw a warning if `normprof` has not been specified. The only difference is that, in functions accepting `normprof` which call other functions accepting `normprof`, there won't be a double warning anymore (it's not needed anyway).

We can now resolve this, and the ones below.

pyccl/halos/halo_model.py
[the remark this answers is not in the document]
nikfilippas: Resolved above.

pyccl/halos/halo_model.py
nikfilippas: Resolved above.

pyccl/halos/halo_model.py
nikfilippas: Resolved.

nikfilippas:
The commits above simplify `__repr__` in `Cosmology`. At the same time, it justifies the output string for clean-looking and easily readable output. Method `__str__` drops the parameters which are the same as the default ones, while still giving an `eval`-able output string. This addition is useful for debugging purposes.

Doing that, I also noted that `extra_parameters` weren't parsed by the `__repr__` method, so `eval(repr(cosmo))` would be inconsistent. I fixed that.

```
# change this
pyccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7, n_s=0.96, sigma8=None,
A_s=2.1e-09, Omega_k=0.0, Omega_g=None, Neff=3.046, w0=-1.0, wa=0.0, T_CMB=None,
bcm_log10Mc=14.079181246047625, bcm_etab=0.5, bcm_ks=55.0, mu_0=[0.02, 0.1, 0.05], 
sigma_0=0.0, c1_mg=1.0, c2_mg=1.0, lambda_mg=0.0, extra_parameters=None, m_nu=0.0, 
m_nu_type='list', z_mg=[0.0, 1.0] , df_mg=[0.01, 0.0], transfer_function='boltzmann_camb', 
matter_power_spectrum='halofit', baryons_power_spectrum='nobaryons', 
mass_function='tinker10', halo_concentration='duffy2008', emulator_neutrinos='strict')

# to this (using `repr(cosmo)` or just `>>> cosmo` in the console)

pyccl.Cosmology(
   Omega_c                 =                0.25 ,
   Omega_b                 =                0.05 ,
   h                       =                 0.7 ,
   n_s                     =                0.96 ,
   sigma8                  =                None ,
   A_s                     =             2.1e-09 ,
   Omega_k                 =                 0.0 ,
   Omega_g                 =                None ,
   Neff                    =               3.046 ,
   m_nu                    =   [0.02, 0.1, 0.05] ,
   m_nu_type               =              'list' ,
   w0                      =                -1.0 ,
   wa                      =                 0.0 ,
   T_CMB                   =                None ,
   bcm_log10Mc             =  14.079181246047625 ,
   bcm_etab                =                 0.5 ,
   bcm_ks                  =                55.0 ,
   mu_0                    =                 0.0 ,
   sigma_0                 =                 0.0 ,
   c1_mg                   =                 1.0 ,
   c2_mg                   =                 1.0 ,
   lambda_mg               =                 0.0 ,
   z_mg                    =          [0.0, 1.0] ,
   df_mg                   =         [0.01, 0.0] ,
   extra_parameters        =                None ,
   transfer_function       =    'boltzmann_camb' ,
   matter_power_spectrum   =           'halofit' ,
   baryons_power_spectrum  =         'nobaryons' ,
   mass_function           =          'tinker10' ,
   halo_concentration      =         'duffy2008' ,
   emulator_neutrinos      =            'strict' )


# and to this (using `print(cosmo)`, `str(cosmo)` or just `In [1]: cosmo` in the iPython console)

pyccl.Cosmology(
   Omega_c    =               0.25 ,
   Omega_b    =               0.05 ,
   h          =                0.7 ,
   n_s        =               0.96 ,
   A_s        =            2.1e-09 ,
   m_nu       =  [0.02, 0.1, 0.05] ,
   m_nu_type  =             'list' ,
   z_mg       =         [0.0, 1.0] ,
   df_mg      =        [0.01, 0.0] )
```

P.S.: I meant to write `replace long expression with xor` in the commit above; there was nothing wrong and I can't amend the commit message.

nikfilippas:
The commits below address #886 : they change the name of `cls.py` to `cells.py` and the name of `bcm.py` to `baryons.py`. Everything is still backwards-compatible, through overriding `__getattr__` in `pyccl.__init__.py`.

nikfilippas:
As this PR has been open for a long time, new things were being merged to master. The commits above don't make any significant new changes - they merely implement the star to the newly-introduced code from master. Some polishing, and a little bit more coverage.

nikfilippas:
Approximate changelog for the commits above:

- from_something become class methods to improve functionality
  * Pk2D.pk_from_model --> Pk2D.from_model
  * concentration_from_name --> Concentration.from_name
  * mass_function_from_name --> MassFunc.from_name
  * halo_bias_from_name --> HaloBias.from_name
  * mass_def_from_name --> MassDef.from_name
  * (deprecate old functions, preserve API)

- empty Pk2D contains extrap attributes
  * these take the default values
  * missing attributes cause problems in binary operations with Pk2D

- science methods of Pk2D in _pk2d
  * Pk2D was getting very long. Avoid lengths similar to `core.py`.
  * `pk_from_model`, `apply_halofit`, `include_baryons`

- some Pk2D classmethods are now instance methods
  * it makes much more sense to call a `Pk2D` method via its object's instance, rather than via the class
  * current class method usage makes code unnecessarily complicated
  * `apply_halofit`, `apply_nonlin_model`, `include_baryons`
  * _Pk2D_descriptor to preserve compatibility

- concatenate Pk2D.eval and Pk2D.eval_dlPk_dlk
  * code was identical apart from the C-level function
  * deriv can now be called via Pk2D.eval(..., derivative=True)
  * preserve compatibility

- make Pk2D.eval.cosmo an optional argument
  * cosmo is used in extrapolation, which is not always necessary
  * users often have to create a dummy cosmo
  * make cosmo optional; TypeError if requested s.f. is out of range
  * inherently, the data in Pk2D should not depend on cosmo

- disable mamba banner in gha

- introduce CCLDeprecationWarning throughout CCL

- deprecate superfluous parameters in Cosmology 
  * depreacte z_mg, df_mg
  * deprecate direct use of `bcm` args  - put them in extra_parameters

- deprecate bcm_correct_pk2d (avoid direct operations on Pk2D objects)
  * operating directly on `Pk2D` objects can be dangerous
  * newly-implemented `Pk2D.copy` method can make copies of `Pk2D` objects
  * `bcm_correct_pk2d` becomes a private function
  * all baryon corrections handled through `pyccl.baryons.baryon_correct`

pyccl/halos/hmfunc.py
nikfilippas: I've reverted those changes.

pyccl/halos/hmfunc.py
nikfilippas: I added an extra test and un-commented those lines, so this is resolved now.

pyccl/halos/hmfunc.py
nikfilippas: Update on this: I discovered the reason I was getting interpolation-out-of-range errors. In `Cosmology`, the default halo model mass definition is virial. For some concentrations, the Delta-translation of the virial mass definition is below 200, which is the lower interpolation boundary in Tinker08 and Tinker10.
This normally wouldn't pass the tests, but it was left uncovered. In this PR I increased coverage, hence the tests were failing.

What I'm going to do is change the halo model default mass definition from virial to 200m, which should solve the issue.

PS: I know the the halo model non-linear power spectrum is deprecated, but I tried to save it by moving the relevant parameters to `extra_parameters`.

nikfilippas:
Update: I rebased `fix_cosmo` into this branch. I had to do it manually because the commit history was so complicated that git just gave up and started randomly deleting files.

nikfilippas:
I've now added caching the calls to the Boltzmann solver.

nikfilippas:
Work towards v3 was done more strategically in PR1038.

Issue #890, opened 2021-06-17
halomod_power_spectrum not optimized


nikfilippas:
Generally, the calculation of the halo model power spectrum will be the single most expensive operation (by a large margin) in a likelihood. If we want to autocorrelate a profile (and potentially also normalize it), passing anything else other than `None` in `prof2` (and `normprof2`) of `halomod_power_spectrum` will re-calculate the `I_1_1` integral:
```
if prof2 is None:
    i11_2 = i11_1
else:
    i11_2 = hmc.I_1_1(cosmo, k_use, aa, prof2)
```

In the current implementation, users are expected to handle this at the likelihood level (i.e. set `prof2=None`) when they autocorrelate. This complicates things slightly, as the same function needs to be called differently for autocorrelations than it does for cross-correlations.

If they choose not to do this it will slow down their likelihood by a considerable amount.
Timing it on my laptop for an autocorrelation of an HOD profile on a single core, I get:
```
## get_1h = True ##
|            | norm=True        | norm=False       |
|------------|------------------|------------------|
| prof2=p1   | 229 ms ± 5.26 ms | 208 ms ± 1.85 ms |
| prof2=None | 153 ms ± 2.85 ms | 144 ms ± 4.3 ms  |
                 33% faster          30% faster


## get_1h = False ##
|            | norm=True        | norm=False        |
|------------|------------------|-------------------|
| prof2=p1   | 138 ms ± 1.28 ms | 121 ms ± 1.16 ms  |
| prof2=None | 92 ms ± 2.5 ms   | 83.4 ms ± 771 µs  |
                 33% faster          31% faster
```

For a 3x2pt likelihood calculation (where one of them is an autocorrelation) that lasts ~ 1 second per core, this translates to a speed increase/decrease of ~10% in the total runtime.

We could absorb this into CCL by checking whether the passed profiles are equivalent and only calculate the second `I_1_1` (and the normalization) if they are not.

PR #891, opened 2021-06-18
Optimize HM autocorrelation power spectra

nikfilippas:
**Update**: This now relies on `CCLObject` (PR934), so leaving it as draft until PR934 is reviewed. In particular, it makes use of the `__eq__` framework built there.

Addresses #890 .

Checklist of optimizations:
- [x] Halo Model Calculator integrals (`I_x_2`)
- [x] Halo Model power spectrum (`halomod_power_spectrum`)
- [x] Halo Model 1-halo trispectrum (`halomod_trispectrum_1h`)
- [x] SSC (`halomod_Tk3D_SSC`)
- [x] 2-point Fourier (`Profile2pt.fourier_2pt`)

In short, calculating power spectra and trispectra for autocorrelations can be 30% faster if we do some basic checks on the passed profiles. With this PR, functions in the checklist above will now check if the passed profiles are equivalent (through a newly-implemented `__eq__` method in `HaloProfile`), and it will save time by not calculating unnecessary integrals if they are.

----


c-d-leonard:
Just raising a question - do we want profiles with different FFTlog precision parameters to be treated as equivalent? Is there a chance this could induce some kind of numerical difference from what we are expecting to get at a level we would care about?

pennalima:
Hi @c-d-leonard, @nikfilippas  and @damonge, 

Ready to review! :)

nikfilippas:
Right, so I added an FFTLog precision parameters check just in case. It probably wouldn't make any difference in the calculation of the power spectra unless the different precision parameters were way too different, but better to be on the safe side.
Under normal use the precision params will simply be the default ones anyway. It's okay if the calculation is a bit slower in the very few cases where users set same profile types with different precision params.

nikfilippas:
**Update**
I added a full check of the profiles, to simplify things. As it was before, the two profiles could have the same parameters and FFTLog parameters, but different mass-concentration relations. So I replaced the way we check profile equivalence by doing a full check:
- mass-concentration relation name (where applicable)
- mass-concentration relation mass definition
- parameters
- FFTLog parameters
- added optimizations in case we find the two profiles to be equivalent before checking the parameters individually.

pyccl/halos/profiles.py
```diff
@@ -147,6 +147,36 @@ def _get_plaw_projected(self, cosmo, a):
...
+        if type(self) != type(prof2):
+            # catch different profile types
+            return False
+
+        params, params2 = self.__dict__, prof2.__dict__
+
+        if params == params2:
+            # same profiles
+            return True
+
```
damonge, line 161: Humm, I'm confused, it seems the only outcome after this line is for this function to return `False`, right? If so, why the extra checks?

damonge (commented):
One preliminary question before anything else.

pyccl/halos/profiles.py
nikfilippas: Good catch, I meant to return `True` if all the checks have passed. I added an extra if block, which if passes, means that the two profiles are indeed equivalent.

c-d-leonard:
Is this ready to be looked at again? @nikfilippas @damonge

pyccl/halos/profiles.py
```diff
@@ -147,6 +147,38 @@ def _get_plaw_projected(self, cosmo, a):
...
+        for key, val in params.items():
+            if key == "cM":
+                cm2 = params2.get(key)
+                if val.name != cm2.name:
+                    return False
+                if not val.mdef.__eq__(cm2.mdef):
+                    return False
```
damonge, line 168: OK, this is all good as long as c-M relations are fully defined by their name and mass definition. However, for instance the Ishiyama model you implemented also depends on whether you pick relaxed halos or use the Vmax method. E.g. you could have two profiles generated with the same Ishiyama cM relation and mass definition, but one for relaxed and the other for all halos, and this would identify both profiles as the same, right?

damonge (commented):
OK, some comments below.

My main potential worry is whether further down the line we start implementing profiles that include complex parameters (e.g. objects, just like we currently have concentration-mass relations) and we forget to update the `__eq__` method. Thoughts?

pyccl/halos/halo_model.py
```diff
@@ -283,14 +283,17 @@ def I_0_2(self, cosmo, k, a, prof1, prof_2pt, prof2=None):
                 a profile covariance object
                 returning the the two-point moment of the two profiles
                 being correlated.
-            prof2 (:class:`~pyccl.halos.profiles.HaloProfile`): a
-                second halo profile. If `None`, `prof` will be used as
+            prof2 (:class:`~pyccl.halos.profiles.HaloProfile` or None):
+                a second halo profile. If `None`, `prof1` will be used as
```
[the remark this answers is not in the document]
nikfilippas: Yes, there is inconsistent use of `prof` vs `prof1` throughout CCL. I've taken care of all these things in the `star` PR, but I also changed it here for completeness.

pyccl/halos/halo_model.py
```diff
@@ -315,7 +318,7 @@ def I_1_2(self, cosmo, k, a, prof1, prof_2pt, prof2=None):
             cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
             k (float or array_like): comoving wavenumber in Mpc^-1.
             a (float): scale factor.
-            prof (:class:`~pyccl.halos.profiles.HaloProfile`): halo
+            prof1 (:class:`~pyccl.halos.profiles.HaloProfile`): halo
```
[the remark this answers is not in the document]
nikfilippas: Done.

pyccl/halos/halo_model.py
```diff
@@ -420,8 +435,8 @@ def halomod_mean_profile_1pt(cosmo, hmc, k, a, prof,
         `k` and `a` respectively. If `k` or `a` are scalars, the
         corresponding dimension will be squeezed out on output.
     """
-    a_use = np.atleast_1d(a)
-    k_use = np.atleast_1d(k)
+    a_use = np.atleast_1d(a).astype(float)
+    k_use = np.atleast_1d(k).astype(float)
```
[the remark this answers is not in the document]
nikfilippas: (I often test CCL functions that take `(k, a)` with `func(1, 1)` and it's annoying when I get an error doing that and I have to replace by `func(1., 1.)`)

pyccl/halos/halo_model.py
```diff
@@ -380,14 +386,23 @@ def I_0_22(self, cosmo, k, a,
...
         if prof34_2pt is None:
             prof34_2pt = prof12_2pt
 
         self._get_ingredients(a, cosmo, False)
         uk12 = prof12_2pt.fourier_2pt(prof1, cosmo, k, self._mass, a,
                                       prof2=prof2, mass_def=self._mdef).T
-        uk34 = prof34_2pt.fourier_2pt(prof3, cosmo, k, self._mass, a,
-                                      prof2=prof4, mass_def=self._mdef).T
+
+        if prof3.__eq__(prof1) and prof2.__eq__(prof1):
```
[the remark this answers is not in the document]
nikfilippas: I guess that's also doable. Your suggestion is slightly more generic (it requires that `prof4` is already defined) which the current implementation doesn't take for granted. To keep things simple I originally only considered the 4-pt function where all profiles are the same. I've now followed your suggestion to implement it properly.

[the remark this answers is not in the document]
nikfilippas: Also, if we look at the docs:
```
prof4 (:class:`~pyccl.halos.profiles.HaloProfile`): a second halo profile. If `None`, `prof3` will be used as prof4`.
```
I don't understand why it has to be this way? If we want to do `abab`, we should either be able to do it using `func(prof1=a, prof2=b, prof3=a, prof4=b)` or `func(prof1=a, prof2=b)`. The way it's currently implemented, doing `func(prof1=a, prof2=b)` will compute `abaa`. Isn't this weird?

pyccl/halos/profiles.py
```diff
@@ -147,6 +147,38 @@ def _get_plaw_projected(self, cosmo, a):
...
+            elif key == "precision_fftlog":
+                if val != params2[key]:
+                    return False
+            else:
+                if val != params2.get(key):
+                    return False
```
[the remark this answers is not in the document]
nikfilippas: Yes, good catch, I didn't know dictionary equality can check more than one level in depth. I removed this clause.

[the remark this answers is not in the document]
nikfilippas: Yes indeed, good catch, I removed it.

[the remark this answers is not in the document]
nikfilippas: Yes, the first check is for speed. In case the `__dict__` of the profiles only contain numerical values and no objects (such as `cM`), this check speeds up the entire process.

pyccl/halos/profiles.py
nikfilippas: I see your point. There are two alternatives we can choose from:

1. Either I include a `__dict__` check in the statement so that it also checks the other attributes on top of the mass definition;
2. Or I delete the long for-loop-check altogether and if the two `cM` relations are not identical, it returns False.

These are all edge cases, so I don't think it will make a difference no matter which one we choose. 1 is more complete, 2 is simpler.

pyccl/halos/halo_model.py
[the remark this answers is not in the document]
damonge: Humm, that might be more natural, yes... Do you want to make that change here?

pyccl/halos/halo_model.py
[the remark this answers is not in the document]
damonge: ok, let's leave it as it is then. I suspect there'll be other instances of this in the code, but this is fine

pyccl/halos/profiles.py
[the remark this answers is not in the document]
damonge: OK, I see.

pyccl/halos/profiles.py
damonge: OK, I have another proposal, which I'll leave in the main comment for this review.

pyccl/halos/halo_model.py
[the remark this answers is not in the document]
nikfilippas: I could, but strictly speaking, this counts as API breakage. I highly doubt anyone will have ever used it that way, but still, the proposed behavior is different from the current one.

nikfilippas:
I think that even if we go down this route, the check will still not be 100% failsafe.

`Concentration` depends on `MassDef`, but `MassDef` also depends on `Concentration` (to translate between masses). So, to check everything properly, we'd have to check concentration, then go to mass def, then, if the mass def is defined with a default concentration, check that concentration also, then check the mass def of that concentration etc. This is too many lines of code for a simple check of profile equality.

Of course, it doesn't really matter which concentration prescription you use to convert between mass definition (as long as it's something sensible), but checking the `__dict__` 2-levels deep is already too complicated.

nikfilippas:
My point is that if we define an `__eq_same` method, and add so many lines of code, then it better do a comprehensive check. It won't though. So we resort to the original `__eq__` method, and either choose a strict but 100% safe check, or a softer (which will still give correct results) check.

damonge:
OK, there are two points:
a) As we create new profiles, it will be annoying to have to add more and more if statements to the central `__eq__` method to capture all possible cases. An `__eq_same` method for the specific difficult profiles would sort that.
b) Comparing two cMs is annoying. In this case I'd be happy if we just check: a) that they are the same type, b) that the massdefs have the same Delta and rho_type, c) that any other elements of __dict__ are the same (e.g. in the case of Ishiyama).

nikfilippas:
a) Likewise, as we create new profiles, it will be annoying to have to repeat the same lines of code for the same checks in every new profile. Because this all happens in a for-loop, we can't cut the loop in two pieces and have the numerical checks happening at the base and the difficult checks ad hoc. So this issue boils down to:

- introduce new checks at base as they come OR
- have an equivalence method for each profile separately (which for the most part does the same checks --> repeated lines of code --> harder to maintain)

b) Using the current implementation, `Ishiyama21` works fine. The two additional attributes that are checked are accessed through the `__dict__` method and are a string and a boolean. The equality operator works the same as with numerical values. So I don't think that's an issue.

damonge:
> Likewise, as we create new profiles, it will be annoying to have to repeat the same lines of code for the same checks in every new profile. 

I don't think that's true. You would create a default `__eq_same` in the base class that does the most likely checks (comparing dictionaries and fftlog). Then you only need to rewrite that method for the few profiles for which that is not enough.

nikfilippas:
@damonge have a look at this new implementation - it's simpler than the one we had before and I think it addresses all of your concerns. Basically I implemented an `__eq__` method in `Concentration` which solves the problem. Now, no new code has to be rewritten in the subclasses, unless it's something very specific.

Also look at the one comment that is still open - I made those changes as well.

pyccl/halos/profiles.py
```diff
@@ -147,6 +147,31 @@ def _get_plaw_projected(self, cosmo, a):
         """
         return self.precision_fftlog['plaw_projected']
 
+    def _prof_equiv(self, prof2):
+        """Check profile equivalence. If equivalence between two profiles
+        is not trivial, and simply checking equality of their attributes
+        does not suffice, this method has to be reloaded.
```
nikfilippas, line 153: TODO: reloaded --> overloaded

pyccl/halos/profiles_2pt.py
```diff
@@ -111,7 +111,7 @@ def fourier_2pt(self, prof, cosmo, k, M, a,
             k (float or array_like): comoving wavenumber in Mpc^-1.
             M (float or array_like): halo mass in units of M_sun.
             a (float): scale factor.
-            prof2 (:class:`~pyccl.halos.profiles.HaloProfile`):
+            prof2 (:class:`~pyccl.halos.profiles.HaloProfile` or None):
```
damonge, line 114: I just realised this actually has to be `HaloProfileHOD`, not just `HaloProfile`

pyccl/halos/profiles_2pt.py
```diff
@@ -128,9 +128,14 @@ def fourier_2pt(self, prof, cosmo, k, M, a,
...
-
         if prof2 is not None:
-            if prof2 is not prof:
-                raise ValueError("prof2 must be the same as prof")
+            if not isinstance(prof2, HaloProfileHOD):
+                raise TypeError("prof2 must be of type "
+                                "`HaloProfileHOD` or None")
+        else:
+            prof2 = prof
+
```
damonge, line 137: I think the 6 preceding lines are pointless, right? We don't really use prof2 for anything other than to check if it's `prof`, so why not just include the two lines right after this comment within the `if prof2 is not None:` bracket?

pyccl/tests/test_tkk1h.py
```diff
@@ -14,7 +14,7 @@
 P2 = ccl.halos.HaloProfileHOD(ccl.halos.ConcentrationDuffy08(M200))
 P3 = ccl.halos.HaloProfilePressureGNFW()
 P4 = P1
-Pneg = ccl.halos.HaloProfilePressureGNFW(P0=-1)
+Pneg = ccl.halos.HaloProfilePressureGNFW(P0=1j)  # (1j*1j = -1)
```
damonge, line 17: wait, what?

damonge: Why do you need to do this at all?

pyccl/tests/test_tkkssc.py
```diff
@@ -22,7 +22,7 @@
     P2_nogc.is_number_counts = False
 P3 = ccl.halos.HaloProfilePressureGNFW()
 P4 = P1
-Pneg = ccl.halos.HaloProfilePressureGNFW(P0=-1)
+Pneg = ccl.halos.HaloProfilePressureGNFW(P0=1j)  # (1j*1j = -1)
```
damonge, line 25: same question

damonge (commented):
A first preliminary review. I won't look at the `halo_model` code until the cclobject branch has been merged

pyccl/halos/profiles_2pt.py
nikfilippas: ok, fixed.

pyccl/halos/profiles_2pt.py
nikfilippas: All these kinds of type checks are pointless. Python relies on duck-typing a lot, there is no need to check everything, every time. If the type isn't correct the code will exit and will almost always have a useful error.
There's a lot of stuff to remove in `halo_model.py` too but I suggest to leave it for the v3 branch.

nikfilippas: (I have fixed this particular one)

damonge:
BTW, don't you need to rebase this to master?

nikfilippas:
It was done automatically when I merged CCLObject to master.

pyccl/halos/halo_model.py
```diff
@@ -1267,138 +1291,127 @@ def halomod_Tk3D_SSC(cosmo, hmc,
...
-        prof3_bak = prof1
-    else:
-        prof3_bak = prof3
-    if prof34_2pt is None:
-        prof34_2pt_bak = prof12_2pt
-    else:
-        prof34_2pt_bak = prof34_2pt
+    for prof, norm in profs.items():
+        if prof.is_number_counts and not norm:
+            raise ValueError("normprof must be True if prof.is_number_counts")
```
damonge, line 1328: The previous code told users what the profile number was that was failing this.

pyccl/halos/halo_model.py
```diff
@@ -1267,138 +1291,127 @@ def halomod_Tk3D_SSC(cosmo, hmc,
...
         norm12 = norm1 * norm2
         norm34 = norm3 * norm4
 
         pk = pk2d.eval(k_use, aa, cosmo)
         dpk = pk2d.eval_dlogpk_dlogk(k_use, aa, cosmo)
         # (47/21 - 1/3 dlogPk/dlogk) * I11 * I11 * Pk+I12
-        dpk12[ia, :] = norm12*((2.2380952381-dpk/3)*i11_1*i11_2*pk+i12_12)
-        dpk34[ia, :] = norm34*((2.2380952381-dpk/3)*i11_3*i11_4*pk+i12_34)
+        dpk12[ia, :] = norm12*((47/21 - dpk/3)*i11_1*i11_2*pk + i12_12)
+        dpk34[ia, :] = norm34*((47/21 - dpk/3)*i11_3*i11_4*pk + i12_34)
```
damonge, line 1385: while 47/21 is clearer, it's also slower

damonge (commented):
More comments. Will move to tests after this.

damonge:
(also, tests seem to be failing now)

README.md
```diff
@@ -111,13 +111,13 @@ for more information.
 # Contact
 
 If you have comments, questions, or feedback, please
-[write us an issue](https://github.com/LSSTDESC/CCL/issues). 
+[write us an issue](https://github.com/LSSTDESC/CCL/issues).
 
-The current lead of the LSST DESC CCL Topical Team is Danielle Leonard (c-d-leonard, danielle.leonard at ncl.ac.uk)
+The current lead of the LSST DESC CCL Topical Team is David Alonso (damonge, david.alonso at physics.ox.ac.uk)
```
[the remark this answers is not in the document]
nikfilippas: any time! 😂

pyccl/halos/halo_model.py
nikfilippas: This is a common misconception among Python users who come from different languages. When the interpreter first runs, it sees all that and calculates whatever can be calculated when creating the `.pyc` files in `__pycache__`. So it isn't slower.

damonge:
@nikfilippas tests should be passing now, and I'd be happy to merge as it is, but since I've made modifications, you should take a look. Just look at my last 3 commits.

pyccl/halos/halo_model.py
```diff
@@ -402,31 +402,31 @@ def I_0_22(self, cosmo, k, a,
...
              value of `k`.
         """
-        if (prof3, prof4) == (None, None):
-            prof3, prof4 = prof1, prof2
-        elif (prof3, prof4).count(None) == 1:
-            raise ValueError("prof3 and prof4 should be both defined or None")
+        if prof3 is None:
+            prof3 = prof1
+        if prof4 is None:
+            prof4 = prof2
```
nikfilippas: I suggest leave the error as it was. In what scenario would someone define (prof1, prof2, prof3) but no prof4? Sounds to me like bad use of the function. We should accept either of these:
- (`prof1`,) --> (`prof1`, `prof1`, `prof1`, `prof1`)
- (`prof1`, `prof2`) --> (`prof1`, `prof2`, `prof1`, `prof2`)
- (`prof1`, `prof2`, `prof3`, `prof4`)

pyccl/halos/halo_model.py
```diff
@@ -402,31 +402,31 @@ def I_0_22(self, cosmo, k, a,
...
 
         if prof34_2pt is None:
             prof34_2pt = prof12_2pt
 
         self._get_ingredients(cosmo, a, False)
         uk12 = prof12_2pt.fourier_2pt(prof1, cosmo, k, self._mass, a,
                                       prof2=prof2, mass_def=self._mdef).T
-        if (prof1, prof2) == (prof3, prof4):
+        if (prof3 is prof1) and (prof4 is prof2):
```
nikfilippas, line 429: Add a TODO here too.

nikfilippas:
@damonge there is no point in adding all those TODO items for later. Let's just implement an `__eq__` in `HaloProfile` and one in `Profile2pt` that just returns `return self is other`. This way, we can leave the `==` signs everywhere and not bother about them ever again.

pyccl/halos/halo_model.py
damonge: This is not what the docstring says, and I do not want to change the API. Besides that, I don't want to be prescriptive about what is good or bad use of a function.

pyccl/halos/halo_model.py
```diff
@@ -1390,27 +1400,35 @@ def get_norm(normprof, prof, sf):
...
 
-            b1 = i11_1*norm1 if prof1.is_number_counts else np.zeros_like(k_use)  # noqa
+            if prof1.is_number_counts:
+                b1 = i11_1 * norm1
 
-            if prof2.is_number_counts:
+            # TODO: change to == when __eq__ is properly implemented
+            if prof2 is prof1:
+                b2 = b1
+            elif prof2.is_number_counts:
```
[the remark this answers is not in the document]
damonge: This just mirrors what we're doing with b3 and b4. I could imagine this being slower if i11_2 is a large array.

pyccl/halos/halo_model.py
damonge: done

damonge:
I do not want to leave anything that could affect perfomance negatively on master. It's best if you get rid of the TODOs in the `__eq__` branch

damonge (approved):

PR #893, opened 2021-07-05
Allow access to gsl spline data

tilmantroester:
At the moment, data that is used to create splines (e.g., in `Pk2D`) is not accessible from python after the spline is created. This PR allows access to the underlying GSL splines. The new functions return copies of the underlying data, so should be harmless. 

This is infrastructure for another PR I'm working on that allows adding `Pk2D` objects. For example, `pk_1h + pk_2h`. Beyond that, I imagine it being useful for debugging purposes.


pyccl/ccl_utils.i
```diff
@@ -49,4 +56,80 @@ void spline_integrate(int n_integ,
...
+  }
+  if(x_size != spline->interp->size) {
+    *status = CCL_ERROR_INCONSISTENT;
+    return;
+  }
+  if(y_size != spline->interp->size) {
+    *status = CCL_ERROR_INCONSISTENT;
+    return;
+  }
+  for(int i=0; i<x_size; i++) {
```
damonge, line 98: would a `memcpy` be faster here?

damonge: same below for y

pyccl/ccl_utils.i
```diff
@@ -49,4 +56,80 @@ void spline_integrate(int n_integ,
...
+  if(x_size != spline2d->interp_object.xsize) {
+    *status = CCL_ERROR_INCONSISTENT;
+    return;
+  }
+  if(y_size != spline2d->interp_object.ysize) {
+    *status = CCL_ERROR_INCONSISTENT;
+    return;
+  }
+  for(int i=0; i<x_size; i++) {
+    xarr[i] = spline2d->xarr[i];
```
damonge, line 125: as above about `memcpy`

damonge (commented):
Couple of comments below. Looks good to me otherwise @tilmantroester

pyccl/ccl_utils.i
tilmantroester: Switched to `memcpy`, makes the code shorter as well. 
I played a bit around with godbolt (https://godbolt.org/z/13jKM6GE9) and for some compilers, `memcpy` and the loop end up with the same number of instructions. Interestingly, the loop gets more and more instructions with more recent compiler versions.
Since this gets copied to the python layer, I don't think any speed differences matter, if there are any.

pyccl/ccl_utils.i
tilmantroester: Done.

pyccl/ccl_utils.i
damonge: oh, that's interesting. Yes, I don't think this would ever be a bottleneck. Code looks neater though. Thanks!

damonge (approved):
LGTM!

PR #894, opened 2021-07-05
__add__ operator of Pk2D

tilmantroester:
This PR adds the `__add__` operator to `Pk2D` to allow the addition of two `Pk2D` objects. A specific application where this comes in handy are halo models where the 1h and 2h terms have different halo profile contributions. It's mostly syntactic sugar and checks around a new `get_spline_data` method.
To prevent any funny business when the support of the original Pk2D data doesn't match between the operands, the addition is only defined when the support of the 2nd operand is a subset of that of the 1st operand. 

This could in principle be extended to other operators like `__mul__` and `__pow__`, for a neat, if somewhat inefficient, implementation of the HMCode 1h->2h interpolation scheme.

The PR relies on #893


nikfilippas:
In Python 3.9 they added a dictionary union operator, (`|`). The proposal originally revolved around introducing the `+` to make usage simpler, but many objected because of the usage of `+` in Python. For example, in a list or in a string, `+` means "extend".
It might not be immediately clear what the addition operation means in the `Pk2D` context. I personally am fine with this, but just leaving it as a counterargument / food for thought.
Discussion on using the `+` operator in `dict` is [here (PEP)](https://www.python.org/dev/peps/pep-0584/#use-the-addition-operator) and [here (mailing list)](https://mail.python.org/archives/list/python-ideas@python.org/thread/BHIJX6MHGMMD3S6D7GVTPZQL4N5V7T42/).

pyccl/pk2d.py
```diff
@@ -207,13 +211,92 @@ def eval(self, k, a, cosmo):
...
+        a_arr_a, lk_arr_a, pk_arr_a = self.get_spline_arrays()
+
+        a_arr_b, lk_arr_b, pk_arr_b = other.get_spline_arrays()
+        if not (a_arr_a.size == a_arr_b.size and lk_arr_a.size == lk_arr_b.size
+                and np.allclose(a_arr_a, a_arr_b)
+                and np.allclose(lk_arr_a, lk_arr_b)):
+            # Since the power spectrum is evalulated on a smaller support than
+            # where it was defined, no extrapolation is necessary and the
+            # dependence on the cosmology in moot.
+            # CosmologyVanillaLCDM is being imported here instea of the top of
```
nikfilippas, line 275: typo

tilmantroester:
The `+` here really means addition in the math sense. As in `Pk_halomodel = Pk_1h + Pk_2h`. Since `Pk2D` is essentially a fancy wrapper around a 2D array, the operation should be well-defined, except for the edge cases that the checks in the implementation hopefully catch.

nikfilippas:
Coming back to this one, I think `__mul__` will be very useful, particularly for pk correction factors. It would be great if this PR introduced that as well.

tilmantroester:
The `__add__` part should be ready to review. If I have some time I'd also implement the other operators.

tilmantroester:
I've now added the `__mul__` and `__pow__` operators, so now you can do something like `(Pk1 + 0.5* Pk2)**1.8`, if you're so inclined. Currently there are now checks being done on the power operator, so if you take a non-positive power spectrum to a negative or fractional power, weird things might happen.

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+        if not self.has_psp or not other.has_psp:
+            raise ValueError("Pk2D object does not have data.")
+        if (self.psp.lkmin < other.psp.lkmin
+                or self.psp.lkmax > other.psp.lkmax):
+            raise ValueError("The 2nd operand has its data defined over a "
+                             "smaller k range than the 1st operand.")
+        if (self.psp.amin < other.psp.amin
+                or self.psp.amax > other.psp.amax):
+            raise ValueError("The 2nd operand has its data defined over a "
+                             "smaller a range than the 1st operand.")
```
nikfilippas, line 297: I see the point of throwing an exception whenever one operand has less support than the other operand (instead of, for example, extrapolating). But is there a particular reason this operation is not commutative? Could we not operate on the one with the widest support, in both `a` and `k`?

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+
+        a_arr_a, lk_arr_a, pk_arr_a = self.get_spline_arrays()
+        a_arr_b, lk_arr_b, pk_arr_b = other.get_spline_arrays()
+        if not (a_arr_a.size == a_arr_b.size and lk_arr_a.size == lk_arr_b.size
+                and np.allclose(a_arr_a, a_arr_b)
+                and np.allclose(lk_arr_a, lk_arr_b)):
+            # Since the power spectrum is evalulated on a smaller support than
+            # where it was defined, no extrapolation is necessary and the
+            # dependence on the cosmology in moot.
+            # CosmologyVanillaLCDM is being imported here instea of the top of
```
nikfilippas, line 307: instead*

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+            # the module due to circular import issues there.
+            from .core import CosmologyVanillaLCDM
+            dummy_cosmo = CosmologyVanillaLCDM()
+            pk_arr_b = np.array([other.eval(k=np.exp(lk_arr_a),
+                                            a=a_,
+                                            cosmo=dummy_cosmo)
+                                 for a_ in a_arr_a])
+        return a_arr_a, lk_arr_a, pk_arr_a, pk_arr_b
```
nikfilippas, line 315: The function name is `check_compatibility` but it doesn't only check stuff, it returns some compatible arrays.

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+    def __add__(self, other):
+        """Adds two Pk2D instances.
+
+        The a and k ranges of the 2nd operand need to be the same or smaller
+        than the 1st operand.
+        The returned Pk2D object uses the same a and k arrays as the first
+        operand.
+        """
+        a_arr_a, lk_arr_a, pk_arr_a, pk_arr_b = \
+            self._check_Pk2D_compatability(other)
```
nikfilippas, line 326: function name

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+        if isinstance(other, (float, int)):
+            a_arr_a, lk_arr_a, pk_arr_a = self.get_spline_arrays()
+            pk_arr_new = other * pk_arr_a
+        elif isinstance(other, Pk2D):
+            a_arr_a, lk_arr_a, pk_arr_a, pk_arr_b = \
+                self._check_Pk2D_compatability(other)
+            pk_arr_new = pk_arr_a * pk_arr_b
+        else:
+            raise ValueError("Multiplication of Pk2D is only defined for "
+                             "floats, ints, and Pk2D objects.")
```
nikfilippas, line 360: Would it be useful to do the same for `__add__`?
i.e. make `my_pk + 0.002` a valid call?

Also, would it be useful to accept ready-made 2d arrays (of narrower support maybe) and multiply them without accessing the data from the 2nd operand?
i.e. `Pk * my_2d_array_k_a`

pyccl/tests/test_pk2d.py
```diff
@@ -260,3 +260,99 @@ def test_pk2d_parsing():
...
+        pk2d_a + empty_pk2d
+
+    pk2d_c = ccl.Pk2D(a_arr=x, lk_arr=log_y, pk_arr=zarr_b,
+                      is_logp=False)
+
+    # This raises an error because addition is only defined for Pk2D + Pk2D
+    with pytest.raises(ValueError):
+        pk2d_a + 1
+    with pytest.raises(ValueError):
+        1 + pk2d_a
```
nikfilippas, line 304: As above. Maybe this could be a valid call.
Otherwise, this should be a `TypeError`.

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
+
+    def __pow__(self, exponent):
+        """Take a Pk2D instance to a power.
+        """
+        if not isinstance(exponent, (float, int)):
+            raise ValueError("Exponentiation of Pk2D is only defined for "
+                             "floats and ints.")
```
nikfilippas, line 380: This should be a `TypeError`.

pyccl/pk2d.py
nikfilippas, line 360: This should be a `TypeError`.

pyccl/tests/test_pk2d.py
```diff
@@ -260,3 +260,99 @@ def test_pk2d_parsing():
...
+    zarr_b = np.outer(-1*x, 4*np.exp(log_y))
+
+    pk2d_a = ccl.Pk2D(a_arr=x, lk_arr=log_y, pk_arr=np.log(zarr_a),
+                      is_logp=True)
+    pk2d_b = ccl.Pk2D(a_arr=x, lk_arr=log_y, pk_arr=zarr_b,
+                      is_logp=False)
+
+    # This raises an error because multiplication is only defined for
+    # float, int, and Pk2D
+    with pytest.raises(ValueError):
```
nikfilippas, line 336: `TypeError`

pyccl/tests/test_pk2d.py
```diff
@@ -260,3 +260,99 @@ def test_pk2d_parsing():
...
+        pk2d_a*np.array([0.1, 0.2])
+
+    # This raises an error because exponention is only defined for
+    # float and int
+    with pytest.raises(ValueError):
```
nikfilippas, line 341: `TypeError`

pyccl/pk2d.py
```diff
@@ -248,13 +253,145 @@ def eval_dlogpk_dlogk(self, k, a, cosmo):
...
+                   extrap_order_hik=self.extrap_order_hik)
+        return new
+
+    def __rmul__(self, other):
+        return self.__mul__(other)
+
+    def __pow__(self, exponent):
+        """Take a Pk2D instance to a power.
+        """
+        if not isinstance(exponent, (float, int)):
```
nikfilippas, line 378: Would it make sense to add a check and warn users that "weird things might happen" if the power spectrum is non-positive and the exponent is noninteger or negative?

nikfilippas (changes requested):
A very useful addition to the library, these operations just make sense to have!
My main comment has to do with commutativity. I think looking for the `Pk2D` with the widest support and operating on that one would make `+` and `*` commutative at very little extra cost.

pyccl/pk2d.py
tilmantroester: Done.

pyccl/pk2d.py
tilmantroester: Changed to `_get_binary_operator_arrays`.

pyccl/pk2d.py
tilmantroester: Done.

pyccl/pk2d.py
nikfilippas: Great, thank you!
Agreed on the second point, so I'm now resolving this.

nikfilippas:
One final comment regarding the error message in something that I foresee will be raised a lot.
Otherwise, LGTM. Great to have this in CCL!

pyccl/pk2d.py
nikfilippas: Except if there is a very particular reason where, if we have `B` with narrower support than `A`, we would expect B += A to return `B` with the same support as the original `B`?

tilmantroester: My issue is that these operations would change the internal state of the Pk2D object in a way that might be opaque to the user. For example, a user creates a Pk2D object over some known a and k range and then adds other Pk2D objects to it. Unless the user carefully track the support of these objects, they might try to access the Pk2D object at k and a values that were inside the support originally but now have fallen outside of it. 
The alternative would be to disallow any operation if the supports don't match and provide some helper function to create new Pk2D object, given some support ranges.

I see the point about more verbose error messages and have now implemented that.

tilmantroester: >  This is well-defined. A+B should return a Pk2D with the largest support (that of A). No extrapolation necessary.

I disagree. The operation is only well-defined on the intersection. Outside of that you either need to extrapolate `B` or assume `B` is zero, which leads to discontinuities in the power spectrum, neither of which are desirable.

nikfilippas: Right, I agree that extrapolation is tricky and should be avoided. But at the same time, I would trust that the user has sampled the `Pk2D` well enough so that we can assume it's zero elsewhere in the case of addition, and one in the case of multiplication. This practice is nothing new to CCL; `Pk2D` can accept a user-defined `pk_arr` where we specify that "if you pass your own Pk array, you are responsible of making sure that it is sufficiently well-sampled.
What is the rationale behind only using the intersection? I think we are being too strict, judging from similar situations where we are lenient and allow users to do what they want.

c-d-leonard:
Looks like this might be ready to merge, @nikfilippas @tilmantroester ?

pyccl/pk2d.py
damonge: If I'm allowed to weigh in, I'd advocate for Keeping It Simple (S). In this sense I see two options:
1. You only allow operations between pks with exactly the same support and throw an exception otherwise.
2. You only allow to add, to a given pk2d A, another pk2d B that requires no extrapolation (i.e. A fits in B), and the result is only defined on the original pk2d's support (i.e. that of A). I think this is what's currently implemented. This breaks commutativity, but it is the simplest implementation. Also, extrapolation would happen only when leaving known ranges (the ones used to define the first pk2d), and never silently (which would happen if we allow for B to fit into A). This behaviour is also easy to document for users.

If we go with number 2 I'd advocate for throwing a warning if someone tries to add two pk2ds with different supports (but compatible in the sense defined above).

I also think commutativity would be great to have, but we can try to get that in afterwards. The way I see it, most people will use this function to operate on power spectra they have created themselves on the same ranges, so in most situations this discussion will be moot.

nikfilippas: On point 2, which is what we are discussing, how would the users know that `A + B` gives a `Pk2D` with the common range? Wouldn't it feel more natural to expect the resulting pk to have the widest support of the two?

Suppose we have defined a P(k,a) correction factor which is only relevant in `k∈[0.5, 5.0]`. If we want to use this for all `Pk2D` objects down a pipeline we would have to (a) either define a `Pk2D` with an extreme range which contains the tiny section of useful values we want (b) or manually copy and expand the range for each different `Pk2D` we want to operate on.

By the way, I agree with adding a message warning the users of the narrower support if the two supports don't match.

Issue #892, opened 2021-06-25
c500 mass-concentration relation


nikfilippas:
A mass-concentration relation calibrated from simulations, for halo mass definitions relative to 500 times the critical density.
Very useful for analyses including tSZ, as the latter is measured from X-ray data and halos there look very concentrated.
It would be good to have an option for this in CCL. [arXiv:2007.14720](https://arxiv.org/pdf/2007.14720.pdf)

PR #895, opened 2021-07-08
Implement mass-concentrations from Ishiyama et al. (c500)

nikfilippas:
Addresses #892 .

Checklist:
- [x] mass-concentration relation
- [x] M500 mass definition
- [x] unit tests
- [x] benchmarks


pyccl/halos/concentration.py
```diff
@@ -382,6 +382,171 @@ def _concentration(self, cosmo, M, a):
...
+                        self.c_alpha = 0.22
+                    else:  # now it's 500c
+                        self.kappa = 1.10
+                        self.a0 = 2.30
+                        self.a1 = 1.64
+                        self.b0 = 1.72
+                        self.b1 = 3.60
+                        self.c_alpha = 0.32
+
+    def _dlsigmaR(self, cosmo, R, a, *, eps=1e-6, num=2):
```
nikfilippas, line 505: TODO: `num` is leftover from the previously-used `np.gradient` method. Remove it and replace it.

pyccl/halos/concentration.py
```diff
@@ -353,6 +354,169 @@ def _concentration(self, cosmo, M, a):
...
+        check(status)
+        return -3/np.log(10) * dlns_dlogM
+
+    def _G(self, x, n_eff):
+        fx = np.log(1 + x) - x / (1 + x)
+        G = x / fx**((5 + n_eff) / 6)
+        return G
+
+    def _G_inv(self, Arg, n_eff):
+        from scipy.optimize import root_scalar
```
damonge, line 493: Would it be worth importing this at the top of the file?

pyccl/halos/concentration.py
```diff
@@ -353,6 +354,169 @@ def _concentration(self, cosmo, M, a):
...
+        M_use = np.atleast_1d(M)
+
+        nu = 1.686 / sigmaM(cosmo, M_use, a)
+        n_eff = -2 * self._dlsigmaR(cosmo, M_use, a) - 3
+        alpha_eff = growth_rate(cosmo, a)
+
+        A = self.a0 * (1 + self.a1 * (n_eff + 3))
+        B = self.b0 * (1 + self.b1 * (n_eff + 3))
+        C = 1 - self.c_alpha * (1 - alpha_eff)
+        Arg = A / nu * (1 + nu**2 / B)
```
damonge, line 511: pedantic point: can we call this with a lower-case initial?

benchmarks/test_concentration_Ishiyama21.py
```diff
@@ -0,0 +1,64 @@
...
+
+    hmd = ccl.halos.MassDef(Delta, "critical")
+    cm = ccl.halos.ConcentrationIshiyama21(mdef=hmd,
+                                           relaxed=pars["relaxed"],
+                                           Vmax=pars["Vmax"])
+
+    for i, zz in enumerate(Z):
+        dat = data[:, i+1]
+        mod = cm.get_concentration(COSMO, M_use/H100, 1/(1+zz))
+        assert np.allclose(mod, dat, rtol=UCHUU_DATA_SCATTER)
```
damonge, line 64: OK, so things match at the 10% level because that's the data scatter. But how well does our implementation match the Ishiyama implementation?

damonge (commented):
Good stuff @nikfilippas !
See below a few minor comments. I'll check timing on my end while you address these

damonge:
Just a quick note: I just checked and I don't think the computational bottleneck comes from the calculation of dlnsigma/dlnR. If anything it may be dominated by the root finding, but that it's quite fast in itself, so I don't think we should worry.
In summary: ready to go as soon as the previous comments are addressed.

pyccl/halos/massdef.py
```diff
@@ -258,6 +258,19 @@ def __init__(self, c_m='Duffy08'):
                                           c_m_relation=c_m)
 
 
+class MassDef500c(MassDef):
+    """`MassDef` class for the mass definition
+    with Delta=200 times the critical density.
```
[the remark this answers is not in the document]
nikfilippas: Done

pyccl/halos/concentration.py
```diff
@@ -353,6 +354,169 @@ def _concentration(self, cosmo, M, a):
...
+
+    def _check_mdef(self, mdef):
+        if mdef.Delta != 'vir':
+            if isinstance(mdef.Delta, str):
+                return True
+            elif mdef.rho_type != 'critical':
+                return True
+            elif mdef.Delta not in [200, 500]:
+                return True
+            elif (mdef.Delta == 500) and self.Vmax:
```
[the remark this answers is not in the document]
nikfilippas: Indeed, it wasn't. https://arxiv.org/pdf/2007.14720.pdf#table.2

pyccl/halos/concentration.py
nikfilippas: Minimal overhead either way, but I moved it.

nikfilippas:
You are right, the bottleneck is the computation of `sigmaM`, in case these haven't already been computed. Then, root finding also takes time but I can't think of any other way. I profiled this one and `Duffy08`, and this one is an order of magnitude slower. Have a look at the attached images.
`Ishiyama21`
Screenshot from 2022-01-12 18-27-20
`Duffy08`
Screenshot from 2022-01-12 18-28-22

benchmarks/test_concentration_Ishiyama21.py
nikfilippas: They agree at the 10% level. I've actually made a comment on that point. All but the 500c concentrations agree at the 5% level. All the discrepancies between our code and the benchmarks come from the very low masses, so I assume this is a precision error or something. In practice, it will barely make any difference because such low masses contribute very little to the power spectra. See attached plots for an example of what I mean (200c, relaxed, no Vmax).

Concentration across different redshifts (red=CCL, black=bench)
cm

Relative error across different redshifts
error

nikfilippas:
@damonge ready to be looked at again.

nikfilippas:
@damonge ready to be looked at again.

benchmarks/test_concentration_Ishiyama21.py
damonge: OK, good. I'd suggest cutting the mass range to 10^8-10^16 and lowering the tolerance to 0.02. That way we will actually notice if this ever breaks.

damonge:
OK, you'll have to teach me how to  generate those timing diagrams ;-)

Just one comment above, and we can merge as soon as that's done.

benchmarks/test_concentration_Ishiyama21.py
nikfilippas: Unfortunately, 0.035 is as low as it can go. Otherwise, I'd have to clip at `log10(M) ~ 9.5`. I think this still gives a good indication of whether the code is broken. Below I attach the summary plot of all errors in all parametrizations. The ones that don't match below 2% are the 500c relaxed.
clipped

damonge: @nikfilippas ok, that's fine. 10^9 is still a pretty low mass, especially if this is M500. In any case, drop it to 0.035 and we're done.

damonge: oh, you did that already, then we're done!

damonge (approved):
LGTM

PR #898, opened 2021-07-14
Eisenstein & Hu power spectrum without wiggles

damonge:
Eisenstein & Hu 1997 also provide formulas for the linear transfer function without wiggles (Eqs. 30 and 31), which can be quite useful for BAO analyses. This PR adds this functionality.


pyccl/pk2d.py
```diff
@@ -123,6 +123,7 @@ def pk_from_model(Pk2D, cosmo, model):
             model (:obj:`str`): model to use. Three models allowed:
                 `'bbks'` (Bardeen et al. ApJ 304 (1986) 15).
                 `'eisenstein_hu'` (Eisenstein & Hu astro-ph/9710252).
+                `'eisenstein_hu_nowiggles'` (Eisenstein & Hu astro-ph/9710252).
```
carlosggarcia, line 126: Is this the correct reference? I think you mean astro-ph/9709112, which is what is also cited in https://github.com/LSSTDESC/CCL/blob/c3d04fc11e3eaa5c54135717f31ed5dfa80ed5b4/src/ccl_eh.c#L139-L140

carlosggarcia (changes requested):
Just a comment about the paper referenced

pyccl/pk2d.py
damonge: You're right, good catch!

carlosggarcia (approved):

Issue #897, opened 2021-07-14
Cosmic emulator implementation issue


patricialarsen:
There is an implementation issue for the cosmic emulator matter power spectrum involving massive neutrino models. 

In line 323 of ccl_power.c the matter density is defined as follows:

      emu_par[0] = (cosmo->params.Omega_c+cosmo->params.Omega_b)*cosmo->params.h*cosmo->params.h;

The lack of neutrinos in the total mass can cause inconsistencies in the matter power spectrum for higher neutrino masses, these can be particularly noticeable at linear scales. A correction can be fudged in the cosmology definitions by defining omega_c as omega_m - omega_b in the cosmology input for the emulator matter power spectrum, and not for the halofit/linear matter power spectrum, which gives results at the emulation points as follows 

![image](https://user-images.githubusercontent.com/20701830/125565208-18b14551-0e74-4721-85e2-d2d6bd8aa5e6.png)

(This is now consistently accurate to the percent level, which is what we expect at the emulation points within the emulator k range)

To make the linear theory and emulator implementations consistent this line should be changed to something like:

      emu_par[0] = (cosmo->params.Omega_c+cosmo->params.Omega_b + cosmo->Omega_nu_mass)*cosmo->params.h*cosmo->params.h;
      
      (the omega_nu_mass here should should be calculated as for emu_para[7] without the h factors in the following lines, there is an if else catch for the neutrino definitions in line 330)
      
This is somewhat related to the issue here:  #708 

@elisachisari

patricialarsen:
This issue has been fixed, the current benchmarks updated and a new benchmark test (test_emu_lin) added to catch issues like this in future.

PR #899, opened 2021-07-20
fix for issue 897, cosmic emulator omega_M missing massive neutrinos

patricialarsen:
Fix for https://github.com/LSSTDESC/CCL/issues/897

This changes two things 

- Alters the calculation of Omega_M used to call the emulator from omega_c+omega_b to omega_c+omega_b+omega_nu
- Alters the cosmology Omega_M values for the benchmark tests to the same definition

This should fix the issue without breaking the benchmark, and make the linear theory spectra more consistent with the other implementations when incorporating massive neutrinos.


elisachisari (commented):
The changes you proposed makes sense: adding neutrinos to emu_par[0], which was incorrectly interpreted as not including neutrinos from Table 1 of https://arxiv.org/pdf/1705.03388.pdf. This requires changing the benchmark cosmologies as well, and I've checked this is done consistently. 
The only thing I can suggest is adding the check you did to discover this issue to the benchmarks. Would that be possible?

patricialarsen:
Interesting suggestion, I did two checks to discover the issue - one was the same as the benchmark, but checking over the cosmologies for the input points and confirming the definition of omega_M, so that is now consistent with what is there. 

The other is comparing the low-k values against the halofit/linear power spectra with the same cosmology from CCL. At the emulation points at redshift 0 (restricting to cosmologies without a w=-1 crossing) you should see agreement within about 1%

image

So we could add a check that between k= 10^-3 and 10^-2 the power spectra should be consistent to within, say, 3% of the linear power spectrum at a selection of emulation points. At higher redshifts time evolving dark energy implementations can have differences at very linear scales, so this should be implemented at z=0. It would also catch any bugs in how you specify w0/wa.

patricialarsen:
Note: new commit is just adding a line to the changelog

patricialarsen:
I have added the new benchmark test for a selection of 3 emulator input cosmologies (taken from the Lawrence et al 2017 paper and written in benchmarks/data/emu_input_cosmologies.txt), as suggested above. This is added within test_emu.py and called test_emu_lin. It compares the emulator power spectrum between k=10^-3 and k=10^-2 to the linear power spectrum and enforces a <3% agreement. 

This would catch the issue above, and enforces agreement between the linear power spectrum and the emulator. These benchmarks all pass when I run the test. 

Further commit is fixing lint issues in the benchmark test.

patricialarsen:
Note: update is changing the linear comparison to boltzmann_camb rather than class. The emulator was built off a modified camb implementation and so this is a closer comparison to the input.  You can see the difference in the fractional error for the model that was failing the integration test below. 

I suspect the difference is due to differences in the way time evolving dark energy is handled as this model has a high wa value. The emulator should be very consistent with the camb fluid dark energy implementation, while class by default uses the PPF approximations which can diverge at very low k (but have the benefit of allowing w=-1 crossings). If you look for instance at https://camb.info/ppf/ you can see the known issues 

camb_class

benchmarks/test_emu.py
```diff
@@ -73,3 +73,47 @@ def test_emu(model):
...
+    a = 1
+    k = np.logspace(-3, -2, 50)
+
+    # Catch warning about neutrino linear growth
+    if (np.sum(mnu) > 0):
+        pk = ccl.pyutils.assert_warns(ccl.CCLWarning,
+                                      ccl.nonlin_matter_power,
+                                      cosmo, k, a)
+    else:
+        pk = ccl.nonlin_matter_power(cosmo, k, a)
```
elisachisari, line 111: Small thing - should we use ccl.linear_matter_power here? Same above in 108

patricialarsen: I'm not sure what you mean: these two lines are intended to get the non-linear matter power from the emulator for the comparison, and the linear_matter_power for the comparison is in line 114.

elisachisari: Ah, ok, sorry I was thinking of the line 114 input indeed

elisachisari (approved):
Nice work by @patricialarsen catching this issue and fixing it. We iterated a bit on the changes and I think the PR is now ready to merge. The changelog has been updated and new benchmarks have been provided.

PR #901, opened 2021-08-02
Additional PT biases

### LINKED ISSUE #900: b3nl and bk in CCL
Add the interface with fastpt for b3nl and bk bias terms.

damonge:
Adds third-order and non-local bias.
Closes #900

### DIFF
--- a/pyccl/nl_pt/power.py
+++ b/pyccl/nl_pt/power.py
@@ -190,7 +190,7 @@ class PTCalculator(object):
                0.25*(b21*b22)[None, :] * (Pd2d2 - 2.*s4) +
                0.5*(b11*bs2 + b12*bs1)[None, :] * Pd1s2 +
                0.25*(b21*bs2 + b22*bs1)[None, :] * (Pd2s2 - (4./3.)*s4) +
-               0.25*(bs1*bs2)[None, :] * (Ps2s2 - (8./9.)*s4)+
+               0.25*(bs1*bs2)[None, :] * (Ps2s2 - (8./9.)*s4) +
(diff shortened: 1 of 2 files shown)


c-d-leonard:
Nice, I see you've tagged Jonathan and Shivam but let me know if you want me to take a look too.

c-d-leonard:
@damonge @jablazek @shivamp1495 what's the status here?

benchmarks/data/codes/pt_bm.py
```diff
@@ -45,6 +45,8 @@
 Pd1s2 = g4 * dd_bias[4][None, :]
 Pd2s2 = g4 * dd_bias[5][None, :]
 Ps2s2 = g4 * dd_bias[6][None, :]
+Pd1d3 = g4 * dd_bias[8][None, :]
```
jablazek, line 48: I think we should be careful with notation. This is b3nl, which is not the classic b3 (which gets renormalized) and isn't the factor in front of delta^3.

pyccl/nl_pt/power.py
```diff
@@ -51,11 +51,22 @@ class PTCalculator(object):
         C_window (float): `C_window` parameter used by FAST-PT
             to smooth the edges and avoid ringing. See FAST-PT
             documentation for more details.
+        k_cutoff (float): exponential cutoff scale. All power
+            spectra will be multiplied by a cutoff factor of the
+            form :math:`\\exp(-(k/k_*)^n)`, where :math:`k_*` is
+            the cutoff scale. This may be useful when using the
```
jablazek, line 57: This is useful functionality, but it seems it is more logically placed in the transform part of the code. Why have it here?

PR #904, opened 2021-09-10
LPT bias expansions

damonge:
Implements functionality to compute power spectra using Convolution Lagrangian Effective Field Theory expansions ([this](https://arxiv.org/pdf/1706.03173.pdf))

@ntbfin00 should get cookie points for this


c-d-leonard:
Should I think about this as an alternative PT bias implementation that can be used instead of the fastpt implementation?

c-d-leonard:
Status, @damonge?

damonge:
This is heavily outdated and superseded by #1097

PR #905, opened 2021-10-20
CIB profiles

damonge:
Adds a halo profile for the Cosmic Infrared Background. Still WIP:

- [x] Finish profile implementation
- [x] Add tracer
- [x] Find benchmark and add tests


damonge:
This is ready for review now. Whoever ends up reviewing this should bear the following in mind:
- This PR includes a halo profile model for the CIB following the model of [Shang et al.](https://arxiv.org/abs/1109.1522). It is benchmarked against the same calculation by @borisbolliet 's [class_sz](https://github.com/borisbolliet/class_sz).
- You can ignore anything to do with readthedocs (which is the vast majority of the new files). I recompiled that documentation because it was severely out of date.
- Besides this, I've fixed a couple of issues with old documentation (in covariance.py and cls.py).

borisbolliet (commented):
checked on my side, for 217, 545 and 3000GHz auto-power spectra. 
class_sz and ccl match perfectly. Minor diffs at 3000Ghz with Fiona McCarthy's results (probably due to flux cut not taken into account in ccl/class_sz in this particular computation, see https://arxiv.org/abs/2010.16405).
class_sz script available at this link
https://github.com/borisbolliet/class_sz/commit/e66bc589f8e53175f315c6845a5739795b8513e7

borisbolliet (approved):
checked on my side, for 217, 545 and 3000GHz auto-power spectra.
class_sz and ccl match perfectly. Minor diffs at 3000Ghz with Fiona McCarthy's results (probably due to flux cut not taken into account in ccl/class_sz in this particular computation, see https://arxiv.org/abs/2010.16405).
class_sz script available at this link
borisbolliet/class_sz@e66bc58

borisbolliet:
here is the comparison plot at 3000GH
[cib_comparison_3000_3000GHz.pdf](https://github.com/LSSTDESC/CCL/files/7556941/cib_comparison_3000_3000GHz.pdf)
z.

pyccl/halos/profiles_cib.py
```diff
@@ -0,0 +1,316 @@
...
+        gamma (float): high frequency slope.
+        s_z (float): luminosity evolution slope.
+        log10meff (float): log10 of the most efficient mass.
+        sigLM (float): logarithmic scatter in mass.
+        Mmin (float): minimum subhalo mass.
+        L0 (float): luminosity scale (in
+            :math:`{\\rm Jy}\\,{\\rm Mpc}^2\\,M_\\odot^{-1}`).
+    """
+    one_over_4pi = 0.07957747154
+
```
nikfilippas, line 81: might be useful to give these profiles a name, like in all other halo profiles

nikfilippas: also, `one_over_4pi` could become a private attribute as it is only stored here to save lines of code and doesn't serve any other useful function within the profile

damonge: both good points! Done.

borisbolliet:
Just implemented the flux cut (i.e., corresponding to point sources that are masked in the intensity maps - see sec. II.A.6 [here](https://arxiv.org/pdf/2010.16405.pdf)), it is indeed what solves the small difference that there was between class_sz with Fiona's results.
See plot [with_cib_flux_cut.pdf](https://github.com/LSSTDESC/CCL/files/7592340/with_cib_flux_cut.pdf) and class_sz commit [there](https://github.com/borisbolliet/class_sz/commit/972190b1071ceb8d4ef31688c5a8b49a964062d9). 
Just something to keep in mind!
(and many thanks to Fiona McCarthy and Mat Madhavacheril (@msyriac) for all the help with this implementation.)

yanzastro:
Can we also add the model from arXiv: 2006.16329?

Also, I think the _fourier_variance() function for CIB needs some modification if we cross-correlate CIB with a HOD profile.

damonge:
Adding Maniyar et al. would be great!

yanzastro:
The function `_Lumsat` to calculate satellite luminosity is indeed a bit slow. Maybe we can change it to:

`    def _Lumsat(self, M, a, nmsub=10):
        Mmin = self.Mmin
        Ms = np.ones((M.size, nmsub)) * M[:,None]        
        msubs = np.geomspace(Mmin, M, nmsub).T  # integrated variables. If M<Mmin, this generates a decreasing array
        Leff_sub = self._Lum(np.log10(msubs).reshape(-1), a)
        nmsubs = (self.dNsub_dlnM_TinkerWetzel10(msubs.reshape(-1), Ms.reshape(-1))*Leff_sub).reshape(M.size, nmsub)
        Lumsat  = simps(nmsubs, np.log(msubs), axis=1)
        Lumsat[Lumsat<0] = 0 # if M<Mmin, the decreasing integrated variable will yield a negative Lumsat and we can just change them to be zero.
        return Lumsat`

The intervals between integrated variables for different M are different, but that will not change the result very much. One can adjust nmsub to make the integral more precise.  This speeds up the function at a factor of ~3

damonge:
@yanzastro I was looking into this now. I don't think this does exactly what we want, right?
I don't see why `msubs` needs to start at `Mmin`. `Mmin` is the minimum parent halo mass that can host subhalos, right? If so, subhalo masses can be smaller than `Mmin`, I think.

damonge:
Humm, from Eq. 15 of https://arxiv.org/pdf/1109.1522.pdf it may actually be both!

yanzastro:
> 

Actually I agree with you. I'm also confused why they (subhalo mass and host halo mass) had to both start with `Mmin`. In https://www.aanda.org/articles/aa/pdf/2021/01/aa38790-20.pdf, the authors use an arbitrary `Mmin` (=10^5M_sun) for subhalo mass and argue that this does not affect much at a range 10^5~10^7 Msun. I guess we can also do something like this?

yanzastro:
It might also help to add the CIB model in my recent paper https://arxiv.org/abs/2204.01649? It is similar to Maniyar et al. 2021, with a different HOD model.

PR #906, opened 2021-10-28
Emulator support for CCL

nikfilippas:
**Update**: This now relies on `CCLObject` (PR934), so leaving it as draft until PR934 is reviewed. In particular, it makes use of the caching framework implemented there.

I have built some infrastructure for CCL to work with emulators, be it power spectrum emulators, or halo model emulators.
This loads the emulator to memory only once (per thread) given a current config. It should be straightforward to use.

I have added 2 examples: `baccoemu` for the linear, non-linear, and baryons-non-linear matter power spectrum (which can be loaded and used directly from `Cosmology` using its own emulated transfer function: `ccl.CosmologyVanillaLCDM(transfer_function='bacco')` and a halo mass function emulator (`Bocquet20`), which can be directly accessed just like every other mass function in the halo model.

The base infrastructure lives in `pyccl.emulator` with some additional definitions (for power spectra) in `pyccl.pk2d`.

----


damonge:
@nikfilippas I'd like to decouple this from the v3 PR, as this requires a more involved discussion among others in CCL, but the v3 branch is a lot more urgent (since it's blocking the addition of new science as per our plan). Can you do that?

nikfilippas:
No problem, I'll decouple it once we get to the v3 branch review.
Note, this merely provides a framework for emulators to work, i.e. if a user wants they can use our code instead of having to build an external adapter module to link the emulator they want to use with CCL. The two I am testing here are only examples, and I'm happy to remove them (also for licensing reasons) and only leave the framework.

damonge:

PR #914, opened 2021-12-07
typo in param naming

nikfilippas:
In `halomod_power_spectrum` I had written `k_1h_supress` for the HM-Code 1-halo suppression factor, instead of "suppress".
This breaks API but since no one has complained so far about my bad english, I assume no one has used it, so we can merge it, break API, and still get away with it.


damonge:
I did notice XD. How about adding this to all your other modifications so we merge it together with those as part of the v3 overhaul?

damonge:
I should say, this is actually my fault, since I reviewed that PR. Probably best not to break the API though.

nikfilippas:
Okay, closing this now because I've merged this branch to `star`.

PR #917, opened 2022-01-29
10x speed increase in CIB profile

nikfilippas:
Using `numpy` broadcasting rules we can get at least a 10-fold speed increase (and up to 100x) in the computation of `_Lumsat` in `HaloProfileCIBShang12`, thus avoiding having to write this in C.

### DIFF
--- a/pyccl/halos/profiles_cib.py
+++ b/pyccl/halos/profiles_cib.py
@@ -200,7 +200,7 @@ class HaloProfileCIBShang12(HaloProfile):
         dnsubdlnm = self.dNsub_dlnM_TinkerWetzel10(10**msub, M_use)
         integ = dnsubdlnm * Lum
         Lumsat = simps(integ, x=np.log(10)*msub)
-        res[-len(Lumsat): ] = Lumsat
+        res[-len(Lumsat):] = Lumsat
         return res
 
     def _real(self, cosmo, r, M, a, mass_def):
--- a/pyccl/halos/profiles_cib.py
+++ b/pyccl/halos/profiles_cib.py
@@ -189,18 +189,19 @@ class HaloProfileCIBShang12(HaloProfile):
         return Lumcen
 
     def _Lumsat(self, M, a):
-        logM = np.log10(M[None, :])
-        # log-spaced integration points (10 per dex)
+        res = np.zeros_like(M)
+        M_use = M[M >= self.Mmin, None]
+        logM = np.log10(M_use)
         LOGM_MIN = 10
         nm = max(2, 10*int(np.max(logM) - LOGM_MIN))
-        msub = np.linspace(LOGM_MIN, np.max(logM), nm+1)[:, None]
+        msub = np.linspace(LOGM_MIN, np.max(logM), nm+1)[None, :]
 
-        dnsubdlnm = self.dNsub_dlnM_TinkerWetzel10(10**msub, M)
         Lum = self._Lum(msub, a)
+        dnsubdlnm = self.dNsub_dlnM_TinkerWetzel10(10**msub, M_use)
         integ = dnsubdlnm * Lum
-        Lumsat = simps(integ, x=np.log(10)*msub, axis=0)
-        Lumsat[M < self.Mmin] = 0.
-        return Lumsat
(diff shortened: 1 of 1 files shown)


pyccl/halos/profiles_cib.py
```diff
@@ -189,21 +189,19 @@ def _Lumcen(self, M, a):
...
+        M_use = M[M >= self.Mmin, None]
+        logM = np.log10(M_use)
+        LOGM_MIN = 10
+        nm = max(2, 10*int(np.max(logM) - LOGM_MIN))
+        msub = np.linspace(LOGM_MIN, np.max(logM), nm+1)[None, :]
+
+        Lum = self._Lum(msub, a)
+        dnsubdlnm = self.dNsub_dlnM_TinkerWetzel10(10**msub, M_use)
+        integ = dnsubdlnm * Lum
+        Lumsat = simps(integ, x=np.log(10)*msub)
```
damonge, line 202: I am not 100% sure this is doing the same calculation. Note that the upper limit of the integral being calculated in the old loop changes with each element of the loop (Mparent). Here, the upper limit is always the same.

nikfilippas: Yes, it's not the same, but we can exploit the fact that the subhalo mass function drops extremely rapidly after `Msub > Mparent`, so the high-mass contributions are next to zero.
To be precise, `Lum` changes by some orders of magnitude as mass increases, but `dnsubdlnm` instantly drops 8 orders of magnitude for a mass change of 0.1 dex, which is the sampling rate either way.
So unless I didn't read it right, the high-mass contributions are negligible.

pyccl/halos/profiles_cib.py
```diff
@@ -189,21 +189,19 @@ def _Lumcen(self, M, a):
...
-                # Sample integrand
-                dnsubdlnm = self.dNsub_dlnM_TinkerWetzel10(msub, Mparent)
-                Lum = self._Lum(np.log10(msub), a)
-                integ = dnsubdlnm*Lum
-                Lumsat[iM] = simps(integ, x=np.log(msub))
-        return Lumsat
+        res = np.zeros_like(M)
+        M_use = M[M >= self.Mmin, None]
+        logM = np.log10(M_use)
+        LOGM_MIN = 10
```
damonge, line 195: It seems the model actually uses `Mmin` as the lower limit of this integral too (see Eq. 15 of https://arxiv.org/pdf/1109.1522.pdf). Can you make this modification?

pyccl/halos/profiles_cib.py
```diff
@@ -189,21 +189,19 @@ def _Lumcen(self, M, a):
...
+        nm = max(2, 10*int(np.max(logM) - LOGM_MIN))
+        msub = np.linspace(LOGM_MIN, np.max(logM), nm+1)[None, :]
```
damonge, line 197: I think you probably need to check that `np.max(M) > self.Mmin`, and if that's not the case, just return 0 everywhere.

damonge (changes requested):
@nikfilippas ok, I like this (in fact I find the high-mass sharp cutoff rather unnatural - the model should have use the subhalo mass function's natural cutoff).

Just a couple of tiny comments and this should be ready to merge.

damonge:
One more comment though: I don't find this to be 10x faster when it comes to computing a C_ell, for example. In that case I find a ~x2 speedup, which is good! I guess the 10x claim is just when computing `Lumsat` itself?

pyccl/halos/profiles_cib.py
nikfilippas: Which line does this refer to? I can't see where `Mmin` isn't used as the lower limit.

pyccl/halos/profiles_cib.py
nikfilippas: ok done

pyccl/halos/profiles_cib.py
damonge: self.Mmin

nikfilippas: oh okay, done

damonge (approved):
LGTM!

Issue #588, opened 2019-03-01
Pass (more) CLASS accuracy parameters from pyccl


c-d-leonard:
At the moment we cannot pass any accuracy parameters to CLASS from the python side, and even from within C we only pass two: z_max_pk and P_k_max_1/Mpc. CLASS has many more accuracy parameters.

Related to this: It has been shown (see e.g. this [notebook)](https://github.com/xzackli/Cosmology-Notebooks/blob/master/CLASS%20vs.%20CAMB.ipynb)) that with (very) high accuracy settings, P(k)'s from CLASS and CAMB basically agree to 1e-4 for all k we care about. However, I checked this high-accuracy-setting CAMB output against P(k) as computed via CCL with our default accuracy settings (see a modified version of that notebook, [here](https://github.com/c-d-leonard/Cosmology-Notebooks)) and these agree only at >1e-3. 

The result is that for functionality for which we need to benchmark against CAMB or codes based on CAMB (see e.g. PR #442 for mu / Sigma support) we currently need to relax the benchmark agreement requirements to this level.

c-d-leonard:
I think this issue may be obsolete because we now include multiple boltzmann codes so we are no longer benchmarking against camb (since we integrated it). We may still want to pass these accuracy parameters though. Thoughts? @beckermr @damonge @elisachisari

beckermr:
Yes we should keep this open. Also, we in general need to clean up the accuracy parameters and create some simpler-to-understand options for people.

PR #918, opened 2022-02-07
Tunable CCL global settings at the Python-level

nikfilippas:
This PR decouples the (otherwise connected at the C level) CCL accuracy parameters from `Cosmology` objects, at the Python level.

In a new module `parameters.py` I have implemented the class `CCLParameters` which wraps the SWIG-generated classes for the spline parameters, the GSL parameters, and the physical constants.

In the spirit of making `Cosmology` objects truly immutable (which they ought to be) I have deprecated the option to mutate the accuracy parameters of any instantiated `Cosmology`: all new instances are populated with the global parameters at the time they are created.

Users can access (and set) the global parameters using `ccl.gsl_params`, `ccl.spline_params`, and `ccl.physical_constants`.

To do this I removed the `const` type definition of the C-level spline and GSL parameters, so they are added in `ccllib.cvar` by SWIG. Physical constants were already mutable (because `T_CMB` is temporarily mutated in `Cosmology`).

This feature comes with the advantage that we can remove all dummy cosmologies from the codebase. Every time we need one just to look up the accuracy parameters, we can just peek at the parameters at the time of instantiation, if a `Cosmology` is not provided.

Maybe it's also worth considering having something like `ccl.spline_params.set_accuracy("high" | "medium" | "low")` with pre-made parameter sets. High accuracy would be what is expected from LSST-like analyses with slower computation, while low accuracy will give a considerable increase in speed. This is one of the things mentioned in #588.


tilmantroester:
@nikfilippas Do you want to rebase/merge master and I then rewrite the warning to match the new `gsl_params` setup?

nikfilippas:
@tilmantroester okay, merged and adjusted the warning as required. Have you managed to have a look? Would you like to review this?

pyccl/errors.py
```diff
@@ -20,3 +23,23 @@ def __eq__(self, other):
...
+    def __hash__(self):
+        return hash(repr(self))
+
+    @classmethod
+    def enable(cls):
+        warnings.simplefilter("always")
+
+    @classmethod
+    def disable(cls):
+        warnings.filterwarnings(action="ignore", category=cls)
```
damonge, line 45: Why do we need this in this PR? Is this a leftover from the v3 PRs?

pyccl/pk2d.py
```diff
@@ -56,10 +57,8 @@ class Pk2D(object):
              logarithm of the power spectrum. Otherwise, the true value
              of the power spectrum is expected. Note that arrays will be
              interpolated in log space if `is_logp` is set to `True`.
-        cosmo (:class:`~pyccl.core.Cosmology`): Cosmology object. The cosmology
-             object is needed in order if `pkfunc` is not `None`. The object is
-             used to determine the sampling rate in scale factor and
-             wavenumber.
+        cosmo (:class:`~pyccl.core.Cosmology`, optional): Cosmology object.
+             Used to determine sampling rates in scale factor and wavenumber.
```
damonge, line 61: Is this still the case? (I.e. is this still used to determine sampling rates?)

pyccl/tests/test_cclerror.py
```diff
@@ -34,3 +39,27 @@ def test_cclwarning_not_equal():
...
+    assert w[0].category == pyccl.CCLDeprecationWarning
+
+    # switch off CCL (future) deprecation warnings
+    pyccl.CCLDeprecationWarning.disable()
+    with warnings.catch_warnings(record=True) as w:
+        warnings.warn("test", pyccl.CCLDeprecationWarning)
+    assert len(w) == 0
+
+    # switch back on
+    pyccl.CCLDeprecationWarning.enable()
```
damonge, line 65: Should all these live in a different PR?

src/ccl_core.c
```diff
@@ -597,3 +615,18 @@ void ccl_get_pk_spline_lk_array(ccl_cosmology *cosmo,int ndout,double* doutput,i
...
+    d = ccl_log_spacing(spline_params->K_MIN, spline_params->K_MAX, ndout);
+    if (d == NULL)
+      *status = CCL_ERROR_MEMORY;
+  }
+  if (*status == 0) {
+    for(int ii=0; ii < ndout; ii++)
+      doutput[ii] = log(d[ii]);
+  }
+  free(d);
+}
```
damonge, line 632: Since you created these two nice functions, should we put them to use in `src/ccl_power.c` to reduce the clutter there?

pyccl/parameters.py
```diff
@@ -0,0 +1,120 @@
...
+                f"Direct assignment of {name} is deprecated. "
+                f"Set via `pyccl.{name}.{key}` before instantiation.",
+                CCLDeprecationWarning)
+            object.__setattr__(self, key, value)
+
+        cls._instance.__class__.__setattr__ = _new_setattr
+
+    def __init__(self):
+        # Keep a copy of the default parameters.
+        object.__setattr__(self, "_bak", CCLParameters.get_params_dict(self))
```
damonge, line 49: same

pyccl/parameters.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def __getattribute__(self, name):
+        get = object.__getattribute__
```
damonge, line 52: just explain this to me

damonge (changes requested):
OK, thanks a lot @nikfilippas . First review. Main thing is you need to explain to me this `object` business you're using in parameters.py

pyccl/__init__.py
```diff
@@ -70,4 +70,4 @@
 # Expose function to toggle debug mode
 from .pyutils import debug_mode, resample_array
 
-from .errors import CCLError, CCLWarning
+from .errors import CCLError, CCLWarning, CCLDeprecationWarning
```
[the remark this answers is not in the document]
nikfilippas: Yes, in the spirit of making Cosmology objects truly immutable, I am now allowing `spline_params` and `gsl_params` of a particular Cosmology to change, but a `CCLDeprecationWarning` will be thrown.
I think this is good enough for the time being until v3, where the deprecation cycle will come to an end and trying to change any params after instantiation will throw an error.

pyccl/boltzmann.py
```diff
@@ -118,7 +119,7 @@ def get_camb_pk_lin(cosmo, nonlin=False):
     # thus we set T_i_eff = T_i = g^(1/4) * T_nu and solve for the right
     # value of g for CAMB. We get g = (TNCDM / (11/4)^(-1/3))^4
     g = np.power(
-        lib.cvar.constants.TNCDM / np.power(11.0/4.0, -1.0/3.0),
+        physical_constants.TNCDM / np.power(11.0/4.0, -1.0/3.0),
```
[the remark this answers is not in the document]
nikfilippas: uhmm I have no opinion on that. I just copied what was already there. I suggest let's leave it as is for now.

src/ccl_core.c
nikfilippas: These are very similar to the 2 functions below it. Instead of calculating `na` and `nk` from `cosmo`, they calculate `na` and `nk` from a `ccl_parameters` struct. So, all 4 functions belong to the same file. Shall they live in `ccl_core.c` or in `ccl_power.c`? Your call.

pyccl/parameters.py
```diff
@@ -0,0 +1,120 @@
...
+        def _new_setattr(self, key, value):
+            # Make instances of the SWIG-level class immutable
+            # so that everything is handled through this interface.
+            # SWIG only assigns `this` via the low level `_ccllib`;
+            # we therefore disable all other direct assignments.
+            if key == "this":
+                return object.__setattr__(self, key, value)
+            name = self.__class__.__name__
+            # TODO: Deprecation cycle for fully immutable Cosmology objects.
+            # raise AttributeError(f"Direct assignment in {name} not supported.")  # noqa
```
damonge, line 38: Add a comment that this is what will happen in v3 (i.e. an error will be thrown)

benchmarks/test_tracers.py
```diff
@@ -20,15 +20,17 @@ def get_prediction(ells, chi_i, chi_f, alpha, beta, gamma,
 
 @pytest.fixture(scope='module')
 def set_up():
+    ccl.physical_constants.T_CMB = 2.7
```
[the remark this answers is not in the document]
damonge: OK, I see. I'd like to get rid of this altogether. It seems extremely silly. I'll open an issue.

pyccl/pk2d.py
damonge: Can you add "Will be deprecated in the next CCL version"?

pyccl/parameters.py
damonge, line 49: Just so I understand: here you're using `object.__setattr__` instead of `self._bak` because you changed `self`'s setattr below, right?

src/ccl_core.c
damonge: No, no, they can live here or wherever you want. What I mean is why not use them in `ccl_power.c` to generate the k and a arrays? E.g. [here](https://github.com/LSSTDESC/CCL/blob/c01d3b032bff936cc752a7f2b6b94e23758d9d9d/src/ccl_power.c#L59) or [here](https://github.com/LSSTDESC/CCL/blob/c01d3b032bff936cc752a7f2b6b94e23758d9d9d/src/ccl_power.c#L76).

damonge (commented):
Ok, a couple of new comments and one old one remaining.

One overall comment: in an ideal world I'd have preferred to start by cleaning up the C layer so actually none of this need to live there. This would simplify a lot of the code you've had to add here, but c'est la vie. I think this is what we'll do in the future.

pyccl/pk2d.py
nikfilippas: Let's not deprecate this. `cosmo` is still used to extrapolate backwards in time using the linear power spectrum, when the requested scale factor is outside of the range of the `Pk2D` object.
With my changes, `eval` will work as before, but users won't be able to extrapolate backwards.

nikfilippas: Oh wait no! I got my PRs mixed up. No, we won't be deprecating this. `cosmo` will stay there, but I think I've addressed your point in another PR of mine, #923 .

pyccl/parameters.py
nikfilippas: Done.

pyccl/parameters.py
nikfilippas: Yes. Before `__init__` is called, `self`'s dictionary doesn't contain the entry `_bak`. So if you try to do `self._bak` directly, it will `raise KeyError(f"Parameter _bak does not exist.")`.

src/ccl_core.c
nikfilippas: Oh, you said "put them to use in", I read "put them in".

nikfilippas: Okay, I know what you mean but I can't see which lines should be replaced exactly. And why would they have to be replaced with my functions that accept a `ccl_parameters` struct? The functions you showed me take `cosmo` as an argument. The C code already has versions of my functions that accept `cosmo`. So why bother with my functions?

nikfilippas:
@damonge back to you. I agree with your overall comment. CCL has to be re-written in C++ like @tilmantroester has suggested, but changing the entire internal structure will require a lot of work. This PR does the job in what I believe is an elegant way, but at some point how CCL handles the parameters needs to change.

pyccl/ccl_core.i
```diff
@@ -4,7 +4,7 @@
 /* put additional #include here */
 %}
 
-// SWIG black magic. Change the behaviour of setting A_SPLINE_MAX to thowing an
+// SWIG black magic. Change the behaviour of setting A_SPLINE_MAX to throwing an
 // error.
 %typemap(memberin) double A_SPLINE_MAX {
```
tilmantroester, line 9: Can't we remove that now and disallow changing `A_SPLINE_MAX` in the `__setattribute__` method?

pyccl/parameters.py
```diff
@@ -0,0 +1,121 @@
...
+            return get(self, name)
+
+    def __setattr__(self, key, value):
+        if self._frozen and key != "T_CMB":
+            # `T_CMB` mutates in Cosmology.
+            name = self.__class__.__name__
+            raise AttributeError(f"Instances of {name} are frozen.")
+        if not hasattr(self._instance, key):
+            raise KeyError(f"Parameter {key} does not exist.")
+        if (key, value) == ("A_SPLINE_MAX", 1.0):
```
tilmantroester, line 66: `A_SPLINE_MAX` should be immutable.

pyccl/ccl_core.i
nikfilippas: Yes we can, but I didn't want to touch the C-code for that one in case someone changes the Python-code in the future and forgets to reinstate that check. There are a couple of places this check appears in the C-code.

pyccl/parameters.py
nikfilippas: It is. Here, for example I am making use of the C-code. If `A_SLINE_MAX` gets set to 1, then the function returns doing nothing. If not, the C-code takes over and raises.

damonge:
@tilmantroester @nikfilippas are we happy to leave C-level modifications for a future PR and merge this one? If so, I can approve.

pyccl/parameters.py
tilmantroester: Could you make it raise an exception even when being set to 1? `A_MAX_SPLINE` shouldn't be touched, even when trying to assign the "correct" value.

nikfilippas: I would argue it's okay to set it to its correct value. It is actually touched (set) whenever the parameters are reloaded to their default values, through `spline_params.reload()`.

Also in the (near) future, we might want to relax this requirement to see how things evolve at scale factors larger than 1 (i.e. the far future).

nikfilippas: Jokes aside, if we have it raise an error, I would have to fall back to `object.__setattr__` in `reload`, which is slightly slower.

tilmantroester: Fine. I'd just caution that the C code implicitly assumes `A_MAX_SPLINE==1` in a couple of places and allowing setting `A_MAX_SPLINE` needs a careful review of much of the C code base to make sure this is okay.

nikfilippas:
@damonge I think there is consensus now.

damonge:
@nikfilippas can you resolve the conflicts? I'll approve then when the tests pass.

nikfilippas:
@damonge Done.

damonge (approved):
LGTM!

PR #921, opened 2022-02-18
Update docs and tests to use mead2020_feedback.

tilmantroester:
The docs suggest that `halofit_version=mead2020` accounts for baryon feedback, but that should be `mead2020_feedback`.
This also fixes the tests to get the agreement with CAMB to O(1e-5) by using the same setup as in `boltzmann.py` (namely setting T_CMB and using the same k-range).


nikfilippas (approved):
LGTM!

PR #922, opened 2022-03-23
IA halo model

chrgeorgiou:
A draft pull request with the implementation of the halo model for intrinsic alignments within CCL.


damonge:
@chrgeorgiou I see you marked this as ready for review a couple of weeks ago. Are you still looking for one? If so, can you provide an example of what this tries to achieve? I see you mentioned a cclx notebook.

chrgeorgiou:
Hi @damonge, indeed the PR is ready for review as soon as I fix the coverage issue (which decreased 2%). I think its best if I create a CCLX dev branch with a new notebook showcasing what this CCL development is trying to achieve. I will link this here once I have finished it.

damonge:
OK, great. Feel free to ping any of us for a review when you think it's ready!

chrgeorgiou:
The development of the branch is now completed and I am happy to have it reviewed. To make this easier, I am happy to book a few hours for answering questions about what has been developed here.

doc/0000-ccl_note/main.tex
```diff
@@ -772,6 +772,16 @@ \subsubsection{Galaxy-IA cross correlation}
...
+ \begin{align}
+	P_{\delta \mathrm{I}, 1 \mathrm{~h}}^{s}(\mathbf{k}, z)&=\int \mathrm{d} M n(M) \frac{M}{\bar{\rho}_{m}} f_{s}(z) \frac{\left\langle N_{s} | M\right\rangle}{\bar{n}_{s}(z)}\left|\hat{\gamma}_{s}^{I}(\mathbf{k} | M)\right| u(k, M)\\
+	P_{\mathrm{II}, \mathrm{h}}^{s s}(\mathbf{k}, z)&=\int \mathrm{d} M n(M) f_{s}^{2}(z) \frac{\left\langle N_{s}\left(N_{s}-1\right) | M\right\rangle}{\bar{n}_{s}^{2}(z)}\left|\hat{\gamma}_{s}^{I}(\mathbf{k} | M)\right|^{2},
+  \label{eq:IAHaloModelPowerSpectra}
+\end{align}
+where $n(M)$ is the halo mass function, $f_s(z)$ is the fraction of satellites as a function of redshift, $\langle N_s|M\rangle$ is the HOD of the satellites, $\hat{\gamma}_s^I(\mathbf{k}|M)$ is the Fourier transform of the intrinsic shear field, and $\bar{n}_s$ is the average number density of satellites.
```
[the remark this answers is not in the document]
chrgeorgiou: Yes, good point. I added some text following the equations in Fortuna et al. 2021.

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1553,314 @@ def _Ns(self, M, a):
...
+
+    def _I_integral(self, a, b, e=1E-10):
+        '''
+        Computes the integral
+        .. math::
+            I(a,b) = \\int_{-1}^1 \\mathrm{d}x (1-x^2)^{a/2}x^b.
+        '''
+        from scipy.integrate import simps
+        x_range = np.linspace(-1 + e, 1 - e, 1024)
+        return simps((1 - x_range ** 2) ** (a / 2.) * x_range ** b, x_range)
```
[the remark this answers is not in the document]
chrgeorgiou: Good point here as well, I have implemented this in the code. I checked that the results from the two methods were the same.

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1553,314 @@ def _Ns(self, M, a):
...
+            for j in range(0, m + 1):
+                m_sum += (binom(m, j) * gj[j] *
+                          self._I_integral(j + var1_add, m - j) *
+                          np.sin(thk) ** j * np.cos(thk) ** (m - j))
+            l_sum += binom(l, m) * binom((l + m - 1) / 2, l) * m_sum
+        if phik is not None:
+            l_sum *= np.exp(1j * 2 * phik)
+        return 2 ** l * l_sum
+
+    def gamma(self, r, r_vir):
```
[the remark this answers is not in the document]
chrgeorgiou: Done!

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1553,314 @@ def _Ns(self, M, a):
...
+        For now, it assumes a NFW profile.
+        '''
+        M_use = np.atleast_1d(M)
+        r_use = np.atleast_1d(r)
+
+        rvir = self._rvir(cosmo, M_use, a, mass_def)
+        # TODO: make it possible to specify the halo density profile?
+        density_prof = HaloProfileNFW(self.cM, truncated=True,
+                                      fourier_analytic=True)
+        rho = density_prof.real(cosmo, r_use, M_use, a, mass_def)
```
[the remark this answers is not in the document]
chrgeorgiou: I changed this as well, and tested that the results are the same.

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1553,314 @@ def _Ns(self, M, a):
...
+
+    def _fftlog_wrap(self, cosmo, k, M, a, mass_def,
+                     l=0,
+                     fourier_out=True,
+                     large_padding=True):
+        '''
+        Adapted from pyccl.halos.profiles.HaloProfile._fftlog_wrap, but
+        modified to allow computation of the Hankel transform with variable
+        spherical bessel function index.
+        '''
```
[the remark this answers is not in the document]
chrgeorgiou: Initially, I did not want to touch code that was already there. But I see that there should be no harm from the changes I made, so I edited the `_fftlog_wrap` directly now.

pyccl/halos/profiles.py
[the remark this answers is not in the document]
damonge: ok, one question: will b always be an integer? I'm wondering about the -1^b part

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1554,262 @@ def _Ns(self, M, a):
...
+                                         for l in range(2, self.lmax+1, 2)])\
+                .reshape(self.lmax//2, 1)
+        self.cM = c_M_relation
+        super(SatelliteShearHOD, self).__init__(c_M_relation,
+                                                fc_0=0.0,
+                                                ns_independent=True,
+                                                **kwargs)
+        self.update_precision_fftlog(padding_lo_fftlog=1E-2,
+                                     padding_hi_fftlog=1E3,
+                                     n_per_decade=512*1,
```
damonge, line 1640: Do you always need this many samples, or is this only for a high-precision benchmark? Also, I guess `512*1=512` :-P

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1554,262 @@ def _Ns(self, M, a):
         M1 = 10.**(self.lM1_0 + self.lM1_p * (a - self.a_pivot))
         alpha = self.alpha_0 + self.alpha_p * (a - self.a_pivot)
         return np.heaviside(M-M0, 1) * (np.fabs(M-M0) / M1)**alpha
+
+
+class SatelliteShearHOD(HaloProfileHOD):
+    def __init__(self, c_M_relation, a1h=0.001, b=-2,
+                 lmax=6, integration_method='FFTLog',
+                 **kwargs):
+        '''
```
damonge, line 1563: The docstring describing the init method usually goes after the class declaration. Check how we do it here for instance https://github.com/LSSTDESC/CCL/blob/d5408a777ab5fc4b2ae6fada2114a168b9bacee4/pyccl/halos/profiles.py#L624

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1554,262 @@ def _Ns(self, M, a):
...
+        With :math:`a_{1\\mathrm{h}}` the amplitude of intrinsic alignments on
+        the 1-halo scale, :math:`b` the index defining the radial dependence
+        and :math:`\\theta` the angle defining the projection of the
+        semi-major axis of the galaxy along the line of sight.
+
+        Args:
+            c_M_relation: concentration-mass relation for the halo model.
+            a1h: Amplitude of the satellite intrinsic shear profile.
+            b: Power-law index of the satellite intrinsic shear profile.
+            If zero, the profile is assumed to be constant inside the halo.
```
damonge, line 1582: see the syntax for Args here https://github.com/LSSTDESC/CCL/blob/d5408a777ab5fc4b2ae6fada2114a168b9bacee4/pyccl/halos/profiles.py#L645
My understanding is that you need to leave 4 spaces when starting a new line on the same argument. It helps with readability, but it may also be that otherwise rtd won't be able to parse them.

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1554,262 @@ def _Ns(self, M, a):
...
+        prof = self.gamma_I(r_use, rvir) * u
+
+        if np.ndim(r) == 0:
+            prof = np.squeeze(prof, axis=-1)
+        if np.ndim(M) == 0:
+            prof = np.squeeze(prof, axis=0)
+
+        return prof
+
+    def _usat_fourier(self, cosmo, k, M, a, mass_def):
```
damonge, line 1754: OK, I'm trying to understand the logic for this one. So you are overloading the `_usat_fourier` method in the HOD profile so that it now returns the satellite shear profile, right?

But then, this will go into [this line](https://github.com/LSSTDESC/CCL/blob/d5408a777ab5fc4b2ae6fada2114a168b9bacee4/pyccl/halos/profiles.py#L1507), which also contains a contribution from centrals to the total number density. What's happening with that contribution?

damonge (changes requested):
Second review!

One overall question: I imagine it's hard to find a benchmark for this. Can you say something about what you've done to ensure that the calculation implemented here is correct?

pyccl/halos/profiles.py
```diff
@@ -1551,3 +1553,314 @@ def _Ns(self, M, a):
         M1 = 10.**(self.lM1_0 + self.lM1_p * (a - self.a_pivot))
         alpha = self.alpha_0 + self.alpha_p * (a - self.a_pivot)
         return np.heaviside(M-M0, 1) * (np.fabs(M-M0) / M1)**alpha
+
+
+class SatelliteShearHOD(HaloProfileHOD):
```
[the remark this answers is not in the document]
damonge: ok, see my question below about centrals

pyccl/halos/profiles.py
chrgeorgiou: That is correct. The contribution from centrals is going to be neglected because of [this line](https://github.com/LSSTDESC/CCL/blob/1ce402c63ca7b01c739b32c55c1fdfce7450a922/pyccl/halos/profiles.py#L1634), where the `fc_0=0` and `ns_independent=True`.

The idea behind this is that, for the alignments between centrals we can use the NLA model. There might be a better way to do this, however, so I am happy to discuss this.

chrgeorgiou:
> Second review!
> 
> One overall question: I imagine it's hard to find a benchmark for this. Can you say something about what you've done to ensure that the calculation implemented here is correct?

True, the benchmark is a bit tricky. I have tested against a development version of CosmoSIS, originally developed by Maria Cristina Fortuna, that the computation of $\gamma^I(k,M)$ is correct. Taking this further is difficult mostly because of the different HOD the two codes employ.

pyccl/halos/profiles.py
chrgeorgiou: I reduced this to 100, which produces accurate enough angular power spectra. I tested this against a highly sampled Simpsons integration.

damonge:
@chrgeorgiou just checking that you saw my last review!

chrgeorgiou:
Hi @damonge, here is my reply to the last points you mentioned.

> OK, final items for review:
> 
>     1. HOD parameters: right now users could mess things up if they pass HOD parameters for centrals. 

I have changed the script to throw a warning if the user ever sets ```fc_0``` or ```fc_p``` at a value different than zero. These are forced to zero. 

>     2. Benchmarks: if you checked against a prediction by Maria Cristina Fortuna, could we add that as a benchmark?

Yes, I believe we can add this as a benchmark, let me know how I can do that. 

Note I have also added a bit of documentation based on a couple of tests I have carried out. If needed, I can add figures to support those tests or share a documented notebook that carries them out but maybe it's okay to leave things as they are.

chrgeorgiou:
@damonge pinging here for PR status. Changes have been completed, let me know if there is anything else needed for this PR.

pyccl/halos/profiles.py
```diff
@@ -1711,3 +1730,282 @@ def _Ns(self, M, a):
...
+                .reshape((6, 1))
+        else:
+            self._angular_fl = np.array([self._fl(l, b=self.b)
+                                         for l in range(2, self.lmax+1, 2)])\
+                .reshape(self.lmax//2, 1)
+        self.cM = c_M_relation
+        if 'fc_0' in kwargs.keys() or 'fc_p' in kwargs.keys():
+            warnings.warn(
+                'Provided fractions for central galaxies are not zero.'
+                'Setting them to zero (fc_0=fc_p=0).',
```
damonge, line 1813: I guess you either need to check that they are not zero or change the message to say that these parameters are not tweakable (I'm thinking of an overzealous user who decides to pass fc_0=0 and fc_p=0 instead of not passing anything)

pyccl/halos/profiles.py
```diff
@@ -1711,3 +1730,282 @@ def _Ns(self, M, a):
...
+        Compute the prefactor due to the satellite intrinsic shear
+        halo model, which is the satellite galaxy fraction f_s=ns/(nc+ns)
+        divided by the satellite galaxy number density ns, making it
+        just 1/ngal.
+        '''
+        from scipy.integrate import simps
+        M_use = hmc._mass
+        ngal = simps(self._Nc(M_use, a) * (1+self._Ns(M_use, a)) *
+                     hmc._massfunc.get_mass_function(cosmo, a=a, M=M_use),
+                     np.log10(M_use))
```
damonge, line 1857: you can use `hmc._lmass` and avoid taking an extra logarithm.

benchmarks/test_haloprofile.py
```diff
@@ -177,3 +177,28 @@ def test_weak_lensing_functions():
...
+
+    sat_gamma_HOD_simps = ccl.halos.SatelliteShearHOD(cM, lmax=6,
+                                                      a1h=0.000989)
+    gamma_k_CCL = sat_gamma_HOD_simps._usat_fourier(cosmo,
+                                                    k_arr,
+                                                    mass_eval,
+                                                    1 / (1 + z_eval),
+                                                    hmd_200m)
+
+    assert np.all(np.fabs(-gamma_k_CCL / gamma_k_CosmoSIS - 1) < 0.8)
```
damonge, line 204: Humm, isn't this precision very bad? Also, what's with the preceding -1 sign in gamma_k_CCL?

damonge (changes requested):
@chrgeorgiou very sorry for the delay. Here's a review. I mostly want to check I understand the logic behind the text.

benchmarks/test_haloprofile.py
damonge: Can you add a comment for us to improve this benchmark in the future?

chrgeorgiou:
Hi @damonge, I have pushed a new PR addressing the final comments on this. 
- On testing the new `_prefactor`, I wasn't sure what to add and I hope what I developed is sufficient. 
- On the benchmarks, the external benchmark from a development version of CosmoSIS is not accurate enough to be trusted quantitatively but serves only as a qualitative comparison. I developed code to rigorously test the accuracy of the FFTLog integration method and added the code I used for this. 

Let me know if there is anything else!

PR #924, opened 2022-04-10
Allow access to Tk3D spline data

nikfilippas:
This PR implements the `Tk3D` equivalent of `Pk2D.get_spline_arrays`.

### DIFF
--- a/pyccl/ccl_utils.i
+++ b/pyccl/ccl_utils.i
@@ -144,18 +144,17 @@ void get_spline3d_arrays(gsl_spline2d **spline2d,
                          int *status)
 {
   // check for inconsistencies //
-  if(spline2d == NULL) {
+  if (spline2d == NULL) {
     *status = CCL_ERROR_MEMORY;
     return;
   }
-  // check only the first array element; the others must be consistent //
-  if(x_size != spline2d[0]->interp_object.xsize) {
-    *status = CCL_ERROR_INCONSISTENT;
-    return;
-  }
-  if(y_size != spline2d[0]->interp_object.ysize) {
-    *status = CCL_ERROR_INCONSISTENT;
-    return;
+
+  for (int ia = 0; ia < na; ia++) {
+    if (x_size != spline2d[ia]->interp_object.xsize
+        || y_size != spline2d[ia]->interp_object.ysize) {
+      *status = CCL_ERROR_INCONSISTENT;
+      return;
+    }
   }
 
   for (int ia = 0; ia < na; ia++) {
@@ -163,6 +162,7 @@ void get_spline3d_arrays(gsl_spline2d **spline2d,
       tarr[ia*x_size*y_size + ik] = spline2d[ia]->zarr[ik];
     }
   }
+
   // no need to do this for every scale factor //
   memcpy(xarr, spline2d[0]->xarr, sizeof(double)*x_size);
   memcpy(yarr, spline2d[0]->yarr, sizeof(double)*y_size);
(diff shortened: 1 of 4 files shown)


pyccl/ccl_utils.i
```diff
@@ -122,4 +136,48 @@ void get_spline2d_arrays(gsl_spline2d *spline2d,
...
+                         int t_size, double* tarr,
+                         int na,
+                         int *status)
+{
+  // check for inconsistencies //
+  if(spline2d == NULL) {
+    *status = CCL_ERROR_MEMORY;
+    return;
+  }
+  // check only the first array element; the others must be consistent //
```
tilmantroester, line 151: Maybe add a check for the other elements to the for loop, since the function accepts an array of arbitrary spline pointers, for which the dimensions could differ.

pyccl/ccl_utils.i
```diff
@@ -122,4 +136,48 @@ void get_spline2d_arrays(gsl_spline2d *spline2d,
...
+      tarr[ia*x_size*y_size + ik] = spline2d[ia]->zarr[ik];
+    }
+  }
+  // no need to do this for every scale factor //
+  memcpy(xarr, spline2d[0]->xarr, sizeof(double)*x_size);
+  memcpy(yarr, spline2d[0]->yarr, sizeof(double)*y_size);
+}
+
+
+void get_array(double *arr,
```
tilmantroester, line 172: Is that just to copy the data? I thought python could already access raw arrays.

pyccl/pyutils.py
```diff
@@ -570,3 +570,33 @@ def _get_spline2d_arrays(gsl_spline):
     check(status)
 
     return yarr, xarr, zarr.reshape(y_size, x_size)
+
+
+def _get_spline3d_arrays(gsl_spline, length):
+    """Get array data from a 3D GSL spline.
```
tilmantroester, line 576: An array of 2D GSL splines, instead of a 3D GSL spline?

pyccl/tests/test_tk3d.py
```diff
@@ -137,3 +137,38 @@ def test_tk3d_eval(is_product):
...
+        assert np.allclose(np.log(out[0]), tkka_arr, rtol=1e-15)
+
+
+def test_tk3d_spline_arrays_raises():
+    (a_arr, lk_arr, fka1_arr, fka2_arr, tkka_arr) = get_arrays()
+    tsp = ccl.Tk3D(a_arr, lk_arr, tkk_arr=tkka_arr)
+
+    if hasattr(tsp.__class__, "has_tsp"):
+        # `has_tsp` is a property
+        delattr(tsp, "tsp")
```
tilmantroester, line 168: Does this cause memory leaks?

tilmantroester (changes requested):

pyccl/ccl_utils.i
nikfilippas: Yes, just to copy the data. I can't access them as raw arrays:
```
In [10]: tsp.tsp.a_arr
Out[10]: <Swig Object of type 'double *' at 0x7f63880ae600>
```

pyccl/pyutils.py
nikfilippas: Done.

pyccl/ccl_utils.i
nikfilippas: Done. I thought I wouldn't loop through every scale factor to save a bit of runtime as this is a very specialized function, but better to be on the safe side.

nikfilippas:
@tilmantroester good to go for a second pass I think.

pyccl/ccl_utils.i
tilmantroester: Ah, maybe I'm confusing it with `ctypes`.

pyccl/tests/test_tk3d.py
tilmantroester: Maybe add a comment, just in case this gets flagged by some checks couple years down the road.

nikfilippas: Okay, I tweaked that bit just in case PR923 doesn't make the cut, and I also added a line to free the C-memory.

tilmantroester (approved):
LGTM!

PR #926, opened 2022-04-13
Make tracer spline errors more verbose

tilmantroester:
This catches the case where the tracer redshift support is wider than the interal background spline. This is mostly useful to prevent opaque error in calculator mode. A similar check should probably be made for the power spectrum as well.

### DIFF
--- a/pyccl/tracers.py
+++ b/pyccl/tracers.py
@@ -25,7 +25,7 @@ def _Sig_MG(cosmo, a, k=None):
     return _vectorize_fn6(lib.Sig_MG, lib.Sig_MG_vec, cosmo, a, k)
 
 
-def _check_background_spline_compatability(cosmo, z):
+def _check_background_spline_compatibility(cosmo, z):
     """Check that a redshift array lies within the support of the
     CCL background splines.
     """
@@ -34,7 +34,7 @@ def _check_background_spline_compatability(cosmo, z):
 
     if a.min() < a_bg.min() or a.max() > a_bg.max():
         raise ValueError(f"Tracer defined over wider redshift range than "
-                         f"interal CCL splines. Tracer: "
+                         f"internal CCL splines. Tracer: "
                          f"z=[{1/a.max()-1}, {1/a.min()-1}]. Background "
                          f"splines: z=[{1/a_bg.max()-1}, {1/a_bg.min()-1}].")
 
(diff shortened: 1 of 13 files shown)


pyccl/tracers.py
```diff
@@ -24,6 +25,20 @@ def _Sig_MG(cosmo, a, k=None):
     return _vectorize_fn6(lib.Sig_MG, lib.Sig_MG_vec, cosmo, a, k)
 
 
+def _check_background_spline_compatability(cosmo, z):
```
damonge, line 28: I think it should be "compatibility" ;-)

pyccl/tracers.py
```diff
@@ -24,6 +25,20 @@ def _Sig_MG(cosmo, a, k=None):
...
+    """Check that a redshift array lies within the support of the
+    CCL background splines.
+    """
+    a_bg, _ = _get_spline1d_arrays(cosmo.cosmo.data.chi)
+    a = 1/(1+z)
+
+    if a.min() < a_bg.min() or a.max() > a_bg.max():
+        raise ValueError(f"Tracer defined over wider redshift range than "
+                         f"interal CCL splines. Tracer: "
```
damonge, line 37: interal -> internal

damonge (changes requested):
This looks good to me. Have you checked if this adds a lot of time to the usual life cycle of a tracer? I assume not, but I forget if `_get_spline1d_array` was a costly function.

damonge:
Also, could we add a unit test that makes this fail?

pyccl/tracers.py
tilmantroester: Done.

pyccl/tracers.py
tilmantroester: Done.

tilmantroester:
Added a unit test. `_get_spline1d_array` just copies the memory of the C-level spline data, so should be negligible in terms of computational effort. On the other hand it shouldn't be necessary. Maybe I'll add a `a_min` and `a_max` to the `cosmo.data` C-struct for bookkeeping.

damonge:
Thanks a lot!
I'm happy to approve this without the C-level logging of amin/max as soon as the final tests pass.

damonge (approved):
LGTM

Issue #927, opened 2022-04-14
Integration errors with certain n(z)


tilmantroester:
CCL fails to create (weak lensing) tracers for certain n(z). A minimal example is given below. Enabling debug mode shows that the issue appears when trying to normalise the n(z). Curiously, there should be an error message being set (https://github.com/LSSTDESC/CCL/blob/master/src/ccl_tracers.c#L85) but for some reason that gets lost somewhere.

```python
import numpy as np
import pyccl as ccl

z, nz = np.loadtxt("bad_nz.txt", unpack=True)

ccl.debug_mode(True)
ccl_cosmo = ccl.CosmologyVanillaLCDM()
ccl_tracer = ccl.WeakLensingTracer(ccl_cosmo, (z, nz))
```
[bad_nz.txt](https://github.com/LSSTDESC/CCL/files/8489362/bad_nz.txt)

nikfilippas:
When I tried to reproduce this some days ago, adding `cosmo` as an extra argument in the checks of `tracers.py` brings the erro message back. Maybe it's worth adding those to your PR as well.

tilmantroester:
Further investigation showed that the fixed order quadrature used in the tracer creation in CCL doesn't cope with this n(z). Doing the normalisation integral in python independently of CCL shows that fixed order quadrature fails miserably and adaptive quadrature fails to converge as well with default settings: 

```python
import scipy.integrate
import scipy.interpolate

z, n = np.loadtxt("bad_nz.txt", unpack=True)
nofz = scipy.interpolate.InterpolatedUnivariateSpline(z, n)

norm = {}
norm["trapz"] = np.trapz(n, z)
norm["spline_integral"] = nofz.integral(z[0], z[-1])
norm["adaptive_quad"] = scipy.integrate.quad(nofz, z[0], z[-1])[0]
for i in range(1, 8):
    norm[f"fixed_quad_order_{i}"] = scipy.integrate.fixed_quad(nofz, z[0], z[-1], n=i)[0]

for name, value in norm.items():
    print(f"{name:<20}: {value}")
```
```
trapz               : 1.0
spline_integral     : 1.000000000000001
adaptive_quad       : 0.9999361333214825
fixed_quad_order_1  : 0.08137690272918875
fixed_quad_order_2  : 3.7549939231261926
fixed_quad_order_3  : 0.03843028804193624
fixed_quad_order_4  : 0.39222769522755374
fixed_quad_order_5  : 1.371894742387018
[...]

Issue #919, opened 2022-02-08
delta-like redshift distribution error


ShimingGu:
Hi CCL Team,
I am working on a project that may need a very narrow redshift range of foreground galaxy lenses. If I run the following codes:
```
# Generate a delta-like n(z)
n_bin_z_lens = 101
z_lens = np.linspace(0.0,3.0,n_bin_z_lens)
n_z_lens = np.zeros_like(z_lens)
n_z_lens[50] = 1
# Compute the galaxy-galaxy lensing C_\ell
tracer_lens = ccl.NumberCountsTracer(cosmo, has_rsd=False, dndz=(z_lens,nz_lens), bias=(z_lens,bias_z_lens), mag_bias=None)
tracer_source = ccl.WeakLensingTracer(cosmo, dndz=(z_source,n_z_source), has_shear=True, ia_bias=None)
cl_ggl = ccl.angular_cl(cosmo, tracer_lens, tracer_source, l_arr)
```
it brings the error
`ccl_cls.c: ccl_angular_cls_limber(); integration error`
This error disappears when I also put
`n_z_lens[51] = 1` 
if given a relatively smaller number of redshift bins like 
`n_bin_z_lens = 101`
in the above example. However, that problem still persists when the number of redshift bins goes to a larger number -- when my redshift distribution approach to a Dirac-delta function even if it has more than one non-zero value. It looks like an intrinsic characteristic in pyccl and I don't know how to deal with it if I want better approximations to a very narrow redshift distribution. 

I am very grateful if anyone would be interested to have a look at this problem and tell me how to deal with it.

Have a great day and Thank you so much!

tilmantroester:
Internally, CCL uses spline interpolation for the window functions and then adaptive quadrature for the integral. Neiter is particuarly suited to deal with delta functions. Since the window function for the lenses is just the lens n(z), you could write skip the Limber integral all together and write it as prefector * W(z_lens) * P(k(l, z_lens), z_lens).

damonge:
#928 seems to have solved this (I don't get an error anyore)

tilmantroester:
I don't remember if there's a tests for such a pathological n(z). So I would check if the results make sense first.

damonge:
Looks good to me
![clfine](https://user-images.githubusercontent.com/8809493/183121159-6af8b659-963b-4a51-ba01-10b7e7dffb09.png)

ShimingGu:
[...]

Issue #672, opened 2019-08-19
ENH clean out the C layer warnings


beckermr:
Right now we have a lot of C layer warnings which are 

 - not descriptive
 - not correct
 - override other warnings so that the error is not clear at the python layer

We need to clean this up and make sure everything is correct. 

One thing that might be good to do is to use various C macros to denote the file and line number. This will ease the burden of listing the function name in the string. Also, C99 has a standard function name macro that we can use as well.

c-d-leonard:
@beckermr is this resolved by the above merged PR?

beckermr:
Nope. This is still a big issue in the library. The above PR is where we made this issue.

c-d-leonard:
Ah okay thanks, sorry.

PR #928, opened 2022-04-14
Make tracer creation more robust

tilmantroester:
This is an attempt to make CCL more robust to bad n(z). See e.g. #927, #919.

- Added spline integration for the n(z) normalisation integral and the lensing kernel integral
- Added flags in `gsl_params` to switch between spline and GSL quadrature integration for the two cases. Since spline integration is more robust and seems to be faster in the case where the n(z) is not super smooth or not very finely sampled, this is chosen as the default.
- When using GSL quadrature for the lensing integral, the integration error is compared to the peak of the lensing kernel, instead of just the relative accuracy of the integral. This solves issues with numerical convergence of the integrals in the tails of the kernel where the absolute values are small but the integration method struggles.
- Added option to choose lower number of samples for the lensing kernel than samples in the n(z). This is done using the `n_samples` argument, like in `CMBLensingTracer`. Right now there is an inconsitency in the API, in that the tSZ, CIB, and ISW tracers use `n_chi` for the same thing. This should be unified but as this involves a change to existing API, I've not included it here. @damonge Let me know what you think.
- Added the option to call `Tracer.get_kernels` without specifying `chi`. In this case, the internal spline representation is returned. This is mostly useful for debugging purposes.
- Enabled warnings by default (can be turned off with `pyccl.debug_mode(False)`. Many of the warnings are there for good reason and warning the user that internal computations might produce garbage results is probably preferable to silent failures.
- Added a warning when the status message gets overwritten. Right now there's no error stack or traceback. If multiple errors get written to `cosmo.status_message`, earlier messages get lost, which makes debugging difficult. Partially addresses #672. But this should really be handled cleaner (let's just switch to C++ and get exceptions and smart pointers!)


pyccl/tracers.py
```diff
@@ -188,33 +190,46 @@ def get_kernel(self, chi):
...
-                The shape will be `(n_tracer, chi.size)`, where \
-                `n_tracer` is the number of tracers. The last \
-                dimension will be squeezed if the input is a \
+            array_like: list of radial kernels for each tracer.
+                The shape will be `(n_tracer, chi.size)`, where
+                `n_tracer` is the number of tracers. The last
+                dimension will be squeezed if the input is a
                 scalar.
+                If no chi was provided, returns two arrays, the radial kernels
+                and the comoving radial distances.
```
damonge, line 204: "... corresponding to the internal values used for interpolation." or something like that?

pyccl/tracers.py
```diff
@@ -616,9 +637,13 @@ class WeakLensingTracer(Tracer):
             normalization. Set to False to use the raw input amplitude,
             which will usually be 1 for use with PT IA modeling.
             Defaults to True.
+        n_samples (int, optional): number of samples over which the kernel
+            is desired. These will be equi-spaced in radial distance.
```
damonge, line 641: Specify this is for the lensing kernel (i.e. this is not used for the IA contribution)

src/ccl_tracers.c
```diff
@@ -58,40 +61,62 @@ static double get_nz_norm(ccl_cosmology *cosmo, ccl_f1d_t *nz_f,
...
+        w, &nz_norm, &nz_enorm);
+
+      if (gslstatus != GSL_SUCCESS) {
+        ccl_raise_gsl_warning(gslstatus, "ccl_tracers.c: get_nz_norm():");
+        *status = CCL_ERROR_INTEG;
+        ccl_cosmology_set_status_message(
+          cosmo,
+          "ccl_tracers.c: get_nz_norm(): "
+          "integration error when normalizing N(z)\n");
+      }
```
damonge, line 113: Why are we raising both a warning and an error here?

src/ccl_tracers.c
```diff
@@ -167,38 +204,239 @@ static double lensing_kernel_integrand(double z, void *pars) {
...
+        #pragma omp for
+        for(int ichi=0; ichi<nchi; ichi++) {
+          chi_end = chi_arr[ichi];
+          a = ccl_scale_factor_of_chi(cosmo, chi_end, &local_status);
+          z_end = 1./a-1;
+
+          // We don't need to start at 0 but finding the index corresponding to chi_end would make things more complicated
+          for(int i=0; i<nz; i++) {
+            if(chi_of_z_array[i] < chi_end) {
+              integrand_array[i] = 0.0;
```
damonge, line 403: OK, if I understand this correctly, this will work well assuming that the N(z) is more finely sampled than the lensing kernel, correct? Is there a real-world situation where this may not be the case?

pyccl/tests/test_tracers.py
```diff
@@ -151,3 +155,90 @@ def test_tracer_nz_support():
...
+
+    cosmo.cosmo.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION = True
+    tr_wl = ccl.WeakLensingTracer(cosmo, dndz=(z, n))
+    w_wl_spline, _ = tr_wl.get_kernel(chi=None)
+
+    cosmo.cosmo.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION = False
+    tr_wl = ccl.WeakLensingTracer(cosmo, dndz=(z, n))
+    w_wl_gsl, _ = tr_wl.get_kernel(chi=None)
+
+    assert np.allclose(w_wl_spline[0], w_wl_gsl[0])
```
damonge, line 205: This has an absolute tolerance of 1E-8. Can you remind me what the order of magnitude is for the lensing kernel?

damonge (commented):
Thanks a lot @tilmantroester . A few comments below

damonge:
On cleaning up warnings: I agree... we should try to sort that out before V3
On nchi vs. nsamples: I also agree, but let's not break the API yet. Perhaps this could be added to #889

nikfilippas:
Unrelated to the title of this PR but definitely related to C-level warnings: currently, GSL raises an error when a value outside of the interpolation range is requested. For some reason, this is caught and converted to a warning in CCL, and the code ends up failing with an unexplained error `CCLError: Error 1: `. Would we consider promoting GSL out-of-bounds warnings to errors (perhaps in this PR)?

Try, for example, `ccl.CosmologyVanillaLCDM().sigmaM(M=1e5, a=1)`.

As the vast majority (citation needed?) of GSL failures in CCL are caused by the queried values being out of bounds, maybe it's worth either catching this case specifically and promoting to an error (`GSL_EDOM=-1` [here](https://github.com/LuaDist/gsl/blob/master/err/gsl_errno.h) or just promoting every GSL failure to an error.

pyccl/tracers.py
tilmantroester: Done.

src/ccl_tracers.c
tilmantroester: No reason. Just some left-over I had forgotten to remove earlier.

src/ccl_tracers.c
tilmantroester: That's a good point. If the n(z) is sampled very sparsely, then this could become inaccurate. In that case the quadrature methods would work better. One option would be to choose quadrature when `n_chi` > `nz.size`, while allowing the user to force either spline integration or quadrature.

pyccl/tests/test_tracers.py
tilmantroester: About 1e-5. I've added absolut tolerances now.

src/ccl_tracers.c
damonge: How about just complementing the spline integral with a single trapezoidal integral for the piece between `chi_end` and `chi_of_z_array[i_first]`, where `chi_of_z_array[i_first]` is the first value of  `chi_of_z_array` satisfying `chi_of_z_array[i_first] >= chi_end`?

damonge: It'd require assuming something about what `Nz(chi_end)` is, but this could be calculated from the spline interpolator.

damonge:
OK, a more complete solution to error handling is for another PR. @tilmantroester see my suggestion above for the poorly-sampled case. Not sure if it's clear. I can provide a diagram if not.

src/ccl_tracers.c
tilmantroester: The integrand at chi_end is 0, so that was simple to account for. There's a subtlety when z[0] >> 0 and nz[0] > 0 but that was luckily caught by a test (and I added another one for that specific edge case).

tilmantroester:
@damonge The poorly-sampled case should be addressed now.
@nikfilippas I added a warning if the n(z) is poorly sampled that points used to changing the integration method, which is set by a flag in `gsl_params`. Once your PR is in that warning needs to be adjusted accordingly.

damonge (approved):
LGTM!

nikfilippas:
@tilmantroester I noticed while running some tests that this PR introduces a bunch of unnecessary warnings in the python unit tests and in the benchmarks. It would be nice to get rid of them in a separate PR to keep the output as clean as possible.

Issue #929, opened 2022-04-18
ConcentrationIshiyama21 is very slow


nikfilippas:
The way we implement the Ishiyama et al. 2021 concentrations, we use the root-finding function `root_scalar` which slows down the computation a lot. At the same time, this functional form of the c-M relation, which makes use of an inverse function is physically motivated and shown to be more accurate than other (perhaps simpler) prescriptions.

I have been running a likelihood that uses `ConcentrationIshiyama21`. Below are some benchmarks.
Using `ConcentrationIshiyama21`:
![Screenshot from 2022-04-18 19-10-31](https://user-images.githubusercontent.com/22447385/163837603-7d0d6c9a-c3f7-4789-8fa7-90c718920fb8.png) --> 8.2 s/iteration
Compare to `ConcentrationDuffy08`:
![Screenshot from 2022-04-18 19-11-06](https://user-images.githubusercontent.com/22447385/163837701-593f11d8-fd3b-4c82-b441-a2184da74f76.png) --> 1.47 s/iteration.

This is obviously a huge bottleneck, as, for a chain of ~100 000 steps, using `ConcentrationIshiyama21` adds ~7 days of CPU time!

PR #930, opened 2022-04-18
c(M) Ishiyama21 speedup

nikfilippas:
Addresses #929. Replacing `root_scalar` with `brentq` to run the code more than an order of magnitude faster.

**Update:** In a textbook case of K.I.S.S. @damonge suggested we use `brentq` which is much more efficient than `root_scalar`. This has similar performance to the emulator (for completeness I include its timing benchmark below).

**Old implementation:**

I built an RBF emulator to speed up the c(M) computation by 7 times.

The bottleneck of the computation is the `_G_inv` method so I performed the following steps:
- Ran this method locally with a bunch of different cosmologies (some of them extreme), different masses, and different scale factors to see what is the range of the arguments of `_G_inv`;
- I then widened the `[vmin, vmax]` boundaries of those ranges by 20% on each side to be on the very-safe side;
- Picked 1024 training points in the 2-dimensional parameter space of that method, using Latin Hypercube sampling;
- Verified that the training data are well-behaved (span one order of magnitude, no need for rescaling);
- Fitted the data using a Radial Basis Function interpolant with a modified kernel that only looks at the 20 nearest neighbors in parameter space.

Data live in `CCL/pyccl/data`, and the emulator is automatically loaded the first time `ConcentrationIshiyama21` is instantiated.
I verified that the emulator reproduces very well the `_G_inv` behavior, and added an additional benchmark test.

Speed comparison for the same likelihood I mention in the quoted issue:
before:
Screenshot from 2022-04-18 19-10-31 --> 8.2 s/iteration
after:
Screenshot from 2022-04-18 20-00-46 --> 2.3 s/iteration
(= 3.5x speed increase)

c(M)-specific timing benchmark (without and with emu):
Screenshot from 2022-04-18 20-04-05
(=7x speed increase).
Using `brentq`:
Screenshot from 2022-04-19 13-10-55
(=9x speed increase).


nikfilippas:
Code to reproduce the training (uploading this just so it exists in the repo):
```python
""" Trains the Ishiyama21 `_G_inv` emulator. """
import pyccl as ccl
from pyccl import halos as hal
import numpy as np
from scipy.interpolate import RBFInterpolator
from scipy.stats import qmc
from tqdm import tqdm


# First we sample `_G_inv` to see what the range of `arg` and `n_eff` is.
# LatinHypercube here is an overkill; this might as well have been random sampling.
sampler = qmc.LatinHypercube(d=7)
sample = sampler.random(n=128)

# parameters: Oc   Ob    h  n_s sigma8 M   a
lo_bounds = [0.2, 0.02, 0.5, 0.9, 0.6, 6, 0.01]
hi_bounds = [0.5, 0.08, 0.8, 1.1, 1.1, 16, 1.00]
sample = qmc.scale(sample, lo_bounds, hi_bounds)

# These are all the options the Concentration class accepts.
options = {
    "mass_def": [hal.MassDefVir(), hal.MassDef200c(), hal.MassDef500c()],
    "relaxed": [True, False],
    "Vmax": [True, False]}

# For each combination of options, run 128 times in the 7-dimensional param
# space of Omega_c, Omega_b, h, n_s, sigma8, M, a.
for mass_def in options["mass_def"]:
    for relaxed in options["relaxed"]:
        for Vmax in options["Vmax"]:
            try:
                cm = hal.ConcentrationIshiyama21(
                    mass_def=mass_def, relaxed=relaxed, Vmax=Vmax, emu=False)
            except ValueError:
                # some option combinations don't exist (e.g. 500c, vmax=True)
                continue
            for node in tqdm(sample, desc=f"{mass_def}, {relaxed}, {Vmax}"):
                cosmo = ccl.Cosmology(
                    **dict(zip(["Omega_c", "Omega_b", "h", "n_s", "sigma8"],
                               node[:5])))
                try:
                    _ = cm.get_concentration(cosmo, M=10**node[-2], a=node[-1])
                except ccl.CCLError:
                    # C-level error associated with sigmaM splines out of bounds
                    continue

# I hacked `ConcentrationIshiyama21._G_inv` to save [arg, n_eff, root]
# at every function call. This was saved in `Concentration.saved_data`.
data = np.asarray(hal.Concentration.saved_data)
mins, maxes = np.min(data, axis=0)[:-1], np.max(data, axis=0)[:-1]

# Widen the ranges of the sampling parameters by 40%.
lo_bounds = (1 - 0.2 * np.sign(mins)) * mins
hi_bounds = (1 + 0.2 * np.sign(maxes)) * maxes

# Build a Latin Hypercube for `arg` and `n_eff`.
sampler = qmc.LatinHypercube(d=2)
sample = sampler.random(n=1024)
sample = qmc.scale(sample, lo_bounds, hi_bounds)
func = cm._G_inv
d = func(*sample.T)

# Disregard the nans (when `root_scalar` reaches max tries and exits).
sample = sample[~np.isnan(d)]
d = d[~np.isnan(d)]

# RBF Interpolation with thin-plate kernel truncated to 20 nearest neighbors.
rbf = RBFInterpolator(sample, d, neighbors=20)

# Save, reload.
import pickle
with open("Ishiyama21_Ginv", "wb") as f:
    pickle.dump(rbf, f)
with open("Ishiyama21_Ginv", "rb") as f:
    rbfl = pickle.load(f)
```

pyccl/halos/concentration.py
```diff
@@ -378,7 +377,7 @@ class ConcentrationIshiyama21(Concentration):
     def __init__(self, mdef=None, relaxed=False, Vmax=False):
         self.relaxed = relaxed
         self.Vmax = Vmax
-        super(ConcentrationIshiyama21, self).__init__(mass_def=mdef)
+        super().__init__(mass_def=mdef)
```
damonge, line 380: why this change?

damonge (commented):
Mostly LGTM. Just one quick question and I'll approve.

pyccl/halos/concentration.py
nikfilippas: `super` is smart in python 3, this was only required in python 2, so just cleaning up the code.

damonge (approved):
LGTM!

PR #932, opened 2022-04-21
TST add test for high-z halofit

### LINKED ISSUE #907: Weird behavior of halofit non linear scale computation at high z
Just a note on a fun behavior I found out when working on jax-cosmo's halofit implementation with @jecampagne .

Obviously halofit is not meant to be used for z >= 50, but still it looks like the CCL logic handling the computation of non-linear scales is inconsistent, so just wanted to point that out.

Here is a ratio of Pk_nl/Pk_lin for different scale factors, illustrating what the issue is:
```python
for i,a in enumerate(np.logspace(-2,np.log10(0.4), 12)):
    pk_ccl = ccl.nonlin_matter_power(cosmo_ccl, k, a)
    pk_lin = ccl.linear_matter_power(cosmo_ccl, k, a)
    loglog(k,pk_ccl/pk_lin, label='z = %2.f'%a2z(a),
           color=plt.cm.rainbow(i*25), alpha=0.6)
xlim(1, 100) ; legend() ; title('pk_halofit/pk_lin')
```
image

So, it looks like the non linear scales are clipped when you start approaching z=10, and then something weird happens at z=71 where the non linear power increases, and finally at z=99 halofit is completely disabled and you recover the linear pk.

So my proposal would be this: just completely disable halofit at after a given scale factor more or less consistent with the maximum k_nl and/or raise warning if halofit is used beyond this limit.

beckermr:
This PR adds a fix and test for high-z halofit.

closes #907

### DIFF
--- a/pyccl/tests/test_halofit_highz.py
+++ b/pyccl/tests/test_halofit_highz.py
@@ -1,28 +1,25 @@
 import numpy as np
 import pyccl as ccl
-import pytest
 
 COSMO = ccl.Cosmology(
     Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
     transfer_function='bbks', matter_power_spectrum='halofit')
 
 
-@pytest.mark.parametrize("zl,zh", [
-    (0, 1),
-    (25, 75),
-])
-def test_halofit_highz(zl, zh):
-    al = 1.0/(1 + zl)
-    ah = 1.0/(1 + zh)
+def test_halofit_highz():
+    vals = [(25, 75)] + list(zip(range(0, 90), range(1, 99)))
+    for zl, zh in vals:
+        al = 1.0/(1 + zl)
+        ah = 1.0/(1 + zh)
 
-    k = np.logspace(0, 2, 10)
-    pkratl = (
-        ccl.nonlin_matter_power(COSMO, k, al)
-        / ccl.linear_matter_power(COSMO, k, al)
-    )
-    pkrath = (
-        ccl.nonlin_matter_power(COSMO, k, ah)
-        / ccl.linear_matter_power(COSMO, k, ah)
-    )
+        k = np.logspace(0, 2, 10)
+        pkratl = (
+            ccl.nonlin_matter_power(COSMO, k, al)
+            / ccl.linear_matter_power(COSMO, k, al)
+        )
+        pkrath = (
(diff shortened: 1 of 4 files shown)


pyccl/tests/test_halofit_highz.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+import pyccl as ccl
+
+COSMO = ccl.Cosmology(
+    Omega_c=0.27, Omega_b=0.045, h=0.67, sigma8=0.8, n_s=0.96,
+    transfer_function='bbks', matter_power_spectrum='halofit')
+
+
+def test_halofit_highz():
+    vals = [(25, 75)] + list(zip(range(0, 90), range(1, 99)))
```
beckermr, line 10:
```suggestion
    vals = [(25, 75)] + list(zip(range(0, 98), range(1, 99)))
```

beckermr:
@damonge @c-d-leonard IDK who to tag to review this one. I've fixed the high-z halofit behavior.

damonge (commented):
Ok, so you've enlarged the range of redshifts over which we can find the nonlinear scale without resorting to sketchy interpolations for normal cosmologies. I have two comments/questions:
- since the bracketing interval is now larger, has this increased the halofit computation time too much?
- I'm worried about removing the sketchy interpolation altogether. When using poor data, our mcmcs may wander into low sigma8 regions where even with these changes it's not possible to find the nonlinear scale at z~10. This will currently throw an error and that cosmology will be discarded, which I think is wrong (most likely halofit at z=10 isn't even used in the calculation, so nothing is really wrong). Thoughts?

beckermr:
The old sketchy thing was so wrong that I think an error is more useful. 

I did not do detailed timing but the tests are not taking much longer so I am not too worried.

beckermr:
BTW, I switched the root-finding to work in log space since the dynamic range is so big. I am guessing this is helping with the larger bracketing interval.

damonge:
Can we check for what kinds of sigma8 this fails? If it's ~0.1 I won't worry!

beckermr:
Def fails! I have an idea to fix. Will push later.

beckermr:
ok @damonge - it works for sigma8 = 0.1 now. Do we think this will be robust enough?

damonge (approved):
OK, great. I just checked that this only adds at most ~10% to the usual halofit computation time, so LGTM!

beckermr:
Thank you!

Issue #114, opened 2017-02-17
Save CLASS transfer functions when calculating matter power


dkirkby:
This could either be triggered by an option to `ccl_cosmology_compute_power_class()` or we could always calculate and save the transfer functions (since they are presumably always being calculated anyway) to simplify the API.  Any opinions?

c-d-leonard:
I believe we still do not save the transfer functions. Is this something we still want? I can see it being useful, especially for modified gravity applications. @damonge @dkirkby?

damonge:
Yes, this is eventually gonna bite us in the ass. We have a relatively nice formalism to store these things (the generalized Pks), but we're not using it for it.

dkirkby:
Yes, I think we still want this.  Is it easier to implement now?

damonge:
I think so, and it could be done at the python level (the same way it's done for the power spectrum now)

PR #933, opened 2022-04-28
Caching framework.

nikfilippas:
**Note**: We could integrate #114 in here as well.

This PR implements a framework to cache calls to expensive routines. It is part of a larger refactoring proposal of `pyccl`, but this exists as a separate PR in order to make review easier.

**Hashing**
With this, we introduce a new module, `base.py`. In there, class `Hashing` contains a set of routines that can hash anything (even objects which are otherwise unhashable because they are mutable) consistently. The main idea behind being able to hash mutable types is that these are sometimes passed as arguments to functions in CCL, so we'd like to hash them in order to enable caching.

**Caching**
This class contains a set of routines that enable caching calls to expensive functions. Examples: Boltzmann solvers, non-linear power spectra, loading emulators. This functionality draws upon and extends the functionality of `functools.lru_cache` in that, (1) it can cache anything, (2) argument order is not important, (3) specifying a default value vs. not specifying it at all makes no difference. This extra functionality comes with a small speed disadvantage, but it is still orders of magnitude faster than the cached functions.

Instead of having a global caching dictionary, caches live as attributes of their parent function, like so:
`ccl.Cosmology._compute_linear_power.cache_info._caches`. `cache_info` is a `ccl.base.CacheInfo` object, which contains the cache dictionary. The dictionary contains ordered key-value pairs of the hash and the cached item. The cached item is of type `ccl.base.CachedObject`. The actual cache lives in `CachedObject.item`.

----


damonge:
@nikfilippas since I think there's consensus on hashing, but caching might be more debatable, do you want to split this into 2 PRs? The hashing one is probably straightforward (or more so), so we could get things moving faster.

I haven't looked at the code yet, so not sure if the split can be done easily.

pyccl/base.py
```diff
@@ -0,0 +1,301 @@
...
+    def _pop(cls, dic, policy):
+        """Remove one cached item as per the implemented caching policy."""
+        if policy == "lfu":
+            keys = list(dic)
+            idx = np.argmin([item.counter for item in dic.values()])
+            dic.move_to_end(keys[idx], last=False)
+        dic.popitem(last=False)
+
+    @classmethod
+    def _decorator(cls, func, maxsize, policy):
```
damonge, line 142: Please, explain in a lot of detail what this is doing.

pyccl/tests/test__caching.py
```diff
@@ -0,0 +1,146 @@
+# We use double underscore to make it the first test alphabetically.
```
damonge, line 1: I think there's one too many `_` in this file's name

nikfilippas: Just checked, it has 2 underscores. This is intentional, as the comment reads. It is for this test to be the first one alphabetically, so it is run first.

damonge (changes requested):

pyccl/base.py
```diff
@@ -43,3 +46,254 @@ def hash_(obj):
     """Generic hash method, which changes between processes."""
     digest = hash(repr(_to_hashable(obj))) + sys.maxsize + 1
     return digest
+
+
+class _ClassPropertyMeta(type):
+    """Implement `property` to a `classmethod`."""
+    # TODO: in py39+ decorators `classmethod` and `property` can be combined
```
damonge, line 53: can you make it more explicit that we want to drop this by the time we make 3.9 the minimum requirement?

pyccl/tests/test__caching.py
damonge: Kind of embarrassing I didn't read the comment

pyccl/base.py
```diff
@@ -43,3 +46,254 @@ def hash_(obj):
...
+    Caching is used for pre-computed objects that are expensive to compute.
+
+    Attributes:
+        maxsize (``int``):
+            Maximum number of caches to store. If the dictionary is full, new
+            caches are assigned according to the set cache retention policy.
+        policy (``'fifo'``, ``'lru'``, ``'lfu'``):
+            Cache retention policy.
+    """
+    _enabled: bool = True
```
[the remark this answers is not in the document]
nikfilippas: Done.

pyccl/base.py
nikfilippas: Looks like we're going to have to stick with it, the Python devs removed the ability to wrap `property` inside `classmethod` in Python 3.11.
https://docs.python.org/3.11/library/functions.html#classmethod
https://github.com/python/cpython/issues/89519

nikfilippas: Update: Yes, so it looks like after Py3.11, metaclasses are the only reliable way to combine `classmethod` and `property`.

damonge: hahaha

damonge (approved):
LGTM

PR #934, opened 2022-04-28
Refactor pyccl under an abstract CCLObject.

nikfilippas:
This is part 2 of the refactoring proposal. It implements a `CCLObject` which all other base classes inherit from. The aim of this is to be able to centrally control the behavior of every type defined in pyccl, as well as to be able to adopt a uniform framework for object comparison.

The main idea is that the developer only needs to define a `__repr__` for the class, and `__hash__` and `__eq__` will be automatically added. This way, we can check equivalence of any object to any other object, and it is up to the developer to decide the equivalence criteria when writing up the `__repr__` method.

Because some objects contain a large amount of data (such as `Pk2D` - contains huge arrays), hashing is a relatively expensive operation. Therefore, `hash` and `repr` for the object are cached the first time they are requested.

But, what if we want to update some parameters? Won't the cached `repr` be wrong then?

All `CCLObject`s are immutable by default, unless an `update_parameters` method is defined. Only through there may the instance mutate, and when it does, the caches are deleted and recomputed on request. Alternatively, there are two ways to mutate an object: either via the `UnlockInstance` context manager, or via the `@unlock_instance` decorator.

Docstrings in `base.py` explain more thoroughly the details of this implementation.

**Secondary changes**
In an attempt to improve the OOP code structure, I made two more major changes:
1. All template classes are now abstract. So, `HaloProfile`, `Concentration`, `HaloBias`, and `MassFunc` (and `Emulator` in PR906) cannot be instantiated because they contain abstract methods.
2. Converted all "special implementation" subclasses to simple functions. These are subclasses that only overload the `__init__` method of the parent class, as there is no point that these live as different types. The subclasses converted to functions are (i) `CosmologyVanillaLCDM`, (ii) all subclasses of the `MassDef` superclass, (iii) all subclasses of the `Tracer` superclass. These changes don't affect how users interact with CCL; this is all internal restructure of the codebase.

**How to review this**
I suggest starting from `CCLObject` in `base.py`, then having a look at `_repr.py`. Everything else relies on these two things.
There are more than 100 changed files, but most of them are in `tests/` and `benchmarks/`, where I've re-written the imports to take advantage of caching and run them faster. This version passes all the tests on my PC and I have been using it for a while, so if you are a developer feel free to install it and play around.

**Note**: This PR relies on 918, 923, 925, and 933. Hopefully once these are through the review will be much more straightforward as `git diff` from the default branch will only contain the relevant changes.

----


damonge (commented):
A preliminary review, just trying to understand if the changes to tests/benchmarks are related to the meat of this PR or just incidental.

pyccl/halos/__init__.py
```diff
@@ -95,7 +95,7 @@
     'MassFunc', 'MassFuncPress74', 'MassFuncSheth99', 'MassFuncJenkins01',
     'MassFuncTinker08', 'MassFuncTinker10', 'MassFuncWatson13',
     'MassFuncAngulo12', 'MassFuncDespali16', 'MassFuncBocquet16',
-    'MassFuncBocquet20', 'mass_function_from_name',
+    'mass_function_from_name',
```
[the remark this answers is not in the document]
nikfilippas: Accidental leftover from a forward branch.

benchmarks/__init__.py
```diff
@@ -0,0 +1,4 @@
+import pyccl
+pt = pyccl.nl_pt
+
+__all__ = ('pyccl', 'pt',)
```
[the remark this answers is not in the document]
nikfilippas: reverted

pyccl/tests/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+get_isitgr_pk_lin = pyccl.boltzmann.get_isitgr_pk_lin
+_spline_integrate = pyccl.pyutils._spline_integrate
+assert_warns = pyccl.pyutils.assert_warns
+UnlockInstance = pyccl.base.UnlockInstance
+
+__all__ = (
+    'pyccl', 'lib', 'CCLWarning', 'CCLError', 'CosmologyVanillaLCDM',
+    '_fftlog_transform', 'get_isitgr_pk_lin', '_spline_integrate',
+    'assert_warns', 'UnlockInstance',
+)
```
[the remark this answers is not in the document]
nikfilippas: we don't actually, reverted

benchmarks/test_bbks.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
-import pyccl as ccl
+from . import pyccl as ccl
```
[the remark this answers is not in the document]
nikfilippas: no we don't, reverted.

damonge:
One question about the abstract classes, and subclasses now being generated via functions. Say we now wanted to subclass a subclass in the previous implementation (e.g. making NFW a subclass of GNFW - not that this is a very good example), because they would share a lot of the code. Have we lost the ability to do that?

pyccl/tracers.py
```diff
@@ -311,9 +308,9 @@ def get_transfer(self, lk, a):
...
         Returns:
-            array_like: list of transfer functions for each tracer. \
-                The shape will be `(n_tracer, lk.size, a.size)`, where \
-                `n_tracer` is the number of tracers. The other \
+            array_like: list of transfer functions for each tracer.
+                The shape will be `(n_tracer, lk.size, a.size)`, where
+                `n_tracer` is the number of tracers. The other
                 dimensions will be squeezed if the inputs are scalars.
         """
         if not hasattr(self, '_trc'):
```
damonge, line 316: should we not remove this the same way you did for `get_kernel` and `get_f_ell`?

damonge: Same question for `get_bessel_derivative`

damonge (commented):
OK, these are from looking at tracers.py

nikfilippas:
Abstract classes is a different thing to turning subclasses into functions. In general, it is considered bad practice in OOP to subclass just to make special implementations of the parent class, i.e. `CosmologyVanillaLCDM` is just `Cosmology` with some pre-selected numbers. If you subclass, you add unnecessary things to the inheritance tree. These cases are better handled as simple functions which use the parent class, which acts as a factory.

On the other hand, abstract classes *have* to be subclassed to specific implementations (and extending their usage). This is the case with the NFW/GNFW profiles that you mentioned. That's why each profile implementation is a subclass on its own, but each mass definition is just a function which calls the factory class.

nikfilippas:
As a rule of thumb, if your subclass only has an `__init__` method, it's probably better written as a function.

pyccl/tests/test_profiles.py
```diff
@@ -44,17 +45,8 @@ def smoke_assert_prof_real(profile, method='_real'):
...
-def test_defaults():
-    p = ccl.halos.HaloProfile()
-    with pytest.raises(NotImplementedError):
-        p.real(None, None, None, None, None)
-    with pytest.raises(NotImplementedError):
-        p.fourier(None, None, None, None, None)
-
-
 @pytest.mark.parametrize('prof_class',
-                         [ccl.halos.HaloProfileNFW,
```
damonge, line 56: Why did we drop NFW from this test?

pyccl/tests/test_hmfunc.py
```diff
@@ -121,17 +121,6 @@ def test_nM_from_string_raises():
...
-    nM = ccl.halos.MassFunc(COSMO)
-    with pytest.raises(NotImplementedError):
-        nM._get_fsigma(COSMO, 1., 1., 8)
-
-    M_in = 1E12
-    lM_out = nM._get_consistent_mass(COSMO,
-                                     M_in, 1., nM.mdef)
-    assert np.fabs(np.log10(M_in) - lM_out) < 1E-10
-
-
```
damonge, line 134: Why did we drop these tests?

pyccl/tests/test_hbias.py
```diff
@@ -54,14 +54,3 @@ def test_bM_from_string(name):
...
-
-def test_bM_default():
-    bM = ccl.halos.HaloBias(COSMO)
-    with pytest.raises(NotImplementedError):
-        bM._get_bsigma(COSMO, 1., 1.)
-
-    M_in = 1E12
-    lM_out = bM._get_consistent_mass(COSMO,
-                                     M_in, 1., bM.mdef)
-    assert np.fabs(np.log10(M_in) - lM_out) < 1E-10
```
damonge, line 67: Why did we drop this test?

pyccl/tests/test_cosmology.py
```diff
@@ -189,42 +184,6 @@ def test_cosmology_p18lcdm_raises():
...
-    assert_(
-        ccl.comoving_radial_distance(cosmo, 0.5) ==
-        ccl.comoving_radial_distance(cosmo2, 0.5))
-
-    cosmo3 = eval(repr(cosmo))
-    assert_(
-        ccl.comoving_radial_distance(cosmo, 0.5) ==
-        ccl.comoving_radial_distance(cosmo3, 0.5))
-
-
```
damonge, line 227: Should we test the new repr?

pyccl/tests/test_concentration.py
```diff
@@ -20,12 +20,6 @@
 M400 = ccl.halos.MassDef(400, 'critical')
 
 
-def test_cM_default_mass_def():
-    cM = ccl.halos.Concentration()
-    assert cM.mdef.__eq__(MDEF)
-    assert not cM._check_mdef(MDEF)
-
-
```
damonge, line 28: Why did we drop this?

pyccl/tests/test__caching.py
```diff
@@ -12,11 +12,6 @@
 # runs; normally this is is expected to be another order of magnitude faster
 SPEEDUP = 50
 
-# enable caching if not already enabled
-DEFAULT_CACHING_STATUS = ccl.Caching._enabled
-if not ccl.Caching._enabled:
-    ccl.Caching.enable()
-
```
damonge, line 19: Why are we not doing this anymore?

pyccl/tests/__init__.py
```diff
@@ -0,0 +1,4 @@
+import pyccl as ccl
+
+# Enable caching for testing.
+ccl.Caching.enable()
```
damonge, line 4: why?

damonge (commented):
These ones are from looking at `pyccl/tests` (except the cclobject ones)

benchmarks/__init__.py
```diff
@@ -0,0 +1,4 @@
+import pyccl as ccl
+
+# Enable caching for testing.
+ccl.Caching.enable()
```
damonge, line 4: Why? I'd rather do the tests without caching to avoid potential bugs.

pyccl/tests/test_hmfunc.py
nikfilippas: Abstract classes can't be instantiated in Python. They will never be instantiated if they are abstract.

pyccl/tests/test_hbias.py
nikfilippas: abstract class

pyccl/tests/test__caching.py
nikfilippas: We want to use caching in testing - it's much, much faster. It is disabled by default, but I temporarily enabled it for testing in `__init__.py`.

pyccl/tests/test_concentration.py
nikfilippas: abstract classes

pyccl/tests/__init__.py
nikfilippas: I will demonstrate to you why caching is a good idea in testing, and we can then discuss it.

damonge (commented):
This from benchmarks. I'll leave the hardcore stuff for tomorrow.

pyccl/tests/test_profiles.py
nikfilippas: I'll revert that change.

pyccl/tests/test_cosmology.py
nikfilippas: It is implicitly checked in `test_cclobject.py` where we test if 2 cosmologies are equivalent,

pyccl/tests/__init__.py
damonge: I understand it may be faster, but I don't want our tests to give false positives if hashing/caching has a bug (which may not be the case now, but it may be in the future). For the tests, I want the code to do exactly what it says, without any internal caching.

E.g.: I have a test that checks that function A with parameters X=X1 and Y=Y1 gives me the same result as function A with parameter X=X1*0.5 and parameter Y=Y1*2. But now someone introduced a bug when setting up the caching of some internal complex object, and it actually identifies both calls as the same, so it just reuses the first result and the test passes with flying colors.

pyccl/tests/test__caching.py
damonge: Again, see my other comment. This really worries me.

pyccl/tests/__init__.py
nikfilippas: I would use the same line of reasoning to argue that if someone introduced a bug, then the test will fail because the assert statement will compare two things that are not equal. I think this is happens way more often in the tests, than the example you mentioned, but I guess it's indeed more clear if we disable it.

Caching is only used in saving the Boltzmann solver results, and will later be used to save emulators.
But okay, I guess if it has even the slightest chance of a false negative, better to disable it altogether.
I will disable it.

damonge: Not sure if there's a way to tell pytest that you'd like to use caching, so people could exploit it locally to make the tests go faster. If you figure out a way of doing that, go for it, but by default I'd rather it was off.

pyccl/tests/test__caching.py
nikfilippas: Addressed.

pyccl/tests/__init__.py
nikfilippas: I now disable it in tests but give users the option to use it via `pytest tests/ --use-cache`.

benchmarks/__init__.py
nikfilippas: Addressed above.

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        bound = sig.bind_partial(self, *args, **kwargs)
+        bound.apply_defaults()
+        bound.arguments.pop("self")
+        for name, value in bound.arguments.items():
+            setattr(self, name, value)
+        # func may now override the attributes we just set
+        func(self, *args, **kwargs)
+    return wrapper
```
damonge, line 326: Why do we always want to do this. Won't this bloat certain objects (e.g. Pk2Ds) by making them store the input arrays both at the python and c levels?

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+            cls.__init__ = auto_assign(cls.__init__, sig=cls.__signature__)
+
+        # 3. Replace repr (if implemented) with its cached version.
+        if "__repr__" in vars(cls):
+            # If the class defines a custom `__repr__`, this will be the new
+            # `_repr` (which is cached). Decorator `cached_property` requires
+            # that `__set_name__` is called on it.
+            bmethod = functools.cached_property(cls.__repr__)
+            cls._repr = bmethod
+            bmethod.__set_name__(cls, "_repr")
```
damonge, line 560: Why do we need both of the two lines above? Wouldn't `cls._repr` already assign a method called `_repr` to `cls`?

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+            # Fall back to using `__ccl_repr__` from `CCLObject`.
+            cls.__repr__ = cls.__ccl_repr__
```
damonge, line 562: I'm confused at the difference between `__repr__` and `_repr` for a given class, and so I'm not sure I understand what this line is trying to accomplish (you already started in line 558 by assuming that `cls` has a `__repr__`.

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+    """
+
+    def __init__(self, instance, mutate=True):
+        self.instance = instance
+        self.mutate = mutate
+        # Define these attributes for easy access.
+        self.id = id(self)
+        self.thread_lock = RLock()
+        # We want to catch and exit if the instance is not a CCLObject.
+        # Hopefully this will be caught downstream.
```
damonge, line 379: What?

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+        # do nothing; otherwise reset.
+        if self.id != self.object_lock._lock_id:
+            return
+
+        with self.thread_lock:
+            # Reset `repr` if the object has been mutated.
+            if self.mutate:
+                try:
+                    delattr(self.instance, "_repr")
+                    delattr(self.instance, "_hash")
```
damonge, line 412: This deletes the `_repr` and `_hash`, but shouldn't they be recreated? When does this happen?

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+
+
+def unlock_instance(func=None, *, argv=0, mutate=True):
+    """Decorator that temporarily unlocks an instance of CCLObject.
+
+    Arguments:
+        func (``function``):
+            Function which changes one of its ``CCLObject`` arguments.
+        argv (``int``):
+            Which argument should be unlocked. Defaults to the first argument.
```
damonge, line 428: Took me a while. Can you specify here that said argument is meant to be a CCLObject?

damonge: Also, can we only use this for one CCLObject at a time? E.g. I could imagine a function that modifies both a tracer and a profile.

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+        # 4. Unlock instance on specific methods.
+        def Funlock(cls, name, mutate):
+            # Allow instance to change or mutate if method `name` is called.
+            func = vars(cls).get(name)
+            if func is not None:
+                newfunc = unlock_instance(mutate=mutate)(func)
+                setattr(cls, name, newfunc)
+
+        Funlock(cls, "__init__", False)
+        Funlock(cls, "update_parameters", True)
```
damonge, line 573: So this unlocks them permanently or just instantaneously?

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+
+    def __new__(cls, *args, **kwargs):
+        # Populate every instance with an `ObjectLock` as attribute.
+        instance = super().__new__(cls)
+        object.__setattr__(instance, "_object_lock", ObjectLock())
```
damonge, line 578: This assigns the lock object, but when are objects locked for the first time? I didn't see that anywhere (I see them being locked after being unlocked, but not where they are first locked)

pyccl/base.py
```diff
@@ -298,3 +299,332 @@ def increment(self):
...
+    def __eq__(self, other):
+        # Two same-type objects are equal if their representations are equal.
+        if self.__class__ is not other.__class__:
+            return False
+        return repr(self) == repr(other)
+
+
+class CCLHalosObject(CCLObject):
+    """Base for halo objects. Automatically assign all ``__init__``
+    parameters as attributes.
```
damonge, line 621: Again, it worries me that we're doing this.

damonge (commented):
First batch of comments on `base.py`.

pyccl/tests/__init__.py
nikfilippas: Tested the `--use-cache` option and it works fine.
In a testing directory I cloned the following file twice:
```python
import pyccl as ccl

new_cosmo = lambda solver: ccl.CosmologyVanillaLCDM(transfer_function=solver)

def test_camb():
    solver = "boltzmann_camb"
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    assert True

def test_class():
    solver = "boltzmann_class"
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    new_cosmo(solver).compute_nonlin_power()
    assert True
```
and here are the timings for simple tests and tests with caching:
```
nick@canopus:~/Desktop$ pytest test/
============================= test session starts ==============================
platform linux -- Python 3.11.0, pytest-7.2.1, pluggy-1.0.0
rootdir: /home/nick/Desktop
collected 4 items                                                              

test/test_1.py ..                                                        [ 50%]
test/test_2.py ..                                                        [100%]

============================== 4 passed in 11.40s ==============================

nick@canopus:~/Desktop$ pytest test/ --use-cache 
============================= test session starts ==============================
platform linux -- Python 3.11.0, pytest-7.2.1, pluggy-1.0.0
rootdir: /home/nick/Desktop
collected 4 items                                                              

test/test_1.py ..                                                        [ 50%]
test/test_2.py ..                                                        [100%]

============================== 4 passed in 1.71s ===============================
```

pyccl/base.py
nikfilippas: `auto_assign` is a helper and is only used in `CCLHalosObject`. I figured, since we already save the passed arguments of initialized objects in the `halos` subpackage as attributes, we might as well control that centrally by a purpose-made object which does that automatically.
This also helps with the `repr` of the object. Users can see what arguments where used to initialize, say, an `HMCalculator`.

pyccl/base.py
nikfilippas: Python doesn't bind methods to an object from outside `__init__`. There are two kinds of objects: functions, and bound methods. If we were to omit L560, the function wouldn't become a bound method, and the implicit first argument (`self`) wouldn't refer to an instance of that class. So, in this way, you're telling the class dict that what you give it is actually a bound method, not a function.

pyccl/base.py
nikfilippas: Foul language maybe, I meant if the object that we try to unlock comes from outside of the CCL library, the devs should catch it downstream of course.

pyccl/base.py
nikfilippas: This just deletes them. This is inside the context manager `UnlockInstance`, which means that the object has been mutated. When the object is mutated, we want to discard the previously saved `_repr` and `_hash`. No need to recreate them, unless the user asks to. They are only recreated ad-hoc (i.e. when `==`, `print()`, `repr()` etc. are used).

nikfilippas: *Once they are recreated, how are they stored?*:
This happens inside `CCLObject`. Note that `_repr` and `_hash` are cached properties, which means that they are saved when they are computed.

nikfilippas: The best way to understand this is that we un-cache them using the context manager. Otherwise, they'd be stored forever and wouldn't allow us to mutate objects.

pyccl/base.py
nikfilippas: There's the context manager `UnlockInstance` which does the bulk of the work, and then there's the helper function `unlock_instance` which can be used as a decorator. Currently, `unlock_instance` is only meant to be used as a function decorator, and is agnostic to the variable name of what is being mutated. Implicitly, this is assumed to be the first argument of the function (`self`), because the only reason I created this one was to use it inside class methods which mutate their objects.

If multiple `CCLObject`s need to be mutated at a time, this can simply be done using stacked context managers. In theory, I could extend the functionality to accept multiple objects, but I think it would unnecessarily complicate the code, especially as `mutate` needs to be specified for all of them.

Let me know what you think.

pyccl/base.py
nikfilippas: Answering the second part of the question: every `CCLObject` has a `__repr__`. If it doesn't, we simply fall back to Python's native one.

pyccl/base.py
nikfilippas: These two lines basically say the following:
*If `__init__` and/or `update_parameters` exist, unlock the object before the functions are run, and lock them again on exit.*

pyccl/base.py
nikfilippas: For the first part of the question, let me demonstrate it to you first, because it might seem tricky. I think this is the simplest way to do it, but you might come up with something better.

pyccl/base.py
nikfilippas: They are automatically locked when `__init__` finishes. See the `Funlock` comment I made.

pyccl/base.py
damonge: And even for `CCLHalosObject` it worries me that we're doing this. It may not be the case right now, but in the future I can see us writing profiles whose initialisation depends on a large amount of data that never gets used again. One example is my plans for a future CIB profile that takes in a bunch of infrared spectra at different redshifts and then interpolates through them to get a single spectrum. Storing all the spectra is unnecessary.

nikfilippas: > And even for CCLHalosObject it worries me that we're doing this.

I see. We can discuss it. I wasn't too keen on doing that either. Let's discuss it and find a better way to do this.

pyccl/base.py
nikfilippas: Imagine for example the following scenario:
```python
cosmo0 = ccl.CosmologyVanillaLCDM()
cosmo1 = ccl.CosmologyVanillaLCDM()
```
The following will work
```python
# Nested context managers for multiple attribute assignments.
with ccl.UnlockInstance(cosmo0):
    with ccl.UnlockInstance(cosmo1):
        cosmo0.test = 0
        cosmo1.test = 1
```
and the following (neater) way will work as well
```python
# Can be sequenced like so
with ccl.UnlockInstance(cosmo0), ccl.UnlockInstance(cosmo1):
    cosmo0.test = 2
    cosmo1.test = 3
```

If what you are proposing is something like
```python
with ccl.UnlockInstance(cosmo0, cosmo1, mutate=[True, False]):
    cosmo0.test = 2
    cosmo1.test = 3
```
I don't think there's much benefit to it.

pyccl/tracers.py
nikfilippas: Indeed, I removed them.

pyccl/tests/__init__.py
damonge: ok, great

pyccl/base.py
damonge: OK, to summarise what we discussed on slack: you'll remove `auto_assign`

pyccl/base.py
damonge: ok, specify that args[argv] is meant to be a CCLObject in the comment and we're done with this

nikfilippas: I have, and I have pushed it. Not sure why it is not showing.

nikfilippas: It is showing now.

pyccl/base.py
nikfilippas: I have now introduced `__repr_attrs__` which acts as a class variable (hook) in place of the `auto_assign` decorator.
The idea is that instead of having to implement a tailor-made `__repr__` method for their class, developers can just specify which instance variables they need to be included in the `repr` via the `__repr_attrs__` hook, and the rest is done automatically.
As always, this can be overridden if they want to implement their own new shiny `repr`.

nikfilippas: For this to work, devs should make sure that the attribute names they pass in `__repr_attrs__` are defined as instance variables in `__init__`. However, there is no easy way to check down the whole MRO of a class to verify that all variables are indeed defined.

In order to not leave this hanging, I implemented a test to catch this kind of error during development, before it reaches the users. Before all the tests are run with `pytest`, I decorate every `__init__` which in subclasses of `CCLHalosObject` with a function that verifies that, once `__init__` has finished running, all variable names specified in `__repr_attrs__` exist. If they don't some tests will fail with a useful error. The test that will fail depends on where the erroneous class is used first.

nikfilippas: Note, that this check only happens in testing. The actual code doesn't have that. This check is developer-oriented, and it flags potential errors before they are merged to master.

nikfilippas: The test I'm talking about is in `tests/conftest.py` and uses the decorator I wrote in `tests/test_cclobject.py`.

pyccl/base.py
nikfilippas: I think this has been addressed with the new commits that get rid of `auto_assign`.

pyccl/halos/massdef.py
```diff
@@ -246,69 +246,43 @@ def from_name(cls, name):
         Returns:
             MassDef subclass corresponding to the input name.
         """
-        mass_defs = {m.name: m for m in cls.__subclasses__()}
-
-        if name in mass_defs:
-            return mass_defs[name]
-        else:
+        try:
+            return eval(f"MassDef{name.capitalize()}")
```
damonge, line 250: why capitalize?

damonge (changes requested):
OK, a few more comments and I'll start with the individual classes and then tests.

damonge:
Also, have you checked what's happening to the tests?

pyccl/base.py
```diff
@@ -298,3 +299,345 @@ def increment(self):
 
     def reset(self):
         self.counter = 0
+
+
+def auto_assign(func, sig):
```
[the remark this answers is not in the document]
nikfilippas: I forgot to remove it.

pyccl/base.py
```diff
@@ -298,3 +299,345 @@ def increment(self):
...
+            linked = cls.__linked_abstractmethods__
+
+            def is_abstract(cls, method):
+                # Check if the `method` of class `cls` is abstract.
+                meth = getattr(cls, method)
+                return getattr(meth, "__isabstractmethod__", False)
+
+            if not all([is_abstract(cls, method) for method in linked]):
+                abstracts = set(cls.__abstractmethods__) - set(linked)
+                cls.__abstractmethods__ = frozenset(abstracts)
```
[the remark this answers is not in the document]
nikfilippas: Actually I agree. It was already in my to-do list of things to eliminate so for the past few days I'd been trying to think of ways to do this more elegantly.
The `abc` module introduced the really convenient `update_abstractmethods` which can be used as a class decorator and could achieve what I was trying to implement. The problem is this only exists in Python 3.10. 🙄

So I'll probably be dropping this altogether and just add a check after the class is defined, that verifies that `_real` or `_fourier` exist. Do you agree?

pyccl/base.py
```diff
@@ -298,3 +299,345 @@ def increment(self):
...
+    may be called with ``@unlock_instance(mutate=False)`` (or equivalently
+    for the context manager ``UnlockInstance(..., mutate=False)``). Otherwise,
+    the instance is assumed to have mutated.
+    """
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+
+        # 1. Store the signature of the constructor on import.
+        cls.__signature__ = signature(cls.__init__)
```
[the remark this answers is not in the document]
nikfilippas: `self.__signature__` is used in `_repr.py` to determine which default parameters are passed, so that they aren't printed in the repr.

pyccl/base.py
```diff
@@ -298,3 +299,345 @@ def increment(self):
...
+        # do nothing; otherwise reset.
+        if self.id != self.object_lock._lock_id:
+            return
+
+        with self.thread_lock:
+            # Reset `repr` if the object has been mutated.
+            if self.mutate:
+                try:
+                    delattr(self.instance, "_repr")
+                    delattr(self.instance, "_hash")
```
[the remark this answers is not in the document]
nikfilippas: Every time `==`, `hash()`, `repr()`, `print()` is used.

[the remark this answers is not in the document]
nikfilippas: Under 'everyday' CCL use, only comparisons are used. If no comparisons between objects are used, the `repr` will not be calculated, so it won't make any difference to what is happening under the hood.

pyccl/halos/massdef.py
nikfilippas: `capitalize()` capitalizes only the first letter of every word, if it starts with a letter. Basically there for `MassDefVir`.

pyccl/base.py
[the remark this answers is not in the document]
nikfilippas: I had a very close look at this and (unfortunately) metaclasses is the only obvious option if we want to link abstract methods. This can't be done any other way (e.g. within `__init_subclass__`) because `abc` only determines which methods are abstract after all this has run. So the only way to tap into the inner workings of `abc` is to use a metaclass (abstract methods are defined using `ABCMeta` in the first place).

Seeing as we only want this to link `_real` and `_fourier` I also find no point in having all this extra code for these two methods, so I removed them completely.

nikfilippas:
@damonge tests should be fine now.

pyccl/halos/profiles.py
```diff
@@ -1256,7 +1270,7 @@ def update_parameters(self, mass_bias=None, P0=None,
             re_fourier = True
             self.x_out = x_out
 
-        if re_fourier and (self._fourier_interp is not None):
+        if re_fourier and (self._fourier is not None):
```
damonge, line 1273: is this right?

pyccl/_repr.py
```diff
@@ -0,0 +1,383 @@
+import numpy as np
```
damonge, line 1: I think all this looks ok, but I'd like us to have a library-level flag that determines whether to use these fancy reprs or just the standard "this is my id" repr. The default for this flag would be "Use the fancy reprs", but I'd like to easily switch to the simple option if needed (or if anyone determines that the performance hit actually matters for them).

damonge (changes requested):
Last few comments before I move on to the tests.

pyccl/halos/profiles.py
nikfilippas: It is not, thanks for catching it.

pyccl/tests/test_cclobject.py
```diff
@@ -0,0 +1,189 @@
...
+    COSMO2 = ccl.CosmologyVanillaLCDM(**kwargs)
+    assert COSMO1 != COSMO2
+
+    # 2. Using a Pk2D object.
+    cosmo = ccl.CosmologyVanillaLCDM(transfer_function="bbks")
+    cosmo.compute_linear_power()
+    PK1 = cosmo.get_linear_power()
+    PK2 = ccl.Pk2D.pk_from_model(cosmo, "bbks")
+    assert PK1 == PK2
+    assert ccl.Pk2D(empty=True) == ccl.Pk2D(empty=True)
```
damonge, line 30: can you add a test when they're not the same?

pyccl/tests/test_cclobject.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # 3.1. Using a factorizable Tk3D object.
+    a_arr, lk_arr, pk_arr = PK1.get_spline_arrays()
+    TK1 = ccl.Tk3D(a_arr=a_arr, lk_arr=lk_arr,
+                   pk1_arr=pk_arr, pk2_arr=pk_arr, is_logt=False)
+    TK2 = ccl.Tk3D(a_arr=a_arr, lk_arr=lk_arr,
+                   pk1_arr=pk_arr, pk2_arr=pk_arr, is_logt=False)
+    assert TK1 == TK2
```
damonge, line 38: As above

pyccl/_repr.py
nikfilippas: That's a good idea, I infused the `CCLObject` with an object similar to `_object_lock`, which I call `_fancy_repr`. It can be enabled/disabled using e.g. `ccl.CCLObject._fancy_repr.disable()` to fall back to Python's default memory-address-id.

damonge (commented):
Quick comments on tests

pyccl/_repr.py
nikfilippas: This is inherited, but we ideally want users who want to disable it (i.e. they know what they are doing) to only access it through `CCLObject` and *not* through any of its subclasses, or their instances. To prevent this from happening, I replace the `_fancy_repr` object in all subclasses of `CCLObject` with a descriptor that exits and throws a useful message of how to properly enable/disable the framework.

damonge: Would it be so bad if users disabled it through the subclasses? I imagine in that case it's only disabled for that subclass (and its subclasses) right? And that might have its advantages (e.g. only disabling some of the expensive ones).

pyccl/base.py
```diff
@@ -298,3 +299,316 @@ def increment(self):
...
+        # `_repr` (which is cached). Decorator `cached_property` requires
+        # that `__set_name__` is called on it.
+        bmethod = functools.cached_property(method)
+        cl._repr = bmethod
+        bmethod.__set_name__(cl, "_repr")
+        # Fall back to using `__ccl_repr__` from `CCLObject`.
+        cl.__repr__ = cl.__ccl_repr__
+
+
+class _DisableGetMethod:
```
damonge, line 474: See my comment about why this may not be a bad thing to have at the subclass level.

pyccl/base.py
```diff
@@ -298,3 +299,316 @@ def increment(self):
...
+        for cl in self._classes.keys():
+            cl.__repr__ = object.__repr__
+        self._enabled = False
+
+    @classmethod
+    def bind_and_replace(cls, cl, method):
+        """Bind a method to a class, and replace original with default.
+        This helper only works for binding and replacing ``__repr__`` methods
+        for ``CCLObjects``.
+        """
```
damonge, line 463: Can you document what `cl` is here?

pyccl/_repr.py
nikfilippas: Yes, it would be bad. There is only one instance of `FancyRepr` in the entire library. The classes that have been fancy-repred are stored in `FancyRepr._classes` so if someone were to do `ccl.Cosmology._fancy_repr.disable()` or even `cosmo._fancy_repr.disable()` they would in fact disable everything.

I would leave it if there were multiple instances of `FancyRepr` like they are with `ObjectLock` but I think this is an overkill for what we are trying to do.

pyccl/tests/test_cclobject.py
```diff
@@ -0,0 +1,189 @@
+import pytest
+import numpy as np
+import pyccl as ccl
+import functools
+
+
```
damonge, line 6: Can you test that disabling fancy_repr does the right thing?

pyccl/_repr.py
nikfilippas: What I mean is that
```
id(ccl.Cosmology._fancy_repr) == id(cosmo._fancy_repr) == id(ccl.CCLObject._fancy_repr) == id(ccl.HaloProfileGaussian()._fancy_repr)
```

damonge (commented):
Alright, few comments on fancy_repr. Thanks for implementing it.

pyccl/_repr.py
damonge: Right, ok. Then, can you add a test that shows this failing?

nikfilippas: Ofc, I am writing all the tests at the moment.

nikfilippas: Done.

pyccl/tests/test_cclobject.py
nikfilippas: Added in the tests.

pyccl/tests/test_cclobject.py
nikfilippas: done

pyccl/tests/test_cclobject.py
nikfilippas: done

pyccl/base.py
nikfilippas: Done.

pyccl/tests/test_cclobject.py
```diff
@@ -4,6 +4,25 @@
...
 
 
+def test_fancy_repr():
+    # Test fancy-repr controls.
+    cosmo1 = ccl.CosmologyVanillaLCDM()
+    cosmo2 = ccl.CosmologyVanillaLCDM()
+
+    ccl.CCLObject._fancy_repr.disable()
+    assert "False" in repr(ccl.CCLObject._fancy_repr)
+    assert repr(cosmo1) == object.__repr__(cosmo1)
```
damonge, line 13: Humm, since cosmo1 was created before disabling fancy_repr, shouldn't it have the fancy repr? Or do they only get created when you first do a comparison?

pyccl/tests/test_cclobject.py
```diff
@@ -4,6 +4,25 @@
...
+    assert cosmo1 != cosmo2
+
+    ccl.CCLObject._fancy_repr.enable()
+    assert "True" in repr(ccl.CCLObject._fancy_repr)
+    assert repr(cosmo1) != object.__repr__(cosmo2)
+    assert cosmo1 == cosmo2
+
+    with pytest.raises(AttributeError):
+        cosmo1._fancy_repr.disable()
```
damonge, line 21: Would `ccl.Cosmology._fancy_repr.disable()` also fail? (I.e. calling from the class instead of the instance)

pyccl/base.py
```diff
@@ -392,68 +392,75 @@ def __exit__(self, type, value, traceback):
             # Lock the instance on exit.
             self.object_lock.lock()
 
+    @classmethod
+    def unlock_instance(cls, func=None, *, argv=0, mutate=True):
+        """Decorator that temporarily unlocks an instance of CCLObject.
 
-def unlock_instance(func=None, *, argv=0, mutate=True):
-    """Decorator that temporarily unlocks an instance of CCLObject.
```
damonge: why did you bring it inside `UnlockInstance`? Same for `Funlock`.

pyccl/tests/test_cclobject.py
```diff
@@ -0,0 +1,212 @@
...
+    # Test fancy-repr controls.
+    cosmo1 = ccl.CosmologyVanillaLCDM()
+    cosmo2 = ccl.CosmologyVanillaLCDM()
+
+    ccl.CCLObject._fancy_repr.disable()
+    assert repr(cosmo1) == object.__repr__(cosmo1)
+    assert cosmo1 != cosmo2
+
+    ccl.CCLObject._fancy_repr.enable()
+    assert repr(cosmo1) != object.__repr__(cosmo2)
```
damonge, line 17: Humm, what is this testing exactly? Won't this always be the case, regardless what you've done to fancy_repr?

damonge (commented):
Alright, final few comments before merging!

pyccl/tests/test_cclobject.py
nikfilippas: First of all, this test is outdated, my `git` says "nothing to commit" and once again, github hasn't updated the changes.
Secondly, it turns out I was wrong yesterday when I told you on Slack that already-instantiated objects won't change. Every time `__repr__` is called, the interpreter re-calculates what the `__repr__` for that object should be, looking at its class. So if you change the class of an object, you also automatically change the object's method. Which means that it even works with instantiated objects.

pyccl/tests/test_cclobject.py
nikfilippas: Correct, should have been `cosmo1` instead of `cosmo2`.

pyccl/tests/test_cclobject.py
nikfilippas: Yes, every instance of `CCLObject` would fail. Instances might be subclasses, or instances of subclasses.

nikfilippas: I'll add a test for the class.

pyccl/tests/test_cclobject.py
nikfilippas: Okay, it looks like if you go to the "Files Changed" tab, it shows the correct (synced) file. So I can only assume that Github has forgot to put an `(outdated)` badge in the "Conversation" tab.

pyccl/tests/test_cclobject.py
nikfilippas: Done, check and we can merge.

pyccl/tests/test_cclobject.py
damonge: OK, but if _repr is cached, it's not recalculated, right?

damonge (commented):
OK, can you merge master after the latest GHA branch merge? Then, when all tests pass, I'll approve.

pyccl/tests/test_cclobject.py
nikfilippas: It is. If `_repr` is cached, and in the meantime you disable and re-enable fancy-reprs, the `_repr` that was cached gets replaced (on enable):
```python
    def enable(cls):
        """Enable fancy representations if they exist."""
        for cl, method in cls._classes.items():
            FancyRepr.bind_and_replace(cl, method)
        cls._enabled = True
```
which calls
```python
    @classmethod
    def bind_and_replace(cls, cl, method):
        """Bind ``method`` to class ``cl``, and replace original with default.
        This helper only works for binding and replacing ``__repr__`` methods
        for ``CCLObjects``.
        """
        # If the class defines a custom `__repr__`, this will be the new
        # `_repr` (which is cached). Decorator `cached_property` requires
        # that `__set_name__` is called on it.
        bmethod = functools.cached_property(method)
        cl._repr = bmethod
        bmethod.__set_name__(cl, "_repr")
        # Fall back to using `__ccl_repr__` from `CCLObject`.
        cl.__repr__ = cl.__ccl_repr__
```

nikfilippas: See here for an example:
```python
import pyccl as ccl
cosmo.compute_linear_power()
pk = cosmo.get_linear_power()
print(pk)
ccl.CCLObject._fancy_repr.disable()
print(pk)
pk += 1
ccl.CCLObject._fancy_repr.enable()
print(pk)
```
The first print:
```
<pyccl.pk2d.Pk2D>
	+==============+=================================================================+
	| a \ log10(k) | -4.30e+00 -4.16e+00 -4.01e+00 ...  8.37e-01  9.29e-01  1.02e+00 |
	+==============+=================================================================+
	|  1.00e-02    |  1.84e-01  2.53e-01  3.49e-01 ...  1.09e-04  6.25e-05  3.56e-05 |
	|  1.26e-02    |  2.98e-01  4.11e-01  5.66e-01 ...  1.72e-04  9.83e-05  5.60e-05 |
	|  1.58e-02    |  4.82e-01  6.63e-01  9.13e-01 ...  2.71e-04  1.55e-04  8.81e-05 |
	|     ...      |                               ...                               |
	|  9.54e-01    |  1.19e+03  1.64e+03  2.26e+03 ...  6.06e-01  3.47e-01  1.97e-01 |
	|  9.77e-01    |  1.22e+03  1.68e+03  2.32e+03 ...  6.22e-01  3.56e-01  2.02e-01 |
	|  1.00e+00    |  1.25e+03  1.72e+03  2.37e+03 ...  6.37e-01  3.64e-01  2.07e-01 |
	+==============+=================================================================+
	| is_log = True , extrap_orders = (1, 2)                                         |
	| HASH_ARRS = 0x1b938ef61e5e1edb9                                                |
	+==============+=================================================================+
```

The second print
```
<pyccl.pk2d.Pk2D object at 0x7f6f21779450>
```

The third print
```
<pyccl.pk2d.Pk2D>
	+==============+=================================================================+
	| a \ log10(k) | -4.30e+00 -4.16e+00 -4.01e+00 ...  8.37e-01  9.29e-01  1.02e+00 |
	+==============+=================================================================+
	|  1.00e-02    |  1.18e+00  1.25e+00  1.35e+00 ...  1.00e+00  1.00e+00  1.00e+00 |
	|  1.26e-02    |  1.30e+00  1.41e+00  1.57e+00 ...  1.00e+00  1.00e+00  1.00e+00 |
	|  1.58e-02    |  1.48e+00  1.66e+00  1.91e+00 ...  1.00e+00  1.00e+00  1.00e+00 |
	|     ...      |                               ...                               |
	|  9.54e-01    |  1.19e+03  1.64e+03  2.26e+03 ...  1.61e+00  1.35e+00  1.20e+00 |
	|  9.77e-01    |  1.22e+03  1.68e+03  2.32e+03 ...  1.62e+00  1.36e+00  1.20e+00 |
	|  1.00e+00    |  1.25e+03  1.73e+03  2.37e+03 ...  1.64e+00  1.36e+00  1.21e+00 |
	+==============+=================================================================+
	| is_log = True , extrap_orders = (1, 2)                                         |
	| HASH_ARRS = 0x10759946ba24489cb                                                |
	+==============+=================================================================+
```

nikfilippas: So essentially by disabling we simply unlink `__repr__` from `_repr`. Note that `_repr` is the one that carries the workload when using fancy-reprs.

damonge: right, but that's only if you change fancy_repr, if you don't, _repr is cached

damonge (approved):
LGTM

PR #935, opened 2022-05-06
Fix status message string initialisation

tilmantroester:
`ccl_cosmology_set_status_message` checks if `cosmo->status_message` is empty, i.e a 0-length string. If `cosmo->status_message` is initially garbage, that check triggers the warning about overwriting an existing message and prints garbage. This PR explicitly initialises `cosmo->status_message` as a 0-length string.

### DIFF
--- a/src/ccl_core.c
+++ b/src/ccl_core.c
@@ -286,7 +286,8 @@ ccl_cosmology * ccl_cosmology_create(ccl_parameters params, ccl_configuration co
   cosmo->computed_growth = false;
   cosmo->computed_sigma = false;
   cosmo->status = 0;
-  ccl_cosmology_set_status_message(cosmo, "");
+  // Initialise as 0-length string
+  cosmo->status_message[0] = "\0";
 
   if(cosmo->spline_params.A_SPLINE_MAX !=1.) {
     cosmo->status = CCL_ERROR_SPLINE;
@@ -511,8 +512,8 @@ void ccl_cosmology_set_status_message(ccl_cosmology * cosmo, const char * messag
   #pragma omp critical
   {
     if(strlen(cosmo->status_message) != 0) {
-      ccl_raise_warning(CCL_ERROR_OVERWRITE, "Status message being overwritten.");
-      ccl_raise_warning(cosmo->status, cosmo->status_message);
+      ccl_raise_warning(CCL_ERROR_OVERWRITE, "Status message being overwritten:");
+      fprintf(stderr, "STATUS: %d. %s\n", cosmo->status, cosmo->status_message);
     }
     vsnprintf(cosmo->status_message, trunc, message, va);
 
--- a/src/ccl_core.c
+++ b/src/ccl_core.c
@@ -287,7 +287,7 @@ ccl_cosmology * ccl_cosmology_create(ccl_parameters params, ccl_configuration co
   cosmo->computed_sigma = false;
   cosmo->status = 0;
   // Initialise as 0-length string
-  cosmo->status_message[0] = "\0";
+  cosmo->status_message[0] = '\0';
 
   if(cosmo->spline_params.A_SPLINE_MAX !=1.) {
     cosmo->status = CCL_ERROR_SPLINE;


nikfilippas:
I tested this and while it solves the issue of throwing bogus warnings on a Cosmology instantiation, sometimes the status message is still nonempty:
```
In  [1]: ccl.CosmologyVanillaLCDM().cosmo.status_message
Out[1]: ' \udce2\udcb3\x11\udc93\x7f'
```
and setting a real warning triggers the overwrite warning:
```
In  [2]: ccl.lib.cosmology_set_status_message(cosmo.cosmo, "actual warning")
WARNING 1064: Status message being overwritten:
STATUS: 0.  (some bogus unicode that breaks github markdown rendering)
```

Also, compiling this yields this warning:
```
/home/nick/CCL/src/ccl_core.c:290:28: warning: assignment to 'char' from 'char *' makes integer from pointer without a cast [-Wint-conversion]
  290 |   cosmo->status_message[0] = "\0";
      |                            ^
```
with some perhaps useful help from [stackoverflow](https://stackoverflow.com/questions/5122231/warning-assignment-makes-integer-from-pointer-without-a-cast).

tilmantroester:
I'm away and without laptop until next week, at which point I can have another  look at this.

damonge:
Just a quick ping about this one @tilmantroester

tilmantroester:
I think I've fixed it. I forgot (although I'm not sure I was ever aware of this fact) that it matters in C whether you use single or double quotes.

src/ccl_core.c
```diff
@@ -511,8 +512,8 @@ void ccl_cosmology_set_status_message(ccl_cosmology * cosmo, const char * messag
   #pragma omp critical
   {
     if(strlen(cosmo->status_message) != 0) {
-      ccl_raise_warning(CCL_ERROR_OVERWRITE, "Status message being overwritten.");
-      ccl_raise_warning(cosmo->status, cosmo->status_message);
+      ccl_raise_warning(CCL_ERROR_OVERWRITE, "Status message being overwritten:");
+      fprintf(stderr, "STATUS: %d. %s\n", cosmo->status, cosmo->status_message);
```
damonge, line 516: Maybe you already discussed this. Why are we changing this from a warning to always printing an error message?

damonge: Right, but it's enclosed by:
```
if (_ccl_debug_mode_policy == CCL_DEBUG_MODE_ON) {
    fprintf(stderr, "WARNING %d: %s\n", err, message);
  }
```
meaning that we can switch them off, right? Don't we want the same here?

tilmantroester: I'd say no, since this "warning" only gets printed when there's an error message being overwritten. So in that sense it should be an error but our current error handling can't deal with more than one error occurring.

damonge:
I can't reproduce this either. @nikfilippas could it be that you have an old CCL version lingering around?

nikfilippas:
I can consistently reproduce this after removing every trace of CCL and reinstalling in py37 and py39 on my laptop and on the cluster, with python and ipython. @tilmantroester @damonge  could you confirm that you don't see anything when you run
```
python -c "import pyccl as ccl; ccl.lib.cosmology_set_status_message(ccl.CosmologyVanillaLCDM().cosmo, 'actual warning')"
```
?

I get something like
```
WARNING 1064: Status message being overwritten:
STATUS: 32622. h��~n
```
as output.

tilmantroester:
`python setup.py clean && pip install -e .`
and then
`python -c "import pyccl as ccl; ccl.lib.cosmology_set_status_message(ccl.CosmologyVanillaLCDM().cosmo, 'actual warning')"`
Produces no outputs for me. `python setup.py clean` removes `ccllib.so`, which otherwise can stick around and cause issues.

nikfilippas:
Doesn't work for me when I delete `_ccllib.so` and reinstall (I get some weird seg faults) but I guess if both of you can't reproduce that then the issue must be on my part. Happy for this to be merged if David can confirm no output.

tilmantroester:
I just tried on a clean conda env and clone of the repo and still works on my machine.

PR #943, opened 2022-06-12
Fix WL functions in pyccl.halos.profiles

hsinfan1996:
In pyccl.halos.profiles, convergence, shear, reduced_shear and magnification can now handle array_like a_source.

### DIFF
--- a/benchmarks/test_haloprofile.py
+++ b/benchmarks/test_haloprofile.py
@@ -146,6 +146,8 @@ def test_weak_lensing_functions():
         np.log(rmax/rmin) * np.arange(data.shape[0]) / (data.shape[0]-1))
 
     r_al = r / a_lens
+    len_r = len(r)
+    a_source = [a_source]*len_r
 
     mdef = ccl.halos.MassDef(mDelta, 'matter')
     c = ccl.halos.ConcentrationConstant(c=concentration, mdef=mdef)
--- a/pyccl/halos/profiles.py
+++ b/pyccl/halos/profiles.py
@@ -292,7 +292,7 @@ class HaloProfile(object):
             M (float or array_like): halo mass in units of M_sun.
             a_lens (float): scale factor of lens.
             a_source (float or array_like): scale factor of source.
-                If array_like, it must have the same shape as r.
+                If array_like, it must have the same shape as `r`.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
 
@@ -326,7 +326,7 @@ class HaloProfile(object):
             M (float or array_like): halo mass in units of M_sun.
             a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
-                If array_like, it must have the same shape as r.
+                If array_like, it must have the same shape as `r`.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
 
@@ -359,7 +359,7 @@ class HaloProfile(object):
             M (float or array_like): halo mass in units of M_sun.
             a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
-                If array_like, it must have the same shape as r.
+                If array_like, it must have the same shape as `r`.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
 
@@ -387,7 +387,7 @@ class HaloProfile(object):
             M (float or array_like): halo mass in units of M_sun.
             a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
-                If array_like, it must have the same shape as r.
+                If array_like, it must have the same shape as `r`.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
 
--- a/pyccl/halos/profiles.py
+++ b/pyccl/halos/profiles.py
@@ -290,7 +290,7 @@ class HaloProfile(object):
             cosmo (:class:`~pyccl.core.Cosmology`): A Cosmology object.
             r (float or array_like): comoving radius in Mpc.
             M (float or array_like): halo mass in units of M_sun.
-            a_lens (float or array_like): scale factor of lens.
+            a_lens (float): scale factor of lens.
             a_source (float or array_like): scale factor of source.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
@@ -323,7 +323,7 @@ class HaloProfile(object):
             cosmo (:class:`~pyccl.core.Cosmology`): A Cosmology object.
             r (float or array_like): comoving radius in Mpc.
             M (float or array_like): halo mass in units of M_sun.
-            a_lens (float or array_like): lens' scale factor.
+            a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
@@ -355,7 +355,7 @@ class HaloProfile(object):
             cosmo (:class:`~pyccl.core.Cosmology`): A Cosmology object.
             r (float or array_like): comoving radius in Mpc.
             M (float or array_like): halo mass in units of M_sun.
-            a_lens (float or array_like): lens' scale factor.
+            a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
             mass_def (:class:`~pyccl.halos.massdef.MassDef`):
                 a mass definition object.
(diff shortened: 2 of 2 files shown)


damonge:
@hsinfan1996 thanks for this. Just so I'm clear, what fails exactly when you make `a_source` an array?

hsinfan1996:
> @hsinfan1996 thanks for this. Just so I'm clear, what fails exactly when you make `a_source` an array?

In convergence and shear, the first step `self.projected(cosmo, r, M, a_lens, mass_def)` requires `a_lens` to be a float. The next step `sigma_critical(cosmo, a_lens, a_source)` requires `a_lens` and `a_source` to have the same dimension, so both of them have to be floats given that `a_lens` is a float. If `a_source` is an array, `sigma_critical` throws an error.

benchmarks/test_haloprofile.py
```diff
@@ -146,6 +146,8 @@ def test_weak_lensing_functions():
         np.log(rmax/rmin) * np.arange(data.shape[0]) / (data.shape[0]-1))
 
     r_al = r / a_lens
+    len_r = len(r)
+    a_source = [a_source]*len_r
```
damonge, line 150: I don't understand this. Why do you need `a_source` to be of the same shape as `r`?

hsinfan1996:
> OK, I see! For practical applications, how inconvenient is it that `a_lens` can only be a float? Just trying to figure out if it's worth trying to enable that in this PR.

From what I know, that `a_lens` can only be a float does not affect any relevant calculations in CLMM.

benchmarks/test_haloprofile.py
hsinfan1996: The benchmark data `data` has the same length as `r` (500 values), so requiring`a_source` to have the same length as `r` ensures the outputs from the calculations below have the correct dimension. Otherwise, there will be an error on numpy broadcasting.

damonge:
OK, but the only two possibilities are:
a) `a_source` is a float
b) `a_source` is an array of the same shape as `r`.

I think option b) above is not useful at all (and is not documented as such), so I can see two ways forward:
1. We force `a_source` and `a_lens` to be always floats.
2. We allow `a_source` to be an array, but then the output of these functions should be a 2D array of shape `[len(a_source), len(r)]`, so that `r` and `a_source` don't need to be linked up.

Option 1 above seems to be the simplest one, so I'd advocate for that unless you feel like giving the second one a go.

hsinfan1996:
> OK, but the only two possibilities are: a) `a_source` is a float b) `a_source` is an array of the same shape as `r`.
> 
> I think option b) above is not useful at all (and is not documented as such), so I can see two ways forward:
> 
> 1. We force `a_source` and `a_lens` to be always floats.
> 2. We allow `a_source` to be an array, but then the output of these functions should be a 2D array of shape `[len(a_source), len(r)]`, so that `r` and `a_source` don't need to be linked up.
> 
> Option 1 above seems to be the simplest one, so I'd advocate for that unless you feel like giving the second one a go.

Actually, b) is what I wanted to achieve in this PR and is already very useful to us. In addition, it only requires a few lines of code added to pyccl to allow array_like `a_source`, which is said to be supported in the documentation. I agree that we can make the functions more versatile by allowing different shapes of `a_source` and `r`, but I think it will need c-level changes (not 100% sure).

damonge:
Humm, I don't really understand how b) is useful at all. Why is it useful to have the function return e.g. convergence as a function of (r,a_source) where each r corresponds to a different a_source?

Wouldn't it be more useful to have it as a function of both r and a_source independently of each other?

damonge:
(I don't think the latter would require c-level modifications, BTW, that's why I'm asking)

eduardojsbarroso:
> 

For what I understand, in CLMM, we want to make computations considering several sources with different projected radius but in the same redshift, therefore same a_source, or to make computations for several sources in different redshifts. That is why we wanted the case b). Either each projected radius must have a pair, thus len(r) = len(a_source), or we are giving several projected radius but we want to analyze all sources in the same redshift, which then requires `a_source = [a_source]*len_r`.


I feel like the case where, for example, r = [1,2,3] and a_source = [0.5, 0.4] is not very usefull because it not clear what this stands for. When we give an array `r` and a float `a_source`, we can consider that all the sources are in the same redshift, but in the case where we give an array `a_source` that does not match the array `r`, how can we know in which redshift is the all the sources represented by `r`? 

We did implement these changes in the CLMM but we've come up with two problems: 1) the vectorization on our end let the code very slow. 2)Even though the calculations were feasible, we got a failled test because 'a_source' was an array. And that is why we opened this issue.

damonge:
Thanks @eduardojsbarroso 
OK, just wanna make sure I understand this (and I'm afraid I still don't), so when you say:

>I feel like the case where, for example, r = [1,2,3] and a_source = [0.5, 0.4] is not very usefull because it is not clear what this stands for

This would give you the profile at r=[1, 2, 3] for two sources, one at a=0.5 and one at a=0.4. Isn't this more useful than e.g. having the profile for e.g. r=1 at a=0.4, r=2 at a=0.5, and r=3 at a=0.6?

I had assumed you'll want to do some line-of-sight integrals that involve integrating over source redshifts for the same lens, and that's why I thought that having r and a_source be independent would be useful, but I'm probably wrong. Can I check what calculation is it that benefits from having r and a_source coupled?

damonge:
(I should say, I don't want to delay this getting merged, just want to make sure I'm on the same page!)

pyccl/halos/profiles.py
```diff
@@ -300,6 +300,9 @@ def convergence(self, cosmo, r, M, a_lens, a_source, mass_def=None):
                 :math:`\\kappa`
         """
         Sigma = self.projected(cosmo, r, M, a_lens, mass_def) / a_lens**2
+        if np.iterable(a_source):
```
nikfilippas, line 303: Small technicality, but could we replace `np.iterable` with a builtin (such as `hasattr(a_source), "__iter__")`)? The latter is twice as fast, and numpy has started to deprecate old helper functions that can easily be replaced by builtins.

pyccl/halos/profiles.py
```diff
@@ -331,6 +334,9 @@ def shear(self, cosmo, r, M, a_lens, a_source, mass_def=None):
         """
         Sigma = self.projected(cosmo, r, M, a_lens, mass_def)
         Sigma_bar = self.cumul2d(cosmo, r, M, a_lens, mass_def)
+        if np.iterable(a_source):
```
nikfilippas, line 337: Same about numpy vs builtin here.

pyccl/halos/profiles.py
hsinfan1996: Fixed. Thanks for the explanation.

pyccl/halos/profiles.py
hsinfan1996: Fixed.

pyccl/halos/profiles.py
nikfilippas: Thanks!

pyccl/halos/profiles.py
```diff
@@ -290,7 +290,7 @@ def convergence(self, cosmo, r, M, a_lens, a_source, mass_def=None):
             cosmo (:class:`~pyccl.core.Cosmology`): A Cosmology object.
             r (float or array_like): comoving radius in Mpc.
             M (float or array_like): halo mass in units of M_sun.
-            a_lens (float or array_like): scale factor of lens.
+            a_lens (float): scale factor of lens.
             a_source (float or array_like): scale factor of source.
```
damonge, line 294: Can you add to the docs that "If an array, it must have the same shape as `r`"?

pyccl/halos/profiles.py
```diff
@@ -320,7 +323,7 @@ def shear(self, cosmo, r, M, a_lens, a_source, mass_def=None):
             cosmo (:class:`~pyccl.core.Cosmology`): A Cosmology object.
             r (float or array_like): comoving radius in Mpc.
             M (float or array_like): halo mass in units of M_sun.
-            a_lens (float or array_like): lens' scale factor.
+            a_lens (float): scale factor of lens.
             a_source (float or array_like): source's scale factor.
```
damonge, line 327: Same as above

pyccl/halos/profiles.py
```diff
@@ -349,7 +355,7 @@ def reduced_shear(self, cosmo, r, M, a_lens, a_source, mass_def=None):
...
```
damonge, line 359: Same as above

pyccl/halos/profiles.py
```diff
@@ -376,7 +382,7 @@ def magnification(self, cosmo, r, M, a_lens, a_source, mass_def=None):
...
```
damonge, line 386: Same as above

damonge (changes requested):
Thanks @hsinfan1996 @eduardojsbarroso , if this is actually what you guys need, let's move on with it. Thanks for indulging me!

Just one final query to fix the docs while we're at it and we can merge

pyccl/halos/profiles.py
hsinfan1996: Done. I was trying to not make the line too long, but I am not sure if I did it correctly.

hsinfan1996:
> Thanks @hsinfan1996 @eduardojsbarroso , if this is actually what you guys need, let's move on with it. Thanks for indulging me!
> 
> Just one final query to fix the docs while we're at it and we can merge

@damonge Thank you for reviewing and the conversation. @nikfilippas Thank you.

damonge (approved):
LGTM!

PR #944, opened 2022-06-12
CIB speedup

damonge:
This speeds up the calculation of the satellite luminosity in the CIB profile (as per @yanzastro 's suggestion)


damonge:
Just realised #917 is probably a better idea

PR #946, opened 2022-06-23
Allow user defined biases for SSC trispectrum

carlosggarcia:
As a first step towards the SSC implementation in TJPCov I think it would be useful to avoid having to input many halo model parameters, that will be sample specific, to compute the SSC. Instead, the user could compute the SSC using the galaxy biases that one gets fitting the power spectrum.

I have also fixed the docstring.


carlosggarcia:
I realized this is not going to solve the issue. I will open a new PR

Issue #876, opened 2021-05-16
Too simple extrapolation of matter power spectra in cosmology calculator mode


nikfilippas:
The cosmology calculator is an excellent way to experiment with passing matter power spectra produced by emulators into CCL, which has the advantage of increasing the speed of the calculations. Emulators constructed on neural networks have some unavoidable scatter (i.e. the derivatives are not that smooth).

In the current implementation of the cosmology calculator, when the user wishes to probe scales (and scale factors) outside of the boundaries of the `P(k,a)` passed, CCL extrapolates high-k's with a quadratic and low-k's linearly.

While this has some physical motivation (large scales are linear), if a neural-net emulator's boundaries are not quite at the linear regime (and also due to the emulator's scatter) extrapolation doesn't work very well. While this is mostly on the emulator, I think CCL could be a bit more elaborate with extrapolation.

I have been testing `baccoemu` (`arXiv:2104.14568`), in the neural-net mode, using the nonlin matter power spectrum. The NN has been trained in `k \in [1e-2, 5] h/Mpc`. At the large scales (low k) this is roughly at the knee of the power spectrum.

This showcases the failed extrapolation:
![fig1](https://user-images.githubusercontent.com/22447385/118381809-894ce680-b5f7-11eb-8c63-c5752c8500b6.png)

Unfortunately, changing the order of the extrapolation from 1 to 2 or 3 doesn't do much, mainly because of this NN-related scatter. This can be more easily demonstrated using the derivative of P(k). In the figures below, extrapolation should aim for the emu curve to approach the CCL curve.
1) Here's the linear extrapolation:
![fig2](https://user-images.githubusercontent.com/22447385/118381921-a2a26280-b5f8-11eb-9674-ef321ed9ec3c.png)
2) The quadratic (effectively gives the same result as the linear):
![fig3](https://user-images.githubusercontent.com/22447385/118381930-b9e15000-b5f8-11eb-8574-68f728326481.png)
3) The cubic (well...):
![fig4](https://user-images.githubusercontent.com/22447385/118381952-cc5b8980-b5f8-11eb-9904-6151f48cae65.png)

In all cases, CCL fails to correctly predict the power spectrum even slightly outside the input ranges.

nikfilippas:
Possible solution: instead of only taking into account the n points at the edges for the n-th order extrapolation, we could try a `savgol_filter` or something similar near the edges so the scatter is smoothed out.

[...]

PR #951, opened 2022-07-06
Extrapolation in log space for non-log Pk2Ds

### LINKED ISSUE #816: Improve extrapolation in Pk2D
Extend the functionality built-in to Pk2D to allow for power-law (or similar) extrapolation when is_logp = False. This will prevent differences in Pk that comes from nl_pt vs directly from the main cosmology object, and it will provide a physically reasonable extrapolation for PT calculations. This closes #773, since it will solve the underlying issue.

Something like this:

if is_logp = True
continue as normal
(could add an option for all neg values, i.e. log(-P) was passed)

if is_logp = False
if log_ext = False(new option)
... original behavior
if log_ext = True
- use the last two values on both ends (or three, depending on desired order of taylor expansion. or use more if we are worried about noise)
- confirm all positive or negative
- take log of P (or -P)
- perform 1st or 2nd order extrapolation
- (could also use the original Pk2D extension for all pos or all neg values, but we might want to allow a purely flexible power law extension without assuming n_s at low k)

damonge:
Closes #816

### DIFF
--- a/benchmarks/ccl_test_f2d.c
+++ b/benchmarks/ccl_test_f2d.c
@@ -89,7 +89,7 @@ CTEST2(f2d,constant) {
 		    2,
 		    2,
 		    ccl_f2d_constantgrowth,
-		    1,0,2,
+		    1,0,0,2,
 		    ccl_f2d_3,
 		    &status);
   ASSERT_TRUE(status==0);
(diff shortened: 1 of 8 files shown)


nikfilippas:
@damonge let me know once this is done so I can test it using the failed example I showcased in #876 .

damonge:
This PR is very much behind the current master, so I will close it. The code within the branch could be valuable though, so I will not delete it for now.

PR #952, opened 2022-07-14
Action to publish on pypi

damonge:
Follow-up to #948 after checking that it works on test.pypi


joezuntz (approved):
I'd probably delete the old commented out test bit, but looks good.

damonge:
OK, awesome! I thought I'd leave it in case we want to use test.pypi for checks in the future, but that should be straightforward.
Thanks, I'll merge when tests pass.
Thanks Joe for doing this!

PR #957, opened 2022-07-22
Check for range of background splines when using halofit with wa != 0.

tilmantroester:
This issue recently came up when people tried using the CCL halofit with wa != 0 cosmologies in calculator mode and caused unhelpful error messages. Specifically, halofit translates (w0, wa) to a w0_eff. This requires computing the comoving distance to the CMB, which requires the background splines being sampled to sufficiently high redshifts. This PR adds a check for this case and gives a more informative error message.

### DIFF
--- a/pyccl/pk2d.py
+++ b/pyccl/pk2d.py
@@ -174,6 +174,7 @@ class Pk2D(object):
             # HALOFIT translates (w0, wa) to a w0_eff. This requires computing
             # the comoving distance to the CMB, which requires the background
             # splines being sampled to sufficiently high redshifts.
+            cosmo.compute_distances()
             _, a = _get_spline1d_arrays(cosmo.cosmo.data.achi)
             if min(a) > 1/(1 + 3000):
                 raise CCLError("Comoving distance spline does not cover "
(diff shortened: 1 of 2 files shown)


pyccl/pk2d.py
```diff
@@ -170,6 +170,20 @@ def apply_halofit(Pk2D, cosmo, pk_linear):
             pk_linear (:class:`Pk2D`): a :class:`Pk2D` object containing
                 the linear power spectrum to transform.
         """
+        if cosmo["wa"] != 0:
+            # HALOFIT translates (w0, wa) to a w0_eff. This requires computing
+            # the comoving distance to the CMB, which requires the background
+            # splines being sampled to sufficiently high redshifts.
+            _, a = _get_spline1d_arrays(cosmo.cosmo.data.achi)
```
damonge, line 177: I think there's no guarantee that cosmo will have computed distances yet. Can you add a `cosmo.compute_distances()` above this?

damonge (changes requested):
One potential bug

pyccl/pk2d.py
tilmantroester: Right! Added.

damonge (approved):
LGTM!

PR #958, opened 2022-07-23
Add analytic _projected, _cumul2d and _fourier to Hernquist profile

hsinfan1996:
The expressions are given by WolframAlpha and are tested with CLMM's example notebook `examples/Paper_v1.0/validation_tests.ipynb` on the [CLMM repo](https://github.com/LSSTDESC/CLMM).


hsinfan1996:
_Fourier can also be analytic, but it will only make the precision of the projected densities worse.

damonge:
Thanks @hsinfan1996 
Not sure I understand? Why would implementing `_fourier_analytic` affect the precision of the projected profiles?

hsinfan1996:
> Thanks @hsinfan1996 Not sure I understand? Why would implementing `_fourier_analytic` affect the precision of the projected profiles?

If _projected and _cumul2d are not available, aren't they calculated by FFT?

damonge:
@hsinfan1996 just checking: I assume you're still working on this one, right? If not, we do need more tests (coverage decreased significantly)

hsinfan1996:
> @hsinfan1996 just checking: I assume you're still working on this one, right? If not, we do need more tests (coverage decreased significantly)

Tests are added.

pyccl/halos/profiles.py
```diff
@@ -919,21 +919,48 @@ class HaloProfileHernquist(HaloProfile):
...
             truncated at :math:`r = R_\\Delta` (i.e. zero at larger
             radii.
     """
     name = 'Hernquist'
 
-    def __init__(self, c_M_relation, truncated=True):
+    def __init__(self, c_M_relation,
+                 fourier_analytic=False,
+                 projected_analytic=False,
+                 cumul2d_analytic=False,
```
damonge, line 940: Can you preserve the order of `truncated`? otherwise, if someone had a code that did
```
HaloProfileHernquist(cM, False)
```
it will silently not do what they're thinking! (thus breaking the API)

pyccl/halos/profiles.py
```diff
@@ -919,21 +919,48 @@ class HaloProfileHernquist(HaloProfile):
...
+        if cumul2d_analytic:
+            if truncated:
+                raise ValueError("Analytic cumuative 2d profile not supported "
+                                 "for truncated Hernquist. Set `truncated` or "
+                                 "`cumul2d_analytic` to `False`.")
+            self._cumul2d = self._cumul2d_analytic
         super(HaloProfileHernquist, self).__init__()
         self.update_precision_fftlog(padding_hi_fftlog=1E2,
-                                     padding_lo_fftlog=1E-2,
+                                     padding_lo_fftlog=1E-4,
```
damonge, line 963: why was this necessary?

damonge (changes requested):
Thanks @hsinfan1996 . Just one comment about API breaking

pyccl/halos/profiles.py
```diff
@@ -919,21 +919,48 @@ class HaloProfileHernquist(HaloProfile):
...
         if not isinstance(c_M_relation, Concentration):
             raise TypeError("c_M_relation must be of type `Concentration`)")
 
         self.cM = c_M_relation
         self.truncated = truncated
+        if fourier_analytic:
+            self._fourier = self._fourier_analytic
+        if projected_analytic:
+            if truncated:
+                raise ValueError("Analytic projected profile not supported "
```
damonge, line 951: One more thing, can you add a test that this error gets thrown when trying to use it with truncated?

pyccl/halos/profiles.py
```diff
@@ -919,21 +919,48 @@ class HaloProfileHernquist(HaloProfile):
...
+                                 "for truncated Hernquist. Set `truncated` or "
+                                 "`projected_analytic` to `False`.")
+            self._projected = self._projected_analytic
+        if cumul2d_analytic:
+            if truncated:
+                raise ValueError("Analytic cumuative 2d profile not supported "
```
damonge, line 957: and this one

damonge (changes requested):
And an afterthought

hsinfan1996:
@damonge Could you wait for this PR before tagging a new release of CCL? We want to use it in CLMM but some people cannot install CCL from the source.

damonge:
yes, I was indeed waiting for it. No worries.

pyccl/halos/profiles.py
hsinfan1996: To get line 469 of `pyccl/tests/test_profiles.py` pass with a relative tolerance of 5E-3. Otherwise, the relative tolerance would need to be 0.12.

pyccl/halos/profiles.py
hsinfan1996: Done.

pyccl/halos/profiles.py
hsinfan1996: Done. The same test is also added for NFW.

pyccl/halos/profiles.py
hsinfan1996: I see. Fixed.

damonge (approved):
LGTM!

damonge:
(feel free to press the green button yourself @hsinfan1996 !)

hsinfan1996:
@damonge should I choose squash and merge?

damonge:
Always squash and merge, yes. No worries.

PR #963, opened 2022-08-01
Bug in magnification bias

### LINKED ISSUE #962: Unexpected change in mag bias behaviour
It seems between v2.3 and v2.4 the behaviour of mag bias kernels changed. In particular, I would expect a value of `s = 0.4` to return a kernel of 0 and this was the case in v <= 2.3

For v2.4 (and now v2.5) it seems to be `s = 0` which returns a kernel of 0.

Attached is a minimal example, which gives:

```
2.3.0
mag bias kernel at s=0.0 are not zero!
mag bias kernel at s=0.4 all zero
```

```
2.4.0
mag bias kernel at s=0.0 all zero
mag bias kernel at s=0.4 are not zero!
```

```
2.5.0
mag bias kernel at s=0.0 all zero
mag bias kernel at s=0.4 are not zero!
```
[ccl_mag_bug.py.txt](https://github.com/LSSTDESC/CCL/files/9234464/ccl_mag_bug.py.txt)

damonge:
Closes #962

### DIFF
--- a/pyccl/tests/test_tracers.py
+++ b/pyccl/tests/test_tracers.py
@@ -41,6 +41,38 @@ def get_tracer(tracer_type, cosmo=None, **tracer_kwargs):
     return tr, ntr
 
 
+def test_tracer_mag_0p4():
+    z = np.linspace(0., 1., 2000)
+    n = dndz(z)
+    # Small bias so the magnification contribution
+    # is significant (if there at all)
+    b = np.ones_like(z)*0.1
+    s_no = np.ones_like(z)*0.4
+    s_yes = np.zeros_like(z)
+    # Tracer with no magnification by construction
+    t1 = ccl.NumberCountsTracer(COSMO, True,
+                                dndz=(z, n),
+                                bias=(z, b))
+    # Tracer with s=0.4
+    t2 = ccl.NumberCountsTracer(COSMO, True,
+                                dndz=(z, n),
+                                bias=(z, b),
+                                mag_bias=(z, s_no))
+    # Tracer with magnification
+    t3 = ccl.NumberCountsTracer(COSMO, True,
+                                dndz=(z, n),
+                                bias=(z, b),
+                                mag_bias=(z, s_yes))
+    ls = np.array([2, 200, 2000])
+    cl1 = ccl.angular_cl(COSMO, t1, t1, ls)
+    cl2 = ccl.angular_cl(COSMO, t2, t2, ls)
+    cl3 = ccl.angular_cl(COSMO, t3, t3, ls)
+    # Check cl1 == cl2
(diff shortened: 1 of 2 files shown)


pyccl/tests/test_tracers.py
```diff
@@ -41,6 +41,38 @@ def get_tracer(tracer_type, cosmo=None, **tracer_kwargs):
     return tr, ntr
 
 
+def test_tracer_mag_0p4():
+    z = np.linspace(0., 1., 2000)
+    n = dndz(z)
+    # Small bias so the magnification contribution
+    # is significant (if there at all)
+    b = np.ones_like(z)*0.1
+    s_no = np.ones_like(z)*0.4
```
tilmantroester, line 50: I thought no magnification was s=0. Am I missing something?

src/ccl_tracers.c
```diff
@@ -350,31 +350,31 @@ static void integrate_lensing_kernel_spline(ccl_cosmology *cosmo,
...
 
   if(*status == 0) {
-    // Fill chi and sz arrays
+    // Fill chi and qz arrays
     for(int i=0; i<nz; i++) {
       double a = 1./(1+z_arr[i]);
       chi_of_z_array[i] = ccl_comoving_radial_distance(cosmo, a, status);
       if(sz_f != NULL) {
-        sz_array[i] = ccl_f1d_t_eval(sz_f, z_arr[i]);
+        qz_array[i] = 1-2.5*ccl_f1d_t_eval(sz_f, z_arr[i]);
```
tilmantroester, line 367: Ah, that's embarrassing. No idea how that got missed...

pyccl/tests/test_tracers.py
damonge: no magnification is s=2/5, I'm afraid :-)

tilmantroester: Ah, right. Might want to update the docstring here then: https://github.com/LSSTDESC/CCL/blob/master/pyccl/tracers.py#L94

damonge: Hmm, but this is correct, right? If s is `None`, then it assumes s=0 (i.e. q=1) and returns the usual weak lensing kernel.

itrharrison (approved):
LGTM!

FYI we noticed this when some tests in soliket started failing ;-)

damonge:
That's good to know. If you have an independent benchmark for magnification, we've been looking for one for a while!

pyccl/tests/test_tracers.py
tilmantroester: You're right. I was confused by the `mag_bias` argument in `NumberCountsTracer`, where `mag_bias=None` effectively gives q=0, and `mag_bias` in `get_lensing_kernel`, where `mag_bias=None` gives q=1.

PR #967, opened 2022-08-03
Fixed new flake8 errors

damonge:


hsinfan1996 (approved):

hsinfan1996:
Very trivial, but it seems that the script does not accept tabs there. Need to add spaces all the way to align with the previous line.

PR #968, opened 2022-08-04
Hashing

nikfilippas:
This PR only contains the new functionality that is relevant to hashing, and no caching. See #933 for more details.
Hashing is essential for creating representations of objects in CCL and to be able to compare them.

Note: To properly implement `__repr__`,  #934 needs to go through, because there, I make use of the central controls for all `CCLObjects` to store the `__init__` signatures.

(I have branched from caching, hence the long commit history).


nikfilippas:
@damonge okay, this is the hashing branch. You can have a look here first if that makes it any easier.

pyccl/core.py
```diff
@@ -763,11 +763,8 @@ def compute_growth(self):
         status = lib.cosmology_compute_growth(self.cosmo, status)
         check(status, self)
 
-    def compute_linear_power(self):
-        """Compute the linear power spectrum."""
-        if self.has_linear_power:
-            return
-
+    def _compute_linear_power(self):
+        """Return the linear power spectrum."""
```
damonge, line 767: why do we need these?

pyccl/halos/halo_model.py
```diff
@@ -115,37 +116,35 @@ def _integ_spline(self, fM, lM):
...
-                                         self._lmass)) / self._m0
             self._a_current_mf = a
+            rho0 = cosmo.rho_x(1., "matter", is_comoving=True)
+            self._mf = self._massfunc.get_mass_function(
+                cosmo, self._mass, a, mdef_other=self._mdef)
+            self._mf0 = (rho0 - self._integrator(
+                self._mf*self._mass, self._lmass)) / self._m0
 
         if get_bf:
+            self._a_current_mf = a
```
damonge, line 130: why?

pyccl/halos/halo_model.py
```diff
@@ -115,37 +116,35 @@ def _integ_spline(self, fM, lM):
...
-                self.bf = self._hbias.get_halo_bias(cosmo, self._mass, a,
-                                                    mdef_other=self._mdef)
-                self.mbf0 = (rho0 -
-                             self._integrator(self.mf * self.bf * self._mass,
-                                              self._lmass)) / self._m0
-            self._a_current_bf = a
+                self._bf = self._hbias.get_halo_bias(
+                    cosmo, self._mass, a, mdef_other=self._mdef)
+                self._bf0 = (rho0 - self._integrator(
+                    self._mf*self._bf*self._mass, self._lmass)) / self._m0
```
damonge, line 137: why did you change the name from `mbf0` to `bf0`? Note that this is meant to represent b*mass_function, not b

pyccl/halos/halo_model.py
```diff
@@ -115,37 +116,35 @@ def _integ_spline(self, fM, lM):
...
+            self._a_current_mf = a
             if a != self._a_current_bf:
                 if rho0 is None:
                     rho0 = cosmo.rho_x(1., "matter", is_comoving=True)
-                self.bf = self._hbias.get_halo_bias(cosmo, self._mass, a,
-                                                    mdef_other=self._mdef)
-                self.mbf0 = (rho0 -
-                             self._integrator(self.mf * self.bf * self._mass,
-                                              self._lmass)) / self._m0
-            self._a_current_bf = a
```
damonge, line 138: aren't you missing the previous line?

pyccl/tests/test_hashing.py
```diff
@@ -0,0 +1,34 @@
...
+
+OBJECTS = [ccl.Cosmology,  # class
+           (0, 1, 2),      # tuple
+           [0, 1, 2],      # list
+           set([0, 1, 2]),   # set
+           np.arange(3),   # array
+           {0: None, 1: None, 2: None},                    # dict
+           {0: None, 1: None, 2: {2.1: None, 2.2: None}},  # nested dict
+           OrderedDict({0: None, 1: None, 2: None}),       # OrderedDict
+           ccl.CosmologyVanillaLCDM(),  # something else
```
damonge, line 15: humm, can you walk me through how this hashes a cosmology?

damonge (changes requested):
OK, see comments. So this isn't really doing much hashing, right? I.e. this only implements the possibility of hashing things.

How long do we need to wait (i.e. which PRs do we need to go through) before we get there?

Since we haven't agreed that caching is a good idea, I want to make sure we can keep things separate.

pyccl/core.py
nikfilippas: oops, leftover from the other branch, have fixed it now.

pyccl/halos/halo_model.py
nikfilippas: yes I meant to put it under the if-statement, good that you noticed it

pyccl/halos/halo_model.py
nikfilippas: I changed it back.

pyccl/halos/halo_model.py
nikfilippas: If you meant the scale factor change, I put it where it belongs now.

pyccl/tests/test_hashing.py
nikfilippas: Sure, so it falls under the "something else" category of the `hash_` function, i.e. it can't be reliably hashed by that function. So, it simply calls the `__hash__` method of this object:
```
    elif hasattr(obj, "__hash__"):
        # Hashables: Just return the object.
        return obj
```

The idea is that we use the representation of the object for hashing, since, representations should strictly be unique, as per Python guidelines. And we simply hash the representation string.

nikfilippas: Now, in the `CCLObject` PR, I've created a module `_repr.py` which contains repr methods for all objects in CCL. You can have a look there to see specifically how I'm thinking of implementing hashing for `Cosmology` objects, but this is up to discussion.

nikfilippas:
@damonge first pass done

pyccl/core.py
```diff
@@ -914,7 +914,7 @@ def compute_nonlin_power(self):
 
         if mps == "camb" and self._has_pk_nl:
             # Already computed
-            return
+            return self._pk_nl['delta_matter:delta_matter']
```
damonge, line 917: This function isn't meant to return anything.

pyccl/halos/halo_model.py
```diff
@@ -115,37 +116,35 @@ def _integ_spline(self, fM, lM):
...
             self._a_current_mf = a
+            rho0 = cosmo.rho_x(1., "matter", is_comoving=True)
+            self._mf = self._massfunc.get_mass_function(
+                cosmo, self._mass, a, mdef_other=self._mdef)
+            self._mf0 = (rho0 - self._integrator(
+                self._mf*self._mass, self._lmass)) / self._m0
 
         if get_bf:
             if a != self._a_current_bf:
+                self._a_current_mf = a
```
damonge, line 131: this should be `self._a_current_bf`.

pyccl/base.py
```diff
@@ -0,0 +1,47 @@
...
+        if isinstance(obj, np.ndarray):
+            # Numpy arrays: Convert the data buffer to a byte string.
+            return obj.tobytes()
+
+        elif isinstance(obj, dict):
+            # Dictionaries: Build a tuple from key-value pairs,
+            # where all values are converted to hashables.
+            out = dict.fromkeys(obj)
+            for key, value in obj.items():
+                out[key] = _to_hashable(value)
```
damonge, line 25: This seems a bit verbose. Couldn't you do `out = {key: _to_hashable(value) for key_value in obj.items()}`?

pyccl/base.py
```diff
@@ -0,0 +1,47 @@
...
+            # Sort unordered dictionaries for hash consistency.
+            if isinstance(obj, OrderedDict):
+                return tuple(obj.items())
```
damonge, line 28: Wait, shouldn't you be using `out` here?

pyccl/base.py
```diff
@@ -0,0 +1,47 @@
...
+            return tuple(sorted(obj.items()))
```
damonge, line 29: same here, shouldn't this be `out`?

pyccl/base.py
```diff
@@ -0,0 +1,47 @@
...
+
+        else:
+            # Iterables: Build a tuple from values converted to hashables.
+            out = [_to_hashable(item) for item in obj]
+            return tuple(out)
+
+    elif hasattr(obj, "__hash__"):
```
damonge, line 36: Why do we not want to just do this for iterables? Or in other words, why don't we do this before checking for iterables?

damonge (changes requested):
@nikfilippas let's try to wrap this one up!

pyccl/core.py
nikfilippas: Fixed in merging.

pyccl/halos/halo_model.py
nikfilippas: I think this has changed with #983 so should be alright now, check.

pyccl/base.py
nikfilippas: Sure why not.

pyccl/base.py
nikfilippas: Correct. Fixed.

pyccl/base.py
nikfilippas: Fixed.

pyccl/base.py
nikfilippas: Not all iterables are hashable. Iterables which are mutable (e.g. lists) aren't. We have to safely convert them to an immutable first.

nikfilippas: Why we don't do it first. Consider the following example: `tuple([[1, 2, 3], [4, 5]]). This is a tuple (immutable, hashable), but it contains lists. If we try to call `__hash__` on it, we get the following error: `TypeError: unhashable type: 'list'`.

nikfilippas: So we first need to iterate through all iterable items, in all levels of depth, and convert everything that's mutable, to an immutable.

nikfilippas: The function will convert the example I provided to:
```
In[10]: ccl.base._to_hashable(tuple([[1, 2, 3], [4, 5]]))
Out[10]: ((1, 2, 3), (4, 5))
```
which can be safely hashed.

damonge: but my point is, if whatever the object is has `__hash__`, then we should just use that, correct? If so, why not check that first. Or are you trying to replace the hash of iterables by your own hash?

pyccl/core.py
```diff
@@ -913,7 +913,8 @@ def compute_nonlin_power(self):
         if (mps != 'emu') and (mps is not None):
             self.compute_linear_power()
 
-        if mps == "camb":
+        if mps == "camb" and self._has_pk_nl:
+            # Already computed
```
damonge, line 917: I think this is gonna make the tests fail.

pyccl/base.py
```diff
@@ -0,0 +1,46 @@
...
+        # Encapsulate all the iterables to quickly discard as needed.
+
+        if isinstance(obj, np.ndarray):
+            # Numpy arrays: Convert the data buffer to a byte string.
+            return obj.tobytes()
+
+        elif isinstance(obj, dict):
+            # Dictionaries: Build a tuple from key-value pairs,
+            # where all values are converted to hashables.
+            out = dict.fromkeys(obj)
```
damonge, line 23: you don't need this line

damonge (changes requested):

pyccl/base.py
nikfilippas: It is normal to overload `__hash__` if the built-in implementation does not cover all cases that the developers need. In our case, the default `__hash__` wouldn't work.

damonge: ok, that's what I wanted to know

pyccl/base.py
nikfilippas: OK

pyccl/core.py
nikfilippas: Fixed.

damonge (approved):
Alright, let's wait for the tests to pass and we can merge

PR #969, opened 2022-08-04
`HaloProfileGNFW` bugfix for `c500` update.

nikfilippas:
The integration interpolation template should also be recalculated when `c500` changes.

### DIFF
--- a/pyccl/halos/profiles.py
+++ b/pyccl/halos/profiles.py
@@ -1107,15 +1107,17 @@ class HaloProfilePressureGNFW(HaloProfile):
     The parametrization is:
 
     .. math::
+
        P_e(r) = C\\times P_0 h_{70}^E (c_{500} x)^{-\\gamma}
        [1+(c_{500}x)^\\alpha]^{(\\gamma-\\beta)/\\alpha},
 
     where
 
     .. math::
+
        C = 1.65\\,h_{70}^2\\left(\\frac{H(z)}{H_0}\\right)^{8/3}
        \\left[\\frac{h_{70}\\tilde{M}_{500}}
-       {3\\times10^{14}\\,M_\\odot}\\right]^{2/3+0.12},
+       {3\\times10^{14}\\,M_\\odot}\\right]^{2/3+\\alpha_{\\mathrm{P}}},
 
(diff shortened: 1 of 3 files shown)


damonge:
It actually sucks if we do it this way, because we actually don't really need to. You could instead just change `R->R/c500` and remove the c500 dependence in `_form_factor`.

nikfilippas:
@damonge how much would the `qrange` need to change?

damonge:
I bet the qrange you currently have is fine, since c500 is an O(1) quantity

nikfilippas:
Okay, I absorbed the c500 in the form factor, I think it's good to go now.

damonge:
How about the `_fourier` method? I'm surprised the benchmarks are passing...

nikfilippas:
It looks like somehow, the integral used in the Fourier method cancels out `c500`. The proof, however, is left to the interested reader.

damonge:
No, we need to understand this, I'm afraid

pyccl/halos/profiles.py
```diff
@@ -1176,25 +1176,31 @@ def __init__(self, mass_bias=0.8, P0=6.41,
...
+            Concentration parameter.
+        alpha, beta, gamma : float
+            Profile shape parameter.
+        alpha_P : float
+            Additional mass-dependence exponent.
+        P0_hexp : float
+            Power of ``h`` with which the normalization scales.
+            SZ-based normalizations: -1. X-ray-based normalizations: -3/2.
+        x_out : float
+            Profile threshold (as a fraction of r500c).
```
damonge, line 1203: Note that, since `x_out` is in units of R500, you need to update the integration limits now that you've taken c500 out.

damonge: Which means that you probably need to update the fourier transform if changing c500 anyway...

damonge:
so perhaps just undo all the changes after my first comment :-|

damonge (approved):
LGTM. Merge if the tests pass.

nikfilippas:
@damonge I think you need to approve again.

damonge (approved):

PR #971, opened 2022-08-05
Useful error for transfer functions with non-monotonic scale factors

### LINKED ISSUE #823: Generic Cell integration failure with generalized tracers
Here's a case of generic failure of Cell integation for pretty standard inputs. It would be great if we could investigate what drives this. @vitenti @damonge 

[Failed_integr_Cl.pdf](https://github.com/LSSTDESC/CCL/files/5309776/Failed_integr_Cl.pdf)

damonge:
Closes #823 

This is actually documented, but at least now we'll throw a more useful error.


nikfilippas:
LGTM

nikfilippas (approved):
LGTM!

PR #974, opened 2022-08-05
Better docs for correlations

### LINKED ISSUE #826: Correlation function calculation at small angular separation
Hi, I'm working on a project which involves measuring the dark matter distribution in galaxies using data from the HSC survey. My advisors, Professor Rachel Mandelbaum, and Simon Samuroff suggested that I perform a preliminary fitting on the lensing and clustering correlation functions using CCL and I ran into a problem generating these functions at angular separation scales around 10^-2 degree. 

Following [this](https://github.com/LSSTDESC/CCLX/blob/master/CellsCorrelations.ipynb) notebook, if I change `theta_deg` under In[8] to `theta_deg = np.logspace(-3, np.log10(5.), 40)`, I get the following plots for the lensing and clustering correlation functions:
lensing
clustering

Are there any changes that I could make to improve the accuracy of correlation calculation at angular separation scales around 10^-2 deg? Thank you!

damonge:
We need to warn users that calculating correlations may depend on the values of the spline parameters (particularly if they're going down to inordinately small scales). This adds a note on this in the documentation.

Closes #826

### DIFF
--- a/pyccl/correlations.py
+++ b/pyccl/correlations.py
@@ -31,6 +31,21 @@ def correlation(cosmo, ell, C_ell, theta, type='NN', corr_type=None,
                 method='fftlog'):
     """Compute the angular correlation function.
 
+    .. note::
+
+        The accuracy of the Hankel transform depends on the range of
+        multipoles over which the power spectra are integrated. If you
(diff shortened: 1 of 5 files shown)


nikfilippas (approved):
lgtm

Issue #949, opened 2022-07-01
setuptools_scm requires python3.7+


carlosggarcia:
I have found that GitHub actions fail (in a different repo) when you do the tests with python3.6 and you try to install pyccl (both with `pip install pyccl` or `pip install git+https://github.com/LSSTDESC/CCL`). The error comes from `setuptools_scm` requiring python3.7+ (https://pypi.org/project/setuptools-scm/). I haven't been able to reproduce this in my laptop with a fresh conda environment, though.

```
Collecting git+https://github.com/LSSTDESC/CCL
  Cloning https://github.com/LSSTDESC/CCL to /tmp/pip-req-build-qwscu05t
  Running command git clone --filter=blob:none -q https://github.com/LSSTDESC/CCL /tmp/pip-req-build-qwscu05t
  Resolved https://github.com/LSSTDESC/CCL to commit d5408a777ab5fc4b2ae6fada2114a168b9bacee4
  Preparing metadata (setup.py): started
  Preparing metadata (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /opt/hostedtoolcache/Python/3.6.15/x64/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-qwscu05t/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-qwscu05t/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-7eabotjw
       cwd: /tmp/pip-req-build-qwscu05t/
  Complete output (24 lines):
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/pip-req-build-qwscu05t/setup.py", line 108, in <module>
      'Topic :: Scientific/Engineering :: Physics'
    File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/site-packages/setuptools/__init__.py", line 143, in setup
      return distutils.core.setup(**attrs)
    File "/opt/hostedtoolcache/Python/3.6.15/x64/lib/python3.6/distutils/core.py", line 108, in setup
[...]

PR #975, opened 2022-08-05
Stop supporting Python 3.6

nikfilippas:
As mentioned in #949 , `setuptools` won't work anymore because it requires py37+.
Python 3.6 is past its end of life so with this PR we stop supporting it and require that py37+ is used, in order to install CCL.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -92,7 +92,7 @@ setup(
     include_package_data=True,
     use_scm_version=True,
     distclass=Distribution,
-    python_requires='>3.7.0',
+    python_requires='>=3.7.0',
     setup_requires=['setuptools_scm'],
     install_requires=['numpy', 'pyyaml'],
     cmdclass={'build_py': Build, 'develop': Develop, 'clean': Clean},
--- a/setup.py
+++ b/setup.py
@@ -92,6 +92,7 @@ setup(
     include_package_data=True,
     use_scm_version=True,
     distclass=Distribution,
+    python_requires='>3.7.0',
     setup_requires=['setuptools_scm'],
     install_requires=['numpy', 'pyyaml'],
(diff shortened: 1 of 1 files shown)


damonge:
Yeah, we may want to hear people out a bit before we actually do this.

nikfilippas:
CCL won't install if you don't have py37+ anyway.

damonge:
ah, right. OK, let's do it then.

hsinfan1996:
Using `setuptools_scm` v6.x solves #949 I think. And I can install CCL from source with python 3.6.13 in a conda environment. Is there anything wrong with my setup?

hsinfan1996:
I actually don't mind bumping up the minimum python version. Just don't think #949 is an unsolvable issue.

nikfilippas:
`setuptools_scm=7.0.0` was released on June 21st this year. Earlier versions required `py36+` but v7 requires `py37+`.
The problem is that if users have updated their `setuptools` installation they won't be able to install CCL. I'm sure we can patch this, but we'd need to check which `setuptools` version every user has, and I think this is too much of a hassle for little benefit since `py36` has reached its end of life a long time ago.

nikfilippas:
An additional hurdle is that `setuptools` is the very first module loaded when installing CCL, so I think there isn't an easy way to force-install the older version in case the user has `setuptools v7+` installed. We list the package requirements *with* that very package!

setup.py
```diff
@@ -92,6 +92,7 @@ def run(self):
     include_package_data=True,
     use_scm_version=True,
     distclass=Distribution,
+    python_requires='>3.7.0',
```
hsinfan1996, line 95: Do you need '>=3.7.0' instead of '>3.7.0' here?

nikfilippas: yes! thank you!

damonge (approved):

PR #979, opened 2022-08-06
Compute non-linear power spectrum for `(trf, mps) == (not camb, camb)` combinations.

### LINKED ISSUE #978: `pk_nl` not computed with `mps == "camb"` and `tf != "boltzmann_camb"`
Slightly odd usage, but if we want to get the camb non-linear power spectrum with a non-camb transfer function, the non-linear matter power spectrum never computes.
``` python
cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.67, n_s=0.96, A_s=2.34e-9,
                      transfer_function="boltzmann_class",
                      matter_power_spectrum="camb")
cosmo.compute_nonlin_power()  # doesn't complain
assert cosmo.get_nonlin_power() is None
```

nikfilippas:
Closes #978 . 

That was a very peculiar issue, and it has never popped up before because the only combination of `transfer_function` and `matter_power_spectrum` that would result in the non-linear power spectrum not being calculated would be `boltzmann_class` and `camb`. I guess no one has ever wanted to get the linear power spectrum through `CLASS`, but then query `CAMB` for the non-linear model. 

Seeing as in the future we might want to add other options for the linear power spectra, this ensures that the error won't show.

### DIFF
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -813,7 +813,7 @@ class Cosmology(object):
             if abs(self["mu_0"]) > 1e-14:
                 warnings.warn("CAMB doesn't compute non-linear power spectra "
                               "consistently with mu_0 > 0.", CCLWarning)
-            if self["A_s"] is None:
+            if np.isnan(self["A_s"]):
                 raise CCLError("CAMB doesn't rescale non-linear power spectra "
                                "consistently without A_s.")
 
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -388,6 +388,10 @@ class Cosmology(object):
                              "method. Available options are: %s"
                              % (emulator_neutrinos,
                                 emulator_neutrinos_types.keys()))
+        if (matter_power_spectrum == "camb"
+                and transfer_function != "boltzmann_camb"):
+            raise CCLError("To compute the matter power spectrum with CAMB "
+                           "the transfer function should be 'boltzmann_camb'.")
 
         # Assign values to new ccl_configuration object
         config = lib.configuration()
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -818,7 +818,7 @@ class Cosmology(object):
             if abs(self["mu_0"]) > 1e-14:
                 warnings.warn("CAMB doesn't compute non-linear power spectra "
                               "consistently with mu_0 > 0.", CCLWarning)
-            if np.isnan(self["A_s"]):
+            if not np.isfinite(self["A_s"]):
                 raise CCLError("CAMB doesn't rescale non-linear power spectra "
                                "consistently without A_s.")
 
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -909,7 +909,7 @@ class Cosmology(object):
             self.compute_linear_power()
 
         if mps == "camb":
-                return
+            return
 
         if mps is None:
             raise CCLError("You want to compute the non-linear power "
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -390,8 +390,9 @@ class Cosmology(object):
                                 emulator_neutrinos_types.keys()))
         if (matter_power_spectrum == "camb"
                 and transfer_function != "boltzmann_camb"):
-            raise CCLError("To compute the matter power spectrum with CAMB "
-                           "the transfer function should be 'boltzmann_camb'.")
+            raise CCLError(
+                "To compute the non-linear matter power spectrum with CAMB "
+                "the transfer function should be 'boltzmann_camb'.")
 
         # Assign values to new ccl_configuration object
         config = lib.configuration()
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -252,9 +252,7 @@ class Cosmology(object):
 
         self._build_cosmo()
 
-        self._has_pk_lin = False
         self._pk_lin = {}
-        self._has_pk_nl = False
         self._pk_nl = {}
 
     def _build_cosmo(self):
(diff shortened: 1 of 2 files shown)


nikfilippas:
Incidentally, having this explicitly tested would require running CLASS (linear) + CAMB (linear+nonlinear) which would take ~3 s, so I haven't added any tests, but I ensured it works on my computer.

damonge:
I wouldn't support this. I'd rather throw an error if someone asks for this.

nikfilippas:
I agree that with the choices currently available it is inefficient, but it may crop up in the future it we add more transfer function choices.
Users might want to try one of the linear options but use halofit from camb, for example. Why throw an error if they want to do that? The rationale is making the code futureproof.

Also, with this PR we simply re-write the CAMB option (21 lines of code) in a better fashion. Throwing an error would actually increase the number of lines in the code. This is a simple solution.

damonge:
I prefer throwing an error. Otherwise we're silently allowing an iconsistent and inefficient use of CCL. If someone wants an E&Hu linear Pk with the halofit version of camb, they can use the calculator mode, we are not precluding this.

Precisely because of the calculator mode I do not envisage adding new transfer function models in the future. We came to the conclusion that it was best to limit the integration of boltzmann solvers in CCL.

damonge:
To clarify, I agree this is a bug, and I'm not against the restructuring you've done. I just think it's a bad idea to allow this usage. It will inadvertently lead to inconsistencies or bugs. If users want to experiment, the calculator mode allows them full freedom to do so, while forcing them to understand what they're doing.

Suggest keeping most of the changes you've made, but not allowing this particular combination. It's a simple if statement in the __init__

nikfilippas:
I see. CCL claims a CAMB/CLASS agreement of 1e-5, so I wouldn't say it's inconsistent in this particular case, just bad use. I restructured the code with the idea in mind to make it futureproof. But if it's definitive that there will never be other Boltzmann solvers then we don't need to do that.

I added a check as you suggested, but kept the new structure juuuust in case.

nikfilippas:
As you might have seen, I also cleared up the `has_linear_power` and `has_nonlin_power` properties, which up to now, had to manually be set (which beats the purpose of having properties).

pyccl/core.py
```diff
@@ -390,6 +388,10 @@ def _build_config(
                              "method. Available options are: %s"
                              % (emulator_neutrinos,
                                 emulator_neutrinos_types.keys()))
+        if (matter_power_spectrum == "camb"
+                and transfer_function != "boltzmann_camb"):
+            raise CCLError("To compute the matter power spectrum with CAMB "
+                           "the transfer function should be 'boltzmann_camb'.")
```
damonge, line 394: Can you specify that this is the non-linear matter power spectrum?

pyccl/core.py
```diff
@@ -800,33 +802,33 @@ def compute_linear_power(self):
...
+        # to avoid repeating the code in `compute_nonlin_power`.
+        # Because CAMB power spectra come in pairs with pkl always computed,
+        # we set the nonlin power spectrum first, but keep the linear via a
+        # status variable to use it later if the transfer function is CAMB too.
+        pkl = None
+        if self._config_init_kwargs["matter_power_spectrum"] == "camb":
+            if abs(self["mu_0"]) > 1e-14:
+                warnings.warn("CAMB doesn't compute non-linear power spectra "
+                              "consistently with mu_0 > 0.", CCLWarning)
+            if np.isnan(self["A_s"]):
```
damonge, line 820: Is this equivalent to `not np.isfinite`?

pyccl/core.py
```diff
@@ -800,33 +802,33 @@ def compute_linear_power(self):
...
 
+        # Compute the CAMB nonlin power spectrum if needed,
+        # to avoid repeating the code in `compute_nonlin_power`.
+        # Because CAMB power spectra come in pairs with pkl always computed,
+        # we set the nonlin power spectrum first, but keep the linear via a
+        # status variable to use it later if the transfer function is CAMB too.
+        pkl = None
+        if self._config_init_kwargs["matter_power_spectrum"] == "camb":
+            if abs(self["mu_0"]) > 1e-14:
+                warnings.warn("CAMB doesn't compute non-linear power spectra "
```
damonge, line 818: It seems this is no longer being tested.

damonge (changes requested):
Small things

pyccl/core.py
nikfilippas: It is not. `not np.isfinite(np.nan)` is `True`. `np.isnan(np.nan)` is `True`.
But `not np.isfinite(np.inf)` is `True` and `np.isnan(np.inf)` is `False`.
If not declared, `A_s` is `nan`, so we should be explicitly checking for nans.

pyccl/core.py
nikfilippas: It wasn't tested before either, it just picks it up because I restructured the code. I am a bit reluctant to write a test just to check that the warning is thrown because CAMB takes 1 second to run for the non-linear power spectrum. What do you think?

pyccl/core.py
damonge: Since we set these to NAN at the C level, are we sure that the C NAN is the same as `np.nan` for all architectures or python/numpy versions? I perused a bit but couldn't confirm it.

pyccl/core.py
damonge: ok, yeah fine.

pyccl/core.py
nikfilippas: Done.

pyccl/core.py
nikfilippas: That's a good point, I have no idea. I think this is handled by SWIG. Would SWIG behave differently in different architectures?

nikfilippas: The main question is, if not `np.nan` then what? What is it replaced by? I can't think of anything that would fit in here.

nikfilippas: I don't know... I changed it back to `not np.isfinite` just to be on the safe side. Thanks for pointing that out.

damonge (approved):
LGTM

PR #983, opened 2022-08-16
`HaloModelCalculator` check cosmo in `get_ingredients`.

nikfilippas:
In #925 , I moved the `rho0` calculation inside of `get_ingredients` but forgot to include a check, so that `rho0` is re-computed every time `cosmo` changes. In essence, we didn't fix the problem in 925, we just moved it to a different method.

With this PR we now check for cosmo equality. There is no `__eq__` currently implemented in `Cosmology`, so Python falls back to id-checking (an okay assumption for now).

I moved `rho0` out of the if-statement (so it gets computed every time), or we'd have to carry out checks for every possible combination of (mass_function, halo_bias), (cached a, not cached, or None), (cached cosmo, not cached, or None), which would complicate things a lot.

I simplified the way we compute `rho0`, so that it doesn't call `rho_x` anymore, and so it doesn't have to evaluate any splines. This is now very fast (200x faster than calculating the mass function, 400x faster than calculating mf+hb), so we don't need to worry about the overhead every time `_get_ingredients` is called.

Screenshot from 2022-08-16 19-49-36

### DIFF
--- a/pyccl/halos/halo_model.py
+++ b/pyccl/halos/halo_model.py
@@ -123,13 +123,13 @@ class HMCalculator(object):
             self.mf0 = (rho0 - integ) / self._m0
             self._cosmo_mf, self._a_mf = cosmo, a  # cache
 
-    def _get_halo_bias(self, cosmo, a, ρ0):
+    def _get_halo_bias(self, cosmo, a, rho0):
         # Compute the halo bias at this cosmo and a.
         if cosmo != self._cosmo_bf or a != self._a_bf:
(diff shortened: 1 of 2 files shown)


damonge (changes requested):
Looks good. See couple of minor comments

damonge (approved):
I'm mostly ok with merging this one, but you need to decide if you'd like to do this before or after the hashing branch (since both have touched similar parts of the code

damonge:
Actually, I see this is a lot more complete than what's in #968, so will merge this and then update that other PR.

PR #984, opened 2022-08-19
GitHub Actions: Fix class installation

hsinfan1996:
libomp was not found. Add one more LDFLAGS to fix the problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
       - main
       - master
       - releases/*
+      - 'github_actions'
   pull_request: null
 
 jobs:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         os:
           - macos-11
-          - ubuntu-latest
+          #- ubuntu-latest
         py:
           - 3.8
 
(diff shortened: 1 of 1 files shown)


hsinfan1996:
libomp was not found. Add one more LDFLAGS to fix the problem.

.github/workflows/ci.yml
```diff
@@ -78,17 +78,15 @@ jobs:
 
       - name: install class
         run: |
-          export LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
           if [[ ${MATRIX_OS} == "macos-11" ]]; then
+            export LDFLAGS=-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib
+            LDFLAGS+=" -L/Users/runner/miniconda3/envs/test/lib"
```
damonge, line 83: Don't you need to do
```
export LDFLAGS+=" -L/Users/runner/miniconda3/envs/test/lib"
```

hsinfan1996: I also tried with zsh and I got error when doing `export LDFLAGS+=" -L/Users/runner/miniconda3/envs/test/lib"`

hsinfan1996: interestingly, it is okay with bash.

hsinfan1996: I am using something that works for both shells.

damonge: Interesting! OK, as long as it works I'm happy.

hsinfan1996:
Look like it is working.

damonge (approved):
Thanks a lot. LGTM!

PR #987, opened 2022-08-24
Add option to make einasto alpha a free parameter

hsinfan1996:
The Einasto alpha parameter can now be assigned by users.

### DIFF
--- a/pyccl/halos/profiles.py
+++ b/pyccl/halos/profiles.py
@@ -842,12 +842,13 @@ class HaloProfileEinasto(HaloProfile):
     """
     name = 'Einasto'
 
-    def __init__(self, c_M_relation, truncated=True):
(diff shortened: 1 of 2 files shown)


damonge:
Great. Can we give it an `update_parameters` method, the same way we do e.g. [here](https://github.com/LSSTDESC/CCL/blob/17fd6423a67957b5075102b905e106b81623bedc/pyccl/halos/profiles.py#L1184)?

hsinfan1996:
> Great. Can we give it an `update_parameters` method, the same way we do e.g. [here](https://github.com/LSSTDESC/CCL/blob/17fd6423a67957b5075102b905e106b81623bedc/pyccl/halos/profiles.py#L1184)?

Sure.

damonge (approved):
LGTM!

PR #989, opened 2022-08-24
Fix free Einasto alpha dimension

hsinfan1996:


pyccl/halos/profiles.py
```diff
@@ -871,15 +874,14 @@ def _get_cM(self, cosmo, M, a, mdef=None):
...
+        if self.alpha == 'cosmo':
             mdef_vir = MassDef('vir', 'matter')
             Mvir = mdef.translate_mass(cosmo, M, a, mdef_vir)
-            print(cosmo, M, a, mdef_vir)
             sM = sigmaM(cosmo, Mvir, a)
             nu = 1.686 / sM
             alpha = 0.155 + 0.0095 * nu * nu
         else:
-            alpha = self.alpha
+            alpha = np.array([self.alpha])
```
nikfilippas, line 884: This will give alpha an extra array dimension no matter what the dimensionality of alpha is. I think what we actually want here is `np.atleast_1d`.

hsinfan1996: Fixed

hsinfan1996: Just found out alpha should have the same shape as `M`.

damonge: How about allowing the possibility for `alpha` to be a function?

damonge: Yes, assuming that's what you guys need?

hsinfan1996: Sorry, I'm a little bit confused. Right now `_get_alpha` is a function of `cosmo`, `M`, `a` and `mdef`. If we want to calculate alpha from cosmology, then all four variables will be used. But if we want to set alpha to our own value, then only the shape of `M` is used. Do you mean defining a new function `def alpha`... to deal with the two cases?

hsinfan1996: Is there a better way to retrieve alpha when alpha is a free parameter? without first going through `_get_alpha`, which requires unused arguments (`cosmo`, `a` and `mdef`). What I can think of is a `if` before `norm = self._norm(M_use, R_s, c_M, alpha)
` to separate the two cases and it returns `alpha = np.array([self.alpha])` directly.

nikfilippas: These are used arguments are in a private function so it doesn't really matter. Either way, `cosmo`, `a`, and `mdef` are always in the scope of `_real`. You could compress the if-statement if you take the special case outside, but then the code wouldn't be optimized.

If it always has to be the same shape as `M`, why not cast it to `M` using `broadcast_to` inside of the `_get_alpha` function? I think all of the `alpha` treatment should be in that function to properly enclose the calculation.

Then, you could add one more clause to the statement checking whether `alpha` is a function of mass, and output whatever `alpha` needs to be used downstream.

hsinfan1996: Great, I see. I just wanted to make sure it was not a bad practice to ignore some of the arguments. I am using `alpha = np.full_like(M, self.alpha)` and it is a little bit faster than `np.broadcast_to(self.alpha, M.shape)`.

nikfilippas:
@hsinfan1996 looks good. Maybe I missed that point, but isn't alpha from a function a useful addition after all?

hsinfan1996:
> @hsinfan1996 looks good. Maybe I missed that point, but isn't alpha from a function a useful addition after all?

I am afraid I don't get what you mean. Do you mean defining a new function inside `_get_alpha`? Is there an example?

nikfilippas:
As David mentioned, accepting alpha as a function of mass for the input. Would that be any useful? If not I'll just approve.

hsinfan1996:
> As David mentioned, accepting alpha as a function of mass for the input. Would that be any useful? If not I'll just approve.

Do you mean the alpha in `_get_alpha` or in `_real`?

nikfilippas:
The alpha here:
`def __init__(self, c_M_relation, truncated=True, alpha='cosmo'):`

So allow it to either be `'cosmo'`, a callable, or a float. Would that make any sense?

hsinfan1996:
> The alpha here: `def __init__(self, c_M_relation, truncated=True, alpha='cosmo'):`
> 
> So allow it to either be `'cosmo'`, a callable, or a float. Would that make any sense?

Thank you for pointing it out. Currently, I cannot think of a use case that needs alpha to be a function of mass, so it is not so useful to me.

nikfilippas:
Okay! lgtm!

nikfilippas (approved):

PR #991, opened 2022-09-29
Ssc terasawa

RyoTerasawa:
I have implemented a function Tk3D_SSC_Terasawa22 to compute super-sample covariance into the tk3d.py.

### DIFF
--- a/pyccl/tk3d.py
+++ b/pyccl/tk3d.py
@@ -3,11 +3,8 @@ from . import ccllib as lib
 from .pyutils import check, _get_spline2d_arrays, _get_spline3d_arrays
 import numpy as np
 
-## Terasawa
-from . import core #Terasawa
-from . import background # Terasawa
-##
-
+from . import core
+from . import background 
 
 class Tk3D(object):
     """A container for \"isotropized\" connected trispectra relevant for
@@ -367,5 +364,5 @@ def Tk3D_SSC_Terasawa22(cosmo,deltah=0.02,
                 pk1_arr=dpk12, pk2_arr=dpk12,
                 extrap_order_lok=extrap_order_lok,
                 extrap_order_hik=extrap_order_hik, is_logt=use_log)
-    return tk3d, pk, T_h, dpk, dpk12 #pk, T_h, dpk, dpk12 are outputs for debug 
+    return tk3d 
     
--- a/pyccl/tk3d.py
+++ b/pyccl/tk3d.py
@@ -3,6 +3,11 @@ from . import ccllib as lib
 from .pyutils import check, _get_spline2d_arrays, _get_spline3d_arrays
 import numpy as np
 
+## Terasawa
+from . import core #Terasawa
+from . import background # Terasawa
+##
+
 
 class Tk3D(object):
     """A container for \"isotropized\" connected trispectra relevant for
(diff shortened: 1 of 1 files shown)


pyccl/tk3d.py
```diff
@@ -213,3 +216,153 @@ def get_spline_arrays(self):
...
+    the super-sample covariance trispectrum, given by the tensor
+    product of the power spectrum responses associated with the
+    two pairs of quantities being correlated. Currently this 
+    function only applicable to matter power spectrum in flat 
+    cosmology. Each response is calculated using the method 
+    developed in Terasawa et al. 2022 (arXiv:2205.10339v2) as:
+
+    .. math::
+        \\frac{\\partial P_{mm}(k)}{\\partial\\delta_L} =
+        \\left(1 + \\frac{26}{21}T_{h}(k)  -\\frac{1}{3}\\frac{d\\log P_{mm}(k)}{d\\log k}\\right)
```
carlosggarcia, line 235: I had a quick look at Terasawa+2022 and couldn't find where the -1/3... term is coming from. If it's easy to derive or I have missed it, it'd be great to point to the Eq where it is (or that you have to use to get it) for future reference.

pyccl/tk3d.py
```diff
@@ -213,3 +216,153 @@ def get_spline_arrays(self):
...
+                            extra_parameters=extra_parameters)
+    
+    cosmo_hm = core.Cosmology(Omega_c=Omega_c_m, Omega_b=Omega_b_m,
+                            h=hm, n_s=n_s, A_s=A_s,
+                            transfer_function="boltzmann_camb",
+                            matter_power_spectrum="camb",
+                            extra_parameters=extra_parameters)
+    
+    # Growth factor
+    Dp = background.growth_factor_unnorm(cosmo_hp,a_arr)
```
carlosggarcia, line 312: You can remove the `background` dependency and use directly `cosmo_hp.growht_factor_unnorm(a_arr)`. Same in the next line

carlosggarcia (changes requested):
Great work. It looks good to me (modulo the lint errors that GitHub is complaining about). Just a few minor comments. 

In addition, it'd be great if you can add an unit test for this function (see https://github.com/LSSTDESC/CCL/blob/master/pyccl/tests/test_tkkssc.py for reference). If you were comparing this Tk with some other code, you could also add some benchmark (see e.g. https://github.com/LSSTDESC/CCL/blob/master/benchmarks/test_covariances.py). Or maybe with the simulations that you have? What do you think @damonge?

pyccl/tk3d.py
RyoTerasawa: The power spectrum response to the background density is consist of two components: growth response and dilation response (and change in the reference density). Terasawa+2022 focused on the growth response and the dilation term is simply the derivative of the power spectrum with respect to the wave number k. The total power spectrum response (including growth and dilation term) is found in Eq. 46 and 47 of Li, Hu, and Takada 2014a (https://arxiv.org/abs/1401.0385), for example.

pyccl/tk3d.py
RyoTerasawa: Thank you for pointing this out. I have removed the background dependency as you mentioned.

RyoTerasawa:
I am going to add a unit test and a benchmark test later.

RyoTerasawa:
The changes in this PR is now included in 
https://github.com/LSSTDESC/CCL/pull/1134

PR #1002, opened 2022-10-19
Enhancement to the perturbative modelling

tilmantroester:
This PR contains enhancement to the perturbative modelling in CCL to make its integration in firecrown easier.

- Allow skipping of calling `ptc.update_pk` in `get_pt_pk2d`

### DIFF
--- a/pyccl/nl_pt/power.py
+++ b/pyccl/nl_pt/power.py
@@ -445,7 +445,7 @@ def get_pt_pk2d(cosmo, tracer1, tracer2=None, ptc=None,
                 nonloc_pk_type='nonlinear',
                 a_arr=None, extrap_order_lok=1, extrap_order_hik=2,
                 return_ia_bb=False, return_ia_ee_and_bb=False,
-                return_ptc=False):
+                return_ptc=False, update_ptc=True):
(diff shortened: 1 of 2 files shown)


tilmantroester:
That sounds good to me. One thing I noticed was that for `nonlin_pk_type == 'nonlinear'` and requesting a matter-matter power spectrum, `get_pt_pk2d` evaluates the existing non-linear power spectrum at the ptc k sampling and then recreates a new `Pk2D` object (with `is_logp=False`). Since the ptc k sampling differs from the internal P(k) splines, the extrapolation (or lack thereof) then is different for the `Pk2D` object you get back than the CCL-internal power spectrum, even though they both use the same underlying P(k).

tilmantroester:
Could we merge this for now? This is currently holding up the firecrown PR.

marcpaterno:
To close the Firecrown PR in question, we will need a release on conda forge with this addition. When is such a release feasible?

damonge (approved):

damonge:
@marcpaterno let me try to tag a release and see if we can get it up within the next couple of days

marcpaterno:
Thanks David!

PR #1005, opened 2022-11-16
Replace m2r2 with sphinx_mdinclude

hsinfan1996:


damonge:
@hsinfan1996 the test currently failing is fixed in #977. If you review it, we can merge it quickly and then this one.

damonge (approved):
LGTM. Let's merge and I'll check that the docs look good once that's done.

PR #1012, opened 2022-12-21
Numpy 1.24 compatibility

hsinfan1996:


pyccl/__init__.py
```diff
@@ -16,6 +16,12 @@
 # SWIG-generated
 from . import ccllib as lib
 
+# monkey patch for isitgr and fast-pt if Numpy>=1.24
+from packaging.version import parse
+import numpy as np
+if bool(parse(np.__version__)>=parse('1.24')):
```
nikfilippas, line 22: space between operator

nikfilippas (changes requested):
Ok, good. In `core.py`, I'd personally get rid of the extra check for the numpy integer and floating types, particularly as those are generic superclasses.

hsinfan1996:
> Ok, good. In `core.py`, I'd personally get rid of the extra check for the numpy integer and floating types, particularly as those are generic superclasses.

`isinstance(np.float(1.), np.floating)` will evaluate to `False`, because `np.float` is the same as python `float`, which is not part of `np.floating`.

nikfilippas:
> `isinstance(np.float(1.), np.floating)` will evaluate to `False`, because `np.float` is identical to python `float`, which is not part of `np.floating`. Same problem with `np.int` and `int`.

As does `isinstance(1., np.floating)`. AFAIK numpy's floating is a generic superclass which is only there for some lower-level stuff. We can't create instances of `np.floating` so the check `isinstance(v, np.floating)` that was there before is a bug.
Also, the class is so generic I'd argue it is akin to checking if something is an instance of `object`.

hsinfan1996:
> > `isinstance(np.float(1.), np.floating)` will evaluate to `False`, because `np.float` is identical to python `float`, which is not part of `np.floating`. Same problem with `np.int` and `int`.
> 
> As does `isinstance(1., np.floating)`. AFAIK numpy's floating is a generic superclass which is only there for some lower-level stuff. We can't create instances of `np.floating` so the check `isinstance(v, np.floating)` that was there before is a bug. Also, the class is so generic I'd argue it is akin to checking if something is an instance of `object`.

Oh I see what you meant.

pyccl/__init__.py
```diff
@@ -16,6 +16,12 @@
 # SWIG-generated
 from . import ccllib as lib
 
+# monkey patch for isitgr and fast-pt if Numpy>=1.24
+from packaging.version import parse
+import numpy as np
+if bool(parse(np.__version__) >= parse('1.24')):
```
nikfilippas, line 22: One final thing I just realized. Comparison with the output of the `parse` function will return a boolean, so there's no need to convert it to one.

pyccl/__init__.py
```diff
@@ -16,6 +16,12 @@
 # SWIG-generated
 from . import ccllib as lib
 
+# monkey patch for isitgr and fast-pt if Numpy>=1.24
+from packaging.version import parse
+import numpy as np
```
nikfilippas, line 21: And one more. Apologies for doing it incrementally! Can we make sure we remove `np` and `parse` from pyccl's namespace after the check?

hsinfan1996:
Just did some tests
`isinstance(np.float64(1.), float)` evaluates to `True`
`isinstance(np.float32(1.), float)` evaluates to `False`
`isinstance(np.float32(1.), float)` evaluates to `False`
`isinstance(np.float32(1.), np.floating)` evaluates to `True`
Are those what we expected?

pyccl/__init__.py
```diff
@@ -16,6 +16,14 @@
...
+if (parse(np.__version__) >= parse('1.24')):
```
nikfilippas, line 22: If it's no big deal, remove the enclosing brackets to keep formatting consistency throughout CCL. I'm approving either way.

nikfilippas (approved):

PR #1017, opened 2023-01-04
Fix OpenMP again

tilmantroester:


CMakeLists.txt
```diff
@@ -111,7 +114,7 @@ endif()
 # The reason for building ccl as a shared library is that we can link it to
 # class directly, and it's not a dependency anymore
 add_library(ccl SHARED $<TARGET_OBJECTS:objlib>)
-target_link_libraries(ccl ${GSL_LIBRARIES} ${FFTW_LIBRARIES} ${CLASS_LIBRARIES} ${ANGPOW_LIBRARIES} m)
+target_link_libraries(ccl ${GSL_LIBRARIES} ${FFTW_LIBRARIES} ${CLASS_LIBRARIES} ${ANGPOW_LIBRARIES} m $<$<BOOL:${OpenMP_C_FOUND}>:OpenMP::OpenMP_C>)
```
damonge, line 117: OK, this notation is very obscure to me. What's going on?

damonge (commented):
I think it looks ok, but can I check how this relates to #1018 and #1016? 
Should the branch associated with #1018 be deleted?

CMakeLists.txt
tilmantroester: This is a generator expression (https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html). I wish there were an easier way to do this but with cmake, nothing ever seems easy to do. I added a comment on what it does.

This was in earlier versions already (e.g. 2.5.1) and just got restored here.

damonge (approved):
LGTM!

PR #1018, opened 2023-01-04
Disable OpenMP for AppleClang<13 by default

hsinfan1996:
Merge into PR #1017


hsinfan1996:
https://github.com/conda-forge/pyccl-feedstock/pull/33

beckermr:
I don't follow this pr at all. We've built openmp on apple for a long time now. The real problem is the cmake changes, right?

beckermr:
Can't we just roll back the changes made originally for cmake?

hsinfan1996:
Building ccl with AppleClang is mostly fine. However, there will be ImportError in python if libomp is found but not linked correctly. For OpenMP to work with AppleClang12, we need to link libomp with link_libraries in cmake (which breaks conda-forge build for arm64) or users need to set LDFLAGS manually. Now I am trying to revert the changes made for cmake while preventing the ImportError problem.

PR #1025, opened 2023-03-11
GHA & Installation improvements

nikfilippas:
**Installation**
-------------------
- We can install CLASS using pip through `classy` for testing.
- It is [not recommended](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html) to use `python setup.py` directly. `pip` can handle this well, so replace CCL installation using `pip install .`.
- Clean up the conda environment that we set up. Many listed dependencies aren't needed any more.
- Precisely set the required installation packages in `setup.py`. We only need `numpy` to install it (nothing will work after but that's not our business - it installs fine). Numpy is needed to form the `SWIG` link.
- Update the docs to reflect that change. To install `pyccl` we need both `cmake` and `swig`, which are not real packages (although they exist in `conda-forge`) and a user can install them in many different ways, so just clarify that if they aren't there, CCL won't build.

**CI Set-up**
-----------------
- There are no C unit tests any more. Save time by not running `make` for them.
- macOS: Save time by not removing `homebrew`, and bump the version to `macos-latest`.
- Cache conda environment to save time setting it up. Caches are updated at 00:00UTC every day or manually upon request.
- Only submit one coverage job (see from `master`: [2 jobs](https://coveralls.io/builds/57767902)). There is no need to do this twice since we don't aggregate coverage results from several builds.
- Only check lint once, concurrently with the matrix builds. This way, if it fails, the builds won't stop.

**Actions**
--------------
- Bump versions of actions [by @hsinfan1996 ].
- Use Actions when possible, which are developed together with the packages and thus save us from having to review gha whenever there is a breaking change. New: `cache-action`, `flake8-action`, `coveralls-action`.

**Lint**
----------
- Flake8 can now be run from the top-level (CCL/) without any additional arguments, as these are now included in the `.flake8` config file (which replaced the obscurely named `setup.cfg`). Simply `flake8 .` will do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,8 @@ on:
 
 env:
   CACHE_NUMBER: 0  # increase to reset cache manually
-  DEPENDENCIES: .github/environment.yml
+  CONDA_ENV: .github/environment.yml
+  TODAY: ${{ date +'%Y%m%d' }}
 
 jobs:
   build:
@@ -66,9 +67,9 @@ jobs:
           export CFLAGS="-I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
           export LDFLAGS="-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"
 
-      - name: Compute cache key
-        run: |
-          echo "DATE=$(date +'%Y%m%d')" >> $GITHUB_ENV
+       # - name: Compute cache key
+       #   run: |
+       #     echo "TODAY=$(date +'%Y%m%d')" >> $GITHUB_ENV
 
       - name: Get cached environment
         uses: actions/cache@v3
@@ -78,16 +79,16 @@ jobs:
           key:
             conda-py${{ matrix.py }}--${{
             matrix.os }}--${{
-            hashFiles(env.DEPENDENCIES) }}-${{
+            hashFiles(env.CONDA_ENV) }}-${{
             hashFiles('setup.py') }}--${{
-            env.DATE }}-${{
+            env.TODAY }}-${{
             env.CACHE_NUMBER }}
 
       - name: Install environment packages
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
           export MAMBA_NO_BANNER=1  # hide mamba banner from output
-          mamba env update --file ${{ env.DEPENDENCIES }} --prune
+          mamba env update --file ${{ env.CONDA_ENV }} --prune
           pip install "classy<3"
 
       - name: Install CCL
@@ -102,7 +103,7 @@ jobs:
         run: |
           OMP_NUM_THREADS=2 pytest -vv benchmarks --cov=pyccl --cov-report=lcov:lcov.info --cov-append
 
-      - name: Coveralls parallel
+      - name: Code coverage
         if: matrix.label == 'linux-64'
         uses: coverallsapp/github-action@v2
         with:
@@ -119,24 +120,6 @@ jobs:
         uses: py-actions/flake8@v2
         with:
           args: "--config .flake8"
-
-  # coverage:
-  #   runs-on: ubuntu-latest
-  #   needs: build
-  #   if: ${{ success() }}
-  #   defaults:
-  #     run:
-  #       shell: bash -l {0}
-
-  #   steps:
-  #     - name: Checkout CCL repository
-  #       uses: actions/checkout@v3
-
-  #     - name: Coveralls finished
-  #       uses: coverallsapp/github-action@v2
-  #       with:
-  #         parallel-finished: true
-  #         carryforward: "linux-64"
 
 
 # Legacy step: uninstall homebrew (CI works without it).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,8 @@ jobs:
           export CFLAGS="-I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
           export LDFLAGS="-L/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/lib"
 
-       # - name: Compute cache key
-       #   run: |
-       #     echo "TODAY=$(date +'%Y%m%d')" >> $GITHUB_ENV
+      - name: Set current date as environment variable
+        run: echo "TODAY=$(date +'%Y.%m.%d')" >> $GITHUB_ENV
 
       - name: Get cached environment
         uses: actions/cache@v3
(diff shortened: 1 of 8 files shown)


nikfilippas:
Apple McTrouble TBC

nikfilippas:
Haaa really cool @hsinfan1996 , thanks for saving it, I had no idea why it wasn't working. 
This is a quick review, so if you have time let me know what you think and we can merge it.

damonge:
Please check why coveralls seems to say coverage has decreased by 2.6% before merging (and make sure this actually hasn't happened)

nikfilippas:
Yea, something funny must be happening because the lines coveralls flags in here are covered in master. However, coveralls doesn't flag them as "new", just as "uncov". So I'll check more closely before we merge anything.

hsinfan1996:
The coverage is 95% on master tested locally on my machine. Weird.

hsinfan1996:
The coverage is the same as that of the master branch. Don't know why the reported percentage is different.
https://github.com/LSSTDESC/CCL/actions/runs/4348221279/jobs/7596478089#step:11:855

hsinfan1996:
The coverage from benchmarks changed from 98% to 95%.

damonge:
@nikfilippas did this happen in one of the branches we merged recently and didn't notice? I see a lot of the Ishiyama concentration code is untested right now.

nikfilippas:
@damonge no, if that was the case, the coverage would drop in the CCLObject PR too.
Here are the reports from the 2 recently merged PRs:

**Hashing**
Screenshot from 2023-03-12 01-36-22

**Caching**
Screenshot from 2023-03-12 01-36-39

Let me investigate what is happening and I'll get back to you.

nikfilippas:
Okay, it looks like once we change the way we install CCL from `python setup.py ...` to `pip install .`, the coverage report fails to account for the benchmarks, although the benchmarks/ cov is appended to the tests/ cov:
```console
      - name: unit tests
        run: |
          OMP_NUM_THREADS=2 pytest -vv pyccl --cov=pyccl

      - name: benchmarks
        run: |
          OMP_NUM_THREADS=2 pytest -vv benchmarks --cov=pyccl --cov-append
```

This was originally written by @beckermr so any tips here Matt would be appreciated.

beckermr:
Don't pip install to conda environments except for doing dev work on a package. 

Setting those env vars for osx should not be needed and is a sign that something else is broken. 

I have no idea why the coverage changed. Maybe try pip install -e .?

nikfilippas:
@beckermr I think `pip`-installing `classy` is fine in our case because it is the final package we install, so even if the dependency tree breaks after using `pip`, we wouldn't care (it is conda's pip, so the package paths are fine). Plus, `pip` is the final command used, [`conda` won't break it after](https://www.anaconda.com/blog/using-pip-in-a-conda-environment). This is not the cause of the problem either way.

I have no clue about osx, @hsinfan1996 why did you have to add those flags to make it work on a mac?

Wow?! Installing CCL in dev mode seems to have worked? Any idea why it worked?

hsinfan1996:
Installing CLASS on macOS has been problematic. See #967 and #984. After GH Actions switches to macOS 11, somehow the Xcode paths (linker and include) has to be set manually. Those two lines aren't new, they were there but in a different step.

nikfilippas:
In the next version we're going to have to write the OS source code too. 🤷

beckermr:
I had this setup properly a long time ago to use the conda-forge compilers:

```
         if [[ ${MATRIX_OS} == "macos-latest" ]]; then
            conda install llvm-openmp
            echo "DYLD_FALLBACK_LIBRARY_PATH=${CONDA_PREFIX}/lib" >> $GITHUB_ENV
            SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
            echo "SDKROOT: ${SDKROOT}"
            echo "SDKROOT=${SDKROOT}" >> $GITHUB_ENV
            echo "CONDA_BUILD_SYSROOT=${SDKROOT}" >> $GITHUB_ENV
```

This is the correct way to interact with them. IDK if the block above got removed or what.

nikfilippas:
@beckermr that's still there.

beckermr:
This should be all that is needed for packages to link correctly.

beckermr:
See this diff for example: https://github.com/LSSTDESC/CCL/blame/ed8764b99da5817f2301ec8e0de5707c282f6e9f/.github/workflows/ci.yml

nikfilippas:
This might have to do with the maxOS 11 issue @hsinfan1996 mentioned? Because in my first commits here these 2 lines didn't exist, but the mac tests failed to install all dependencies.
https://github.com/LSSTDESC/CCL/actions/runs/4392740347/jobs/7692693580

hsinfan1996:
> I had this setup properly a long time ago to use the conda-forge compilers:
> 
> 
> 
> ```
> 
>          if [[ ${MATRIX_OS} == "macos-latest" ]]; then
> 
>             conda install llvm-openmp
> 
>             echo "DYLD_FALLBACK_LIBRARY_PATH=${CONDA_PREFIX}/lib" >> $GITHUB_ENV
> 
>             SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
> 
>             echo "SDKROOT: ${SDKROOT}"
> 
>             echo "SDKROOT=${SDKROOT}" >> $GITHUB_ENV
> 
>             echo "CONDA_BUILD_SYSROOT=${SDKROOT}" >> $GITHUB_ENV
> 
> ```
> 
> 
> 
> This is the correct way to interact with them. IDK if the block above got removed or what. 

I will do more tests on this part to investigate why those lines do not work on macOS 11.

nikfilippas:
Do we know why this ended up working with `pip install -e .` instead of `pip install .`? What does the dev mode have to do with the `--cov-append` argument for the coverage?

nikfilippas:
Alright everyone, with apologies for the spam you must have been getting from github, this is finally complete, and with an updated description to match all the changes. 🚀

.github/workflows/ci.yml
```diff
@@ -7,129 +8,123 @@ on:
       - releases/*
   pull_request: null
 
+env:
+  CACHE_NUMBER: 0  # increase to reset cache manually
+  CONDA_ENV: .github/environment.yml
```
damonge, line 13: what does this do?

damonge:
The macos tests seem to be stuck

.github/workflows/ci.yml
nikfilippas: The block sets environment variables.
`CACHE_NUMBER` is used in the cache key, and if the developer wants to trigger cache reset manually they only need to change this number for the cache key comparison to fail, and the packages to be re-downloaded.
`CONDA_ENV` is a file with the conda environment packages. I took them out of the workflow to declutter.

nikfilippas:
@damonge they have run (it's the first test in the list). The required one hasn't run because I bumped the macOS version to `macos-latest`. So we need to change the required test from `build (macos-11, 3.8)` to `build (macos-latest, 3.8)`.

damonge (approved):
LGTM

PR #1032, opened 2023-03-15
RTD fix

damonge:
Fixes readthedocs issue triggered by new functionality.
I've tested that this runs as it should on readthedocs.


readthedocs/requirements.txt
nikfilippas, line 1: Are all four necessary? I.e. did  you check that removing any one of those will make RTD fail?

.readthedocs.yaml
```diff
@@ -0,0 +1,25 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+build:
+  os: ubuntu-22.04
```
nikfilippas, line 9: Is there an `ubuntu-latest` option like in GHA?

nikfilippas (commented):
I trust that it works, but have a quick look at the comments.

readthedocs/requirements.txt
damonge: yes, hence the many commits

.readthedocs.yaml
damonge: This (and 20) seemed to be the only two options.

damonge:
The generated docs are here https://ccl.readthedocs.io/en/rtd_fix_cached_property/

PR #1033, opened 2023-03-16
Implement separate `__eq__` methods

nikfilippas:
As @rmjarvis and @marcpaterno correctly pointed out, we can save loads of time comparing `CCLObjects` if we compare parameters instead of strings. To that end, I extended some functionality that was already there to do parameter comparisons for most objects.

Just like the `__repr_attrs__` class variable, which automatically builds a repr using a list of specified arguments, `__eq_attrs__` now makes sure that the default `__eq__` is overridden and that the individual attributes are compared one-by-one.
For example, in the `Cosmology` object, we only need to add this line in the beginning:
```python
    __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs", "_accuracy_params",)  # check these

    def __init__(...):
        ...
```

There are only 3 classes where this is not possible (because we have to check the C-level objects): `Pk2D`, `Tk3D`, `Tracer`. For these, I implemented full `__eq__` methods as per the usual way.

Timings show that this way is indeed much faster (`bld` is the `repr`):
Screenshot from 2023-03-16 04-37-34
Screenshot from 2023-03-16 05-02-53

Note, however, that with the current implementation in `master`, the first computation of `_repr` might be slow, but every new call to `__eq__` is of the order of nanoseconds (compare to microseconds with this PR proposal). It is a trade-off. So, if we compare objects sequentially (and discard the old one if they are not equal, like in an MCMC) this PR wins. But if we compare one object with several others, the currently implemented way wins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ cosmo = ccl.Cosmology(
     transfer_function='bbks')
 
 # Define a simple binned galaxy number density curve as a function of redshift
-z_n = np.linspace(0., 1., 200)
+z_n = np.linspace(0., 1., 500)
 n = np.ones(z_n.shape)
 
 # Create objects to represent tracers of the weak lensing signal with this
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -590,12 +590,12 @@ class CCLObject(ABC):
 
     def __eq__(self, other):
         # Two same-type objects are equal if their representations are equal.
-        if self.__class__ is not other.__class__:
+        if type(self) is not type(other):
             return False
         # Compare the attributes listed in `__eq_attrs__`.
         if hasattr(self, "__eq_attrs__"):
             for attr in self.__eq_attrs__:
-                if any(getattr(self, attr) != getattr(other, attr)):
+                if np.any(getattr(self, attr) != getattr(other, attr)):
                     return False
             return True
         # Fall back to repr comparison.
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -597,6 +597,7 @@ class CCLObject(ABC):
             for attr in self.__eq_attrs__:
                 if any(getattr(self, attr) != getattr(other, attr)):
                     return False
+            return True
         # Fall back to repr comparison.
         return repr(self) == repr(other)
 
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -594,8 +594,9 @@ class CCLObject(ABC):
             return False
         # Compare the attributes listed in `__eq_attrs__`.
         if hasattr(self, "__eq_attrs__"):
-            return all([getattr(self, attr) == getattr(other, attr)
-                        for attr in self.__eq_attrs__])
+            for attr in self.__eq_attrs__:
+                if any(getattr(self, attr) != getattr(other, attr)):
+                    return False
         # Fall back to repr comparison.
         return repr(self) == repr(other)
 
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -5,7 +5,6 @@ import numpy as np
 from inspect import signature
 from _thread import RLock
 from abc import ABC
-from operator import attrgetter
 
 
 def _to_hashable(obj):
@@ -595,7 +594,7 @@ class CCLObject(ABC):
             return False
         # Compare the attributes listed in `__eq_attrs__`.
         if hasattr(self, "__eq_attrs__"):
-            return all([attrgetter(attr)(self) == attrgetter(attr)(other)
+            return all([getattr(self, attr) == getattr(other, attr)
                         for attr in self.__eq_attrs__])
         # Fall back to repr comparison.
         return repr(self) == repr(other)
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -440,6 +440,15 @@ class UnlockInstance:
 unlock_instance = UnlockInstance.unlock_instance
 
 
+def is_equal(this, other):
+    """Powerful helper for equivalence checking."""
+    try:
+        np.testing.assert_equal(this, other)
+        return True
+    except AssertionError:
+        return False
+
+
 class FancyRepr:
     """Controls the usage of fancy ``__repr__` for ``CCLObjects."""
     _enabled: bool = True
@@ -595,7 +604,7 @@ class CCLObject(ABC):
         # Compare the attributes listed in `__eq_attrs__`.
         if hasattr(self, "__eq_attrs__"):
             for attr in self.__eq_attrs__:
-                if np.any(getattr(self, attr) != getattr(other, attr)):
+                if not is_equal(getattr(self, attr), getattr(other, attr)):
                     return False
             return True
         # Fall back to repr comparison.
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -607,8 +607,8 @@ class CCLObject(ABC):
                 if not is_equal(getattr(self, attr), getattr(other, attr)):
                     return False
             return True
-        # Fall back to repr comparison.
-        return repr(self) == repr(other)
+        # Fall back to default Python comparison.
+        return id(self) == id(other)
 
 
 class CCLHalosObject(CCLObject):
--- a/pyccl/base.py
+++ b/pyccl/base.py
@@ -5,6 +5,7 @@ import numpy as np
 from inspect import signature
 from _thread import RLock
 from abc import ABC
+from operator import attrgetter
 
 
 def _to_hashable(obj):
@@ -586,6 +587,11 @@ class CCLObject(ABC):
         # Two same-type objects are equal if their representations are equal.
         if self.__class__ is not other.__class__:
             return False
+        # Compare the attributes listed in `__eq_attrs__`.
+        if hasattr(self, "__eq_attrs__"):
+            return all([attrgetter(attr)(self) == attrgetter(attr)(other)
+                        for attr in self.__eq_attrs__])
+        # Fall back to repr comparison.
         return repr(self) == repr(other)
 
 
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -67,6 +67,36 @@ emulator_neutrinos_types = {
 }
 
 
+def _methods_of_cosmology(cls=None, *, modules=[]):
+    """Assign all functions in ``modules`` which take ``cosmo`` as their
+    first argument as methods of the class ``cls``.
+    """
+    import functools
+    from importlib import import_module
+
+    if cls is None:
+        # called with parentheses
+        return functools.partial(_methods_of_cosmology, modules=modules)
+
+    pkg = __name__.rsplit(".")[0]
+    modules = [import_module(f".{module}", pkg) for module in modules]
+    funcs = [getmembers(module, isfunction) for module in modules]
+    funcs = [func for sublist in funcs for func in sublist]
+
+    for name, func in funcs:
+        pars = signature(func).parameters
+        if pars and list(pars)[0] == "cosmo":
+            setattr(cls, name, func)
+
+    return cls
+
+
+_modules = ["background", "bcm", "boltzmann", "cls", "correlations",
+            "covariances", "neutrinos", "pk2d", "power", "pyutils",
+            "tk3d", "tracers", "halos", "nl_pt"]
+
+
+@_methods_of_cosmology(modules=_modules)
 class Cosmology(CCLObject):
     """A cosmology including parameters and associated data.
 
@@ -198,25 +228,8 @@ class Cosmology(CCLObject):
 
     """
     from ._repr import _build_string_Cosmology as __repr__
-
-    # Go through all functions in the main package and the subpackages
-    # and make every function that takes `cosmo` as its first argument
-    # an attribute of this class.
-    from . import (background, bcm, boltzmann, cls,
-                   correlations, covariances, neutrinos,
-                   pk2d, power, pyutils, tk3d, tracers, halos, nl_pt)
-    subs = [background, boltzmann, bcm, cls, correlations, covariances,
-            neutrinos, pk2d, power, pyutils, tk3d, tracers, halos, nl_pt]
-    funcs = [getmembers(sub, isfunction) for sub in subs]
-    funcs = [func for sub in funcs for func in sub]
-    for name, func in funcs:
-        pars = list(signature(func).parameters)
-        if pars and pars[0] == "cosmo":
-            vars()[name] = func
-    # clear unnecessary locals
-    del (background, boltzmann, bcm, cls, correlations, covariances,
-         neutrinos, pk2d, power, pyutils, tk3d, tracers, halos, nl_pt,
-         subs, funcs, func, name, pars)
+    __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
+                    "_accuracy_params",)
 
     def __init__(
             self, Omega_c=None, Omega_b=None, h=None, n_s=None,
@@ -1167,6 +1180,9 @@ class CosmologyCalculator(Cosmology):
             corresponding to the "HALOFIT" transformation of
             Takahashi et al. 2012 (arXiv:1208.2701).
     """
+    __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
+                    "_accuracy_params", "_pk_lin", "_pk_nl",)
+
     def __init__(
             self, Omega_c=None, Omega_b=None, h=None, n_s=None,
             sigma8=None, A_s=None, Omega_k=0., Omega_g=None,
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -1181,7 +1181,7 @@ class CosmologyCalculator(Cosmology):
             Takahashi et al. 2012 (arXiv:1208.2701).
     """
     __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
-                    "_accuracy_params", "_pk_lin", "_pk_nl",)
+                    "_accuracy_params", "_input_arrays",)
 
     def __init__(
             self, Omega_c=None, Omega_b=None, h=None, n_s=None,
@@ -1215,6 +1215,9 @@ class CosmologyCalculator(Cosmology):
         has_pknl = pk_nonlin is not None
         has_nonlin_model = nonlinear_model is not None
 
+        self._input_arrays = {"background": background, "growth": growth,
+                              "pk_linear": pk_linear, "pk_nonlin": pk_nonlin,
+                              "nonlinear_model": nonlinear_model}
         if has_bg:
             self._init_bg(background)
         if has_dz:
--- a/pyccl/halos/massdef.py
+++ b/pyccl/halos/massdef.py
@@ -286,3 +286,12 @@ def MassDefVir(c_m='Klypin11'):
         c_m (string): concentration-mass relation.
     """
     return MassDef('vir', 'critical', c_m_relation=c_m)
+
+
+def MassDefFof(c_m=None):
+    r""":math:`\Delta = \rm FoF` mass definition.
+
+    Args:
+        c_m (string): concentration-mass relation.
+    """
+    return MassDef('fof', 'matter', c_m_relation=c_m)
--- a/pyccl/halos/massdef.py
+++ b/pyccl/halos/massdef.py
@@ -3,6 +3,7 @@ from ..core import check
 from ..background import species_types, rho_x, omega_x
 from ..base import CCLHalosObject
 import numpy as np
+from functools import cached_property
 
 
 def mass2radius_lagrangian(cosmo, M):
@@ -106,7 +107,7 @@ class MassDef(CCLHalosObject):
         else:
             self._concentration_init(c_m_relation)
 
-    @property
+    @cached_property
     def name(self):
         """Give a name to this mass definition."""
         if isinstance(self.Delta, (int, float)):
--- a/pyccl/halos/massdef.py
+++ b/pyccl/halos/massdef.py
@@ -84,7 +84,7 @@ class MassDef(CCLHalosObject):
             If `None`, no c(M) relation will be attached to this mass
             definition (and hence one can't translate into other definitions).
     """
-    __repr_attrs__ = ("name",)
+    __repr_attrs__ = __eq_attrs__ = ("name",)
 
     def __init__(self, Delta, rho_type, c_m_relation=None):
         # Check it makes sense
--- a/pyccl/halos/massdef.py
+++ b/pyccl/halos/massdef.py
@@ -289,10 +289,10 @@ def MassDefVir(c_m='Klypin11'):
     return MassDef('vir', 'critical', c_m_relation=c_m)
 
 
-def MassDefFof(c_m=None):
+def MassDefFof():
     r""":math:`\Delta = \rm FoF` mass definition.
 
     Args:
         c_m (string): concentration-mass relation.
     """
-    return MassDef('fof', 'matter', c_m_relation=c_m)
+    return MassDef('fof', 'matter')
--- a/pyccl/halos/profiles.py
+++ b/pyccl/halos/profiles.py
@@ -651,8 +651,8 @@ class HaloProfileNFW(HaloProfile):
             radii.
     """
     __repr_attrs__ = __eq_attrs__ = (
-        "cM", "fourier_analytic", "projected_analytic",
-        "cumul2d_analytic", "truncated", "precision_fftlog",)
+        "fourier_analytic", "projected_analytic",
+        "cumul2d_analytic", "truncated", "cM", "precision_fftlog",)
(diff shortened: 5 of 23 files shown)


pyccl/core.py
```diff
@@ -198,25 +228,8 @@ class Cosmology(CCLObject):
...
-    funcs = [getmembers(sub, isfunction) for sub in subs]
-    funcs = [func for sub in funcs for func in sub]
-    for name, func in funcs:
-        pars = list(signature(func).parameters)
-        if pars and pars[0] == "cosmo":
-            vars()[name] = func
-    # clear unnecessary locals
-    del (background, boltzmann, bcm, cls, correlations, covariances,
-         neutrinos, pk2d, power, pyutils, tk3d, tracers, halos, nl_pt,
-         subs, funcs, func, name, pars)
```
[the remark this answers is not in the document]
nikfilippas: No, I moved the whole block out of the way into a class decorator so we don't pollute the namespace with useless stuff.

pyccl/base.py
```diff
@@ -586,6 +587,11 @@ def __eq__(self, other):
         # Two same-type objects are equal if their representations are equal.
         if self.__class__ is not other.__class__:
             return False
+        # Compare the attributes listed in `__eq_attrs__`.
+        if hasattr(self, "__eq_attrs__"):
+            return all([attrgetter(attr)(self) == attrgetter(attr)(other)
```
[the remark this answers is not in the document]
nikfilippas: The rationale is also having it work with attributes that are several levels deep.
For example, `getattr(cosmo, "_params.Omega_c")` won't work, but `attrgetter("_params.Omega_c")(cosmo)` will.

pyccl/pk2d.py
```diff
@@ -130,17 +130,34 @@ def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
...
+            return True
+        # Check extrapolation orders.
+        if not (self.extrap_order_lok == other.extrap_order_lok
+                and self.extrap_order_hik == other.extrap_order_hik):
+            return False
+        # Check the individual splines.
+        a1, lk1, pk1 = self.get_spline_arrays()
+        a2, lk2, pk2 = other.get_spline_arrays()
+        return ((a1 == a2).all() and (lk1 == lk2).all()
+                and np.allclose(pk1, pk2, atol=0, rtol=1e-12))
```
nikfilippas, line 148: For this particular line of code we can define a `ccl.spline_params` parameter that sets `rtol`. I chose 1e-12 because when I tested it `np.max(np.abs( 1 - pk / (2*pk/2) ))` was about `5e-13` due to truncation errors, which means that with what we currently have in `master`, `pk == 2*pk/2` will more often than not return `False`.

Let me know what you think about the additional spline parameter.

pyccl/tk3d.py
```diff
@@ -133,10 +133,39 @@ def __init__(self, a_arr, lk_arr, tkk_arr=None,
...
+        # Check extrapolation orders.
+        if not (self.extrap_order_lok == other.extrap_order_lok
+                and self.extrap_order_hik == other.extrap_order_hik):
+            return False
+        # Check the individual splines.
+        a1, lk11, lk12, tk1 = self.get_spline_arrays()
+        a2, lk21, lk22, tk2 = other.get_spline_arrays()
+        return ((a1 == a2).all()
+                and (lk11 == lk21).all() and (lk21 == lk22).all()
+                and np.allclose(tk1, tk2, atol=0, rtol=1e-12))
```
nikfilippas, line 155: @damonge  same here as with `Pk2D`.

pyccl/tests/test_cclobject.py
```diff
@@ -33,111 +29,20 @@ def test_fancy_repr():
...
-    # TODO: uncomment once __eq__ methods are implemented.
-    # P1 = ccl.halos.HaloProfileHOD(c_M_relation=CM1)
-    # P2 = ccl.halos.HaloProfileHOD(c_M_relation=CM2)
-    # assert P1 == P2
-
-    # PCOV1 = ccl.halos.Profile2pt(r_corr=1.5)
-    # PCOV2 = ccl.halos.Profile2pt(r_corr=1.0)
-    # assert PCOV1 != PCOV2
-    # PCOV2.update_parameters(r_corr=1.5)
-    # assert PCOV1 == PCOV2
```
nikfilippas, line 137: There was no point of this test living in here so I distributed it to the respective test modules.

pyccl/pk2d.py
damonge: I'd rather we check pk1 == pk2 to avoid any surprises at all. This is what anyone would expect __eq__ to do. No reason to complicate things.

pyccl/tk3d.py
damonge: Same answer

pyccl/base.py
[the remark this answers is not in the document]
damonge: Will this often fail with the usual numpy error when comparing arrays?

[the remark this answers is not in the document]
damonge: One other point. It occurs to me that the following would be more efficient:
```
for attr in self.__eq_attrs__:
    if attrgetter(attr)(self) != attrgetter(attr)(other):
        return False
```
so not all attributes need to be compared if one of the first ones doesn't match.

pyccl/halos/massdef.py
```diff
@@ -84,7 +84,7 @@ class MassDef(CCLHalosObject):
             If `None`, no c(M) relation will be attached to this mass
             definition (and hence one can't translate into other definitions).
     """
-    __repr_attrs__ = ("name",)
+    __repr_attrs__ = __eq_attrs__ = ("name",)
```
damonge, line 87: In this case `name` is built on the fly every time. Can we not build it once and store it on initialization? I think it's important if we're gonna rely so much on mdef comparisons.

pyccl/pk2d.py
nikfilippas: The problem is that due to machine precision, if we use `rtol=1e-12` this check will hardly ever pass, so all this eq-checking goes to waste. We want eq to return True if the pk will give us the same theoretical prediction, and having an rtol of 1e-12 will.
That question was more on if we are going to put it somewhere as a global tolerance parameter, or if we'll leave it hard-coded like that. I'm more inclined to do the first but not sure.

pyccl/halos/massdef.py
```diff
@@ -286,3 +286,12 @@ def MassDefVir(c_m='Klypin11'):
         c_m (string): concentration-mass relation.
     """
     return MassDef('vir', 'critical', c_m_relation=c_m)
+
+
+def MassDefFof(c_m=None):
```
damonge, line 291: It doesn't make sense to associate a for mdef with a cm relation, since these masses cannot be translated into other masses

pyccl/pk2d.py
nikfilippas: Note, that if we implement a global control like we have for `FancyRepr` at the moment, all these checks can be disabled so we can use Python's default id-checking.

pyccl/halos/profiles.py
```diff
@@ -654,8 +650,9 @@ class HaloProfileNFW(HaloProfile):
             truncated at :math:`r = R_\\Delta` (i.e. zero at larger
             radii.
     """
-    __repr_attrs__ = ("cM", "fourier_analytic", "projected_analytic",
-                      "cumul2d_analytic", "truncated", "precision_fftlog",)
+    __repr_attrs__ = __eq_attrs__ = (
+        "cM", "fourier_analytic", "projected_analytic",
+        "cumul2d_analytic", "truncated", "precision_fftlog",)
```
damonge, line 655: I would put cM last or before precision_fftlog, since this is the slowest one to compare

pyccl/halos/profiles.py
```diff
@@ -851,7 +848,8 @@ class HaloProfileEinasto(HaloProfile):
         alpha (float, 'cosmo'): Set the Einasto alpha parameter or set to
             'cosmo' to calculate the value from cosmology. Default: 'cosmo'
     """
-    __repr_attrs__ = ("cM", "truncated", "alpha", "precision_fftlog",)
+    __repr_attrs__ = __eq_attrs__ = ("cM", "truncated", "alpha",
+                                     "precision_fftlog",)
```
damonge, line 852: As above re cM

pyccl/halos/profiles.py
```diff
@@ -961,8 +959,9 @@ class HaloProfileHernquist(HaloProfile):
             truncated at :math:`r = R_\\Delta` (i.e. zero at larger
             radii.
     """
-    __repr_attrs__ = ("cM", "fourier_analytic", "projected_analytic",
-                      "cumul2d_analytic", "truncated", "precision_fftlog",)
+    __repr_attrs__ = __eq_attrs__ = (
+        "cM", "fourier_analytic", "projected_analytic",
+        "cumul2d_analytic", "truncated", "precision_fftlog",)
```
damonge, line 964: as above re cM

pyccl/halos/profiles_cib.py
```diff
@@ -77,8 +77,9 @@ class HaloProfileCIBShang12(HaloProfile):
         L0 (float): luminosity scale (in
             :math:`{\\rm Jy}\\,{\\rm Mpc}^2\\,M_\\odot^{-1}`).
     """
-    __repr_attrs__ = ("cM", "nu", "alpha", "T0", "beta", "gamma", "s_z",
-                      "l10meff", "sigLM", "Mmin", "L0", "precision_fftlog",)
+    __repr_attrs__ = __eq_attrs__ = (
+        "cM", "nu", "alpha", "T0", "beta", "gamma", "s_z",
+        "l10meff", "sigLM", "Mmin", "L0", "precision_fftlog",)
```
damonge, line 82: cM last

pyccl/pk2d.py
damonge: Also, following the logic in previous comments, split this into three ifs for each of the arrays, so you don't necessarily need to go through all of them if e.g. the first one fails.

pyccl/pk2d.py
```diff
@@ -130,17 +130,37 @@ def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
...
+
+    def __hash__(self):
+        return hash(repr(self))
+
```
damonge, line 152: Why was this not present before?

pyccl/tk3d.py
```diff
@@ -133,10 +133,42 @@ def __init__(self, a_arr, lk_arr, tkk_arr=None,
...
+    def __eq__(self, other):
+        # Check the object class.
+        if self.__class__ is not other.__class__:
+            return False
+        # If the objects contain no data, return early.
+        if not (self or other):
+            return True
+        # If one is factorizable and the other one is not, return early.
+        if self.tsp.is_product ^ other.tsp.is_product:
+            return False
```
damonge, line 145: Why not !=

pyccl/tk3d.py
damonge: See comment in pk2d about splitting this

pyccl/tracers.py
```diff
@@ -185,6 +185,79 @@ def __init__(self):
...
+
+        # If the tracer collections are not the same length, return early.
+        if len(self._trc) != len(other._trc):
+            return False
+
+        # Check `der_angles` & `der_bessel` for each tracer in the collection.
+        bessel = self.get_bessel_derivative(), other.get_bessel_derivative()
+        angles = self.get_angles_derivative(), other.get_angles_derivative()
+        if not (np.array_equal(*bessel) and np.array_equal(*angles)):
+            return False
```
damonge, line 205: Check one at a time

pyccl/tracers.py
```diff
@@ -185,6 +185,79 @@ def __init__(self):
...
+        for t1, t2 in zip(self._trc, other._trc):
+            if bool(t1.kernel) ^ bool(t2.kernel):
+                # only one of them has a kernel
+                return False
+            if t1.kernel is None:
+                # none of them has a kernel
+                continue
+            if not np.allclose(_get_spline1d_arrays(t1.kernel.spline),
+                               _get_spline1d_arrays(t2.kernel.spline),
+                               **kwargs):
```
damonge, line 218: No, please. Check that they are exactly the same

pyccl/tracers.py
```diff
@@ -185,6 +185,79 @@ def __init__(self):
...
+                    # only one of them has this transfer type
+                    return False
+                if spl1 is None:
+                    # none of them has this transfer type
+                    continue
+                # `pts` contain the the grid points and the transfer functions
+                pts1, pts2 = c2py[attr](spl1), c2py[attr](spl2)
+                for pt1, pt2 in zip(pts1, pts2):
+                    # loop through output points of `_get_splinend_arrays`
+                    if not np.allclose(pt1, pt2, **kwargs):
```
damonge, line 252: Check they are exactly the same

damonge (changes requested):
Some comments. Will look at core.py next, then tests

README.md
```diff
@@ -72,7 +72,7 @@ cosmo = ccl.Cosmology(
     transfer_function='bbks')
 
 # Define a simple binned galaxy number density curve as a function of redshift
-z_n = np.linspace(0., 1., 200)
+z_n = np.linspace(0., 1., 500)
```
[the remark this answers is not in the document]
nikfilippas: 200 produced a warning. We don't want users to stumble across a warning in the first ever go they have at CCL.

pyccl/base.py
```diff
@@ -592,6 +593,11 @@ def __eq__(self, other):
...
+                        for attr in self.__eq_attrs__])
+        # Fall back to repr comparison.
```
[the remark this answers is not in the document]
nikfilippas: It doesn't happen actually, I put it there in case we want to do it in the future, but I agree it isn't a good idea after all, I reverted it.

pyccl/halos/massdef.py
nikfilippas: I know, it's only there to preserve the signature, but we will be addressing all of that in 1064, so leave it for now.

pyccl/halos/profiles.py
nikfilippas: Good idea, done.

pyccl/halos/profiles.py
nikfilippas: done

pyccl/halos/profiles.py
nikfilippas: done

pyccl/pk2d.py
damonge: I agree this will only pass when both pk2ds have been initialised with exactly the same arrays, but that's the intended use of eq. Compare them directly.

pyccl/halos/profiles_cib.py
nikfilippas: done

pyccl/halos/massdef.py
damonge: What signature? This function didn't exist before, right?

pyccl/base.py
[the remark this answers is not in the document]
nikfilippas: I have reverted that.

pyccl/pk2d.py
nikfilippas: 1. No need to split them because Python will exit early if one of them is `False` in the chain. For example `True and print("hi")` prints hi, but `False and print("hi")` doesn't print anything.
2. No, eq shows us equivalence, not absolute equality. It is up to us developers to decide whether we want two arrays that are 1e-12 off to be equivalent. And I recall this was suggested in the past in one of the issues, but can't remember where exactly.

nikfilippas: Like I said, at the moment, `pk == 2*pk/2` is `False`, so we want to avoid that.

pyccl/halos/massdef.py
nikfilippas: Good idea, made it a cached property, which is safe since `MassDef` objects are, and will be, immutable.

pyccl/halos/massdef.py
nikfilippas: Signature compared to the other `MassDef`s. I don't mind removing it, as it's obvious it doesn't play a role, but note that we do the same with e.g. `ConcentrationConstant`, where the input `mass_def` is arbitrary, but we keep it as an argument to have uniform signature everywhere.

pyccl/pk2d.py
nikfilippas: Okay, I'm fine to use `np.array_equal` but doesn't it strike bad when you realize that `pk != 2*pk/2`? If not, I'll do it.

pyccl/tracers.py
nikfilippas: no need, so resolving after our discussion above

pyccl/base.py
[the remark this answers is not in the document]
nikfilippas: Okay, I have replaced the vectorized for loop with an enclosed for loop.
I have also added the option to compare numpy arrays by wrapping the comparison with `any()`.

pyccl/pk2d.py
damonge: Yep, array_equal is fine. I think the fact that pk != 2*pk/2 makes perfect sense :-)

pyccl/halos/massdef.py
damonge: Let's just remove this. No need to add more stuff in v2 if we're not sure we'll keep it in v3

damonge: (meaning remove the whole mdeffof, we can introduce it in v3)

pyccl/pk2d.py
nikfilippas: Ok final question. How would it sound to you if we indeed made `rtol` a global var but set it by default to zero so it exactly emulates the behaviour of `np.array_equal`? I am making these changes now.

pyccl/halos/massdef.py
nikfilippas: We will 100% forget, there are too many things we'll have to remember. So I just removed `cm` as an argument for now.

pyccl/tracers.py
nikfilippas: done

pyccl/tracers.py
nikfilippas: done

pyccl/tk3d.py
nikfilippas: done re. tol set to zero

pyccl/pk2d.py
nikfilippas: done, I set a glob var in `parameters.py` because it was the easiest thing to do, rather than change everything to `array_equal`.

nikfilippas:
@damonge all done, back to you now and we can move on with the rest if you are happy with the changes.

pyccl/pk2d.py
damonge: This just adds more complexity to the code that we don'tneed. There aren't so many places where you need to make this change (2?)

pyccl/tk3d.py
damonge: array_equal. We don't need the extra complexity or the extra operations allclose performs

nikfilippas: done

pyccl/pk2d.py
nikfilippas: done

pyccl/base.py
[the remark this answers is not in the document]
damonge: Humm, I don't see it reverted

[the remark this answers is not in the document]
nikfilippas: github being a pain again, it has been reverted, if you pull the branch locally and do a grep you won't find anything

pyccl/halos/profiles_cib.py
damonge: Doesn't show as done

nikfilippas: oops indeed, that one was a stray because it lives in a different file, sorry! fixed now

pyccl/core.py
```diff
@@ -1167,6 +1180,9 @@ class CosmologyCalculator(Cosmology):
             corresponding to the "HALOFIT" transformation of
             Takahashi et al. 2012 (arXiv:1208.2701).
     """
+    __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
+                    "_accuracy_params", "_pk_lin", "_pk_nl",)
+
```
damonge, line 1185: Don't we need to check background and growth as well?

damonge (commented):
Just one (maybe major) comment for core.py

pyccl/core.py
damonge: To make your life easier, I'd be fine if you want to store the corresponding arrays at the python level for the comparison.

pyccl/base.py
```diff
@@ -590,8 +590,15 @@ def __hash__(self):
 
     def __eq__(self, other):
         # Two same-type objects are equal if their representations are equal.
-        if self.__class__ is not other.__class__:
+        if type(self) is not type(other):
```
damonge, line 593: Remind me, how does this compare subclasses?

pyccl/core.py
nikfilippas: Do we? Those arrays would uniquely describe the power spectra, right? So that extra check would be superfluous, but correct me if I'm wrong.

pyccl/tests/test_cosmology.py
```diff
@@ -4,7 +4,47 @@
 import numpy as np
 from numpy.testing import assert_raises, assert_, assert_no_warnings
 import pyccl as ccl
+import copy
 import warnings
+from .test_cclobject import check_eq_repr_hash
+
+
+def test_HMIngredients_eq_repr_hash():
```
damonge, line 12: Not sure the name of this test matches what it's doing...

damonge (commented):
Final comments on tests

pyccl/core.py
nikfilippas: Right, good call on including background and growth then.

This was way trickier than I thought it would be. If there are attributes that nest things which have to be compared, the previous implementation wouldn't work. It was watertight for simple int, float, str, and ndarray, but not for all sequences like list, tuple, str, set etc.

I created a helper function, taking advantage of the power of numpy, which can compare just about everything:
```python
def is_equal(this, other):
    """Powerful helper for equivalence checking."""
    try:
        np.testing.assert_equal(this, other)
        return True
    except AssertionError:
        return False
```

So now the lines in `__eq__` change to
```python
            for attr in self.__eq_attrs__:
                if not is_equal(getattr(self, attr), getattr(other, attr)):
                    return False
```

nikfilippas:
@damonge back to you

pyccl/core.py
damonge: How does `assert_equal` deal with complex CCL-like objects?

nikfilippas: It uses their internal `__eq__` method, however it is important to note that this never happens here, because what we compare is not CCLObjects, but merely arrays, sequences, numbers etc. But even if it were to compare CCLObjects, it would work.

damonge: Hmm, aren't the attributes being compared here sometimes `MassDef`s or `Concentration`s? 
The numpy documentation does say that they must be "two objects (scalars, lists, tuples, dictionaries or numpy arrays),"

pyccl/base.py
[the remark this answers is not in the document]
damonge: Locally I still see these lines:
```
# Fall back to repr comparison.                                                                                                                                                                            
return repr(self) == repr(other)
```

pyccl/core.py
nikfilippas: Yes. Sorry. I answered a whole different thing. The takeaway is that numpy can handle this:
```python
>>> ccl.base.is_equal(*[ccl.CosmologyVanillaLCDM for _ in range(2)])
True
```
but
```python
>>> Mvir = ccl.halos.MassDefVir()
>>> M200m = ccl.halos.MassDef200m()
>>> ccl.base.is_equal(Mvir, M200m)
False
```

pyccl/base.py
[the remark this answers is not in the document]
nikfilippas: As you should, this is the fallback in case `__eq_attrs__` is not specified.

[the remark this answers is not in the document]
nikfilippas: Hold on, what were you referring to in your original comment?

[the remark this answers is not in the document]
nikfilippas: Ok, reading it again I now realise you were referring to the fallback itself. I thought you meant the `attrgetter`.
I updated this to fall back to id-checking.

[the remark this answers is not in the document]
damonge: We wanted the fallback to be `object.__eq__`

damonge (approved):
OK, we're done. Merge when the tests pass.

PR #1039, opened 2023-03-20
Baryons in v3

damonge:
This PR sets up the framework for baryons in v3. More detailed description to come.

- [x] Remove bcm from C layer
- [x] Remove code snippet from rtd documentation (replace with something else).
- [x] Set up generic baryons container.
~~- [ ] Link container in pk2d object.~~
- [x] Replace internal use of old functionality.
- [x] Add deprecation decorator.


nikfilippas:
You might want to use some of the code I've written [here](https://github.com/nikfilippas/myccl/blob/main/myccl/pspec/bcm.py) to make your life easier.

damonge:
Description: baryonic effects are now controlled by `BaryonicEffect` objects. These contain a method `apply_baryons` that takes in a `Pk2D` object and a `Cosmology` object and returns another `Pk2D` object with baryons applied to it.

We provide both the overall superclass and a subclass for the existing BCM model.

This PR also removes all of the BCM code from C and puts it in python.

pyccl/tests/test_bcm.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import numpy as np
 import pyccl as ccl
+from pyccl.pyutils import assert_warns
```
nikfilippas, line 4: FYI: We have an `assert_warns` function in `pyutils` which we use in tests, however I will be removing all that because we test using `pytest` only and `pytest` has the wonderful `with pytest.warns(Warning):` context manager, used in the same way as `pytest.raises`. There is no need to mix in `assert_warns` and `numpy.testing` like we have in some tests, so just keep that it mind next time you add a test for warnings.

`warnings` and `numpy` are good packages, but for other kinds of stuff. We even say we use `pytest` [here](https://ccl.readthedocs.io/en/latest/source/writing_and_running_unit_tests.html) in the docs.

pyccl/baryons.py
nikfilippas, line 1: It would be nice to name every subclass and include a `from_name` class method in the superclass. You can use the following code which I've taken from `myccl`:
```python
    @classmethod
    def _subclasses(cls):
        return set(cls.__subclasses__()).union(
            [sub for cl in cls.__subclasses__() for sub in cl._subclasses()])

    @classmethod
    def from_model(cls, model):
        """
        """
        power_spectra = {p.name: p for p in cls._subclasses()}
        return power_spectra[model]
```

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        kg = 0.7 * bfunc4 * self.eta_b**(-1.6)
+        gf = bfunc / (1 + (kh/kg)**3) + 1. - bfunc
+        scomp = 1 + (kh / self.k_s)**2
+        fka = gf * scomp
+
+        if np.ndim(k) == 0:
+            fka = fka[0]
+        return fka
+
+    def update_parameters(self, log10Mc=None, eta_b=None, k_s=None):
```
nikfilippas, line 104: Cont: add `cosmo`

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+                the stellar component. Defaults to 55.0.
+        """
+        if log10Mc is not None:
+            self.log10Mc = log10Mc
+        if eta_b is not None:
+            self.eta_b = eta_b
+        if k_s is not None:
+            self.k_s = k_s
+
+    @unlock_instance(mutate=True, argv=2)
```
nikfilippas, line 122: Cont: remove decorator.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Args:
+        log10Mc (:obj:`float`): logarithmic mass scale of hot
+            gas suppression. Defaults to 14.08.
+        eta_b (:obj:`float`): ratio of escape to ejection radii (see
+            Teyssier et al. 2015). Defaults to 0.5.
+        k_s (:obj:`float`): Characteristic scale (wavenumber) of
+            the stellar component. Defaults to 55.0.
+    """
+    __repr_attrs__ = ("log10Mc", "eta_b", "k_s")
```
nikfilippas, line 68: Cont: add `cosmo`

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        if log10Mc is not None:
+            self.log10Mc = log10Mc
+        if eta_b is not None:
+            self.eta_b = eta_b
+        if k_s is not None:
+            self.k_s = k_s
+
+    @unlock_instance(mutate=True, argv=2)
+    def _apply_boost(self, cosmo, pk2d, in_place=False):
```
nikfilippas, line 123: Cont: remove `cosmo` and `in_place`

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        # Applies boost factor
+        if not isinstance(pk2d, Pk2D):
+            raise TypeError("pk2d must be a Pk2D object")
```
nikfilippas, line 126: No need to type-check this, it will fail with a useful error in the next line. There's a bunch of needless type-checks all over CCL which I am also planning to remove. If you really need to use type-checking, simply add a type hint in the arguments, so that users can run it through `mypy`.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        a_arr, lk_arr, pk_arr = pk2d.get_spline_arrays()
+        k_arr = np.exp(lk_arr)
+        fka = np.array([self.boost_factor(cosmo, k_arr, a) for a in a_arr])
```
nikfilippas, line 129: Cont: If we migrate C code to Python we have to avoid using for-loops at all costs! With `numpy` you can almost always avoid loops.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        pk_arr *= fka
+
+        logp = np.all(pk_arr > 0)
+        if logp:
+            pk_arr = np.log(pk_arr)
```
nikfilippas, line 134: Assuming the boost factor can't be negative, you can use the original pk's `logp` definition, and replace this with:
```python
        is_log = pk.is_logp
        if is_log:
            np.log(pk_out, out=pk_out)  # in-place log
```

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+
+        return self._new_pk(cosmo, pk2d, a_arr, lk_arr, pk_arr,
+                            logp, in_place)
```
nikfilippas, line 137: Cont: So here, we can simply return a new `Pk2D` object like so:
```python
        return Pk2D(a_arr=a_arr, lk_arr=lk_arr, pk_arr=pk_out, is_logp=is_log,
                    extrap_order_lok=pk.extrap_order_lok,
                    extrap_order_hik=pk.extrap_order_hik)
```

pyccl/baryons.py
nikfilippas, line 123: Let's rename `pk2d` to `pk` keep it simple. For arrays, we use the identifier `_arr` throughout CCL.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        Args:
+            cosmo (:class:`~pyccl.core.Cosmology`): Cosmological parameters.
+            k (float or array_like): Wavenumber; Mpc^-1.
+            a (float): Scale factor.
+
+        Returns:
+            float or array_like: Correction factor to apply to
+                the power spectrum.
+        """
+        k_use = np.atleast_1d(k)
```
nikfilippas, line 88: If we don't use the power of `numpy` here we'll end up using for-loops in Python later, which is a no-go zone, particularly for "fast" code we ported from C. We have to make both input arrays [broadcastable](https://numpy.org/doc/stable/user/basics.broadcasting.html), so replace this with:
```python
        a, k = map(np.atleast_1d, [a, k])
        a, k = a[:, None], k[None, :]
```
and now both inputs can be floats or arrays. Also, to make reading the code easier, avoid using `a_use`, `k_use` etc. since we don't operate on them in-place.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+
+        z = 1./a - 1.
+        kh = k_use / cosmo['h']
+        b0 = 0.105*self.log10Mc - 1.27
+        bfunc = b0 / (1. + (z/2.3)**2.5)
+        bfunc4 = (1-bfunc)**4
+        kg = 0.7 * bfunc4 * self.eta_b**(-1.6)
```
nikfilippas, line 95: Why not concatenate these two lines? `bfunc4` isn't used anywhere else, no need to assign more memory for it.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+            k (float or array_like): Wavenumber; Mpc^-1.
+            a (float): Scale factor.
+
+        Returns:
+            float or array_like: Correction factor to apply to
+                the power spectrum.
+        """
+        k_use = np.atleast_1d(k)
+
+        z = 1./a - 1.
```
nikfilippas, line 90: In some places of the code you use the decimal point, in some others you don't. I propose dropping it fully, as Python3 does float division by default, and code is more easily readable that way.

nikfilippas (changes requested):
Okay, that's a nice improvement, thank you (and apologies for the many comments)! Let's resolve the comments from `baryons.py` and I'll do a final pass to verify that everything we discuss here has been adapted accordingly.

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+            return None
+        else:
+            new = Pk2D(a_arr=a_arr, lk_arr=lk_arr, pk_arr=pk_arr,
+                       is_logp=logp,
+                       extrap_order_lok=pk2d.extrap_order_lok,
+                       extrap_order_hik=pk2d.extrap_order_hik)
+            return new
+
+
+class BaryonicEffectsBCM(BaryonicEffects):
```
nikfilippas, line 48: If we have moved `cosmo` to `__init__` and have defined `self.lk_arr` and `self.a_arr` using the sampling rates in `cosmo`, we can optionally add the following function to the class, which will return a `Pk2D` object with the correction. Users can then simply multiply the two `Pk2D` objects to get the corrected power spectrum.

And the don't have to worry about `pk.get_spline_arrays()` and array operations, because everything is already coded in for them.

```python
    def get_power_spectrum(self):
        pk = self.boost_factor(np.exp(self.lk_arr), self.a_arr)
        return Pk2D(a_arr=self.a_arr, lk_arr=self.lk_arr, pk_arr=np.log(pk),
                    is_logp=True, extrap_order_lok=0, extrap_order_hik=0)
```

So, they'd just do `pk * bar.get_power_spectrum()`

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class BaryonicEffects(CCLAutoreprObject):
+    """`BaryonicEffect` obects are used to include the imprint of baryons
+    on the non-linear matter power spectrum. Their main ingredient is a
+    method `apply_baryons` that takes in a `ccl.Pk2D` and returns another
+    `ccl.Pk2D` object that now accounts for baryonic effects.
+    """
+
+    def apply_baryons(self, cosmo, pk2d, in_place=False):
```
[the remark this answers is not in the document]
damonge: I disagree. I'd rather keep both things decoupled (just like we are doing with Pk2Ds and halo model calculators).

The point of having a superclass right now, is that we will in the future have other subclasses of it. It also holds the `_new_pk` method. If you can propose a useful `__init__()`, I'm all for it, but I'd rather not use cosmo as an excuse to create one.

pyccl/tests/test_bcm.py
damonge: Done

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
+from . import ccllib as lib
+from .base import CCLAutoreprObject
+from .pyutils import check
+from .pk2d import Pk2D
+from .base import unlock_instance
+import numpy as np
+
+
+class BaryonicEffects(CCLAutoreprObject):
```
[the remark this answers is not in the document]
damonge: Done

pyccl/baryons.py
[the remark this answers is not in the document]
damonge: I went with `include_baryonic_effects`

pyccl/baryons.py
damonge: Done

pyccl/baryons.py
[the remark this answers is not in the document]
damonge: I'd rather keep `in_place`. `Pk2D` objects are memory-heavy, and I can see good uses for this (in fact, this is how we actually use BCM in CCL).

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+    def apply_baryons(self, cosmo, pk2d, in_place=False):
+        """Apply baryonic effects to a given power spectrum.
+
+        Args:
+            cosmo (:class:`~pyccl.core.Cosmology`): Cosmological parameters.
+            pk2d (:class:`~pyccl.pk2d.Pk2D`): power spectrum.
+            in_place (:obj:`bool`): if True, the `pk2d` itself is modified,
+                instead of returning a new `ccl.Pk2D` object.
+        """
+        raise NotImplementedError("Do not use base `BaryonicEffects` class.")
```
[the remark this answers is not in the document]
damonge: Done

pyccl/baryons.py
damonge: As above, I disagree on coupling cosmo to this

pyccl/baryons.py
damonge: Disagree as above

pyccl/baryons.py
damonge: Disagree (renamed pk)

pyccl/baryons.py
damonge: Disagree

pyccl/baryons.py
damonge: Disagree

pyccl/baryons.py
damonge: Disagree

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+    """
+    __repr_attrs__ = ("log10Mc", "eta_b", "k_s")
+
+    def __init__(self, log10Mc=14.08, eta_b=0.5, k_s=55.0):
+        self.log10Mc = log10Mc
+        self.eta_b = eta_b
+        self.k_s = k_s
+        self.apply_baryons = self._apply_boost
+
+    def boost_factor(self, cosmo, k, a):
```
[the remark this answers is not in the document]
damonge: Disagree

pyccl/baryons.py
damonge: Done

damonge: and tested

damonge: and tested

damonge:
@nikfilippas note I've renamed BCM to `Schneider15`, which is more appropriate.

pyccl/baryons.py
damonge: Done

pyccl/baryons.py
damonge: Note that we need to keep track of the original shape of k and a to return in the right shape. So partially done (still use `a_use` and `k_use` to keep track of the original `k` and `a`)

pyccl/baryons.py
```diff
@@ -0,0 +1,137 @@
...
+        bfunc = b0 / (1. + (z/2.3)**2.5)
+        bfunc4 = (1-bfunc)**4
+        kg = 0.7 * bfunc4 * self.eta_b**(-1.6)
+        gf = bfunc / (1 + (kh/kg)**3) + 1. - bfunc
+        scomp = 1 + (kh / self.k_s)**2
+        fka = gf * scomp
+
+        if np.ndim(k) == 0:
+            fka = fka[0]
+        return fka
```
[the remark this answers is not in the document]
damonge: In the first case, I'm fairly sure this will return a scalar.

pyccl/baryons.py
```diff
@@ -0,0 +1,155 @@
...
+        else:
+            new = Pk2D(a_arr=a_arr, lk_arr=lk_arr, pk_arr=pk_arr,
+                       is_logp=logp,
+                       extrap_order_lok=pk.extrap_order_lok,
+                       extrap_order_hik=pk.extrap_order_hik)
+            return new
+
+    @classmethod
+    def _subclasses(cls):
+        return set(cls.__subclasses__()).union(
```
nikfilippas, line 56: Add a comment here indicating that this helper returns a set of all subclasses.

pyccl/baryons.py
```diff
@@ -0,0 +1,155 @@
...
+            [sub for cl in cls.__subclasses__() for sub in cl._subclasses()])
+
+    @classmethod
+    def from_name(cls, name):
+        """
+        """
```
nikfilippas, line 62: Write something in the docstring.

pyccl/baryons.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class Baryons(CCLAutoreprObject):
+    """`BaryonicEffect` obects are used to include the imprint of baryons
+    on the non-linear matter power spectrum. Their main ingredient is a
+    method `include_baryonic_effects` that takes in a `ccl.Pk2D` and
+    returns another `ccl.Pk2D` object that now accounts for baryonic
+    effects.
+    """
+    name = 'base'
```
nikfilippas, line 16: No need to name the base, it will never be instantiated.

nikfilippas: To prevent instantiation (emulate abstraction w/o having any abstract methods) you can use the following code:
```python
    def __init__(self):
        if type(self) == A:
            raise TypeError("Base class A cannot be instantiated.")
```

pyccl/baryons.py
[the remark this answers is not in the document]
nikfilippas: I don't think we should support this. Just like `numpy` and most other Python libraries we should add in the docs that "dimensions of size 1 are squeezed in the output" and have this as a rule everywhere throughout CCL. Either way, it is rare that users will pass an array consisting of a single number.

In fixing the docs, I am planning to include a `Returns` section to every docstring, which is extremely helpful for users.

pyccl/baryons.py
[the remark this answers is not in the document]
nikfilippas: First of all, the `_new_pk` method doesn't use `self` at all, so it could well have been a standalone function (if we decide to keep it).

The use-case here is not akin to `Pk2D` or the `HMCalculator`: here, we deal with parameters.

I don't see the point of keeping them decoupled. `cosmo` just holds parameters, and can be treated as such, giving `BaryonModel` the ability to update it if needed. What I propose is for `BaryonModel` to merely be a 'prescription', that needs the baryonic parameters + the cosmological parameters to work. And it acts on `pk`.

Otherwise, what is the purpose of having it in a class? We could simply move the baryonic parameters to `include_baryonic_feeback` as well, and just call it using `include_baryonic_feeback(pk, cosmo, **bar_params)`. This beats the whole purpose of a class though.

pyccl/baryons.py
```diff
@@ -0,0 +1,155 @@
+from . import ccllib as lib
+from .base import CCLAutoreprObject
+from .pyutils import check
+from .pk2d import Pk2D
+from .base import unlock_instance
+import numpy as np
+
+
+class Baryons(CCLAutoreprObject):
```
nikfilippas, line 9: This ought to be an abstract class, so define an `abstractmethod` called `_include_baryonic_effects` which all subclasses should override.

nikfilippas (changes requested):
Some additional things.

nikfilippas:
@damonge back to you, with one general proposal: since you created this abstract class that is supposed to be subclassed for particular models, should we put them all in a submodule, with each implementation in its own file, just like we are doing with `halos`? There is no need to do it right now, but this way we make the code future-proof.

pyccl/baryons.py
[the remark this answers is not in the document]
damonge: We do support it already in the halos module, and I do think it's good to be consistent. When you put this in a likelihood, sometimes it's hard to predict what size arrays you'll be passing around, and I very much prefer the output to be consistent with the input. I don't really care what numpy do.

And, by the way
```
>>> import numpy as np
>>> np.log10(np.array([1.0]))
array([0.])
```

pyccl/baryons.py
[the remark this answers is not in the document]
damonge: I'm not worried about time efficiency. It's mostly about convenience and memory eficiency. There are many examples of libraries that allow in-place operations, and I don't see a strong reason not to do that here.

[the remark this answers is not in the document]
damonge: For cleanliness, modularity, and flexibility. Two examples:
a) Likelihoods that take advantage of slow-fast split will perhaps want to set up cosmological and baryonic parameters in different parts of the code. This allows for a complete decoupling.
b) In the future, there may be other `Baryons` objects with expensive `__init__`s that would be best to run only once, and not every time the cosmology changes.
c) This formalism still allows for developers to create `Baryons` classes that must be initialised with a cosmology if necessary. I do not think it's a good idea, and definitely not for this model. The only reason why we actually use the cosmology is because the model was formulated in h-inverse units, but it could just as well have been defined without that, and the cosmology would be irrelevant.

pyccl/baryons.py
damonge: Done

pyccl/baryons.py
damonge: Done

pyccl/baryons.py
damonge: Doesn't hurt to name it then.

pyccl/baryons.py
damonge: Done

pyccl/baryons.py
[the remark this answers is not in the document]
nikfilippas: Okay, in case this behaviour is desired I am happy to leave dimensional checks as they are.
However, we can clean up a lot of code if we use a dispatcher that automatically does that, and not require that developers do it every time. I will create one and replace accordingly once we are done merging things to `v3_wip`.

Resolving.

pyccl/baryons.py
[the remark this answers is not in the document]
nikfilippas: I have already replied to the memory efficiency argument: memory efficiency is irrelevant because of the `gbc`.

As far as convenience is concerned, I'd say having in-place operations is error-prone and unnecessarily complicates the code. We introduced `CCLObject` to make objects immutable, and the point was to only use `@unlock_instance` where it is unavoidable. I can't stress the immutability argument enough. We should avoid this if we can.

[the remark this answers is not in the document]
damonge: I'd rather keep this, as reasoned above. We do this kind of operation in several places in the library. I am definitely not as worried about object immutability as you are, especially once proper `__eq__`s are in place.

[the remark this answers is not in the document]
nikfilippas: We don't - we only do it where it can't be avoided: dunder methods of `Pk2D`, setting power spectra in `Cosmology`, getting ingredients in `HMCalculator`, and adding new tracers in `Tracer`. Each of these use-cases is special and can't be worked around.

Here, we can work around that. Why would we want to mutate this (and add more lines of code), if we can completely avoid it?

pyccl/baryons.py
nikfilippas: Extra code which is never used.
The reason we rename the other is to be able to do `Baryons.from_name(name)`.

pyccl/baryons.py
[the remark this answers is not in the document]
damonge: I think I've explained my reasons above, and at this point I would just be repeating myself.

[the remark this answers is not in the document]
nikfilippas: We took this discussion on Slack and agreed to leave `cosmo` as is.

pyccl/baryons.py
damonge: Done.

nikfilippas (approved):
Okay, that's a great addition to the library. Approved!

nikfilippas:
@damonge before you merge, I just realized there are 6 uncovered lines in `BaryonsSchneider15`. Can you write a quick test where you update the parameters?

PR #1040, opened 2023-03-21
Tracers v3

damonge:
This PR sets up the framework for tracers in v3.


nikfilippas:
I think you have made a mistake here in branching - you should have checked out from `v3_wip` but it looks like you checked out from `baryons_v3`, resulting in this PR containing the changes from both PRs. Check, clean up the diffs to preserve only the ones related to `Tracers`, and I will then review.

damonge:
this is the same as your interconnected PRs last week. Sart from baryons, then tracers, then the other 2 in any order, and they get simpler as we merge them.

nikfilippas:
Ah okay, but this way the whole review process will be slower because we introduce a bottleneck that is there can only be 1 PR getting reviewed each time. So for now, I am only going to review baryons.

pyccl/tracers.py
```diff
@@ -434,8 +420,9 @@ def _get_MG_transfer_function(self, cosmo, z):
 
         return mg_transfer
 
+    @warn_api
```
nikfilippas, line 423: In `star` I had regrouped them because `is_logt` refers to the transfer function. Up to you if you want to do it.

pyccl/tracers.py
```diff
@@ -577,16 +564,33 @@ def __del__(self):
                 lib.cl_tracer_t_free(t)
 
 
-def NumberCountsTracer(cosmo, has_rsd, dndz, bias=None,
-                       mag_bias=None, n_samples=256):
+class NzTracer(Tracer):
+    """Specific for tracers with an internal `_dndz` redshift
+    distribution interpolator. Currently only used for
+    :class:`NumberCountsTracer` and :class:`WeakLensingTracer`.
```
nikfilippas, line 570: I'd remove those 2 lines, because once we implement a 3rd tracer, we will 99.9% forget to update this docstring.

pyccl/tracers.py
```diff
@@ -744,7 +752,39 @@ def CMBLensingTracer(cosmo, z_source, n_samples=100):
     return tracer
 
 
-def tSZTracer(cosmo, z_max=6., n_chi=1024):
+def ZPowerTracer(cosmo, *, A, alpha, z_min=0., z_max=6., n_chi=1024):
```
nikfilippas, line 755: I tested it and it won't complain if `z_max` < `z_min`. Can the C level handle this? If not, should we include a check?

nikfilippas, line 755: hmmm so Tracers that use this won't be instances of `ZPowerTracer`, but Tracers that use `NzTracer` will be instances of `NzTracer`.

I think the obvious solution, since this is a helper function, used to avoid repeated code, is to write it in the docs ("Helper that creates ..."), and rename it to something like `create_zpower_tracer` so that users won't confuse it for a class.

Or, (even better) add it as a `classmethod` to `Tracer`, so it can be called like `Tracer.from_zPower(...)`.

pyccl/tracers.py
```diff
@@ -8,12 +8,12 @@
                          growth_factor, scale_factor_of_chi, h_over_h0)
```
nikfilippas, line 8: All of these imports from `background` have become methods of `cosmo`. Could you replace all occurrences in the code below? It's cleaner, and the less we import at the top, the fewer the chances we get into cyclic import trouble in the future.

nikfilippas (changes requested):
Okay, here are my main comments.

pyccl/bcm.py
```diff
@@ -1,6 +1,7 @@
 from .base import unlock_instance
 from .baryons import BaryonsSchneider15
-from .pyutils import deprecated, check
+from .base import deprecated
+from .pyutils import check
```
nikfilippas, line 4: You are using `check(status, cosmo)` but `cosmo` isn't used at the C-level, so it won't carry any error messages even if there are. Remove `cosmo` (although this is now deprecated so we shouldn't bother).

pyccl/tracers.py
```diff
@@ -720,7 +727,8 @@ def WeakLensingTracer(cosmo, dndz, has_shear=True, ia_bias=None,
     return tracer
 
 
-def CMBLensingTracer(cosmo, z_source, n_samples=100):
+@warn_api
+def CMBLensingTracer(cosmo, *, z_source, n_samples=100):
```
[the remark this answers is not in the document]
damonge: I'd rather leave it without a default value (as it was before), so users need to know what they're doing. If I set it to 1100 by default it'll be very easy to forget that's what we use, and 1100 is really only an approximation to z_LSS.

pyccl/tracers.py
damonge: I saw that. I think I prefer this order (and it doesn't matter much, since now they're all keyword args.

pyccl/__init__.py
```diff
@@ -117,6 +117,7 @@
 from .cls import angular_cl
 from .tracers import (
     Tracer,
+    NzTracer,
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tracers.py
```diff
@@ -107,14 +110,14 @@ def get_lensing_kernel(cosmo, dndz, mag_bias=None, n_chi=None):
...
+            and cosmo.cosmo.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION):
         warnings.warn(
             f"The number of samples in the n(z) ({len(z_n)}) is smaller than "
             f"the number of samples in the lensing kernel ({n_chi}). Consider "
             f"disabling spline integration for the lensing kernel by setting "
-            f"pyccl.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION "
-            f"= False",
-            category=CCLWarning)
+            f"pyccl.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION = False"
+            f"before instantiating the Cosmology passed.", category=CCLWarning)
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tracers.py
```diff
@@ -107,14 +110,14 @@ def get_lensing_kernel(cosmo, dndz, mag_bias=None, n_chi=None):
...
+    if (n_chi > len(z_n)
+            and cosmo.cosmo.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION):
         warnings.warn(
             f"The number of samples in the n(z) ({len(z_n)}) is smaller than "
             f"the number of samples in the lensing kernel ({n_chi}). Consider "
             f"disabling spline integration for the lensing kernel by setting "
-            f"pyccl.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION "
-            f"= False",
-            category=CCLWarning)
+            f"pyccl.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION = False"
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tracers.py
damonge: Checking now

pyccl/tracers.py
damonge: Done

nikfilippas (changes requested):
One final comment.

nikfilippas (approved):
LGTM!

PR #1041, opened 2023-03-21
Background, boltzmann, and cls in v3

damonge:
V3 changes for `background.py`, `boltzmann.py`, and `cls.py` (now `cells.py`).


pyccl/__init__.py
nikfilippas, line 1: Add this to the file under "deprecations", to preserve API when doing `from pyccl import cls` and the like.
```python

def __getattr__(name):
    rename = {"cls": "cells"}
    if name in rename:
        from .errors import CCLDeprecationWarning
        import warnings
        warnings.warn(f"Module {name} has been renamed to {rename[name]}.",
                      CCLDeprecationWarning)
        name = rename[name]
        return eval(name)
    raise AttributeError(f"No module named {name}.")
```

benchmarks/test_power_nu.py
```diff
@@ -4,6 +4,10 @@
 
 import pyccl as ccl
 
+# NOTE: We now check up to kmax=23
+# because CLASS v3 has a slight mismatch
+# (6e-3) at higher k wavenumbers.
+KMAX = 23
```
nikfilippas, line 10: We install `class < 3` though.

nikfilippas (changes requested):
Just this comment here, plus the addition of `__getattr__` in my other comment.

benchmarks/test_power_nu.py
damonge: This was failing on my own computer, and I imagine it might for some users.

pyccl/__init__.py
damonge: Done

benchmarks/test_power_nu.py
nikfilippas: That's because you used `class > 3` I assume. Are tests and benchmarks supposed to be run by users too?
If yes, should we perhaps create a `test_cll.sh` at the root directory which automatically runs benchmarks and tests?

pyccl/__init__.py
```diff
@@ -150,6 +150,17 @@
...
 # Deprecated & Renamed modules
+def __getattr__(name):
+    rename = {"cls": "cells"}
+    if name in rename:
+        from .errors import CCLDeprecationWarning
+        import warnings
+        warnings.warn(f"Module {name} has been renamed to {rename[name]}.",
+                      CCLDeprecationWarning)
+        name = rename[name]
+        return eval(name)
```
nikfilippas, line 161: These lines are uncovered, can you add a test at the top of `test_cls.py`?

Also, can you rename `test_cls` to `test_cells` (and all occurrences of the word `cls` within)?
I found another mention of `cls`: `tests/test_pk2d.py:def test_pk2d_cls()`.
And in `benchmarks/test_cls.py`.

nikfilippas (changes requested):
Some final comments regarding coverage and renaming.

benchmarks/test_power_nu.py
damonge: Users don't have to run them, but they often do, and this precludes unnecessary issues being opened. I don't think the script is needed.

nikfilippas (approved):
LGTM!

PR #1047, opened 2023-03-24
Neutrinos, parameters, power in v3

damonge:
Well, that


nikfilippas (changes requested):
The main comment has to do with `T_ncdm`. It is a good idea to move it before we ship v3 so we don't have any more breakages in the future.

nikfilippas:
@damonge I created #1049 to save you time, moves `T_ncdm` to `Cosmology` and removes `T_CMB` from `nu_masses`. It has this PR as its base, so have a look at the diffs and if you agree let's merge it here.

damonge:
I think this is back to you @nikfilippas

nikfilippas (approved):
I flaked it - this ended up being trickier than we originally thought, but it looks good now!

PR #1048, opened 2023-03-24
Pk2D and Tk3D in v3

damonge:


pyccl/pk2d.py
```diff
@@ -527,3 +533,45 @@ def parse_pk2d(cosmo, p_of_k_a, is_linear=False):
...
+    except Exception:
+        raise ValueError("Can't use input function")
+
+    # Set k and a sampling from CCL parameters
+    a_arr = get_pk_spline_a(cosmo=cosmo)
+    lk_arr = get_pk_spline_lk(cosmo=cosmo)
+
+    # Compute power spectrum on 2D grid
+    pk_out = np.array([pkfunc(k=np.exp(lk_arr), a=a) for a in a_arr])
+    return a_arr, lk_arr, pk_out
```
nikfilippas, line 577: I know we are doing thing because I proposed it, but I just realized that, contrary to what I mentioned on Slack, we don't in fact need to pass both `pkfunc` and `cosmo`. With #918 we make `cosmo` optional and sampling is taken from the current `ccl.spline_params` state. I was the one to implement it, and I didn't remember it!

So here's an idea. Why don't we allow `pkfunc` for whoever wants to use it, but deprecate `cosmo` instead. If anyone wants to do nitpicking wrt sampling, they can just change `spline_params`. I think this way is simpler and causes the least disturbance in v3.

And there's no need to keep any of the try-except checking either. The function is used directly below that line. If it's wrong, it will exit with a useful message.

nikfilippas (commented):
One primary comment, and will look at the rest later.

pyccl/pk2d.py
damonge: made into classmethod

pyccl/pk2d.py
```diff
@@ -109,26 +120,56 @@ def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
...
+        Returns:
+            :class:`~pyccl.pk2d.Pk2D`
+                Power spectrum object.
+        """
+        # Set k and a sampling from CCL parameters
+        a_arr = get_pk_spline_a(cosmo=cosmo)
+        lk_arr = get_pk_spline_lk(cosmo=cosmo)
+
+        # Compute power spectrum on 2D grid
+        pk_arr = np.array([pkfunc(k=np.exp(lk_arr), a=a) for a in a_arr])
```
nikfilippas, line 168: And since we are creating this all-shiny new function, why don't we force users to vectorize it outside of CCL?
Best-case scenario, we don't have to run this slow loop. Worst-case scenario, users have to add an extra line of code before calling this function.

pyccl/pk2d.py
```diff
@@ -147,15 +188,21 @@ def from_model(cls, cosmo, model):
...
+            model (:obj:`str`)
+                model to use. These models allowed:
+                  - `'bbks'` (Bardeen et al. ApJ 304 (1986) 15)
+                  - `'eisenstein_hu'` (Eisenstein & Hu astro-ph/9709112)
+                  - `'eisenstein_hu_nowiggles'` (Eisenstein & Hu astro-ph/9709112)
+                  - `'emu'` (arXiv:1508.02654).
+        Returns:
+            :class:`~pyccl.pk2d.Pk2D`
+                The power spectrum of the input model.
+        """  # noqa E501
```
nikfilippas, line 203: Don't worry about fixing the docstring style for now, even this won't show right. I've opened a PR for that, but before we do anything we have to read numpy's style guide.

pyccl/pk2d.py
```diff
@@ -182,21 +229,16 @@ def from_model(cls, cosmo, model):
...
-        power spectrum in `pk_linear`.
-
-        Args:
-            cosmo (:class:`~pyccl.core.Cosmology`):
-                A Cosmology object.
-            pk_linear (:class:`Pk2D`, optional):
-                A :class:`Pk2D` object containing the linear power spectrum
-                to transform.
+        Takahashi et al. 2012 (arXiv:1208.2701) on an input linear power
+        spectrum in `pk_linear`. See ``Pk2D.apply_nonlin_model`` for details.
```
nikfilippas, line 241: Well, I don't think this function exists, either you forgot to copy it, or you decided it wasn't worth it. I don't mind either way.

pyccl/tk3d.py
nikfilippas, line 1: Let's deprecate `eval` here as well.

nikfilippas: Also, as a general comment, why is `Tk3D.eval` like this? Since it's a function of (k1, k2, a), shouldn't it be called like `tk(k, k, a)`? Currently, it returns the entire sub-matrix. But what if I simply want it to give me (k1, k2, a)?

nikfilippas (changes requested):
Good stuff, these are my comments.

pyccl/pk2d.py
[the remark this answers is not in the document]
damonge: that's all done

pyccl/pk2d.py
damonge: I'd rather not force users to do this, for the simple reason that we haven't done so so far, and this will only make the transition even more painful than it's already going to be. This can be done later on (check if vectorised, otherwise we vectorise).

pyccl/pk2d.py
```diff
@@ -79,17 +91,16 @@ class Pk2D(CCLObject):
...
-    def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
-                 is_logp=True, extrap_order_lok=1, extrap_order_hik=2,
-                 cosmo=None, empty=False):
+    @warn_api(reorder=["pkfunc", "a_arr", "lk_arr", "pk_arr", "is_logp",
+                       "extrap_order_lok", "extrap_order_hik", "cosmo"])
+    def __init__(self, *, a_arr=None, lk_arr=None, pk_arr=None,
+                 pkfunc=None, cosmo=None, is_logp=True,
+                 extrap_order_lok=1, extrap_order_hik=2, empty=False):
         if empty:
             return
```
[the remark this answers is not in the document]
damonge: Done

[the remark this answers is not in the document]
nikfilippas: Ah, good call on pointing to `CCLObject.__new__()` rather than `object.__new__()`, to keep all other internal attrs! Thanks!

pyccl/tk3d.py
damonge: Deprecated. It's just for convenience. At least so far this has been the single use of these functions.

pyccl/pk2d.py
nikfilippas: That's a fair point, although you were about to completely deprecate `pkfunc`.

FYI, doing it in the future won't work without breaking API though. The way to check if a function is vectorized is either by adding an extra argument that users have to declare very time, or by try/except.
In the try/except block, if the function isn't vectorized and `len(k_arr) == len(a_arr)`, `numpy` will automatically broadcast them together, and do a pairwise calculation, which is really **not** what the user intended for.

So, if we ever want to implement this, and save us from running Python loops, the best time is now. Otherwise it's going to have to stay like this until v4.

pyccl/base/repr_.py
nikfilippas, line 1: There is an `empty=True` in the `Tk3D` representation as well.

pyccl/base/repr_.py
```diff
@@ -206,7 +206,7 @@ def build_string_Pk2D(self, na=6, nk=6, decimals=2):
             +===============+=============================================+
     """
     if not self.has_psp:
-        return "pyccl.Pk2D(empty=True)"
+        return "pyccl.Pk2D(empty)"
```
nikfilippas, line 209: There is an `empty=True` in the `Tk3D` representation as well.

pyccl/pk2d.py
```diff
@@ -109,26 +120,56 @@ def __init__(self, pkfunc=None, a_arr=None, lk_arr=None, pk_arr=None,
...
                                                         int(extrap_order_hik),
                                                         int(is_logp), status)
-        check(status, cosmo=cosmo)
+        check(status)
+
+    @classmethod
+    def from_function(cls, pkfunc, *, is_logp=True, cosmo=None,
+                      extrap_order_lok=1, extrap_order_hik=2):
+        """ Generates a `Pk2D` object from a function that calculates a power
+        spectrum.
```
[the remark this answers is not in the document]
damonge: This involves a lot of changes to pyutils and I think it's not worth it. As it is right now, `cosmo` is an optional argument (so we don't really depend on it), and we clearly state that it's only used to figure out sampling rates.

With the current structure of the library, it is kind of standard for `Cosmology`s to be used as a container for spline parameters (besides everything else), and this is not the only example of it.

damonge:
All done, back to you @nikfilippas

pyccl/pk2d.py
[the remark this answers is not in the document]
nikfilippas: Literally the only thing you need to change in `pyutils.py` is:
```python
def get_pk_spline_lk(cosmo=None, spline_params=spline_params):

def get_pk_spline_a(cosmo=None, spline_params=spline_params):
```
Just these two lines, replace them with what I show here.
If you think it's not worth it, let's leave it.

pyccl/pk2d.py
damonge: Since we're not even sure this functionality is very useful, I'd rather minimize pain for those who use it now.

And we are completely deprecating pkfunc!

pyccl/base/repr_.py
damonge: Yep, fixed.

pyccl/pk2d.py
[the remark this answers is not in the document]
damonge: I've tried. It gets completely confused with the C-level spline parameters. I don't think it's worth it.

nikfilippas:
@damonge as discussed on Slack, I added the option to initialize the `Pk2D` object from `spline_parameters`. Internally, all it really needs is something that has the `__get__` method (i.e. the dot) to access the spline parameters. So, this will work with both the C-level API, and the Python API, which is nice.

Users can either pass `cosmo.cosmo.spline_params` (if they prefer the old way) or an instance of `ccl.parameters.SplineParams`, (such as `ccl.spline_params`).

pyccl/pk2d.py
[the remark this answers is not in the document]
nikfilippas: This is now more efficient than our original implementation as the for-loop runs only at the C/Python interface and bypasses the overhead checks. Resolving.

pyccl/pk2d.py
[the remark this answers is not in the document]
nikfilippas: In my proposal I had confused `cosmo.cosmo.spline_params` with `cosmo._spline_params` (the latter is not to be used either way). I pushed something that works, awaiting your approval so we can resolve.

pyccl/pk2d.py
```diff
@@ -217,7 +271,7 @@ def apply_halofit(self, cosmo, pk_linear=None):
                                "redshifts. If using the calculator mode, "
                                "check the support of the background data.")
 
-        pk2d = Pk2D(empty=True)
+        pk2d = CCLObject.__new__(Pk2D)
```
nikfilippas, line 274: Can we make this `Pk2D.__new__(Pk2D` just in case we add a custom `__new__` method in the future?

pyccl/pk2d.py
```diff
@@ -229,13 +283,26 @@ def apply_halofit(self, cosmo, pk_linear=None):
         return pk2d
 
     def eval(self, k, a, cosmo=None, *, derivative=False):
-        """Evaluate the power spectrum or its logarithmic derivative.
+        warnings.warn("Pk2D.eval is deprecated. Simply use the object's "
+                      "__call__ method.", category=CCLDeprecationWarning)
```
nikfilippas, line 287: I would phrase it as "simply call the instance" or "simply call the object" because users might not know that `__call__` can be used directly like this.

pyccl/pk2d.py
```diff
@@ -229,13 +283,26 @@ def apply_halofit(self, cosmo, pk_linear=None):
...
+        return self.__call__(k, a, cosmo=cosmo, derivative=derivative)
```
nikfilippas, line 288: So, here, you can just do `self(k, a, ...)`

pyccl/pk2d.py
```diff
@@ -229,13 +283,26 @@ def apply_halofit(self, cosmo, pk_linear=None):
...
+
+    def eval_dlogpk_dlogk(self, k, a, cosmo):
+        """Evaluate logarithmic derivative. See ``Pk2D.eval`` for details."""
+        warnings.warn("Pk2D.eval_dlogpk_dlogk is deprecated. Simply use "
+                      "the object's __call__ method with `derivative=True`.",
+                      category=CCLDeprecationWarning)
```
nikfilippas, line 294: Same here about the warning message.

pyccl/pk2d.py
```diff
@@ -229,13 +283,26 @@ def apply_halofit(self, cosmo, pk_linear=None):
...
+        return self.__call__(k, a, cosmo=cosmo, derivative=True)
```
nikfilippas, line 295: Same here about the direct call on `self`.

pyccl/pk2d.py
```diff
@@ -254,56 +321,52 @@ def eval(self, k, a, cosmo=None, *, derivative=False):
...
+    # so we don't have to initialize one every time no `cosmo` is passed.
+    # This is gentle with memory too, as `free` does not work for an empty
+    # cosmology.
+    __call__._cosmo = type("Dummy", (object,), {"cosmo": lib.cosmology()})()
 
     def copy(self):
         """Return a copy of this Pk2D object."""
         if not self:
-            return Pk2D(empty=True)
+            return CCLObject.__new__(Pk2D)
```
nikfilippas, line 369: Same here about `Pk2D.__new__`.

pyccl/tk3d.py
```diff
@@ -138,41 +141,43 @@ def has_tsp(self):
...
-                                            a, nk*nk, status)
-            f = f.reshape([nk, nk])
-        check(status)
-        return f
+        status = 0
+        for ia, aa in enumerate(a_use):
+            f, status = lib.tk3d_eval_multi(self.tsp, lk_use,
+                                            aa, nk*nk, status)
+            check(status)
+            out[ia, :, :] = f.reshape([nk, nk])
```
nikfilippas, line 174: simply `out[ia]`

nikfilippas (changes requested):
Ok this is my final pass with a bunch of very minor comments, mainly having to do with helping users get accustomed to the new API.
Before everything though, check that you are happy with my changes to `Pk2D.from_function`.

pyccl/pk2d.py
damonge: We'll realise then if that happens.

pyccl/pk2d.py
damonge: Done

pyccl/pk2d.py
damonge: Done

pyccl/pk2d.py
damonge: Done

pyccl/pk2d.py
```diff
@@ -254,56 +321,52 @@ def eval(self, k, a, cosmo=None, *, derivative=False):
...
+            f, status = eval_func(self.psp, lk_use, aa,
+                                  cosmo.cosmo, k_use.size, status)
+
+            # Catch scale factor extrapolation bounds error.
+            if status == lib.CCL_ERROR_SPLINE_EV:
+                raise TypeError(
+                    "Pk2D evaluation scale factor is outside of the "
+                    "interpolation range. To extrapolate, pass a Cosmology.")
+            check(status, cosmo)
+            out[ia, :] = f
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tests/test_pkhm.py
```diff
@@ -173,15 +173,15 @@ def test_pkhm_pk2d():
     pk2d = ccl.halos.halomod_Pk2D(COSMO, hmc, P1,
                                   lk_arr=np.log(k_arr),
                                   a_arr=a_arr, normprof1=True)
-    pk_arr_2 = np.array([pk2d.eval(k_arr, a, COSMO)
+    pk_arr_2 = np.array([pk2d(k_arr, a, COSMO)
                          for a in a_arr])
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tests/test_pkhm.py
```diff
@@ -173,15 +173,15 @@ def test_pkhm_pk2d():
...
     assert np.all(np.fabs((pk_arr / pk_arr_2 - 1)).flatten()
                   < 1E-4)
 
     # Standard sampling
     pk2d = ccl.halos.halomod_Pk2D(COSMO, hmc, P1,
                                   normprof1=True)
-    pk_arr_2 = np.array([pk2d.eval(k_arr, a, COSMO)
+    pk_arr_2 = np.array([pk2d(k_arr, a, COSMO)
                          for a in a_arr])
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tk3d.py
```diff
@@ -138,41 +141,43 @@ def has_tsp(self):
         return 'tsp' in vars(self)
 
     def eval(self, k, a):
-        """Evaluate trispectrum. If `k` is a 1D array with size `nk`, the
-        output `out` will be a 2D array with shape `[nk,nk]` holding
-        `out[i,j] = T(k[j],k[i],a)`, where `T` is the trispectrum function
-        held by this `Tk3D` object.
+        warnings.warn("Tk3D.eval is deprecated. Simply use the object's "
+                      "__call__ method.", category=CCLDeprecationWarning)
```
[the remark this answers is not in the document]
damonge: Done

pyccl/pk2d.py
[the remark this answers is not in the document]
damonge: Fine with me

damonge:
Back to you @nikfilippas

pyccl/pk2d.py
nikfilippas: Don't bet on it, usage of `__new__` is kind of obscure and not everyone knows what it does. I am just pointing out a potential future bug, if you think it's not worth it, resolve.

pyccl/pk2d.py
nikfilippas: as above, resolve if you don't think it's a problem.

nikfilippas:
@damonge ok one final review comment remaining, and it's done!

pyccl/pk2d.py
damonge: Done

pyccl/pk2d.py
damonge: Done

damonge:
Back to you @nikfilippas

nikfilippas (approved):
LGTM!

PR #1049, opened 2023-03-24
Make `T_ncdm` a cosmological parameter + (v3)

nikfilippas:
This PR:

1. Makes `T_ncdm` an argument of `Cosmology` with a default value taken from `ccl.physical_constants` so that the C-code is decoupled from the constants and looks at the cosmological parameters.
2. Deprecates `T_CMB` in `ccl.neutrinos.nu_masses` as it is not used anywhere in the function.
3. Removes coupling between C, Python, and the physical constants for `T_CMB` as well.
4. Moves the `Omega_g` calculation to C to avoid tinkering with C code at the Python level.
5. Removes `norm_pk`: we used to determine whether `A_s` or `sigma8` was passed according to the value of `norm_pk`, which is a dangerous way of doing things.


include/ccl_core.h
```diff
@@ -82,6 +82,12 @@ typedef struct ccl_physical_constants {
   /**
    * T_ncdm, as taken from CLASS, explanatory.ini
    */
+  double T_ncdm;
+
+  /**
+   * TNCDM: repeat the value to preserve API because we renamed TNCDM to T_ncdm.
+   * TODO: Remove for CCLv3.
+   */
```
damonge, line 90: Why are you renaming anything if you just want to move this to the cosmological parameters?

damonge: Also, we should remove T_CMB from physical constants.

benchmarks/test_distances.py
```diff
@@ -46,7 +46,7 @@
 def Neff_from_N_ur_N_ncdm(N_ur, N_ncdm):
     """Calculate N_eff from the number of relativistic
     and massive neutrinos."""
-    Neff = N_ur + N_ncdm * ccl.physical_constants.TNCDM**4 / (4./11.)**(4./3.)
+    Neff = N_ur + N_ncdm * ccl.physical_constants.T_ncdm**4 / (4./11.)**(4./3.)
```
damonge, line 49: since this is gonna go away in v3, change so we use the v3 way

benchmarks/test_tracers.py
```diff
@@ -20,6 +20,9 @@ def get_prediction(ells, chi_i, chi_f, alpha, beta, gamma,
 
 @pytest.fixture(scope='module')
 def set_up():
+    # TODO: Replace once #1050 is merged.
+    # ccl.physical_constants.unfreeze()
+    object.__setattr__(ccl.physical_constants, "_frozen", False)
```
damonge, line 25: Do this

benchmarks/test_tracers.py
```diff
@@ -20,6 +20,9 @@ def get_prediction(ells, chi_i, chi_f, alpha, beta, gamma,
...
     ccl.physical_constants.T_CMB = 2.7
```
damonge, line 26: We don't want this

pyccl/core.py
```diff
@@ -179,6 +180,9 @@ class Cosmology(CCLObject):
         extra_parameters (:obj:`dict`, optional): Dictionary holding extra
             parameters. Currently supports extra parameters for CAMB, with
             details described below. Defaults to None.
+        T_ncdm (:obj:`float`): Non-CDM temperature in units of photon
+            temperature. The default is ``pyccl.physical_constants.T_ncdm``.
+            # TODO: Move after T_CMB for CCLv3.
```
damonge, line 185: Don't include # TODO in docstring

pyccl/core.py
```diff
@@ -419,12 +424,18 @@ def _build_config(
...
+            w0=None, wa=None, T_CMB=None, T_ncdm=None,
             bcm_log10Mc=None, bcm_etab=None, bcm_ks=None,
             mu_0=None, sigma_0=None, c1_mg=None, c2_mg=None, lambda_mg=None,
             z_mg=None, df_mg=None, Omega_g=None,
             extra_parameters=None):
         """Build a ccl_parameters struct"""
+        # Fill-in defaults (SWIG converts `numpy.nan` to `NAN`)
+        A_s = np.nan if A_s is None else A_s
+        sigma8 = np.nan if sigma8 is None else sigma8
+        T_CMB = const.T_CMB if T_CMB is None else T_CMB
```
damonge, line 436: I don't want us to use the constants for CMB or for ncdm. There's no reason why those should be there.

pyccl/core.py
```diff
@@ -419,12 +424,18 @@ def _build_config(
...
+        Omega_g = np.nan if Omega_g is None else Omega_g
+        T_ncdm = const.T_ncdm if T_ncdm is None else T_ncdm
```
damonge, line 438: ditto here. Let's write the values here directly. It's the only place in the code where they need to appear

pyccl/core.py
```diff
@@ -1166,13 +1151,17 @@ class CosmologyCalculator(Cosmology):
             computed. The only non-linear model supported is `'halofit'`,
             corresponding to the "HALOFIT" transformation of
             Takahashi et al. 2012 (arXiv:1208.2701).
+        T_ncdm (:obj:`float`): Non-CDM temperature in units of photon
+            temperature. The default is ``pyccl.physical_constants.T_ncdm``.
+            # TODO: Move to after T_CMB for CCLv3.
```
damonge, line 1156: no # TODO in docstring

pyccl/neutrinos.py
```diff
@@ -40,8 +42,11 @@ def Omega_nu_h2(a, *, m_nu, T_CMB=None):
 
     N_nu_mass = len(m_nu)
 
-    # Call function
-    OmNuh2, status = lib.Omeganuh2_vec(N_nu_mass, T_CMB,
+    # Fill-in defaults
+    T_CMB = const.T_CMB if T_CMB is None else T_CMB
+    T_ncdm = const.T_ncdm if T_ncdm is None else T_ncdm
```
damonge, line 47: As before, remove these from const

readthedocs/source/notation_and_other_cosmological_conventions.rst
```diff
@@ -249,7 +249,7 @@ basic physical constants
   - CLIGHT: speed of light in m/s
   - EV_IN_J: conversion factor between electron volts and Joules
   - T_CMB: temperature of the CMB in K
-  - TNCDM: temperature of the cosmological neutrino background in K
+  - T_ncdm: temperature of the cosmological neutrino background in K
```
damonge, line 252: remove this and TCMB

src/ccl_core.c
```diff
@@ -226,6 +226,12 @@ ccl_physical_constants ccl_constants = {
    */
   0.71611,
 
+  /**
+   * TNCDM: repeat the value to preserve API because we renamed TNCDM to T_ncdm.
+   * TODO: Remove for CCLv3.
+   */
+  0.71611,
+
```
damonge, line 234: again, why are you renaming anything in the constants, if you're gonna remove them in v3 anyway?

damonge (changes requested):

include/ccl_core.h
nikfilippas: For v3 we aim for consistency. It's T{underscore}CMB, so it would make sense to have T_ncdm instead of the hard to read `T_ncdm`. The latter is what CLASS uses too.

benchmarks/test_distances.py
nikfilippas: oops I missed that, thanks

nikfilippas: I need your input here. As you might have seen, there are several places where we need the value of `T_ncdm` where `cosmo` is not passed. We could either copy+paste the same value everywhere, which will make it hard to maintain, or we can have it stored somewhere, in a module, and access it from there.

I chose to leave it in `physical_constants` as that was its original place. Maybe a more fitting location would be in `core.py`, i.e. name a variable `T_ncdm = ...` and when we need it, do `from pyccl.core import T_ncdm`. However, I think this option would be a little non-intuitive.

I propose we leave it in the constants, or create a separate class in `parameters.py` which holds things like that, and would be accessed via `ccl.special_params.T_ncdm`.

benchmarks/test_tracers.py
nikfilippas: done

benchmarks/test_tracers.py
nikfilippas: i think that's a double comment, resolve if ok now

pyccl/core.py
nikfilippas: oh yes, what was I thinking! Thanks, good catch!

pyccl/core.py
nikfilippas: There is for `T_CMB`. I tried removing it, I wanted to get rid of it more than you do. However, the way this is currently set up in `core.c`, if you don't do it that way all of CCL breaks.
Re `T_ncdm` I replied above.

pyccl/core.py
nikfilippas: Try that and see what happens, a wormhole will open.

pyccl/core.py
nikfilippas: yes, thanks!

pyccl/neutrinos.py
nikfilippas: as above

readthedocs/source/notation_and_other_cosmological_conventions.rst
nikfilippas: as above

nikfilippas: yes, so we can remove them from here so that people won't find them in the docs any more, and keep them for internal use maybe? what do you think?

src/ccl_core.c
nikfilippas: To preserve API if someone tries to do `ccl.physical_constants.TNCDM` vs the new one `ccl.physical_constants.T_ncdm`.

damonge:
OK, I've marked everything above resolved. Let's start from scratch and do this properly.
a) T_CMB and T_ncdm should only be set from python when creating a cosmology, and their default values should only live in python.
b) We can leave the rubbish copies of them in `physical_constants` but they'll be deprecated in v3.

Such a nightmare.

nikfilippas:
To do this we should port cosmology initialization fully to Python. In `myccl` I have already done that. Would you like me to port the code to a PR that has this one as its base? And we can get rid of `ccl_cosmology_fill_initial` at the C-level.

damonge:
I'd need to look at the code for that, and I cannot guarantee I'll like it.

nikfilippas:
It is the exact same as in C, but written in Python, and lives in `core.py` instead of `ccl_core.c`.

src/ccl_core.c
```diff
@@ -306,26 +306,24 @@ TASK: fill parameters not set by ccl_parameters_create with some initial values
...
   // kg / m^3
   double rho_crit =
     ccl_constants.RHO_CRITICAL *
     ccl_constants.SOLAR_MASS/pow(ccl_constants.MPC_TO_METER, 3) *
     pow(params->h, 2);
-  params->Omega_g = rho_g/rho_crit;
 
   // Get the N_nu_rel from Neff and N_nu_mass
-  params->N_nu_rel = params->Neff - params->N_nu_mass * pow(ccl_constants.TNCDM, 4) / pow(4./11.,4./3.);
+  params->N_nu_rel = params->Neff - params->N_nu_mass * pow(params->T_ncdm, 4) / pow(4./11.,4./3.);
```
damonge, line 326: Isn't all of this happening in python now?

src/ccl_core.c
```diff
@@ -339,18 +337,27 @@ void ccl_parameters_fill_initial(ccl_parameters * params, int *status)
   // If non-relativistic neutrinos are present, calculate the phase_space integral.
   if((params->N_nu_mass)>0) {
     params->Omega_nu_mass = ccl_Omeganuh2(
-      1.0, params->N_nu_mass, params->m_nu, params->T_CMB, status) / ((params->h)*(params->h));
+      1.0, params->N_nu_mass, params->m_nu, params->T_CMB, params->T_ncdm,
+      status) / ((params->h)*(params->h));
```
damonge, line 341: Same question

src/ccl_core.c
```diff
@@ -339,18 +337,27 @@ void ccl_parameters_fill_initial(ccl_parameters * params, int *status)
...
+    // No value passed for Omega_g
+    params->Omega_g = rho_g/rho_crit;
+  }
+  else {
+    // Omega_g was passed - modify Omega_l
+    double total = rho_g/rho_crit + params->Omega_l;
+    params->Omega_l = total - params->Omega_g;
+  }
+
+  // NULL to NAN in case it is not set
```
damonge, line 360: Same question

readthedocs/source/notation_and_other_cosmological_conventions.rst
damonge: Unresolved this. So we agreed to remove them from the dics, right?

benchmarks/test_tracers.py
damonge: Remind me, what's happening here?

damonge: Oh I understand. I think you can do what the comment says, right?

nikfilippas: For one more time github's diff tool isn't showing the diffs right. I have changed that, as you can see from [here](https://github.com/LSSTDESC/CCL/blob/nu_Tncdm_v3/benchmarks/test_tracers.py)

readthedocs/source/notation_and_other_cosmological_conventions.rst
nikfilippas: yes, I removed them

src/ccl_core.c
nikfilippas: Yes, but halofit and musigma at the C-level use these to create copy of the cosmologies.

src/ccl_core.c
nikfilippas: same answer

src/ccl_core.c
nikfilippas: same answer

damonge (approved):
OK, done. Let the tests finish before merging.

Issue #981, opened 2022-08-09
Issues/Pull request templates


nikfilippas:
There have been discussions about the non-uniformity of the current code, which mostly arises because each contributor has their own style. A developer's guide is available but outdated. We can make use of [Github templates](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository) for new Issues and Pull Requests. This can not only streamline review and make Issue categorization faster, but it can also make sure we don't forget anything when new code is added.

Issue #871, opened 2021-04-14
Link equations to functions in docs


elisachisari:
A suggestion I received: It would be good to make it clear which functions correspond to which equations in the docs. For example, can we give equations for each function here? https://ccl.readthedocs.io/en/latest/api/pyccl.correlations.html

damonge:
Yeah, I started doing this with the halo model, but it's far from homogeneous. Having this would be great.

Issue #710, opened 2019-11-22
a description of CCL modules and their inter-connections needed for the Developer Guide


mishakb:
Matt wrote a very useful Developer Guide for CCL that can be found at https://ccl.readthedocs.io/en/v2.0.0/index.html
(this makes work on CCL development much easier!)
It would be great if we can have an additional description there of CCL modules and their inter-connections (which one calls which).  This can be done using a flowchart or an organigramme.  @beckermr @elisachisari

PR #1053, opened 2023-04-01
Docs v3

nikfilippas:
I have started the overhaul of the docstrings:

1. Provide (much) more info about what every piece of code does, with mathematical formulas, definitions, implementation details etc.
2. Uniform style guide following [Numpy](https://numpydoc.readthedocs.io/en/latest/format.html).

We generally need more docstrings and less code. Below is a list of the so far completed docstrings.

**Halos**
- [x] halo concentrations
- [x] halo biases
- [x] halo mass functions
- [x] halo model calculator
- [x] halo model pk Npt
- [x] halo mass definitions
- [x] halo profiles
- [x] halo profile 2-pt correlators
- [x] halo model base classes

**Main**
- [x] baryons & bcm
- [x] base & parameters
- [x] background
- [x] boltzmann
- [x] cells
- [x] cosmology
- [x] correlations
- [x] covariances
- [x] errors
- [x] ~halomodel, haloprofile, massfunction~
- [x] neutrinos
- [x] pk2d
- [x] power
- [x] pyutils
- [x] tk3d
- [x] tracers

**nl_pt**
- [ ] ?

Write a developer guide [draft]
--------------------------------
- [x] Write equations where possible in the docstring. Provide examples with "Examples" section. (#871 )
- [x] Better dev guide. Explain how the modules work together. (#710 ) Things to include: (1) Separate POK/KWO. (2) Expose to public API with `__all__`. (3) Adheres to the style guide. (4) Write in changelog, contributors etc. (5) Docs/Errors/Warnings: Concise language & messages, (dev guide to provide good/bad examples). No filler words. (6) Don't populate namespace needlessly; see if new feature can be integrated into existing API. No boilerplate code; too much repetition prob means consolidate to single func. (7) Don't import if func is a method of Cosmology. (8) "Returns" & "Raises" section for functions and classes where applicable. (9) Declare useful new attributes in new classes. (10) Prefer inheritance over flags. (11) Prefer type-hinting funcs than docs (except complicated ndarray broadcasting - use scipy conventions there). (12) How to avoid circular imports. (13) Code cov check. (14) No new warnings in unit testing. (15) Import system: builtins/3rd libs/local in AB order. (16) Think of public func names (i.e. is `get_blah` really a getter? is there a Python builtin instead e.g. `__call__`). Funcs generally act on something so name should start with a verb. (17) Specify ndarray type covers numbers. (18) Providing references/bibliography. (19) Cross-referencing. (20) Raw (r-) strings for equations. (21) Provide units. (22) Single vs double backticks. (23) Subclasses with just `__init__` should probs be funcs. (24) Avoid dict input for extra func parameters. (25) Maintain consistent arg name & order with rest of library. (26) No extra var declarations before return; use direct expression. (27) Try to deal with errors first and have clean, flowing code later instead of if-elif-elif-...-else chains. (28) Same with if-else-return blocks. (29) Funcs/classes. 1-line quick description first, then the rest. (30) New modules: quick description at the top. (31) Properties right after `__init__`. (32) Docs deprecated/versionadded.
- [ ] Issues/PR templates with checklist to assist both. (#981 ) Box "I have read the dev guide".
- [x] RTD build in CI + check new docs look fine during PR review (use other CCL docs as example).


damonge:

PR #1056, opened 2023-04-02
PT biases in v3

damonge:
The usage of the new PT module would be as follows:

```python

# Initialize a perturbation theory calculator
ptc = ccl.nl_pt.EulerianPTCalculator(with_NC=True, cosmo=cosmo)

# If you need to update the PT P(k) kernels to a new cosmology:
ptc.update_ingredients(cosmo)

# Now you can calculate biased power spectra, e.g.:
tg = ccl.nl_pt.PTNumberCountsTracer(b1=1.0, b2=0.5)
tm = ccl.nl_pt.PTMatterTracer()
pk_gg = ptc.get_pk2d_biased(tg, tracer2=tg)
pk_gm = ptc.get_pk2d_biased(tg, tracer2=tm)

# You can also ask for specific P(k) templates.
# E.g. this will get you the <delta * s^2> template
pk_d1_s2 = ptc.get_pk2d_template('b1:bs')
```


anicola:
Thanks a lot @damonge! I haven't looked much at the IAs as I am not so familiar with it, but pgg ang pgm look good to me! Just to check - this assumes that stochasticity will be handled outside CCL, right?

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,684 @@
...
+
+    .. note:: Terms of the form
+              :math:`\\langle \\delta^2 \\psi_{nl}\\rangle` (and
+              likewise for :math:`s^2`) are set to zero.
+
+    .. note:: The full non-linear model for the cross-correlation
+              between number counts and intrinsic alignments is
+              still work in progress in FastPT. As a workaround
+              CCL assumes a non-linear treatment of IAs, but only
+              linearly biased number counts.
```
tilmantroester, line 42: Can we have a warning that gets generated when trying to compute P_gi with a non-linear galaxy tracer? I'm concerned that this note in the documentation might get missed.

pyccl/nl_pt/__init__.py
```diff
@@ -7,19 +7,22 @@
...
-# Power spectra
-from .power import (
+# Old power spectrum calculator
+from .power_deprecated import (
     PTCalculator,
     get_pt_pk2d,
 )
 
+# Eulerian PT
+from .ept import EulerianPTCalculator
```
tilmantroester, line 17: Does `EulerianPTCalculator` work with any `PTTracer`? Or might there be a situation in the future where some `*PTCalculator` only works with certain`PTTracer`?

pyccl/nl_pt/power_deprecated.py
```diff
@@ -62,6 +64,10 @@ def __init__(self, with_NC=False, with_IA=False, with_dd=True,
                  pad_factor=1, low_extrap=-5, high_extrap=3,
                  P_window=None, C_window=.75,
                  k_cutoff=None, n_exp_cutoff=4):
+        warnings.warn("PTCalculators are deprecated and will be removed in "
+                      "future versions. Use the new perturbation theory "
+                      "calculators (e.g. EulerianPTCalculator).",
+                      CCLDeprecationWarning)
```
nikfilippas, line 70: Not a good idea to rename the module like this and fill it with all those warnings - I think it's best to just wrap the `__init__` function and all other functions that will be deprecated with the `@deprecated` decorator. Otherwise it just breaks API.

damonge: How does it break the API? The exact same functionality is still there. This will just throw warnings whenever users try to call the deprecated function or initialize the deprecated class.

pyccl/nl_pt/__init__.py
damonge: `EulerianPTCalculator` does work with any `PTTracer`, but that may not be the case with future calculators (e.g. I'm fairly certain `velocileptors` doesn't do IAs. What are you worried about?

pyccl/nl_pt/ept.py
damonge: Done

pyccl/nl_pt/power_deprecated.py
nikfilippas: The module is renamed. To preserve the API you're going to have to add it to the renaming dictionary in `__getattr__` in `__init__.py`, which complicates things. Easiest way is to add `@deprecated` to `__init__()`.

damonge:
@anicola Good point. Yes, indeed, I've added a note about stochastic bias in the docstring.

pyccl/nl_pt/power_deprecated.py
nikfilippas: yep that does it

pyccl/nl_pt/__init__.py
tilmantroester: I'm thinking about how to make sure that only valid combinations of `PTCalculator` and `PTTracers` are used in something like firecrown. Right now the assumption is that there is one `PTCalculator` that can be used with all the PT tracers but I wonder if there might be a situation in the future where there are multiple `PTCalculator`s that need to be matched with their corresponding tracers.

damonge: The way I see it, Tracers are fully general, it's just that some calculators will only be able to compute certain quantities. E.g. the LPT calculator will have to throw an error when asked to compute an IA P(k). But all calculators that can compute e.g. Pk_gg will be able to work with the same `PTNumberCountsTracer`s (i.e. there won't be tracers specific for a type of calculator).

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        if self.with_IA:
+            self.ia_ta = self.pt.IA_ta(
+                pklz0, P_window=self.fastpt_par['P_window'],
+                C_window=self.fastpt_par['C_window'])
+            self.ia_tt = self.pt.IA_tt(
+                pklz0, P_window=self.fastpt_par['P_window'],
+                C_window=self.fastpt_par['C_window'])
+            self.ia_mix = self.pt.IA_mix(
+                pklz0, P_window=self.fastpt_par['P_window'],
+                C_window=self.fastpt_par['C_window'])
```
nikfilippas, line 263: Since they all use the same kwargs, instead of copying+pasting the same thing 3 times you can do
```python
kwargs = {"P": pklz0, "P_window": self.fastpt_par["P_window"], "C_window": self.fastpt_par["C_window"]}
self.ia_ta = self.pt.IA_ta(**kwargs)
self.ia_tt = ...
```

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        """
+        self._check_pt_init()
+        # Get Pk templates
+        Pd1d1 = self.pk_b1
+        Pd1d2 = self._g4[:, None] * self.dd_bias[2][None, :]
+        Pd2d2 = self._g4[:, None] * self.dd_bias[3][None, :]
+        Pd1s2 = self._g4[:, None] * self.dd_bias[4][None, :]
+        Pd2s2 = self._g4[:, None] * self.dd_bias[5][None, :]
+        Ps2s2 = self._g4[:, None] * self.dd_bias[6][None, :]
+        Pd1p3 = self._g4[:, None] * self.dd_bias[8][None, :]
```
nikfilippas, line 312: This creates a new axis for `self._g4` every time. Factor it out and do it only once.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        b12 = tr2.b1(self.z_s)
+        b22 = tr2.b2(self.z_s)
+        bs2 = tr2.bs(self.z_s)
+        bk22 = tr2.bk2(self.z_s)
+        b3nl2 = tr2.b3nl(self.z_s)
+
+        s4 = 0.
+        if self.fastpt_par['sub_lowk']:
+            s4 = self._g4 * self.dd_bias[7]
+            s4 = s4[:, None]
```
nikfilippas, line 330: If sub_lok is True, this assigns 0 to s4 first, and then reassigns it with another value. The proper way to do that is to use an if-else statement. This is going to be a waste of space though, that's why I prefer the one-liners:
```python
        s4 = ((self._g4 * self.dd_bias[7])[:, None]
              if self.fastpt_par['sub_lowk']
              else 0.)
```

pyccl/nl_pt/ept.py
nikfilippas, line 1: To avoid all of these `[None, :]` and `[:, None]` which just slow the code down and make it extremely hard to read, why don't we just store them in the PTCalculator in the shape that they are going to be used? Users will interact with the getter methods, not with the attributes themselves.

This will give the whole module a breath of fresh air.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        a0e2, b0e2, d0ee2, d0bb2 = self.ia_mix
+
+        # Get biases
+        b1 = trg.b1(self.z_s)
+        b2 = trg.b2(self.z_s)
+        bs = trg.bs(self.z_s)
+        bk2 = trg.bk2(self.z_s)
+        b3nl = trg.b3nl(self.z_s)
+        if (np.any(b2 != 0) or np.any(bs != 0) or
+                np.any(bk2 != 0) or np.any(b3nl != 0)):
```
nikfilippas, line 377: `if any([b.any() for b in [b2, bs, bk2, b3nl]])`

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+            array_like: 2D array of shape `(N_a, N_k)`, where `N_k` \
+                is the size of this object's `k_s` attribute, and \
+                `N_a` is the size of the object's `a_s` attribute.
+        """
+        self._check_pt_init()
+        # Get Pk templates
+        Pd1d1 = self.pk_b1
+        a00e, c00e, a0e0e, a0b0b = self.ia_ta
+        ae2e2, ab2b2 = self.ia_tt
+        a0e2, b0e2, d0ee2, d0bb2 = self.ia_mix
```
nikfilippas, line 447: These variable names are 
Untitled
ae2200e2aae02ea2ae00ea2

`".join(np.random.choice(["a", "e", "0", "2"], 23))`

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+
+        t1 = tracer1.type
+        t2 = tracer2.type
+
+        if ((t1 == 'NC') or (t2 == 'NC')) and (not self.with_NC):
+            raise ValueError("Need number counts Pk templates, "
+                             "but calculator didn't compute them")
+        if ((t1 == 'IA') or (t2 == 'IA')) and (not self.with_IA):
+            raise ValueError("Need intrinsic alignment Pk templates, "
+                             "but calculator didn't compute them")
```
nikfilippas, line 562: The error messages imply that the calculator is incompetent.
Better write `Can't use IA tracer in EulerianPTCalculator when 'with_IA=False'.` (but replace ' with `)

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
+import numpy as np
+import warnings
+from ..pk2d import Pk2D
+from ..base import CCLAutoreprObject, unlock_instance
+from ..errors import CCLWarning
+
+
+class EulerianPTCalculator(CCLAutoreprObject):
```
nikfilippas, line 8: Can we rename this to just `EulerianPT` which is much simpler and clear? It sets up Eulerian PT!
I am going to do the same with `HMCalculator`: rename it to `HaloModel` because it sets up a halo model, and it's much clearer.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        elif pk_name == 'c2:cdelta_bb':
+            pk = self._g4[:, None] * self.ia_mix[3][None, :]
+        elif pk_name == 'cdelta:cdelta':
+            pk = self._g4[:, None] * self.ia_ta[2][None, :]
+        elif pk_name == 'cdelta:cdelta_bb':
+            pk = self._g4[:, None] * self.ia_ta[3][None, :]
+        elif pk_name == 'zero':
+            # If zero, store None and return
+            self._pk2d_temp[pk_name] = None
+            return None
```
nikfilippas, line 688: Just `return`, no `None`.
If you want to be extra-spicy, since `setattr` returns None, you can combine both lines into 1:
`return self._pk2d_temp[pk_name] = None`. Kind of like Julia.

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    else:
+        t2 = TRS[options[1]]
+    ptc = ccl.nl_pt.EulerianPTCalculator(
+        with_NC=True, with_IA=True, with_matter_1loop=True,
+        extra_params={'sub_lowk': options[3]},
+        cosmo=COSMO)
+    pk = ptc.get_pk2d_biased(TRS[options[0]],
+                             tracer2=t2,
+                             return_ia_bb=options[2])
+    assert isinstance(pk, ccl.Pk2D)
```
nikfilippas, line 49: I have already replaced some of the tests that do this, but this way of writing tests is very hard to read, and if one fails in the future the developer will have trouble debugging it. It is best to unpack `options` into the real parameter names, and use it like so:
```python
@pytest.mark.parametrize("param1,param2,param3,param4",  # replace with real names
    [["TG", "TG", ...],
    ...]])
def test_(param1, param2, param3, param4):
    ...
```

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+                                     ['TI', 'TI', True, False],
+                                     ['TI', 'TM', False, False],
+                                     ['TM', 'TG', False, False],
+                                     ['TM', 'TI', False, False],
+                                     ['TM', 'TM', False, False]])
+def test_ept_get_pk2d_smoke(options):
+    if options[0] == options[1]:
+        t2 = None
+    else:
+        t2 = TRS[options[1]]
```
nikfilippas, line 41: For example, this can be really neatly put into a one-liner, which is easy to read. I don't know why you are so against them: `t2 = TRS[options[1]] if options[0] != options[1] else None`.

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    pk = ptc.get_pk2d_biased(TRS[options[0]],
+                             tracer2=t2,
+                             return_ia_bb=options[2])
+    assert isinstance(pk, ccl.Pk2D)
+
+
+def test_ept_pk2d_bb_smoke():
+    pee = PTC.get_pk2d_biased(TRS['TI'])
+    pbb = PTC.get_pk2d_biased(TRS['TI'], return_ia_bb=True)
+    assert pee.eval(0.1, 0.9, COSMO) != pbb.eval(0.1, 0.9, COSMO)
```
nikfilippas, line 55: `eval` is deprecated, `cosmo` is kw-only.

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+        with_NC=True, with_IA=True, with_matter_1loop=True,
+        b1_pk_kind=nl, bk2_pk_kind=nl, cosmo=COSMO)
+    pk = ptc.get_pk2d_biased(TRS['TG'])
+    assert isinstance(pk, ccl.Pk2D)
+
+
+@pytest.mark.parametrize('typ_nlin,typ_nloc', [('linear', 'nonlinear'),
+                                               ('nonlinear', 'linear'),
+                                               ('nonlinear', 'pt'),
+                                               ('linear', 'pt')])
```
nikfilippas, line 70: See for example here, this is the clearer way to use it: unpack the parameters.

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    ptc2 = ccl.nl_pt.EulerianPTCalculator(
+        with_NC=True, with_IA=True, with_matter_1loop=True,
+        b1_pk_kind=typ_nloc, cosmo=COSMO)
+    pkmm = ptc1.get_pk2d_biased(tm, tracer2=tm)
+    pkmm2 = ptc2.get_pk2d_biased(tm, tracer2=tm)
+    pkgg = ptc1.get_pk2d_biased(tg, tracer2=tg)
+    ks = np.geomspace(1E-3, 1E1, 128)
+    p1 = pkgg.eval(ks, 1., COSMO)
+    p2 = pkmm.eval(ks, 1., COSMO)+ks**2*pkmm2.eval(ks, 1., COSMO)
+    assert np.all(np.fabs(p1/p2-1) < 1E-4)
```
nikfilippas, line 87: use `np.allclose`

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    ptc1 = ccl.nl_pt.EulerianPTCalculator(with_NC=False, with_IA=False,
+                                          with_matter_1loop=True,
+                                          cosmo=COSMO)
+    ptc2 = ccl.nl_pt.EulerianPTCalculator(with_NC=True, cosmo=COSMO)
+    tg = ccl.nl_pt.PTNumberCountsTracer(b1=1.0)
+    tm = ccl.nl_pt.PTMatterTracer()
+    ks = np.geomspace(1E-3, 10, 64)
+    pk1 = ptc1.get_pk2d_biased(tm).eval(ks, 1.0, COSMO)
+    pk2 = ptc2.get_pk2d_biased(tg).eval(ks, 1.0, COSMO)
+    assert np.all(np.fabs(pk1/pk2-1) < 1E-3)
```
nikfilippas, line 194: `np.allclose`

pyccl/nl_pt/tracers.py
```diff
@@ -79,6 +61,8 @@ class PTTracer(object):
     in a perturbation theory framework to provide N-point
     correlations.
     """
+    __repr_attrs__ = ('type', 'biases')
```
nikfilippas, line 64: nitpicking but it is recommended to end tuples with a comma, just in case it only has one element

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        expansion), `'cdelta'` (:math:`s\\delta` term in IA expansion).
+
+        Args:
+            kind (str): string defining the pair of PT operators for
+                which we want the power spectrum.
+            return_ia_bb (bool): if `True`, the B-mode power spectrum
+                for intrinsic alignments will be returned (if both
+                input tracers are of type
+                :class:`~pyccl.nl_pt.tracers.PTIntrinsicAlignmentTracer`)
+                If `False` (default) E-mode power spectrum is returned.
```
nikfilippas, line 618: this is the final arg

nikfilippas (changes requested):
'Some' comments mainly regarding naming, efficiency, and consistency.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        k_cutoff (float): exponential cutoff scale. All power
+            spectra will be multiplied by a cutoff factor of the
+            form :math:`\\exp(-(k/k_*)^n)`, where :math:`k_*` is
+            the cutoff scale. This may be useful when using the
+            resulting power spectra to compute correlation
+            functions if some of the PT contributions do not
+            fall sufficiently fast on small scales. If `None`
+            (default), no cutoff factor will be applied.
+        n_exp_cutoff (float): exponent of the cutoff factor (see
+            `k_cutoff`).
```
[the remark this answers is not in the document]
damonge: Gonna push back here. I don't think the names are that terrible, in which case I'd rather keep them to minimise the annoyance of v3 migration.

pyccl/nl_pt/ept.py
damonge: I don't disagree, but this is very similar to what FASTPT does internally, so I'd rather keep them to make any future comparisons with the raw code of FASTPT easier.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+            'b2:b3nl': 'zero', 'b2:bs': 'b2:bs', 'b2:bk2': 'zero',
+            'b2:c1': 'zero', 'b2:c2': 'zero', 'b2:cdelta': 'zero',
+            'b3nl:b3nl': 'zero', 'b3nl:bs': 'zero',
+            'b3nl:bk2': 'zero', 'b3nl:c1': 'zero', 'b3nl:c2':
+            'zero', 'b3nl:cdelta': 'zero', 'bs:bs': 'bs:bs',
+            'bs:bk2': 'zero', 'bs:c1': 'zero', 'bs:c2': 'zero',
+            'bs:cdelta': 'zero', 'bk2:bk2': 'zero', 'bk2:c1': 'zero',
+            'bk2:c2': 'zero', 'bk2:cdelta': 'zero', 'c1:c1': 'm:m',
+            'c1:c2': 'm:c2', 'c1:cdelta': 'm:cdelta', 'c2:c2': 'c2:c2',
+            'c2:cdelta': 'c2:cdelta', 'cdelta:cdelta': 'cdelta:cdelta'}
```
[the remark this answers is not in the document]
damonge: Not all PT calculators will have access to the same pairs, so for now I'll define just outside this file.

pyccl/nl_pt/ept.py
damonge: We had this discussion, so resolving.

pyccl/nl_pt/ept.py
damonge: This doesn't work (syntax error). Leaving as is to be less cryptic

pyccl/tests/test_ept_power.py
damonge: Subjective, coding style

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    ks = np.array([0.01, 0.1, 1.0])
+    ptc1 = ccl.nl_pt.EulerianPTCalculator(with_matter_1loop=True,
+                                          b1_pk_kind='linear',
+                                          cosmo=COSMO)
+    pk1 = ptc1.get_pk2d_biased(TRS['TM']).eval(ks, 1.0, COSMO)
+    ptc2 = ccl.nl_pt.EulerianPTCalculator(with_matter_1loop=True,
+                                          b1_pk_kind='pt',
+                                          cosmo=COSMO)
+    pk2 = ptc2.get_pk2d_biased(TRS['TM']).eval(ks, 1.0, COSMO)
+    assert np.all(np.fabs(pk1/pk2-1) < 1E-10)
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+    __repr_attrs__ = ('with_NC', 'with_IA', 'with_matter_1loop',
+                      'k_s', 'a_s', 'exp_cutoff',
+                      'b1_pk_kind', 'bk2_pk_kind', 'fastpt_par')
+
+    def __init__(self, *, with_NC=False, with_IA=False,
+                 with_matter_1loop=True, cosmo=None,
+                 log10k_min=-4, log10k_max=2, nk_per_decade=20,
+                 a_arr=None, k_cutoff=None, n_exp_cutoff=4,
+                 b1_pk_kind='nonlinear', bk2_pk_kind='nonlinear',
+                 extra_params=None):
```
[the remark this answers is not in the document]
damonge: I just made all of these init arguments.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        # Set FAST-PT parameters
+        self.fastpt_par = {'pad_factor': 1.0,
+                           'low_extrap': -5.0,
+                           'high_extrap': 3.0,
+                           'P_window': None,
+                           'C_window': 0.75,
+                           'sub_lowk': False}
+        if extra_params is None:
+            extra_params = {}
+        self.fastpt_par.update(extra_params)
```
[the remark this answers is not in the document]
damonge: I just made all of these init arguments, no more update needed.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+
+        # k sampling
+        nk_total = int((log10k_max - log10k_min) * nk_per_decade)
+        self.k_s = np.logspace(log10k_min, log10k_max, nk_total)
+
+        # a sampling
+        if a_arr is None:
+            a_arr = 1/(1+np.linspace(0, 4, 30)[::-1])
+        self.a_s = a_arr.copy()
+        self.z_s = 1/self.a_s-1
```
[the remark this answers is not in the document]
damonge: Using the default pk sampling in a if None now.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        self.with_NC = with_NC
+        self.with_IA = with_IA
+
+        # Set FAST-PT parameters
+        self.fastpt_par = {'pad_factor': 1.0,
+                           'low_extrap': -5.0,
+                           'high_extrap': 3.0,
+                           'P_window': None,
+                           'C_window': 0.75,
+                           'sub_lowk': False}
```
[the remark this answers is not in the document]
damonge: I just made all of these init arguments.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        b1_pk_kind (str): prescription to use for the
+            power spectrum to use for the first-order bias terms
+            in the expansion. `'linear'`: use the linear matter
+            power spectrum. `'nonlinear'`: use the non-linear matter
+            power spectrum. `'pt'`: use the 1-loop SPT matter
+            power spectrum. Default: `'nonlinear'`.
+        bk2_pk_kind (str): prescription to use for the
+            power spectrum to use for the non-local bias terms
+            in the expansion. Same options and default as
+            `b1_pk_kind`.
```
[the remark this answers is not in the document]
damonge: rephrased

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        extra_params (dict): dictionary of precision parameters
+            for FAST-PT. Options described below. If `None`,
+            default values specified below are used.
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+                                      for a in self.a_s])
+        if 'pt' in [self.b1_pk_kind, self.bk2_pk_kind]:
+            if 'linear' in pks:
+                pk = pks['linear']
+            else:
+                pk = np.array([cosmo.linear_matter_power(self.k_s, a)
+                               for a in self.a_s])
+            # Add SPT correction
+            pk += self._g4[:, None] * self.one_loop_dd[0][None, :]
+            pks['pt'] = pk
```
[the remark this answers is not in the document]
damonge: gonna leave it as is, as the structure is different than in those cases

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        # Initialize all expensive arrays to `None`.
+        self._pt_init = False
+        self.pk_b1 = None
+        self.pk_bk = None
+        self.one_loop_dd = None
+        self.dd_bias = None
+        self.ia_ta = None
+        self.ia_tt = None
+        self.ia_mix = None
+        self._g4 = None
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+            'c2:cdelta': 'c2:cdelta', 'cdelta:cdelta': 'cdelta:cdelta'}
+        # All valid Pk pair labels
+        self._pk_valid = list(self._pk_alias.keys())
+        # List of Pk2Ds to fill out
+        self._pk2d_temp = {}
+
+    def _check_pt_init(self):
+        if self._pt_init:
+            return
+        raise RuntimeError("PT templates have not been initialised "
```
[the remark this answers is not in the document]
damonge: Done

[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+                           "for this calculator. Please do so using "
+                           "`update_ingredients`.")
+
+    @unlock_instance
+    def update_ingredients(self, cosmo):
```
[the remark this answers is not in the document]
damonge: I did some of this (now storing cosmo), but I'll keep as `update_ingredients`, since it's more descriptive.

pyccl/nl_pt/ept.py
damonge: Done

pyccl/nl_pt/ept.py
damonge: Done

pyccl/nl_pt/ept.py
damonge: This seems fine to me as it is.

pyccl/nl_pt/ept.py
damonge: Done

pyccl/nl_pt/ept.py
damonge: I prefer to be explicit

pyccl/nl_pt/ept.py
damonge: Done

pyccl/tests/test_ept_power.py
damonge: Done

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+
+    pk2 = PTC.get_pk2d_biased(t1, tracer2=t2)
+
+    if pk1 is None:
+        assert pk2.eval(0.5, 1.0, COSMO) == 0.0
+    else:
+        assert pk1.eval(0.5, 1.0, COSMO) == pk2.eval(0.5, 1.0, COSMO)
+        # Check cached
+        pk3 = PTC._pk2d_temp[PTC._pk_alias[kind]]
+        assert pk1.eval(0.5, 1.0, COSMO) == pk3.eval(0.5, 1.0, COSMO)
```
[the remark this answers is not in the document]
damonge: done

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+            c1=bdict['c1'], c2=bdict['c2'],
+            cdelta=bdict['cdelta'])
+
+    tn1, tn2 = kind.split(':')
+    t1 = get_tr(tn1)
+    t2 = get_tr(tn2)
+
+    pk2 = PTC.get_pk2d_biased(t1, tracer2=t2, return_ia_bb=True)
+
+    assert pk1.eval(0.5, 1.0, COSMO) == pk2.eval(0.5, 1.0, COSMO)
```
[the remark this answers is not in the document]
damonge: done

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    # as matter P(k)
+    ptc1 = ccl.nl_pt.EulerianPTCalculator(with_NC=False, with_IA=False,
+                                          with_matter_1loop=True,
+                                          cosmo=COSMO)
+    ptc2 = ccl.nl_pt.EulerianPTCalculator(with_NC=True, cosmo=COSMO)
+    tg = ccl.nl_pt.PTNumberCountsTracer(b1=1.0)
+    tm = ccl.nl_pt.PTMatterTracer()
+    ks = np.geomspace(1E-3, 10, 64)
+    pk1 = ptc1.get_pk2d_biased(tm).eval(ks, 1.0, COSMO)
+    pk2 = ptc2.get_pk2d_biased(tg).eval(ks, 1.0, COSMO)
```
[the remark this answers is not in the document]
damonge: Done

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    # Wrong type of b1 and bk2 power spectra
+    with pytest.raises(ValueError):
+        ccl.nl_pt.EulerianPTCalculator(b1_pk_kind='non-linear')
+
+    # Wrong type of b1 and bk2 power spectra
+    with pytest.raises(ValueError):
+        ccl.nl_pt.EulerianPTCalculator(bk2_pk_kind='non-linear')
+
+    # Uninitialized templates
+    with pytest.raises(RuntimeError):
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
damonge: Done

pyccl/nl_pt/ept.py
nikfilippas: For explicit checks with 0 you can do `if any([(b != 0).any() for b in [b2, bs, bk2, b3nl]])`.
Also, the method `.any()` is considerably faster than the function `np.any()`, so you want to use it when you can avoid using the function. In this case, we know that these will be numpy arrays.

pyccl/nl_pt/ept.py
nikfilippas: Yes, I didn't suggest changing them, I just commented on how terrible they are xD

pyccl/nl_pt/tracers.py
nikfilippas: Until it does, in which case you'll get a very weird error. Nevermind.

pyccl/nl_pt/ept.py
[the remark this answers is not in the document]
nikfilippas: Yes, that's literally what I suggested: a module-level variable (not package-level).

pyccl/nl_pt/ept.py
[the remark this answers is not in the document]
nikfilippas: great!

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,697 @@
...
+            to_do.append('IA')
+
+        # k sampling
+        nk_total = int((log10k_max - log10k_min) * nk_per_decade)
+        self.k_s = np.logspace(log10k_min, log10k_max, nk_total)
+
+        # a sampling
+        if a_arr is None:
+            a_arr = get_pk_spline_a()
+        self.a_s = a_arr.copy()
```
nikfilippas, line 170: Why are you copying it? There is no need. You are producing a new array inside the function, and even when the user passes one, you are not modifying it. Copying takes time.

pyccl/nl_pt/ept.py
nikfilippas: Replace ' with `, because ' is reserved for strings, and it might confuse people.

pyccl/nl_pt/ept.py
nikfilippas: It is inefficient because if `sub_lok` is `True` memory will be allocated twice. If we're going to be strict about efficiency like in `halos/` we have to do it everywhere, right? xD
I don't care about that to be fair, I'm just pointing it out, you can leave it as it is.

pyccl/tests/test_ept_power.py
nikfilippas: Why did you resolve all of these?

nikfilippas (changes requested):
A final comment, and we have the unresolved ones remaining.

pyccl/nl_pt/ept.py
damonge: I disagree. The message will just read weird.

pyccl/nl_pt/ept.py
damonge: I prefer to be explicit.

pyccl/nl_pt/ept.py
```diff
@@ -0,0 +1,700 @@
...
+        """
+        self._check_pt_init()
+        if self.b1_pk_kind == 'linear':
+            P1loop = self._g4[:, None] * self.one_loop_dd[0][None, :]
+        else:
+            P1loop = 0.
+        pk = self.pk_b1 + P1loop
+        return pk*self.exp_cutoff
+
+    def get_pk2d_biased(self, tracer1, *, tracer2=None, return_ia_bb=False,
```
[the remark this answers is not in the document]
damonge: Done

pyccl/nl_pt/ept.py
damonge: I'm pretty sure this copy is needed. Check below
```
>>> class My(object):
...     def __init__(self, a):
...             self.a = a
... 
>>> a = np.arange(10)
>>> a
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
>>> m = My(a)
>>> 
>>> m.a
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
>>> a[0] = 3
>>> a
array([3, 1, 2, 3, 4, 5, 6, 7, 8, 9])
>>> m.a
array([3, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

damonge: I.e. if the initial array is modified outside the object, it will also be modified inside, since it gets assigned by reference.

pyccl/nl_pt/ept.py
damonge: Changed to your original suggestion.

pyccl/tests/test_ept_power.py
damonge: Because they weren't worth the effort. Now with allclose so we can move on.

pyccl/tests/test_ept_power.py
```diff
@@ -0,0 +1,260 @@
...
+    pk1 = pk2d1.eval(ks, 1.0, COSMO)
+    ptc2 = ccl.nl_pt.EulerianPTCalculator(with_NC=True,
+                                          k_cutoff=10.,
+                                          n_exp_cutoff=2.,
+                                          cosmo=COSMO)
+    pk2d2 = ptc2.get_pk2d_biased(t, tracer2=t)
+    pk2 = pk2d2.eval(ks, 1.0, COSMO)
+
+    expcut = np.exp(-(ks/10.)**2)
+    assert np.all(np.fabs(pk1*expcut/pk2-1) < 1E-3)
```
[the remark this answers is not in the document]
damonge: done

pyccl/tests/test_ept_power.py
damonge: done

damonge:
All done, back to you @nikfilippas

pyccl/nl_pt/ept.py
nikfilippas: We use ` I think throughout the code, but nevermind this is not worth it. Resolving.

pyccl/tests/test_ept_power.py
nikfilippas: Right, just keep that in mind in code you write/review hereafter. allclose is much easier to read, looking down the line

pyccl/nl_pt/ept.py
nikfilippas: You are correct. I didn't think of the chance of the array mutating outside of the object. Resolving.

nikfilippas (approved):
There are 4 uncovered lines. Do you think we can cover the last 3 of them? I'm approving either way.

damonge:
All should be covered now.

nikfilippas:
@damonge you do the honours

PR #1058, opened 2023-04-04
Remove `T_CMB` and `T_ncdm` as constants (reloaded)

nikfilippas:
This one completely decouples `T_CMB` and `T_ncdm` from the physical constants.
It also exploits the `CCLParameters` framework that was there to make the C-level cosmological parameters communicate with Python, and port their setting-up to Python fully.

### DIFF
--- a/pyccl/ccl_core.i
+++ b/pyccl/ccl_core.i
@@ -15,12 +15,34 @@
 
 %include "../include/ccl_core.h"
 
-// Enable vectorised arguments for arrays
+
 %apply (double* IN_ARRAY1, int DIM1) {
-            (double* zarr, int nz),
-            (double* dfarr, int nf),
-            (double* m_nu, int n_m)
+       (double* mass, int num),
+       (double* zarr, int nz),
+       (double* dfarr, int ndf)
 };
+
+%inline %{
+void parameters_m_nu_set_custom(ccl_parameters *params, double *mass, int num) {
+  params->m_nu = (double*) malloc(num*sizeof(double));
+  memcpy(params->m_nu, mass, num*sizeof(double));
+}
+
+void parameters_mgrowth_set_custom(ccl_parameters *params,
+    double* zarr, int nz, double* dfarr, int ndf) {
+  if (nz > 0) {
+    params->has_mgrowth = true;
+    params->nz_mgrowth = nz;
+    params->z_mgrowth = (double*) malloc(nz*sizeof(double));
+    params->df_mgrowth = (double*) malloc(nz*sizeof(double));
+    memcpy(params->z_mgrowth, zarr, nz*sizeof(double));
+    memcpy(params->df_mgrowth, dfarr, nz*sizeof(double));
+  }
+}
+
+%}
+
+
 %apply (int DIM1, double* ARGOUT_ARRAY1) {(int nout, double* output)};
 
 %inline %{
@@ -35,50 +57,4 @@ void parameters_get_nu_masses(ccl_parameters *params, int nout, double* output) 
     }
 }
 
-ccl_parameters parameters_create_nu(
-    double Omega_c, double Omega_b, double Omega_k, double Neff,
-    double w0, double wa, double h, double A_s, double sigma8, double n_s,
-    double T_CMB, double Omega_g, double T_ncdm,
-    double bcm_log10Mc, double bcm_etab, double bcm_ks, double mu_0,
-    double sigma_0, double c1_mg, double c2_mg, double lambda_mg,
-    double* m_nu, int n_m, int* status)
-{
-    return ccl_parameters_create(
-        Omega_c, Omega_b, Omega_k, Neff, m_nu, n_m, w0, wa, h,
-        A_s, sigma8, n_s, T_CMB, Omega_g, T_ncdm,
-        bcm_log10Mc, bcm_etab, bcm_ks,
-        mu_0, sigma_0, c1_mg, c2_mg, lambda_mg,
-        -1, NULL, NULL, status );
-}
-
 %}
-
-%feature("pythonprepend") parameters_create_nu_vec %{
-    if numpy.shape(zarr) != numpy.shape(dfarr):
-        raise CCLError("Input shape for `zarr` must match `dfarr`!")
-%}
-
-%inline %{
-ccl_parameters parameters_create_nu_vec(
-    double Omega_c, double Omega_b, double Omega_k, double Neff,
-    double w0, double wa, double h, double A_s, double sigma8, double n_s,
-    double T_CMB, double Omega_g, double T_ncdm,
-    double bcm_log10Mc, double bcm_etab, double bcm_ks, double mu_0,
-    double sigma_0, double c1_mg, double c2_mg, double lambda_mg,
-    double* zarr, int nz,
-    double* dfarr, int nf, double* m_nu,
-    int n_m, int* status)
-{
-    if (nz == 0){ nz = -1; }
-    return ccl_parameters_create(
-        Omega_c, Omega_b, Omega_k, Neff, m_nu, n_m, w0, wa, h,
-        A_s, sigma8, n_s, T_CMB, Omega_g, T_ncdm,
-        bcm_log10Mc, bcm_etab, bcm_ks,
-        mu_0, sigma_0, c1_mg, c2_mg, lambda_mg,
-        nz, zarr, dfarr, status);
-}
-
-%}
-
-/* The directive gets carried between files, so we reset it at the end. */
-%feature("pythonprepend") %{ %}
--- a/pyccl/ccl_core.i
+++ b/pyccl/ccl_core.i
@@ -48,13 +48,7 @@ void parameters_mgrowth_set_custom(ccl_parameters *params,
 %inline %{
 
 void parameters_get_nu_masses(ccl_parameters *params, int nout, double* output) {
-    output[0] = 0;
-    output[1] = 0;
-    output[2] = 0;
-
-    for (int i=0; i<params->N_nu_mass; ++i) {
-        output[i] = params->m_nu[i];
-    }
+  memcpy(output, params->m_nu, nout*sizeof(double));
 }
 
 %}
--- a/pyccl/core.py
+++ b/pyccl/core.py
@@ -611,13 +611,13 @@ class Cosmology(CCLObject):
         sqrtk = np.sqrt(np.abs(Omega_k)) * h / c.CLIGHT_HMPC
 
         self._fill_params(
-            sum_nu_masses=sum(nu_mass), N_nu_mass=N_nu_mass, N_nu_rel=N_nu_rel,
-            Neff=Neff, Omega_nu_mass=Omega_nu_mass, Omega_nu_rel=Omega_nu_rel,
-            T_ncdm=T_ncdm, Omega_m=Omega_m, Omega_c=Omega_c, Omega_b=Omega_b,
-            Omega_k=Omega_k, sqrtk=sqrtk, k_sign=int(k_sign), T_CMB=T_CMB,
-            Omega_g=Omega_g, w0=w0, wa=wa, Omega_l=Omega_l, h=h, H0=h*100,
-            A_s=A_s, sigma8=sigma8, n_s=n_s, mu_0=mu_0, sigma_0=sigma_0,
-            c1_mg=c1_mg, c2_mg=c2_mg, lambda_mg=lambda_mg,
+            sum_nu_masses=sum(nu_mass), N_nu_rel=N_nu_rel, Neff=Neff,
+            Omega_nu_mass=Omega_nu_mass, Omega_nu_rel=Omega_nu_rel,
+            Omega_m=Omega_m, Omega_c=Omega_c, Omega_b=Omega_b, Omega_k=Omega_k,
+            sqrtk=sqrtk, k_sign=int(k_sign), Omega_g=Omega_g, w0=w0, wa=wa,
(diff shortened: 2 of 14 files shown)


pyccl/core.py
```diff
@@ -577,35 +569,67 @@ def _build_parameters(
...
+        T_nu = T_CMB * (4/11)**(1/3)
+        rho_nu_rel = N_nu_rel * (7/8) * 4 * c.STBOLTZ / c.CLIGHT**3 * T_nu**4
+        Omega_nu_rel = rho_nu_rel / rho_crit
+
+        # For non-relativistic neutrinos, calculate the phase-space integral.
+        self._fill_params(m_nu=nu_mass)
+        from .neutrinos import Omega_nu_h2
+        kw = {"m_nu": nu_mass, "T_CMB": T_CMB, "T_ncdm": T_ncdm}
+        Omega_nu_mass = Omega_nu_h2(1., **kw) / h**2 if N_nu_mass > 0 else 0
+        # Omega_nu_mass = self.OmNuh2(1.,) / h**2 if N_nu_mass > 0 else 0
```
nikfilippas, line 601: Note to self: have to check why replacing with the final line doesn't work.

pyccl/core.py
```diff
@@ -68,6 +69,12 @@
 }
 
 
+class _Defaults:
+    """Default cosmological parameters used throughout the library."""
+    T_CMB = 2.725
```
damonge, line 74: Note that we want to make this 2.7255 in v3. Making it 2.725 was one of the biggest blunders at the start of CCL

pyccl/core.py
```diff
@@ -126,8 +133,7 @@ class Cosmology(CCLObject):
         wa (:obj:`float`, optional): Second order term of dark energy equation
             of state. Defaults to 0.
         T_CMB (:obj:`float`): The CMB temperature today. The default of
-            ``None`` uses the global CCL value in
-            ``pyccl.physical_constants.T_CMB``.
+            is 2.725.
```
damonge, line 136: As above.

pyccl/core.py
damonge: In which way doesn't it work?

pyccl/neutrinos.py
```diff
@@ -56,6 +52,12 @@ def Omega_nu_h2(a, *, m_nu, T_CMB=None, T_ncdm=None):
     return OmNuh2
 
 
+def OmNuh2(cosmo, a):
+    """Like Omega_nu_h2 but it uses the parameters from a Cosmology object."""
+    return Omega_nu_h2(a, m_nu=cosmo["m_nu"],
+                       T_CMB=cosmo["T_CMB"], T_ncdm=cosmo["T_ncdm"])
+
+
```
damonge, line 60: So now we have `Omega_nu_h2`, `Omeganuh2` and `OmNuh2`. Any way we can clean this up?

pyccl/ccl_core.i
```diff
@@ -35,50 +57,4 @@ void parameters_get_nu_masses(ccl_parameters *params, int nout, double* output)
...
-    int n_m, int* status)
-{
-    if (nz == 0){ nz = -1; }
-    return ccl_parameters_create(
-        Omega_c, Omega_b, Omega_k, Neff, m_nu, n_m, w0, wa, h,
-        A_s, sigma8, n_s, T_CMB, Omega_g, T_ncdm,
-        bcm_log10Mc, bcm_etab, bcm_ks,
-        mu_0, sigma_0, c1_mg, c2_mg, lambda_mg,
-        nz, zarr, dfarr, status);
-}
```
damonge, line 79: Is `ccl_parameters_create` ever being called now?

pyccl/neutrinos.py
damonge: Three proposals:
- `omega_nu` (as in little omega)
- `little_Omega_nu`
- Just a private method of ccl. Users should instead use the `omega_x` and `rho_x` methods.

damonge (commented):
Looking good.

pyccl/core.py
nikfilippas: Open an issue, and set it as a v3 milestone, or we'll forget! We'll change `Neff` too, to the newest value.

pyccl/core.py
nikfilippas: It returns a different value for OmNuh2, I have to check why.

pyccl/core.py
damonge: OK, but I'm saying, should we add a deprecation warning here when someone relies on the default value?

pyccl/ccl_core.i
nikfilippas: Yes, only at the C-level by MG stuff and halofit. Otherwise I'd remove it.

damonge: I see, OK

damonge: Oh, hold on, but those will rely on the default values of T_CMB and T_ncdm, which now live in python, right? I know there are copies at the C level, but it'd be good if we can roadblock them completely. See a global comment further down.

nikfilippas:
Replying to both your general and your specific comments:
- `T_ncdm` is already an initialization parameter of `Cosmology`. It doesn't show in the diffs because this PR is based on `nu_Tncdm_v3` where I first implemented it.
- Luckily the C-level functions that use the `T`s all do it through the cosmology object.
- I went through the entire C library to see if there are any rogue uses of these parameters, but I think I have cleared them up. To verify, I set them both to 1e+300 at the C level, and re-ran tests and benchmarks, and they ran fine.

pyccl/core.py
nikfilippas: A deprecation warning for something so central that is used almost everywhere in CCL and is piped to many other functions? That's a tough one, because `_Defaults.T_CMB` gets called on import multiple times, as it is the default argument of many functions. Instead of tapping into `_Defaults.__getattr__` to warn, I think the best plan here is to warn one single time, on import, and that's it. The warning message will be thrown from `pyccl/__init__.py`.

damonge: Sounds good to me. It'll be a bit annoying, but worth it.

pyccl/core.py
damonge: Fixed it.

damonge:
OK, sounds good. See my comment about deprecation message for T_CMB and renaming the new OmNuh2 function

pyccl/neutrinos.py
damonge: My actual final proposal: we make `Omega_nu_h2` a private method of `Cosmology`, only used for initialization, and deprecate all other versions of this function in favour of `omega_x`.

pyccl/core.py
nikfilippas: Embarrassing! Thanks!

pyccl/neutrinos.py
nikfilippas: What about `nu_masses`?

pyccl/core.py
nikfilippas: In the end it looks like `Omega_nu_mass` is ever so slightly different.

pyccl/neutrinos.py
damonge: Let's leave `nu_masses` there, and just deprecate `T_CMB` as an argument.

nikfilippas: After your input and some modifications by me, there's now a private method in `Cosmology` which calculates OmNuh2, with the original implementation deprecated in favor of `omega_x`.

nikfilippas:
@damonge all done + neutrinos (kind of) debugged!
We went from this:
```python
>>> ccl.CosmologyVanillaLCDM(m_nu=0.1, m_nu_type="single")["m_nu"]
array([0.1, 0. , 0. ])
```
to this
```python
[0.1]
```
So now, `len(m_nu)` matches the number of massive neutrinos, which is what this parameter originally intended.

pyccl/core.py
```diff
@@ -577,73 +569,80 @@ def _build_parameters(
...
-                lib.cosmology_free(self.cosmo)
-        if hasattr(self, "_params"):
-            if (self._params is not None and
-                    hasattr(lib, 'parameters_free') and
-                    lib.parameters_free is not None):
-                lib.parameters_free(self._params)
-
-        # finally delete some attributes we don't want to be around for safety
-        # when the context manager exits or if __del__ is called twice
-        if hasattr(self, "cosmo"):
```
damonge, line 644: Why do we not need these extra checks now?

pyccl/neutrinos.py
```diff
@@ -15,17 +20,18 @@
 }
 
 
-def Omega_nu_h2(a, *, m_nu, T_CMB=None, T_ncdm=None):
```
damonge, line 18: Doesn't losing this function break api?

pyccl/tests/test_cosmology.py
```diff
@@ -110,12 +110,6 @@ def test_cosmology_init():
         m_nu_type='normal')
 
 
-def test_cosmology_setitem():
-    cosmo = ccl.CosmologyVanillaLCDM()
-    with pytest.raises(NotImplementedError):
-        cosmo['a'] = 3
-
-
```
damonge, line 118: Why did we lose this?

pyccl/tests/test_neutrinos.py
```diff
@@ -12,7 +12,8 @@
     [0.1, 0.8, 0.3],
     np.array([0.1, 0.8, 0.3])])
 def test_omnuh2_smoke(a, m):
-    om = ccl.Omega_nu_h2(a, m_nu=m)
+    with pytest.warns(ccl.CCLDeprecationWarning):
+        om = ccl.Omeganuh2(a, m_nu=m)
```
damonge, line 16: Can we start doing this properly with omega_x?

pyccl/tests/test_neutrinos.py
```diff
@@ -39,5 +40,6 @@ def test_neutrinos_raises():
 @pytest.mark.parametrize('split', ['normal', 'inverted', 'equal'])
 def test_nu_mass_consistency(a, split):
     m = ccl.nu_masses(Omega_nu_h2=0.1, mass_split=split)
-    assert np.allclose(ccl.Omega_nu_h2(a, m_nu=m),
-                       0.1, rtol=0, atol=1e-4)
+    with pytest.warns(ccl.CCLDeprecationWarning):
+        om = ccl.Omeganuh2(a, m_nu=m)
```
damonge, line 44: Same

pyccl/tests/test_parameters.py
```diff
@@ -398,18 +392,6 @@ def test_parameters_missing():
...
-    Check that a Cosmology object doesn't let parameters be set.
-    """
-    params = ccl.Cosmology(
-        Omega_c=0.25, Omega_b=0.05, h=0.7, A_s=2.1e-9,
-        n_s=0.96)
-
-    # Check that error is raised when unrecognized parameter requested
-    assert_raises(KeyError, lambda: params['wibble'])
-
-
```
damonge, line 412: Why did we lose this?

pyccl/core.py
nikfilippas: Because the way it is set up you can't have something assigned to `cosmo.cosmo` and that something not being a SWIG object:
```python
        self.cosmo = lib.cosmology_create(self._params, self._config)
```
If this step fails, nothing is assigned to `self.cosmo`.

pyccl/neutrinos.py
nikfilippas: `Omega_nu_h2` would be a v3 addition, this doesn't currently exist in the API.

pyccl/tests/test_cosmology.py
nikfilippas: I removed `__setitem__` because Python's native message is good, and we don't want to test the default internal workings of Python.

pyccl/tests/test_parameters.py
nikfilippas: I removed the custom message because Python's default is good! The difference is that it gives you a more relevant `AttributeError` telling you that `'wibble'` doesn't exist.

pyccl/tests/test_neutrinos.py
nikfilippas: Yea this is already tested in `test_background.py`. In v3, we're going to delete those tests, or I can delete them now if you prefer.

damonge: OK, good. Can we add just one test, for my own peace of mind, that verifies that we get the same answer with omega_x and with Omeganuh2?

nikfilippas: They both call the same C function, `ccl_Omeganuh2`:
```C
double ccl_omega_x(ccl_cosmology * cosmo, double a, ccl_species_x_label label, int *status)
{
  // If massive neutrinos are present, compute the phase-space integral and
  // get OmegaNuh2. If not, set OmegaNuh2 to zero.
  double OmNuh2;
  if ((cosmo->params.N_nu_mass) > 0.0001) {
    // Call the massive neutrino density function just once at this redshift.
    OmNuh2 = ccl_Omeganuh2(a, cosmo->params.N_nu_mass, cosmo->params.m_nu,
                           cosmo->params.T_CMB, cosmo->params.T_ncdm, status);
  } else {
    OmNuh2 = 0.;
  }
```
but there is a slight difference. `ccl_omega_x` returns zero if `N_nu_mass <= 1e-4` whereas `ccl_Omeganuh2` returns zero if `N_nu_mass` is exactly zero. I have removed the check from `ccl_omega_x` in order to let `ccl_Omeganuh2` do the job, and now both calls should be identical, so I don't feel there is need of unit testing, since the input of `ccl_omega_x` is directly piped into `ccl_Omeganuh2`.

damonge (approved):

PR #1061, opened 2023-04-07
Enhanced FFTLog options in `HaloProfile`

nikfilippas:
One thing that is currently implemented in an odd way is that if users want to pass their own values to FFTLog when they create a `HaloProfile`, they have to instantiate it first, and then change the values via `update_precision_fftlog`. This is awkward - we used to do that for the `Cosmology` object too, but we have long deprecated that.

This PR enables passing the FFTLog parameters directly into the halo profile constructors.
The parameters can be passed as keyword arguments, and with calls to `super` that we are doing either way, they are eventually handled by `HaloProfile` base.


damonge:
This is not to be reviewed until all existing functionality has been ported to v3. There is new science implementations that take priority.

damonge:

PR #1062, opened 2023-04-08
Completely decouple Python from all parameters in C

nikfilippas:
Does what it says. With this PR all C-structs that hold parameters are managed from Python - even their values.
I did a slight revamp to the `CCLParameters` class that allows all the parameters that are defined as class attributes to be passed to C on import. And I also fixed some loopholes the previous implementation had.

This only contains 5 commits. The commit history looks like this because I merged different bases instead of rebasing.

**Note**: I had to relax some benchmarks slightly, because of the new CODATA values for the physical constants (mainly `GNEWT`). I verified that CCL's former value gives consistent results with this PR.


damonge:
This is not to be reviewed until all existing capabilities have been ported to v3. There is new science implementations that take priority.

PR #1063, opened 2023-04-09
`PowerSpectrum` abc with arbitrary precision options.

### LINKED ISSUE #424: Enable some non-vanilla dark matter options from CLASS
CLASS is able to calculate observables for a number of more sophisticated dark matter models, like decaying CDM and annihilating DM. It would be nice to enable some of those options (in a simple way) through CCL.

### LINKED ISSUE #588: Pass (more) CLASS accuracy parameters from pyccl

nikfilippas:
Just like with `Baryons`, `HaloProfile`, `MassFunc` etc. this PR introduces a `PowerSpectrum` base class which returns a `Pk2D` object containing the power spectrum of a particular model. The functions in `boltzmann.py` are now subclasses of the base class, which allows for easy model extension.

With this one, I ported the C code for the Eisenstein-Hu and BBKS fitting formulas to Python, into their own subclasses.
It also solves a long-standing issue, of passing arbitrary parameters to the models. For example, users can now exploit that to pass any precision parameter CLASS accepts.

Closes #588 , closes #424 .
May also address #114 if slightly modified.


damonge:
This is not to be reviewed until existing capabilities have been moved to v3. There is new science implementations that take priority.

damonge:

PR #1069, opened 2023-04-17
Fix halo inconsistencies v3

nikfilippas:
This PR contains fixes to some design flaws in halos which can ultimately cause memory leaks and model inconsistencies.
Note: at the moment it breaks API (to make review easier), but once review is complete I will clone it into a non API-breaking one.

(5' read)

Problem 1
-------------
When `MassDef` is instantiated with a concentration it inadvertently contains a reference to itself:
```python
mass_def is mass_def.concentration.mass_def is mass_def.[...].mass_def
```
This causes a memory leak because when the object is deleted, the garbage collector doesn't know the order of deletion so the instances hang in there for the rest of the runtime. This can easily be demonstrated:
```python
import pyccl as ccl


class Concentration:

    def __init__(self, mass_def):
        self.mass_def = mass_def
        self.id = hex(id(self))[-5:]
        print(f"Concentration {self.id} created.")

    def __del__(self):
        print(f"Concentration {self.id} destroyed.")


class MassDef:

    def __init__(self):
        self.concentration = Concentration(self)  # line that causes the problem
        self.id = hex(id(self))[-5:]
        print(f"MassDef {self.id} created.")

    def __del__(self):
        print(f"MassDef {self.id} destroyed.")


mass_def = MassDef()
del mass_def  # doesn't do anything

print("program ends")
```
which outputs
```bash
Concentration 36050 created.
MassDef 3af50 created.
program ends
MassDef 3af50 destroyed.
Concentration 36050 destroyed.
```

An easy way to fix that is to use a weakref proxy of `self` when passing it to the `Concentration` constructor:
`import weakref ... self.concentration = Concentration(weakref.proxy(self))`. This allows the reference count of `mass_def` to go to zero and the gbc to purge it. This tactic is not generally advised though.

However, there is a deeper design flaw with the interplay of `MassDef` and `Concentration`.

Problem 2
-------------
Matter profiles require a `Concentration` to initialize. However, this isn't really a parameter; it is akin to setting up the model. As is `mass_def` (which is why we agreed to deprecate it from calls to subclasses of `HMIngredients`). This is how these two are used in the code, in every matter profile:
```python
    def _real(self, cosmo, r, M, a, mass_def):
        # Comoving virial radius
        R_M = mass_def.get_radius(cosmo, M, a) / a
        c_M = self.concentration(cosmo, M, a)
        R_s = R_M / c_M
```
so they aren't used in anything other than finding the comoving virial radius, given `M`. These two work together; they should be together, but one is passed at initialization and the other one at calls. What's more, this creates an inconsistency: the `mass_def` of `concentration` may not be the same as the `mass_def` passed in `_real`. Luckily, most implemented concentrations are `mass_def`-agnostic, and also users have (hopefully) been passing the same one, but this can easily create problems.

Solution
-----------
- One way is to add a check every time e.g. `_real` is called to verify that the mass definitions are consistent. That's a bad solution, no explanation needed.
- Another way is to deprecate `concentration` in `__init__` and pass it as an extra argument in e.g. `_real`. That's a bad solution because not all profiles use a `Concentration`, so we'll end up unnecessarily complicating the code.
- Another way is to move `mass_def` to `__init__` and deprecate it from e.g. `_real`. That's okay, but the non-matter profiles still use `mass_def` to compute the radius, given the mass. That means, in order to be consistent, we'd have to introduce `mass_def` as an `__init__` argument in all profiles.
- I propose the following. Since `MassDef` and `Concentration` are so strongly coupled, it makes sense to combine them into one single object. Then, every profile will accept a `MassConcentration` object in `__init__` which contains all the info it will ever need: mass definition and an optional concentration. This will simplify things.

In practice, the new class would contain `self.mass_def` and `self.concentration` and would avoid self-references and other caveats. It would initialize both the mass definition and the concentration prescription at the same time.

This is what it could look like:
```python
class MassConcentration(MassDef, Concentration):
    def __init__(self, *, mass_def, concentration):
        # careful with initialization: we don't want cyclic references like in MassDef
        self.mass_def = mass_def
        self.concentration = concentration
```
and then, all halo profiles would have:
```python
class HaloProfileNFW(HaloProfile):
    def __init__(self, mass_concentration, ...):
        self.mass_concentration = mass_concentration
        ...

    def _real(self, cosmo, r, M, a):
        R_M = self.mass_concentration.get_radius(cosmo, M, a) / a
        c_M = self.mass_concentration.concentration(cosmo, M, a)
        R_s = R_M / cM
```
We could even go as far as implementing a `get_comoving_virial_radius` method, to do all of that in a single step and avoid having to repeat that in all matter profiles.

It would also make physical sense because the parameters passed in `__init__` would uniquely define a profile 'entity', which would then be called via its methods and `(r, M, a)` (and `cosmo` which is an outlier as we've already discussed).

For `v2.final` we can leave things as they are, but add warnings that these will be deprecated in the future, and point users towards using `MassConcentration` objects instead of separate `MassDef` and `Concentration`.

The solution would _simplify_ the code - for example, in `benchmarks/test_haloprofile.py` we do
```python
    mdef = ccl.halos.MassDef(mDelta, 'matter')
    c = ccl.halos.ConcentrationConstant(c=concentration, mdef=mdef)
    mdef = ccl.halos.MassDef(mDelta, 'matter', concentration=c)
    p = ccl.halos.HaloProfileEinasto(c, truncated=False)
```
so there are two instantiations of a mass definition to get the one we want to use. With the new implementation we won't need that - it would be a single step.

This proposal is a widely used design pattern in OOP called "dependency injection", and is a technique where an object is passed as a dependency to another object, rather than having the dependent object create the dependency itself. It helps to reduce coupling between different objects, and makes the code more modular and testable.


pyccl/halos/profiles/hernquist.py
```diff
@@ -45,18 +44,15 @@ class HaloProfileHernquist(HaloProfileMatter):
     """
     __repr_attrs__ = __eq_attrs__ = (
         "fourier_analytic", "projected_analytic", "cumul2d_analytic",
-        "truncated", "concentration", "precision_fftlog", "normprof",)
+        "truncated", "mass_concentration", "precision_fftlog", "normprof",)
```
damonge, line 47: Just concentration

pyccl/halos/profiles/einasto.py
```diff
@@ -41,17 +40,19 @@ class HaloProfileEinasto(HaloProfileMatter):
...
+                 mass_def):
         self.truncated = truncated
         self.alpha = alpha
-        super().__init__()
+        super().__init__(mass_def=mass_def, concentration=concentration)
+        mvir = MassDef("vir", "matter")
+        self._to_virial_mass = mass_translator(
+            mass_in=self.mass_def, mass_out=mvir,
+            concentration=self.concentration)
         self.update_precision_fftlog(padding_hi_fftlog=1E2,
```
damonge, line 56: You only need to do this if alpha is cosmo

pyccl/halos/profiles/hod.py
```diff
@@ -248,15 +244,15 @@ def update_parameters(self, *, log10Mmin_0=None, log10Mmin_p=None,
...
         r_use = np.atleast_1d(r)
         M_use = np.atleast_1d(M)
 
         # Comoving virial radius
         bg = self.bg_0 + self.bg_p * (a - self.a_pivot)
         bmax = self.bmax_0 + self.bmax_p * (a - self.a_pivot)
-        R_M = mass_def.get_radius(cosmo, M_use, a) / a
-        c_M = self.concentration(cosmo, M_use, a)
+        R_M = self.mass_concentration.get_radius(cosmo, M_use, a) / a
+        c_M = self.mass_concentration.get_concentration(cosmo, M_use, a)
```
damonge, line 255: Just concentration and mass_def

damonge (commented):
This I can live with (just a couple spurious comments). Ping me for review when it's fully implemented

pyccl/halos/profiles/hernquist.py
nikfilippas: done

pyccl/halos/profiles/einasto.py
nikfilippas: That's ok, doesn't hurt to have it anyway, it's a private attribute and takes no time to declare because it is a nested function. The alternative would be a lot more complicated because we'd have to include checks in `update_parameters`.

pyccl/halos/profiles/hod.py
nikfilippas: fixed

pyccl/halos/massdef.py
```diff
@@ -248,49 +218,130 @@ def from_name(cls, name):
...
+            instance = Type.create_instance(value, mass_def=mass_def)
+            out.append(instance)
 
-    Args:
-        concentration (string): concentration-mass relation.
-    """
-    return MassDef(200, 'matter', concentration=concentration)
+        # Check mass definition consistency.
+        if hmd := set([x.mass_def for x in out]):
+            if hmd != set([mass_def]):
```
damonge, line 287: This seems very obscure. Why not `if set([x.mass_def for x in out]) != set([mass_def]):`

pyccl/halos/profiles/profile_base.py
```diff
@@ -66,12 +76,11 @@ def get_normalization(self, cosmo, a, *, hmc=None):
...
         """
         def integ(M):
-            return self.fourier(cosmo=cosmo,
-                                k=hmc.precision['k_min'],
-                                M=M, a=a, mass_def=hmc.mass_def)
+            return self.fourier(cosmo, hmc.precision["k_min"], M, a)
         return hmc.integrate_over_massfunc(integ, cosmo, a)
         # TODO: CCLv3 replace by the below in v3 (profiles will all have a
         # default normalization of 1. Normalization will always be applied).
+        # NK: (cosmo, a) have to take None defaults in v3.
```
damonge, line 83: what?

damonge (commented):
A few minor comments

damonge:
Also, there seem to be conflicts

pyccl/halos/profiles/profile_base.py
nikfilippas: We should be able to call `prof.get_normalization()` with no arguments in the degenerate case where the norm is 1. How it is now, it requires that `cosmo` and `a` are necessarily passed. It makes no sense to have `hmc` as optional and the other two as required arguments. So the signature should be `(self, cosmo=None, a=None, *, hmc=None)`.

pyccl/halos/massdef.py
damonge: I know the operator, I'm just saying this is rather obscure. My suggestion would check both things in one go with no added overhead.

nikfilippas: That's what I'm saying, it can't be done in one go.

benchmarks/test_massdef.py
```diff
@@ -22,13 +22,16 @@
 cs_200c_di = dc[8]
 
 hmd_vir = ccl.halos.MassDefVir()
-hmd_vir_b = ccl.halos.MassDefVir('Bhattacharya13')
+# hmd_vir_b = ccl.halos.MassDefVir('Bhattacharya13')
+
```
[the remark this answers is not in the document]
nikfilippas: forgot to remove them, fixed now

nikfilippas:
@damonge back to you

pyccl/halos/massdef.py
damonge: why not? hmd is not used anywhere else, and if the set is empty, then it won't be equal to `set([mass_def])`

nikfilippas: Because if no halo model ingredients have been passed, doing what you proposed will raise an error. You have to check first, if `out` is empty first. I replaced it with a single-liner, check if you like it now.

damonge: I don't think you need the first bit of that one-liner, but that's fine, I can live with it.

damonge (approved):

pyccl/halos/massdef.py
nikfilippas: I literally tried it and got an error, the first bit does all the work.

PR #1070, opened 2023-04-18
`self is other` shortcut for `__eq__`

nikfilippas:
We didn't include this check in the beginning which would make comparison much faster for same-id objects.


damonge (commented):
Looks good, but it seems tests didn't pass

nikfilippas:
@damonge they pass now, looks like the coveralls server was down.

damonge (approved):
LGTM

PR #1071, opened 2023-04-18
Cosmology v3

nikfilippas:
This PR renames `core.py` to `cosmology.py`, brings it into v3 form, closes the issues listed on the sidebar, and fully fixes all neutrino inconsistencies in CCL.


pyccl/__init__.py
```diff
@@ -23,6 +23,9 @@
 # SWIG-generated
 from . import ccllib as lib
 
+# Name of the default power spectrum.
+DEFAULT_POWER_SPECTRUM = "delta_matter:delta_matter"
```
damonge, line 27: I think I'd rather have "delta_matter:delta_matter" everywhere, rather than this. It's easier for users to understand from the documentation and function signature.

pyccl/cosmology.py
```diff
@@ -701,51 +717,13 @@ def compute_growth(self):
         """Compute the growth function."""
         if self.has_growth:
             return
-        if self['N_nu_mass'] > 0:
-            warnings.warn(
-                "CCL does not properly compute the linear growth rate in "
-                "cosmological models with massive neutrinos!",
-                category=CCLWarning)
```
damonge, line 708: Why remove this warning?

damonge (commented):
OK, some comments. Most are of two types:
a) Clarify why you're doing what you're doing (this is mostly because there was no explanatory message to start this PR, so I'm not sure I follow the logic).
b) Why did many of the warning/error messages disappear. Many of them seem useful.

pyccl/neutrinos.py
```diff
@@ -75,27 +83,46 @@ def nu_masses(*, Omega_nu_h2, mass_split, T_CMB=None):
...
-            Omega_nu_h2, neutrino_mass_splits[mass_split], 3, status)
-    elif mass_split in ['sum', 'single']:
-        mnu, status = lib.nu_masses_vec(
-            Omega_nu_h2, neutrino_mass_splits[mass_split], 1, status)
-        mnu = mnu[0]
 
-    # Check status and return
-    check(status)
-    return mnu
+def get_neutrino_masses(*, m_nu, mass_split):
```
nikfilippas, line 92: TODO: Maybe make this a private func to avoid cluttering the namespace. But it's a nice func to expose so idk.

pyccl/tests/test_parameters.py
```diff
@@ -305,33 +307,34 @@ def test_parameters_nu_unphysical_raises(m_nu, kind):
...
+#         ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7,
+#                       A_s=2.1e-9, n_s=0.96, w0=-0.9)
+#         ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, h=0.7,
+#                       A_s=2.1e-9, n_s=0.96, w0=-0.9, wa=0.1)
+#         # Check that kwarg order doesn't matter
+#         ccl.Cosmology(h=0.7, Omega_c=0.25, Omega_b=0.05,
+#                       A_s=2.1e-9, n_s=0.96)
+#         # Try a set of parameters with non-zero mu0 / Sig0
+#         ccl.Cosmology(h=0.7, Omega_c=0.25, Omega_b=0.05,
+#                       A_s=2.1e-9, n_s=0.96, mu_0=0.1, sigma_0=0.1)
```
damonge, line 337: Wait, so why uncomment for v3? Shouldn't this be uncommented now, since the relevant code will be gone in v3?

damonge (commented):
Looked at tests, just one minor comment. After this I'll approve!

pyccl/neutrinos.py
```diff
@@ -68,28 +67,35 @@ def Omeganuh2(a, *, m_nu, T_CMB=_Defaults.T_CMB, T_ncdm=_Defaults.T_ncdm):
     return OmNuh2
 
 
-@warn_api(pairs=[("OmNuh2", "Omega_nu_h2")])
-def nu_masses(*, Omega_nu_h2, mass_split, T_CMB=None):
+@warn_api(pairs=[("OmNuh2", "m_total")])
+def nu_masses(*, m_total, mass_split, T_CMB=None, is_Omega_nu_h2=True):
```
nikfilippas, line 71: Too convoluted to do it that way. The new signature should be 
```python
def nu_masses(*, Omega_nu_h2=None, mass_split, T_CMB=None, m_nu=None)
```
Then, inside the function, decide what we want to do based on which is not `None`. I don't like flags in functions except for very special cases.

damonge: I thought you didn't like `None`s ;-).
Sure, push this, that's fine with me.

nikfilippas: I don't like `None` defaults when they can be substituted with the default. Generally, `None` isn't bad when there is no other option.

pyccl/tests/test_parameters.py
nikfilippas: this test checks for no warnings, but there are deprecation warnings all over the place now - shall we just get rid of it?

damonge: sure. As long as the coverage stays healthy, I 'm happy

damonge (approved):
LGTM. Merge if the tests pass

PR #1072, opened 2023-04-19
No normprof

damonge:
This PR removes normprof altogether from v3. The idea is that all profiles should define their own normalisation function (unless the normalisation is just 1, which is the default).

### DIFF
--- a/pyccl/base/deprecations.py
+++ b/pyccl/base/deprecations.py
@@ -146,9 +146,8 @@ def warn_api(func=None, *, pairs=[], reorder=[]):
         if any([par.startswith("normprof") and kwargs.get(par) is not None
                 for par in kwargs]):
             warnings.warn(
-                "Argument `normprof` has been deprecated. Change the default "
-                "value only by subclassing. More comprehensive profile "
-                "normalization options will be provided with CCLv3.0.0.",
+                "Argument `normprof` will be deprecated in CCL v3. All "
+                "profiles will carry their own normalisation.",
                 CCLDeprecationWarning)
 
         # API compatibility for deprecated HMCalculator argument k_min.
--- a/pyccl/halos/halo_model.py
+++ b/pyccl/halos/halo_model.py
@@ -156,16 +156,6 @@ class HMCalculator(CCLAutoRepr):
                            self._mass, a, mass_def=self.mass_def).T
         return 1. / self._integrate_over_mf(uk0)
 
-    def get_profile_norm(self, cosmo, a, prof):
-        """Compute the normalization of a profile."""
-        if not prof.normprof:  # TODO: Remove for CCLv3.
-            return 1
-        uk0 = prof._normalization(self)(cosmo=cosmo, a=a)
-        if isinstance(uk0, (int, float)):
-            return 1 / uk0
-        self._get_ingredients(cosmo, a, get_bf=False)
-        return 1 / self._integrate_over_mf(uk0)
-
     @warn_api(pairs=[("sel", "selection"),
                      ("amin", "a_min"),
                      ("amax", "a_max")],
--- a/pyccl/halos/pk_1pt.py
+++ b/pyccl/halos/pk_1pt.py
@@ -1,4 +1,4 @@
-from ..base import UnlockInstance, warn_api
+from ..base import warn_api
 import numpy as np
 
 
@@ -28,11 +28,6 @@ def _Ix1(func, cosmo, hmc, k, a, prof, normprof):
         `k` and `a` respectively. If `k` or `a` are scalars, the
         corresponding dimension will be squeezed out on output.
     """
-    # TODO: Remove for CCLv3.
-    if normprof is not None:
-        with UnlockInstance(prof):
-            prof.normprof = normprof
-
     func = getattr(hmc, func)
 
     a_use = np.atleast_1d(a).astype(float)
@@ -43,8 +38,9 @@ def _Ix1(func, cosmo, hmc, k, a, prof, normprof):
     out = np.zeros([na, nk])
     for ia, aa in enumerate(a_use):
         i11 = func(cosmo, k_use, aa, prof)
-        norm = hmc.get_profile_norm(cosmo, aa, prof)
-        out[ia] = i11 * norm
+        norm = prof.normalization(hmc, cosmo, aa) if normprof else 1
+        # TODO: CCLv3 remove if
+        out[ia] = i11 / norm
 
     if np.ndim(a) == 0:
         out = np.squeeze(out, axis=0)
--- a/pyccl/halos/profiles/profile_base.py
+++ b/pyccl/halos/profiles/profile_base.py
@@ -1,8 +1,8 @@
 from ...pyutils import resample_array, _fftlog_transform
 from ...base import CCLAutoRepr, unlock_instance, warn_api, deprecate_attr
 from ...parameters import FFTLogParams
+from ...parameters import physical_constants as const
 import numpy as np
-from abc import abstractmethod
 import functools
 from typing import Callable
 
@@ -47,35 +47,25 @@ class HaloProfile(CCLAutoRepr):
                             "_real or _fourier implementation.")
         self.precision_fftlog = FFTLogParams()
 
-    @property
-    @abstractmethod
-    def normprof(self) -> bool:
-        """Normalize the profile in auto- and cross-correlations by
-        :math:`I^0_1(k\\rightarrow 0, a|u)`
-        (see :meth:`~pyccl.halos.halo_model.HMCalculator.I_0_1`).
+    def normalization(self, hmc, cosmo, a):
+        """Returns the overall normalisation factor of this profile.
+
+        Args:
+            hmc (:class:`~pyccl.halos.HMCalculator`): a halo model calculator
+                object.
+            cosmo (:class:`~pyccl.core.Cosmology`): a Cosmology object.
+            a (float): scale factor.
+
+        Reurns:
+            float: normalisation of this profile.
         """
-
-    # TODO: CCLv3 - Rename & allocate _normprof_bool to the subclasses.
-
-    def _normprof_false(self, hmc, **settings):
-        """Option for ``normprof = False``."""
-        return lambda *args, cosmo, a, **kwargs: 1.
-
-    def _normprof_true(self, hmc, k_min=1e-5):
-        """Option for ``normprof = True``."""
-        # TODO: remove the first two lines in CCLv3.
-        k_hmc = hmc.precision["k_min"]
-        k_min = k_hmc if k_hmc != k_min else k_min
-        M, mass_def = hmc._mass, hmc.mass_def
-        return functools.partial(self.fourier, k=k_min, M=M, mass_def=mass_def)
-
-    def _normalization(self, hmc, **settings):
-        """This is the API adapter and it decides which norm to use.
-        It returns a function of ``cosmo`` and ``a``. Optional args & kwargs.
-        """
-        if self.normprof:
-            return self._normprof_true(hmc, **settings)
-        return self._normprof_false(hmc, **settings)
+        uk0 = self.fourier(cosmo=cosmo, k=hmc.precision['k_min'],
+                           M=hmc._mass, a=a, mass_def=hmc.mass_def)
+        hmc._get_ingredients(cosmo, a, get_bf=False)
+        return hmc._integrate_over_mf(uk0)
+        # TODO: CCLv3 replace by the below in v3 (profiles will all have a
+        # default normalisation of 1. Normalisation will always be applied).
+        # return lambda *args, cosmo, a, **kwargs: 1.
 
     @unlock_instance(mutate=True)
     @functools.wraps(FFTLogParams.update_parameters)
@@ -482,19 +472,21 @@ class HaloProfile(CCLAutoRepr):
 
 class HaloProfileNumberCounts(HaloProfile):
     """Base for number counts halo profiles."""
-    normprof = True
 
 
 class HaloProfileMatter(HaloProfile):
     """Base for matter halo profiles."""
-    normprof = True
+
+    def normalization(self, hmc, cosmo, a):
+        """Returns the normalization of all matter overdensity
+        profiles, which is simply the comoving matter density.
+        """
+        return const.RHO_CRITICAL * cosmo["Omega_m"] * cosmo["h"]**2
 
 
 class HaloProfilePressure(HaloProfile):
     """Base for pressure halo profiles."""
-    normprof = False
 
 
 class HaloProfileCIB(HaloProfile):
     """Base for CIB halo profiles."""
-    normprof = False
--- a/pyccl/tests/test_halomodel.py
+++ b/pyccl/tests/test_halomodel.py
@@ -69,7 +69,8 @@ def get_pk_new(mf, c, cosmo, a, k, get_1h, get_2h):
                                  mass_def=mdef)
     return ccl.halos.halomod_power_spectrum(cosmo, hmc, k, a, prf,
                                             get_1h=get_1h,
-                                            get_2h=get_2h)
+                                            get_2h=get_2h,
+                                            normprof1=True)
 
 
 @pytest.mark.parametrize('mf_c', [['shethtormen', 'bhattacharya2011'],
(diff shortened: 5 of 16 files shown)


pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,25 @@ def __init__(self):
...
-    @property
-    @abstractmethod
-    def normprof(self) -> bool:
-        """Normalize the profile in auto- and cross-correlations by
-        :math:`I^0_1(k\\rightarrow 0, a|u)`
-        (see :meth:`~pyccl.halos.halo_model.HMCalculator.I_0_1`).
-        """
-
-    # TODO: CCLv3 - Rename & allocate _normprof_bool to the subclasses.
+    def normalization(self, hmc, cosmo, a):
```
nikfilippas, line 50: 1. Can we rename this to `get_normalization` if it's going to be a public function? `normalization` would imply it is a property.
2. Since we depart from CCL's standard arg ordering, can we make all args kw-only?
3. Are we sure this is never going to need any additional args (i.e. the only things it needs are `cosmo` and `a`?

pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,25 @@ def __init__(self):
...
-        if self.normprof:
-            return self._normprof_true(hmc, **settings)
-        return self._normprof_false(hmc, **settings)
+        uk0 = self.fourier(cosmo=cosmo, k=hmc.precision['k_min'],
+                           M=hmc._mass, a=a, mass_def=hmc.mass_def)
+        hmc._get_ingredients(cosmo, a, get_bf=False)
+        return hmc._integrate_over_mf(uk0)
+        # TODO: CCLv3 replace by the below in v3 (profiles will all have a
+        # default normalisation of 1. Normalisation will always be applied).
+        # return lambda *args, cosmo, a, **kwargs: 1.
```
nikfilippas, line 68: What does that mean? Won't the default be the integration over mf any more?

pyccl/tests/test_halomodel.py
```diff
@@ -69,7 +69,8 @@ def get_pk_new(mf, c, cosmo, a, k, get_1h, get_2h):
                                  mass_def=mdef)
     return ccl.halos.halomod_power_spectrum(cosmo, hmc, k, a, prf,
                                             get_1h=get_1h,
-                                            get_2h=get_2h)
+                                            get_2h=get_2h,
+                                            normprof1=True)
```
nikfilippas, line 73: Why don't we get rid of this?

pyccl/tests/test_tkkssc.py
```diff
@@ -85,7 +85,8 @@ def test_tkkssc_linear_bias(isNC1, isNC2, isNC3, isNC4):
 
     # Test with the full T(k1,k2,a) for an NFW profile with bias ~1.
     tkk = ccl.halos.halomod_Tk3D_SSC(
-        COSMO, HMC, prof=NFW, lk_arr=np.log(KK), a_arr=AA)
+        COSMO, HMC, prof=NFW, lk_arr=np.log(KK), a_arr=AA,
+        normprof1=True)
```
nikfilippas, line 89: Get rid of this?

pyccl/halos/profiles/profile_base.py
nikfilippas, line 1: I am thinking, since we make so much use of `hmc._mass`, `hmc._get_ingredients`, and `hmc._integrate_over_blah`, maybe we should make mass a property of `hmc` and the other two public functions so they can be accessed by anyone wanting to subclass the `HaloProfile` and implement their own norm. There is no particular reason these are private right now.

I would rename `_get_ingredients` to `update_ingredients` to make it very clear what it does.
And I would rename `_integrate_over_mf` to `integrate_mass_function`.
And `_integrate_over_mbf` to `integrate_mass_and_bias_function`.

nikfilippas (changes requested):
Okay, here are my comments, which are very minor.
Great to have this implemented now.

benchmarks/test_covariances.py
```diff
@@ -31,9 +30,12 @@ def test_ssc_WL():
...
 
-    tk3D = ccl.halos.halomod_Tk3D_SSC(cosmo=cosmo, hmc=hmc,
-                                      prof=nfw, prof2=nfw, prof12_2pt=None,
-                                      lk_arr=np.log(k), a_arr=a, use_log=True)
+    with pytest.warns():
+        tk3D = ccl.halos.halomod_Tk3D_SSC(cosmo=cosmo, hmc=hmc,
+                                          prof=nfw, prof2=nfw, prof12_2pt=None,
+                                          lk_arr=np.log(k), a_arr=a,
+                                          use_log=True,
+                                          normprof1=True, normprof2=True)
```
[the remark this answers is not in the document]
damonge: We can't remove them, otherwise the function will return the wrong numbers.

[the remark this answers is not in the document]
nikfilippas: Add a TODO comment in that case, in all of these.

pyccl/halos/profiles/profile_base.py
damonge: No, the default will be 1 (no normalisation)

pyccl/tests/test_halomodel.py
damonge: Can't

pyccl/tests/test_tkkssc.py
damonge: Can't

benchmarks/test_covariances.py
```diff
@@ -31,9 +30,12 @@ def test_ssc_WL():
     a = np.linspace(1/(1+6), 1, n_z)
     k = np.geomspace(k_min, k_max, n_k)
 
-    tk3D = ccl.halos.halomod_Tk3D_SSC(cosmo=cosmo, hmc=hmc,
-                                      prof=nfw, prof2=nfw, prof12_2pt=None,
-                                      lk_arr=np.log(k), a_arr=a, use_log=True)
+    with pytest.warns():
```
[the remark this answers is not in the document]
damonge: Done

benchmarks/test_halomod.py
```diff
@@ -43,15 +43,19 @@ def test_halomod(model):
...
+                                                  normprof1=True)
     tol = pk * HALOMOD_TOLERANCE
     err = np.abs(pk_ccl - pk)
     assert np.all(err <= tol)
 
     z = 1.
     k = data_z1[:, 0] * cosmo['h']
     pk = data_z1[:, -1] / (cosmo['h']**3)
-    pk_ccl = ccl.halos.halomod_power_spectrum(cosmo, hmc, k, 1./(1+z), prf)
+    with pytest.warns():
```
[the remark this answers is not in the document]
damonge: Done

benchmarks/test_halomod.py
```diff
@@ -43,15 +43,19 @@ def test_halomod(model):
...
+        pk_ccl = ccl.halos.halomod_power_spectrum(cosmo, hmc, k, 1./(1+z), prf,
+                                                  normprof1=True)
```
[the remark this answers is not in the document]
damonge: Can't

benchmarks/test_hod.py
```diff
@@ -67,7 +67,8 @@ def _nz_2mrs(z):
     # P(k)
     a_arr, lk_arr, _ = cosmo.get_linear_power().get_spline_arrays()
     pk_hod = ccl.halos.halomod_Pk2D(cosmo, hmc, prf, prof_2pt=prf2pt,
-                                    lk_arr=lk_arr, a_arr=a_arr)
+                                    lk_arr=lk_arr, a_arr=a_arr,
+                                    normprof1=True)
```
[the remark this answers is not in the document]
damonge: Can't

pyccl/base/deprecations.py
```diff
@@ -146,9 +146,8 @@ def wrapper(*args, **kwargs):
         if any([par.startswith("normprof") and kwargs.get(par) is not None
                 for par in kwargs]):
             warnings.warn(
-                "Argument `normprof` has been deprecated. Change the default "
-                "value only by subclassing. More comprehensive profile "
-                "normalization options will be provided with CCLv3.0.0.",
+                "Argument `normprof` will be deprecated in CCL v3. All "
+                "profiles will carry their own normalisation.",
```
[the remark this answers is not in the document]
damonge: Done

pyccl/halos/pk_1pt.py
```diff
@@ -43,8 +38,9 @@ def _Ix1(func, cosmo, hmc, k, a, prof, normprof):
     out = np.zeros([na, nk])
     for ia, aa in enumerate(a_use):
         i11 = func(cosmo, k_use, aa, prof)
-        norm = hmc.get_profile_norm(cosmo, aa, prof)
-        out[ia] = i11 * norm
+        norm = prof.normalization(hmc, cosmo, aa) if normprof else 1
+        # TODO: CCLv3 remove if
+        out[ia] = i11 / norm
```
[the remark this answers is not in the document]
damonge: I decided to do it that way, since this will be now a public function that users can call and this seems more intuitive. I've updated the docstrings of the method to make this clearer.

pyccl/halos/profiles/profile_base.py
damonge: 1. Renamed.
2. I've gone with cosmo, a, hmc, which makes more sense.
3. Yes. Think of this as an even more specific version of `fourier`, which does not admit any other arguments.

damonge:
Back to you @nikfilippas

benchmarks/test_covariances.py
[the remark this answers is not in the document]
damonge: Done

pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,31 @@ def __init__(self):
...
-    @property
-    @abstractmethod
-    def normprof(self) -> bool:
-        """Normalize the profile in auto- and cross-correlations by
-        :math:`I^0_1(k\\rightarrow 0, a|u)`
-        (see :meth:`~pyccl.halos.halo_model.HMCalculator.I_0_1`).
-        """
-
-    # TODO: CCLv3 - Rename & allocate _normprof_bool to the subclasses.
+    def get_normalization(self, cosmo, a, hmc):
```
nikfilippas, line 50: This again departs from typical CCL arg ordering, so I think the best course of action is to make everything kw-only. I don't like making these simple things kw-only, that's why in the original version I had encapsulated a function within a function - the outer one accepted `hmc` and the inner one `cosmo` and `a`.

pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,31 @@ def __init__(self):
...
+        """Profiles may be normalised by an overall function of redshift
+        (or scale factor). This function may be cosmology dependent and
```
nikfilippas, line 52: Yet another reason to make everything kw-only, because not everything depends on `cosmo`, someone might want to implement it like `def get_normalization(self, *, a, hmc, **kwargs)`.

pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,31 @@ def __init__(self):
...
+        often comes from integrating certain halo properties over mass.
+        This method returns this normalising factor. For example,
+        to get the normalised profile in real space, one would call
```
nikfilippas, line 55: normaliZed

pyccl/halos/profiles/profile_base.py
```diff
@@ -47,35 +47,31 @@ def __init__(self):
...
+        the `real` method, and then **divide** the result by the value
+        returned by this method.
```
nikfilippas, line 57: I would personally write here in the docstring the function you showed me in the halos PR.
Use `r` in the beginning of the docstring to avoid having to put double \\ in math mode.

pyccl/halos/profiles/profile_base.py
nikfilippas: Hummm... maybe it would be better to default to nothing, and make this an abstract method so that devs are forced to implement a normalization? What do you think? Just bringing this to the table, but I'm happy with the way it is now.

pyccl/halos/profiles/profile_base.py
nikfilippas, line 1: So we don't provide a default option to normalize a profile using `fourier` now, correct? Would it be worth considering adding one?

pyccl/halos/halo_model.py
```diff
@@ -82,8 +82,8 @@ def __init__(self, *, mass_function, halo_bias, mass_def=None,
             'log10M_min': log10M_min, 'log10M_max': log10M_max, 'nM': nM,
             'integration_method_M': integration_method_M, 'k_min': k_min}
         self._lmass = np.linspace(log10M_min, log10M_max, nM)
-        self._mass = 10.**self._lmass
-        self._m0 = self._mass[0]
+        self.M = 10.**self._lmass
```
nikfilippas, line 85: Not like this, it has to be a property, otherwise users will be able to modify it, because arrays are mutable, and this could resort to very weird things happening.
I'd suggest returning a copy of the masses. I checked, and copying takes O(100ns), so it's fine.
```python
@property
def M(self):
    return self._M.copy()
```

It was ok when it was private, but now that we expose it we need to be extra careful.

nikfilippas (changes requested):
Final comments.

pyccl/halos/profiles/profile_base.py
damonge: The very much most common case will be no normalization, so I'd rather make that the default.

pyccl/halos/profiles/profile_base.py
damonge: The normalization applies in real or fourier space, it's a scale-independent factor.

pyccl/halos/halo_model.py
damonge: Yeah, I hate this. I'll come up with a different way.

pyccl/halos/profiles/profile_base.py
damonge: I'll make hmc kwonly and default to None. They shouldn't always be required.

pyccl/halos/halo_model.py
nikfilippas: me too actually, just make `_mass` private again.

pyccl/halos/profiles/profile_base.py
damonge: Again, we shouldn't allow that. This is meant to be a more restrictive version of `fourier` or `real`, for which we don't allow additional dependence.

nikfilippas: yep - what you proposed in your other comment works, so I'm resolving this!

pyccl/halos/profiles/profile_base.py
nikfilippas: Just note that in the (simplest) default case that we return 1, none of them is required.

pyccl/halos/profiles/profile_base.py
damonge: Will do this in the docs PR

pyccl/halos/profiles/profile_base.py
damonge: This is done now

pyccl/halos/profiles/profile_base.py
damonge: done

pyccl/halos/halo_model.py
damonge: I went back to everything private and created a single public function of the hmc that allows us to integrate any function of mass.

nikfilippas: Is this new function any useful, or is it a special-case solution to our problem?

damonge: It could be generally useful, I think. It's an easy way for users to calculate these integrals, which are quite ubiquitous

damonge:
All done, back to you.

pyccl/halos/halo_model.py
nikfilippas: Yea, I see, so you are allocating the integration work back to `hmc` where it belongs. That's fine.
Can we generalize it in that case? Either create one for mbf or add an argument for users to flag if they want an integral over mf or mbf.

nikfilippas: fair

nikfilippas (approved):

PR #1073, opened 2023-04-19
Eq. for EulerianPT calculator

damonge:
This implements comparison between PT calculators beyond id.
The tracers seem to me to be complicated enough that the only practical and safe comparison is via id equality.


nikfilippas (changes requested):
`biases` and `type` uniquely define a tracer, right? Why not use that as an equivalence check? Very easy. Granted, because of the interp objects this will end up doing id checking in the background, but at some point we might implement `__eq__` for interpolators too.

Also, now that I am looking at it again, the tracer base class should be abstract, since `type` is undefined. Make this an abstract property with
```python
@property
@abstractmethod
def type(self):
   """Docstring."""
```

damonge:
OK, done (although it seems a bit silly, since this is effectively a slower way of just checking ids atm)

nikfilippas (approved):
Ok good, you just need to delete the test that fails. I'm approving either way.

PR #1074, opened 2023-04-19
IA profile v3

damonge:
Ports #922 (by @chrgeorgiou) directly to v3 API


carlosggarcia (approved):
As discussed with @damonge, I have checked all the non-IA changes and they look good to me.

nikfilippas:
I'll chime before this gets merged just to point out that having flags indicating the NC profiles is probably a bad idea. OOP works with inheritance, so it would be best to define an HOD superclass and then use composition, instead of reinstating flags for everything.
Having flags like that clutters the code, and makes it error prone.
https://stackoverflow.com/questions/41001932/oo-design-inheritance-vs-type-enum-variable

chrgeorgiou:
Going through the code now, will post my comments here. To start, I see in the [tests](https://github.com/LSSTDESC/CCL/blob/1bbe1562134fb0ccec8a34ce9249d7ad2372c985/pyccl/tests/test_profiles.py#L645) that the halo model is now called by 
`cM = ccl.halos.ConcentrationDuffy08(mass_def="200m")` and `ccl.halos.SatelliteShearHOD(concentration=cM)` but `ccl.halos.SatelliteShearHOD(cM)` throws an error. Is there no way to initialize the class now without keyword arguments?

benchmarks/test_ia_profile.py
```diff
@@ -0,0 +1,94 @@
...
+    # Below we are benchmarking the normalization factor of the
+    # delta-E power spectrum Eq. (17) in Fortuna et al. 2021.
+    # Specifically, the term f_s(z)*<N_s|M>/n_s(z).
+    mass_benchmark, norm_benchmark = np.loadtxt(
+        './benchmarks/data/IA_halomodel_norm_term.dat', unpack=True)
+    cosmo = ccl.Cosmology(Omega_c=0.25, Omega_b=0.05, Omega_k=0,
+                          sigma8=0.81, n_s=0.96, h=1.)
+    cosmo.compute_sigma()
+    Ns_M_mean = sat_gamma_HOD._Ns(mass_benchmark, 1.)
+    norm = Ns_M_mean / sat_gamma_HOD.get_normalization(cosmo, 1, hmc)
```
chrgeorgiou, line 93: The normalisation here divides by the `get_normalization` while in [my PR](https://github.com/LSSTDESC/CCL/blob/3ed354cbfcd2cfbb51f8ce6ccf4f4bd1d480d998/benchmarks/test_haloprofile.py#L264) I multiply. Has the normalization changed in v3 for profiles?

pyccl/halos/profiles/IA.py
```diff
@@ -0,0 +1,426 @@
...
+                          self._I_integral(j + var1_add, m - j) *
+                          np.sin(thk) ** j * np.cos(thk) ** (m - j))
+            l_sum += binom(l, m) * binom((l + m - 1) / 2, l) * m_sum
+        if phik is not None:
+            l_sum *= np.exp(1j * 2 * phik)
+        return 2 ** l * l_sum
+
+    def get_normalization(self, cosmo, a, hmc):
+        """Returns the normalization of this profile, which is the
+        mean galaxy number density.
```
chrgeorgiou, line 312: Currently testing this normalization but this seems to not be doing the right thing. Also, in the `_get_prefactor` method in [my PR](https://github.com/LSSTDESC/CCL/blob/3ed354cbfcd2cfbb51f8ce6ccf4f4bd1d480d998/pyccl/halos/profiles.py#L1871) the normalization is 1/n_gal.

benchmarks/test_ia_profile.py
damonge: yes, the normalization is now the thing you divide by

pyccl/halos/profiles/IA.py
damonge: Here it's n_gal (the galaxy number density).

damonge:
@chrgeorgiou yep, indeed. In v3 we introduced a star operator to force users to pass some keyword-only arguments (this is to prevent some functions doing the wrong thing silently only because users passed arguments in the wrong order - we realised there were many instances where this could happen).

Your IA profile is the first fully v3 piece of code!

pyccl/halos/profiles/IA.py
chrgeorgiou: Okay, I have run some more tests and this does what it should do, but now the user has to explicitly request the normalization of the profiles (which is the correct thing to do). I also noticed this will be the default option in v3, correct?

chrgeorgiou (approved):

PR #1083, opened 2023-05-09
API restoration

damonge:
The current version of the code implements the v3 API while attempting to preserve the v2 API. In this PR, I made a copy of the unit tests and benchmarks (not part of the PR) as they were before any of these modifications took place (in the latest released version, v2.7). I then ran the tests, which use the pure v2 API, and ensured that they pass.

In doing so, I weeded out a couple of bugs and restored some pieces of code that strictly speaking are part of the v2 API.

Now the caveats. Not all old tests passed, but the reasons behind those that didn't pass are all reasonable:
- Some function calls now raise errors with a different error type than in v2. The new error types are more appropriate, and hence this should be interpreted as a bugfix.
- Some function calls don't raise an error or a warning anymore. These were again cases where we realised this was inappropriate (and in fact it helped close a few open issues), so it should be understood as a bugfix.
- Base classes for things like `MassFunction`s , `HaloBias`s etc. cannot be instantiated anymore. It is doubtful that anyone was using them (as they are completely useless), so I think it's worth living with it as a very mild API breakage.
- The array of neutrino masses that `Cosmology`s hold now always holds the right number of masses (before it had a 3-element array, regardless of how many massive neutrinos there were).
- Some parameters that had a default value of `None` now have a default value. In all cases the default value attempts to mimic the behaviour of it being `None` in the old API.

I'll explain some of the changes as comments below.

### DIFF
--- a/pyccl/cosmology.py
+++ b/pyccl/cosmology.py
@@ -694,7 +694,7 @@ class Cosmology(CCLObject):
         prf = hal.HaloProfileNFW(concentration=c)
         hmc = hal.HMCalculator(mass_function=hmf, halo_bias=hbf,
                                mass_def=mdef)
-        return hal.halomod_Pk2D(self, hmc, prf)
+        return hal.halomod_Pk2D(self, hmc, prf, normprof1=True)
 
     @cache(maxsize=3)
     def _compute_nonlin_power(self):
--- a/pyccl/halos/halo_model.py
+++ b/pyccl/halos/halo_model.py
@@ -57,8 +57,12 @@ class HMCalculator(CCLAutoRepr):
                  log10M_min=8., log10M_max=16., nM=128,
                  integration_method_M='simpson', k_min=1E-5):
         # Initialize halo model ingredients.
-        self.mass_def, self.mass_function, self.halo_bias = MassDef.from_specs(
-            mass_def, mass_function=mass_function, halo_bias=halo_bias)
+        out = MassDef.from_specs(mass_def, mass_function=mass_function,
+                                 halo_bias=halo_bias)
+        if len(out) != 3:
+            raise ValueError("A valid mass function and halo bias is "
+                             "needed")
+        self.mass_def, self.mass_function, self.halo_bias = out
 
         self.precision = {
             'log10M_min': log10M_min, 'log10M_max': log10M_max, 'nM': nM,
--- a/pyccl/halos/halo_model.py
+++ b/pyccl/halos/halo_model.py
@@ -83,6 +83,14 @@ class HMCalculator(CCLAutoRepr):
         self._cosmo_mf = self._cosmo_bf = None
         self._a_mf = self._a_bf = -1
 
+    def _fix_profile_mass_def(self, prof):
+        # TODO v3: remove this (in v3 all profiles have a mass_def).
+        # If profile has no mass definition assigned, assign one.
+        if prof.mass_def is None:
+            warnings.warn("In v3 all profiles will need an associated "
+                          "mass definition.", CCLDeprecationWarning)
+            prof.mass_def = self.mass_def
+
     def _integ_spline(self, fM, log10M):
         # Spline integrator
         return _spline_integrate(log10M, fM, log10M[0], log10M[-1])
@@ -160,6 +168,7 @@ class HMCalculator(CCLAutoRepr):
         Returns:
             float or array_like: integral value.
         """
+        self._fix_profile_mass_def(prof)
         self._check_mass_def(prof)
         self._get_ingredients(cosmo, a, get_bf=False)
         uk0 = prof.fourier(cosmo, self.precision['k_min'], self._mass, a).T
@@ -250,6 +259,7 @@ class HMCalculator(CCLAutoRepr):
             float or array_like: integral values evaluated at each
             value of `k`.
         """
+        self._fix_profile_mass_def(prof)
         self._check_mass_def(prof)
         self._get_ingredients(cosmo, a, get_bf=False)
         uk = prof.fourier(cosmo, k, self._mass, a).T
@@ -278,6 +288,7 @@ class HMCalculator(CCLAutoRepr):
             float or array_like: integral values evaluated at each
             value of `k`.
         """
+        self._fix_profile_mass_def(prof)
         self._check_mass_def(prof)
         self._get_ingredients(cosmo, a, get_bf=True)
         uk = prof.fourier(cosmo, k, self._mass, a).T
@@ -317,6 +328,8 @@ class HMCalculator(CCLAutoRepr):
         """
         if prof2 is None:
             prof2 = prof
+        self._fix_profile_mass_def(prof)
+        self._fix_profile_mass_def(prof2)
 
         self._check_mass_def(prof, prof2)
         self._get_ingredients(cosmo, a, get_bf=False)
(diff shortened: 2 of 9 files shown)


pyccl/cosmology.py
```diff
@@ -694,7 +694,7 @@ def _get_halo_model_nonlin_power(self):
         prf = hal.HaloProfileNFW(concentration=c)
         hmc = hal.HMCalculator(mass_function=hmf, halo_bias=hbf,
                                mass_def=mdef)
-        return hal.halomod_Pk2D(self, hmc, prf)
+        return hal.halomod_Pk2D(self, hmc, prf, normprof1=True)
```
damonge, line 697: without this, the halo model power spectrum as a Cosmology power spectrum option (which will be deprecated in v3) gave the wrong answer.

pyccl/halos/halo_model.py
```diff
@@ -57,8 +57,12 @@ def __init__(self, *, mass_function, halo_bias, mass_def=None,
...
                  integration_method_M='simpson', k_min=1E-5):
         # Initialize halo model ingredients.
-        self.mass_def, self.mass_function, self.halo_bias = MassDef.from_specs(
-            mass_def, mass_function=mass_function, halo_bias=halo_bias)
+        out = MassDef.from_specs(mass_def, mass_function=mass_function,
+                                 halo_bias=halo_bias)
+        if len(out) != 3:
+            raise ValueError("A valid mass function and halo bias is "
+                             "needed")
+        self.mass_def, self.mass_function, self.halo_bias = out
```
damonge, line 65: This is just here to throw an actually useful error if any of these arguments are set to None (the corresponding python-level error was extremely obscure)

pyccl/halos/halo_model.py
```diff
@@ -79,6 +83,14 @@ def __init__(self, *, mass_function, halo_bias, mass_def=None,
...
         self._a_mf = self._a_bf = -1
 
+    def _fix_profile_mass_def(self, prof):
+        # TODO v3: remove this (in v3 all profiles have a mass_def).
+        # If profile has no mass definition assigned, assign one.
+        if prof.mass_def is None:
+            warnings.warn("In v3 all profiles will need an associated "
+                          "mass definition.", CCLDeprecationWarning)
+            prof.mass_def = self.mass_def
+
```
damonge, line 93: This was necessary in order to ensure that we could use profiles that were instantiated without a mass definition attached to them from scratch (this will not be possible in v3, but it was in v2).

pyccl/halos/pk_4pt.py
```diff
@@ -506,19 +514,19 @@ def halomod_Tk3D_SSC(
         def _get_counterterm(pA, pB, p2pt, nA, nB, i11_A, i11_B):
             """Helper to compute counter-terms."""
             # p : profiles | p2pt : 2-point | n : norms | i11 : I_1_1 integral
-            bA = i11_A / nA if isinstance(pA, ProfNC) else np.zeros_like(k_use)
-            bB = i11_B / nB if isinstance(pB, ProfNC) else np.zeros_like(k_use)
+            bA = i11_A / nA if pA.is_number_counts else np.zeros_like(k_use)
+            bB = i11_B / nB if pB.is_number_counts else np.zeros_like(k_use)
```
damonge, line 518: Using profile class type broke the API.

pyccl/halos/pk_4pt.py
```diff
@@ -506,19 +514,19 @@ def halomod_Tk3D_SSC(
...
-        if isinstance(prof, ProfNC) or isinstance(prof2, ProfNC):
+        if prof.is_number_counts or prof2.is_number_counts:
             dpk12[ia] -= _get_counterterm(prof, prof2, prof12_2pt,
                                           norm1, norm2, i11_1, i11_2)
 
-        if isinstance(prof3, ProfNC) or isinstance(prof4, ProfNC):
+        if prof3.is_number_counts or prof4.is_number_counts:
             if (prof, prof2, prof12_2pt) == (prof3, prof4, prof34_2pt):
-                dpk34[ia] -= dpk12[ia]
+                dpk34[ia] = dpk12[ia]
```
damonge, line 529: this was a huge bug

pyccl/pyutils.py
```diff
@@ -677,3 +677,22 @@ def check_openmp_threads():
...
+    import warnings
+    # Check that f() raises a warning, but not an error.
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        res = f(*args, **kwargs)
+    assert len(w) >= 1, "Expected warning was not raised."
+    assert issubclass(w[0].category, wtype), \
+        "Warning raised was the wrong type (got %s, expected %s)" % (
+            w[0].category, wtype)
+    return res
```
damonge, line 698: formally speaking, this was part of the API, even though we don't really use it.

pyccl/tests/test_tkkssc.py
```diff
@@ -128,3 +130,119 @@ def test_tkkssc_linear_bias_raises():
     """Test that it raises if the profile is not NFW."""
     with pytest.raises(TypeError):
         ccl.halos.halomod_Tk3D_SSC_linear_bias(COSMO, HMC, prof=GNFW)
+
```
damonge, line 133: This restores a test that was lost in the last few PRs, but which is quite useful (it actually uncovered the bug mentioned above)

pyccl/halos/halo_model.py
```diff
@@ -313,6 +328,8 @@ def I_0_2(self, cosmo, k, a, prof, *, prof2=None, prof_2pt):
         """
         if prof2 is None:
             prof2 = prof
+        self._fix_profile_mass_def(prof)
+        self._fix_profile_mass_def(prof2)
```
carlosggarcia, line 332: How does this work? If `prof2` is not `None`, I think it will overwrite `prof`

carlosggarcia: Same in all cases below

pyccl/halos/pk_2pt.py
```diff
@@ -104,6 +104,8 @@ def halomod_power_spectrum(cosmo, hmc, k, a, prof, *,
         prof2 = prof
     if prof_2pt is None:
         prof_2pt = Profile2pt()
+    hmc._fix_profile_mass_def(prof)
```
carlosggarcia, line 107: As before

pyccl/halos/pk_4pt.py
```diff
@@ -445,6 +449,10 @@ def halomod_Tk3D_SSC(
     # define all the profiles
     prof, prof2, prof3, prof4, prof12_2pt, prof34_2pt = \
         _allocate_profiles(prof, prof2, prof3, prof4, prof12_2pt, prof34_2pt)
+    hmc._fix_profile_mass_def(prof)
+    hmc._fix_profile_mass_def(prof2)
+    hmc._fix_profile_mass_def(prof3)
+    hmc._fix_profile_mass_def(prof4)
```
carlosggarcia, line 455: As in the other files

carlosggarcia (changes requested):
Just a few comments regarding the mass definitions. Other than that, it looks good.

pyccl/halos/halo_model.py
damonge: `_fix_profile_mass_def` only defines the profile's `mass_def` if it is `None`. Otherwise it doesn't do anything. So the second call will not modify `prof`.

carlosggarcia: hmm, and what if they don't have the same `mass_def`?

damonge: An error will be thrown in the next line (334)

carlosggarcia (approved):
LGTM!

PR #1086, opened 2023-05-18
Bug from sed

damonge:
I introduced a bug in #1082 by using `sed` to search-replace. I'm amazed gha didn't pick it up.
This fixes it.


carlosggarcia (approved):
I missed this file too!

PR #1092, opened 2023-06-15
Integration of N5K winners into CCL

slosar:
This is a draft pull request, just to show you what we are doing. There is still cruft that will be removed before we do a full pull request.

A lot of heavy pulling here has been done by @paulrogozenski .

Summary of changes:
 * All reference to angpow has been removed
 * All placeholders for non-limber in the C-layer have been removed -- non-limber methods are by default pure python (but they can enter C-layer if they wish so)
 * interface to `angular_cell` has changed a little (more later)
 * FKEM method has been implemented
 * a pytest benchmark has been implemented in `test_nonlimber.py`

The main interface to non-limber is by setting `l_limber` to a cross-over value or "auto" in call to `angular_cell`. 
We felt the need that non-limber optionally returns some meta value (i.e. you might be interested which cross-over ell did it actually use) and so there is an extra option `return_meta`, in which case it returns both cells and a dictionary with meta info (currently just l_limber, but could be more, like timing, etc). One can also specify which method to use with `limber_integration_method` parameter.

Here are the current questions / comments / issues:
 * Is the new interface to `angular_cell` appropriate?
 * For the benchmark, we took the N5K test, but just the "fullwidth" version of it. The problem with this test is that (unless we're doing something wrong), the Limber actually passes it pretty well (i.e worst chi2 is 0.3 or so). Guidance of which tests would be appropriate for these tests would be useful.
 * Caching into tracer quantities is not yet implemented, but is being worked on.
 * MATTER method, as coded in N5K is very tightly coupled to CLASS, beyond just fftlog. Not clear that reimplementing would not be easier than porting it, but this would be a significant amount of work. We're still looking at it.
 
Comments / suggestions are welcome at this stage.


damonge:
Hey @slosar @paulrogozenski , would you like me to look into this? Right now review is a bit of a mess because so many files appear as modified. This is because I think you're merging into a very old version of master (before the v3 transition). It'd be much easier (for me!) if you could merge the current master, although solving conflicts might be a bit of a nightmare. It may actually be easier to port your changes to a new branch coming from the current master... Happy to help either way.

slosar:
> Hey @slosar @paulrogozenski , would you like me to look into this? Right now review is a bit of a mess because so many files appear as modified. This is because I think you're merging into a very old version of master (before the v3 transition). It'd be much easier (for me!) if you could merge the current master, although solving conflicts might be a bit of a nightmare. It may actually be easier to port your changes to a new branch coming from the current master... Happy to help either way.

@damonge Erhm, I was merging main in quite regularly until recently and this time I got a bunch of conflicts on completelly unrelated files. Not sure why it didn't merge cleanly, didn't someone mess with the main branch. Anyways, all those conflicts are now resolved.

cmake/Modules/BuildAngpow.cmake
damonge, line 1: What are all of these empty files?

slosar: I don't see them. We tried to remove any angpow vestiges, so perhaps left-overs from early builds?

pyccl/cells.py
```diff
@@ -22,35 +33,58 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
...
+            % limber_integration_method
+        )
+    if non_limber_integration_method not in ["FKEM", "MATTER"]:
+        raise ValueError(
+            "Non-Limber integration method %s not supported"
+            % limber_integration_method
+        )
+    if type(l_limber) == str:
+        if l_limber != "auto":
+            raise ValueError("l_limber cannot be a string other than 'auto'")
```
damonge, line 78: A more informative error would be "`l_limber` must be an integer or 'auto'"

pyccl/cells.py
```diff
@@ -69,17 +103,52 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
     if not (np.diff(ell_use) > 0).all():
         raise ValueError("ell values must be monotonically increasing")
 
+    if auto_limber or (type(l_limber) != str and ell_use[0] < l_limber):
```
damonge, line 106: I don't think you need to check for the type of `l_limber` at this point. If it's not a number it will just fail.

pyccl/cells.py
```diff
@@ -22,35 +33,58 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
...
+        auto_limber = True
+    else:
+        auto_limber = False
 
     # we need the distances for the integrals
     cosmo.compute_distances()
 
     # Access ccl_cosmology object
     cosmo_in = cosmo
-    cosmo = cosmo.cosmo
```
damonge, line 53: This whole cosmo_in vs. cosmo makes things confusing (this is on me, probably). Can we just use `cosmo` for the python object and `cosmo.cosmo` for the C structure everywhere?

pyccl/cells.py
```diff
@@ -69,17 +103,52 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
...
+                limber_max_error,
+            )
+        # else:  # it has to be matter, since we checked the input
+        #     l_limber, cl_non_limber, status = implement_MATTER(
+        #         cosmo, clt1, clt2, psp, ell_use, l_limber, limber_max_error
+        #     )
+        check(status, cosmo=cosmo_in)
+    else:
+        cl_non_limber = np.array([])
+    cosmo = cosmo.cosmo
```
damonge, line 124: which would mean removing this

pyccl/pyutils.py
```diff
@@ -522,6 +522,45 @@ def _fftlog_transform(rs, frs,
     return ks, fks
 
 
+def _fftlog_transform_general(
+    rs, frs, mu, q, spherical_bessel, bessel_deriv, window_frac
+):
```
damonge, line 527: Can we add a comment (doesn't have to be a proper docstring) here saying a bit what the expected inputs and outputs are? Nothing long, just so developers in the future can see what this does and perhaps reuse it for other calculations.

pyccl/nonlimber_FKEM.py
```diff
@@ -0,0 +1,234 @@
...
+ to compute integrals over spherical bessel functions
+"""
+import numpy as np
+from . import lib, check
+from .pyutils import integ_types
+from scipy.interpolate import interp1d
+from pyccl.pyutils import _fftlog_transform_general
+import pyccl as ccl
+
+
```
damonge, line 16: One general comment/question before I continue reviewing: since (presumably) we may implement another non-limber method in the future, and the code for angular C_ells now lives in two different files, would it make sense to put all the C_ell code into its own submodule for tidyness and encapsulation? Otherwise I find it weird to have these two files lying around instead of just putting all the code in `cells.py`.

damonge: Associated comment: all functions/methods not starting with `_` are considered public and part of the API, and hence should be properly documented. In this sense, the method below (`get_general_params`) is very specific of FKEM, and the name would suggest otherwise to an external user (especially since it's not documented).

pyccl/nonlimber_FKEM.py
```diff
@@ -0,0 +1,234 @@
...
+    if b <= 0:
+        deriv = 0
+    else:
+        deriv = b
+
+    return best_nu, deriv, plaw
+
+
+def get_average_a(a_arr, dndz):
```
damonge, line 35: Likewise for this function (re. making it private or renaming and documenting it)

pyccl/nonlimber_FKEM.py
```diff
@@ -0,0 +1,234 @@
...
+    for i in range(len(dndz)):
+        if dndz[i] != 0:
+            new_arr.append(dz * z_arr[i] * dndz[i])
+    z_mean = np.sum(np.array(new_arr))
+    return 1.0 / (1.0 + z_mean)
+
+
+def nonlimber_FKEM(
+    cosmo, clt1, clt2, p_of_k_a, ls, l_limber, limber_max_error
+):
```
damonge, line 48: This would need proper documentation if it's meant to be used as part of the API. If it isn't, then it should be made private.

damonge (commented):
Thanks @paulrogozenski and @slosar. This is great. This first review is only to discuss how the API should be structured. I'll then look at the actual implementation of the algorithm and then the tests. The main question right now is where things should live. I think the options are:
- We make a `cells` module that contains the different methods.
- We put all the code into `cells.py`, and it'll only have a single user-facing function (`angular_cl`) with everything else being private functions.

Either would work for me, and would be tidier than the current structure where things are a bit scattered into two different files.

src/nonlimber_matter_utils/matter.c
damonge, line 1: OK, I'll stop here, just remove the whole `nonlimber_matter_utils` directory, since MATTER doesn't seem to be implemented yet, right?

damonge (commented):
Thanks @paulrogozenski . More review comments.

src/ccl_fftlog.c
```diff
@@ -37,6 +51,18 @@ These are the C++ bindings
 #endif
 /* This code is FFTLog, which is described in arXiv:astro-ph/9905191 */
 
+static double complex complex_mult(double complex a, double complex b){
```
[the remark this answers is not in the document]
paulrogozenski: yes, this is unused in the code. Delted!

pyccl/cells.py
```diff
@@ -69,17 +110,53 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
...
+    cl = np.concatenate((cl_non_limber, cl_limber))
     if np.ndim(ell) == 0:
         cl = cl[0]
 
     # Free up tracer collections
     lib.cl_tracer_collection_t_free(clt1)
     lib.cl_tracer_collection_t_free(clt2)
 
+    if return_meta:
+        meta = {"l_limber": l_limber}  # add other things as needed
```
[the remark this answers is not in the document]
paulrogozenski: the user may enter 'auto', meaning the non-limber calculation will end after the limber/non-limber calculations are sufficiently close to one another (also set by the user). This will return the l-value indicating where limber calculations begin

pyccl/cells.py
```diff
@@ -69,17 +110,53 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
     if not (np.diff(ell_use) > 0).all():
         raise ValueError("ell values must be monotonically increasing")
 
+    if auto_limber or (type(l_limber) is not str and ell_use[0] < l_limber):
```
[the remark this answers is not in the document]
paulrogozenski: I was getting some errors evaluating the if statement when l_limber is a string without this statement (the if statement wasn't short-circuiting and the type comparisons were giving errors)

pyccl/cells.py
```diff
@@ -22,35 +34,64 @@ def angular_cl(cosmo, tracer1, tracer2, ell, *,
...
+            category=CCLWarning,
+        )
+
     if limber_integration_method not in integ_types:
         raise ValueError(
-            f"Unknown integration method {limber_integration_method}.")
+            "Limber integration method %s not supported"
+            % limber_integration_method
+        )
+    if non_limber_integration_method not in ["FKEM", "MATTER"]:
```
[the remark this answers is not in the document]
paulrogozenski: No, we don't (yet). Removed!

pyccl/tests/test_cells.py
```diff
@@ -78,8 +89,34 @@ def test_cells_raise_ell_reversed(ells):
...
-                       limber_integration_method='guad')
+        ccl.angular_cl(
+            COSMO, LENS, LENS, ells, limber_integration_method="guad"
+        )
+
+    with pytest.raises(ValueError):
+        LENS_2 = ccl.WeakLensingTracer(COSMO, dndz=(ZZ, np.zeros(len(ZZ))))
+        ccl.angular_cl(
+            COSMO, LENS, LENS_2, ells, limber_integration_method="quad"
+        )
```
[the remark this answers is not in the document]
paulrogozenski: This is testing that pyccl doesn't return nonsense when given flat, zero n(z)'s. CCL did not catch this before and instead gave a scipy integration error.

pyccl/tracers.py
```diff
@@ -722,6 +722,24 @@ class NzTracer(Tracer):
...
+    def get_chi_fft(self, tracer, Nchi, chimin, chimax, ell):
+        temp = self.chi_fft_dict.get((tracer, Nchi, chimin, chimax, ell))
+        if temp is None:
+            return None, None
+        return temp[0], temp[1]
+
+    def set_chi_fft(self, tracer, Nchi, chimin, chimax, ell, ks, fft):
+        self.chi_fft_dict[(tracer, Nchi, chimin, chimax, ell)] = (ks, fft)
+        return ks, fft
+
```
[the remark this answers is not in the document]
paulrogozenski: This stores the non-limber fft integration over chi for the FKEM method to reduce computing the same integrals multiple times. This should live in the nztracer space, as the non-limber computation is only relevant for 2pt statistics.

compare_nonlimber_FKEM.ipynb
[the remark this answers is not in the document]
paulrogozenski: I will remove this notebook

examples/Benchmark_Comparisons.ipynb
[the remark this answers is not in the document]
paulrogozenski: I ran some examples, so the files were edited and some new files were made. There are no actual edits to the 'examples' folder.

pyccl/cells.py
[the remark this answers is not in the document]
damonge: right, great.

examples/pk_spline_bm_v2.npz
damonge, line 1: do we need this file?

pyccl/tests/test_cells.py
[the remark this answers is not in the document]
damonge: OK, great

pyccl/tracers.py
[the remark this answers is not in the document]
damonge: Hmm, so FKEM right now would only work with `NzTracer`s as input? How about the other tracer types?

src/ccl_fftlog.c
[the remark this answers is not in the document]
damonge: Doesn't show as deleted

benchmarks/test_nonlimber.py
```diff
@@ -0,0 +1,269 @@
...
+                                  nk_per_decade=200)
+    ptc.update_ingredients(cosmo)
+    ptc_lin = pt.EulerianPTCalculator(with_NC=True, with_IA=False,
+                                      log10k_min=-4, log10k_max=4,
+                                      nk_per_decade=200,
+                                      b1_pk_kind='linear',
+                                      bk2_pk_kind='linear')
+    ptc_lin.update_ingredients(cosmo)
+    ptt_g = [pt.PTNumberCountsTracer(b1=bias) for bias in tpar["b_g"]]
+    ptt_m = pt.PTMatterTracer()
```
damonge, line 188: Humm, why do you need any perturbation theory here? You're only testing linearly biased things, right? In which case you can just pass the bias in the `NumberCountsTracer` objects. Perhaps I'm missing something.

benchmarks/test_nonlimber.py
```diff
@@ -0,0 +1,269 @@
...
+                tracers2[cross_type][i2],
+                ells,
+                p_of_k_a=p_of_k_a,
+                p_of_k_a_lin=p_of_k_a_lin,
+                l_limber='auto',
+                non_limber_integration_method=method,
+                return_meta=True
+            )
+            l_limber = meta['l_limber']
+        cls *= bias_fact
```
damonge, line 255: Humm, why do you need this?

damonge (commented):
Sorry for the delay @paulrogozenski . A few final comments. I think we're almost there, and I'll be on standby to review quickly.

benchmarks/test_nonlimber.py
paulrogozenski: the user may choose 'auto' for the non-limber code, which means it computes as limber after a convergence criteria is met. This allows the transitional ell mode to be returned to the user, which can be valuable for some science/systematics investigations.

benchmarks/test_nonlimber.py
paulrogozenski: This test confirms that using PTTracers in the non-limber code still works as-expected.

examples/pk_spline_bm_v2.npz
paulrogozenski: I believe this is used to run benchmark tests for code unrelated to this PR. It is still in the main branch, so perhaps it is still necessary to have: https://github.com/LSSTDESC/CCL/blob/master/examples/pk_spline_bm_v2.npz

benchmarks/test_nonlimber.py
damonge: no, sorry, I was referring to the next line. Why do you need to multiply by the biases? Isn't that part of either the C_ell tracers or of the input Pks?

benchmarks/test_nonlimber.py
damonge: but is it right that you could achieve the same result simply passing biases through `NumberCountsTracer`s and using the linear and nonlinear matter power spectra?

slosar: I think we should at some point get to the bottom of this. I grudingly added these PTtracer into tests, but fundamentally I still don't understand why does the non-limber calculate ever needs to know the linear theory.

benchmarks/test_nonlimber.py
paulrogozenski: This is addressed in lines 93-95 of the same file. It is there so that we can use the same approach for both Tracers and PTTracers in the same test function.

benchmarks/test_nonlimber.py
paulrogozenski: This is possible for NZTracers, but not PTTracers. PTTracers can only assume _one_ underlying pk model. This code is as-written for generality regardless of tracer type.

pyccl/_nonlimber_FKEM.py
```diff
@@ -0,0 +1,273 @@
...
+                    nu,
+                    1,
+                    float(deriv),
+                    float(plaw),
+                )
+                check(status, cosmo=cosmo)
+
+                clt1.set_chi_fft(
+                    clt1._trc[i], Nchi, chi_min, chi_max, ell, k, fk1
+                )
```
damonge, line 182: One last question about the algorithm itself: I fail to see what the advantage is of setting and storing these at the level of the tracers, as opposed to just precomputing them and storing them within this function.

The way I see it (and this may be where my misconception lies), `Nchi`, `chi_min`, `chi_max` and perhaps `ell`, will likely change with each new power spectrum you calculate (since e.g. `chi_min` and `chi_max` actually depend on the specific pair of tracers you're correlating, and not on an individual tracer properties). So you end up storing one of these in the tracer for each individual power spectrum you calculate. If that's the case, you don't really reuse these, so there would be no advantage to precomputing them locally within the function.

A parallel worry of storing these within the tracers is that, if what I'm saying above is correct, the associated tracer-level dictionaries may end up being huge if there's a lot of cross-correlations (if you have N tracers, the number of dictionary elements added up across all tracers would grow like N^3), which could cause memory issues.

Might be that I'm missing something, so let me know what you think.

pyccl/tracers.py
```diff
@@ -722,6 +757,13 @@ class NzTracer(Tracer):
     :func:`NumberCountsTracer` and :func:`WeakLensingTracer`.
     """
 
+    def __init__(self):
+        """By default this `Tracer` object will contain no actual
+        tracers
+        """
+        # Do nothing, just initialize list of tracers
+        super().__init__()
+
```
damonge, line 766: Why do you need to add this? This seems equivalent to just letting this class use its parent's init function.

damonge:
@paulrogozenski , thanks a lot again for this. I've put just one final comment above about the need to precompute certain things.

I also added one small commit in which I fixed a couple small things and streamlined the benchmark test script. The logic behind the latter is that the role of unit/benchmark tests is to test, as quickly as possible, that a given part of the code is doing what it should (in this case, in terms of accuracy too). So I think we should separate this (which can be done by limiting ourselves only to the linear bias model and avoiding the Limber test), from an example notebook that showcases how the new functionality interacts with other parts of the code (e.g. PT calculators, or how different Limber and non-Limber C_ells are). So, what I did is shrink a bit the benchmark test code, leaving only what we need to ensure the method performs as it did for the N5K challenge, and then opened a PR in CCLX where I've put your longer test, which we should then transform into a proper notebook once this branch is merged.

pyccl/tracers.py
```diff
@@ -447,6 +448,38 @@ def get_angles_derivative(self):
         """
         return np.array([t.der_angles for t in self._trc])
 
+    def get_chi_fft(self, tracer, Nchi, chimin, chimax, ell):
```
damonge, line 451: Just one final query before I approve: could we make this function name a bit more verbose, and also make it a private function (so it does not form part of the API, in case we want to change it in the near future)? Suggestion: `_get_fkem_fft`.

Same query for `_set_fkem_fft`

damonge (changes requested):

PR #1093, opened 2023-06-29
Projected Einasto profile from numerical integration

hsinfan1996:
Using FFTLog is faster, but using numerical integration is more accurate at large radius.


pyccl/halos/profiles/einasto.py
```diff
@@ -37,16 +38,23 @@ class HaloProfileEinasto(HaloProfileMatter):
             concentration-mass relation to use with this profile.
         truncated (:obj:`bool`): set to ``True`` if the profile should be
             truncated at :math:`r = r_\\Delta`.
+        projected_integrate (:obj:`bool`): set to ``True`` to calculate the
+            projected profile with numerical integration.
```
damonge, line 42: Can we call this `projected_quad` or something like that? Just to make it clear that this is as opposed to FFTLog.

pyccl/tests/test_profiles.py
```diff
@@ -548,6 +551,27 @@ def test_hernquist_cumul2d_accuracy(fourier_analytic):
...
+                                      projected_integrate=False)
+
+    M = 1E14
+    a = 0.5
+    rt = np.logspace(-3, 2, 1024)
+    srt1 = p1.projected(COSMO, rt, M, a)[:500]
+    srt2 = p2.projected(COSMO, rt, M, a)[:500]
+
+    res2 = np.fabs(srt2/srt1-1)
+    assert np.all(res2 < 6e-2)
```
damonge, line 572: humm, this is only 6% accurate. Is the limiting factor here FFTLog or quad?

pyccl/halos/profiles/einasto.py
```diff
@@ -105,3 +113,30 @@ def _real(self, cosmo, r, M, a):
...
+        R_M = self.mass_def.get_radius(cosmo, M_use, a) / a
+        c_M = self.concentration(cosmo, M_use, a)
+        R_s = R_M / c_M
+
+        alpha = self._get_alpha(cosmo, M_use, a)
+
+        def integrand(z):
+            x = np.sqrt(z**2. + r_use[None, :]**2.) / R_s[:, None]
+            return np.exp(-2. * (x**alpha[:, None] - 1.) / alpha[:, None])
+        prof = quad_vec(integrand, 0., np.inf)[0]
```
damonge, line 131: Hmm is this accurate if we use `truncated=True`?

damonge (commented):
Thanks a lot @hsinfan1996 . A few comments below.

pyccl/halos/profiles/einasto.py
hsinfan1996: Yes, I was thinking about the naming too.

pyccl/halos/profiles/einasto.py
hsinfan1996: Exactly the same as `truncated=False`

damonge: humm, but shouldn't the range of integration be different here? Or shouldn't the integrand be truncated?

pyccl/tests/test_profiles.py
hsinfan1996: I am still testing this part.

pyccl/halos/profiles/einasto.py
hsinfan1996: The whole profile is obtained and then truncated just like all the other profiles. And yes, I should make the calculation more efficient by not doing integration for r_use > R_M.

pyccl/tests/test_profiles.py
hsinfan1996: The limiting factor is the truncated profile given by FFTLog. I set `truncated=False` for FFTLog and they agree to 6e-5.

hsinfan1996: A plot showing the difference between truncated and untruncated `_projected_quad`.
I only plot the range without truncation and thee difference is exactly 0.
download-1

pyccl/halos/profiles/einasto.py
hsinfan1996: Fixed.

hsinfan1996: I could not get `truncated` to work due to the array shapes of the broadcasted array and `r_use`. All the other profiles of different halo model  do not support `truncated` either, so this implementation is consistent with the current settings.

damonge: sorry, to clarify, are you suggesting we disable the truncated option for einasto altogether? Not that I'd be opposed.

pyccl/halos/profiles/einasto.py
```diff
@@ -37,16 +38,27 @@ class HaloProfileEinasto(HaloProfileMatter):
             concentration-mass relation to use with this profile.
         truncated (:obj:`bool`): set to ``True`` if the profile should be
             truncated at :math:`r = r_\\Delta`.
+        projected_quad (:obj:`bool`): set to ``True`` to calculate the
+            projected profile with numerical integration.
         alpha (:obj:`float` or :obj:`str`): :math:`\\alpha` parameter, or
             set to ``'cosmo'`` to calculate the value from cosmology.
     """
     __repr_attrs__ = __eq_attrs__ = (
         "truncated", "alpha", "mass_def", "concentration", "precision_fftlog",)
```
damonge, line 47: can you add `"projected_quad"` to the list of attributes here?

pyccl/tests/test_profiles.py
```diff
@@ -548,6 +555,34 @@ def test_hernquist_cumul2d_accuracy(fourier_analytic):
...
+                                      concentration=cM, truncated=False,
+                                      projected_quad=True)
+    # projected profile from FFTLog
+    p2 = ccl.halos.HaloProfileEinasto(mass_def='200c',
+                                      concentration=cM, truncated=False,
+                                      projected_quad=False)
+    # truncated projected profile from FFTLog
+    p3 = ccl.halos.HaloProfileEinasto(mass_def='200c',
+                                      concentration=cM, truncated=True,
+                                      projected_quad=False)
```
damonge, line 571: What's the point of testing against the truncated version now? Should we just remove this bit?

damonge (changes requested):
Couple more things and we're done.

pyccl/halos/profiles/einasto.py
damonge: Nevermind, I see what you did. Good for me.

pyccl/halos/profiles/einasto.py
hsinfan1996: Fixed.

pyccl/tests/test_profiles.py
hsinfan1996: The test coverage may drop. Still waiting for the checks to finish.

hsinfan1996: Test coverage dropped. Somehow the `truncated` flag of _real profile is tested here.

damonge (approved):
LGTM!

hsinfan1996:
Thank you.

PR #1096, opened 2023-07-07
Improve NFW and Hernquist `_fx_projected` and `_fx_cumul2d`

hsinfan1996:
Use `np.arcsinh` and `np.arcsin` for numerical stability when `xx` is close to 1.


hsinfan1996:
Add a negative sign instead of using `np.fabs`, which is ~50 times slower.

damonge (approved):
LGTM!

PR #1099, opened 2023-07-11
Version number on CMakeLists.txt

### LINKED ISSUE #1090: Update version number in CMakelist?
We should probably update the version here before releasing?
https://github.com/LSSTDESC/CCL/blob/d421ca0a5e42ff489fc8208f4b1e01ac6866ab16/CMakeLists.txt#L11

damonge:
Closes #1090


carlosggarcia (approved):
Cannot be automatized so that it takes the value when creating the github tag?

damonge:
Don't know of a way to do this, no. It doesn't really affect the python versioning, so I don't think it's a big issue.

PR #1094, opened 2023-07-04
Emu pk

matteozennaro:
Enable passing linear and nonlinear pk emulators to the cosmology class (baccoemu and euclid emulator 2 implemented for now, generic emulator class introduced extendable to other emulators). Added baccoemu baryonic effect emulator to baryons class.

PR #1102, opened 2023-07-12
CosmicEmu MiraTitanIV

damonge:
Implements the latest version of CosmicEmu (https://github.com/lanl/CosmicEmu/) by effectively implementing their code in python, and including the data as part of the repo (since no self-contained python wrapper exists AFAIK). We use the new emulator class created in #1094.

While doing this, I noticed an inconsistency in the emulators API (using scale factor before cosmo as an argument, which breaks with the general rule in CCL).


matteozennaro (approved):
I agree with the (cosmo, a) <-> (a, cosmo), nothing to say on the cosmic emu implementation

PR #1103, opened 2023-07-13
CosmicEmu 2017

damonge:
This implements the older version of CosmicEmu that we already had at the C level.

Here's a list of changes:
- Implemented the 2017 version of CosmicEmu, using the same framework we used in #1102 . For this, I've created a base class for Cosmicemu, made the 2022 version into a child class, and the new 2017 one into another child class. They both live in the same file. Unfortunately github is noting all code in that file as new, but the 2022 code is actually unchanged.
- Removed the C-level implementation of the same emulator, and removed it as an option in the Cosmology object. From now on, users who want to use CosmicEmu will need to pass it as any of the other emulators. Note that I've left the old benchmark test in place, but now using the python implementation, to demonstrate that we achieve the same precision as with the old implementation.
- I realised there were still a bunch of deprecated flags at the C level (to do with the halo mass function, concentration-mass relation, and baryonic effects). I've removed those.
- I've removed the caching option inside the cosmology object methods (which was not being exploited anyway). This is because it led to unwanted caching of non-linear power spectra generated with different emulators (and is an example of how careful we need to be if we ever need to benefit from this).


benchmarks/test_emu.py
```diff
@@ -3,6 +3,7 @@
 import pyccl as ccl
 
 EMU_TOLERANCE = 3.0E-2
+cemu = ccl.CosmicemuMTIIPk('tot')
```
matteozennaro, line 6: I think that for these cosmological models there is never a test for the 'cb' case, which is a bit strange to me since we also have cb (and in the other file test_cosmicemu.py there are indeed tests for the 'cb' case). Is this intended?

matteozennaro (approved):

benchmarks/test_emu.py
damonge: Yeah, this was the original test for the C-implementation, when we were mostly concerned with cosmic shear. I agree we need to test cb too (and hence the new tests!).

Saying this, I'm just realising we don't have a test for cb and tot in the presence of massive neutrinos. Let me create one now.

damonge: @matteozennaro I've added tests for cosmologies with neutrinos, and I've put the old tests in the same file as the new ones (just so they're all in the same place).

matteozennaro (approved):

PR #1105, opened 2023-07-16
Add new halo models to convert_concentration

hsinfan1996:
Add support for the Einasto and Hernquist profiles, although the Einasto profile itself uses the equation for NFW when getting the alpha parameter from cosmology. Now the function is in pure python, and it runs slower. For comparison, converting one concentration value takes 5 μs before vs 270 μs now and converting an array of 10 concentration values  takes 6 μs before vs 600 μs now. I will close this PR if this is unacceptably slow.


hsinfan1996:
Still need to update the equations in the docstrings.

hsinfan1996:
I cannot seem to fix the tests.

damonge:
Humm, although this is not something that is currently heavily used, I would rather not replace something in the code by a version that is so much slower. I can see two options:

- Leave the existing NFW code in place for NFW, and use the method you implemented for the other two profiles.
- Even better, implement the other two in C. A quick look at the code tells me it's just a matter of replacing the function whose root we need to find, so perhaps it's not so complicated.

hsinfan1996:
Not planning on touching c-level files. Since CLMM already had this functionality, will close this PR.

PR #1106, opened 2023-07-18
Allow sigma8 as an input when using CAMB for nonlinear power spectra

tilmantroester:
This PR rescales the primordial power spectrum in CAMB before computing the non-linear power spectrum, allowing using sigma8 as an input.

By default, `INTEGRATION_SIGMAR_EPSREL` is set to 1e-5, which results in a relative error of 1e-3 on sigma8. We might want to push this a bit.


pyccl/boltzmann.py
```diff
@@ -140,30 +139,39 @@ def get_camb_pk_lin(cosmo, *, nonlin=False):
         wa=cosmo['wa']
     )
 
+    cp.set_for_lmax(extra_camb_params.get("lmax", 5000))
```
damonge, line 142: why do we need this?

pyccl/boltzmann.py
```diff
@@ -140,30 +139,39 @@ def get_camb_pk_lin(cosmo, *, nonlin=False):
...
+            **options)
+    else:
+        assert camb_res.Params.NonLinear == camb.model.NonLinear_none
 
-    cp.set_for_lmax(extra_camb_params.get("lmax", 5000))
-    cp.InitPower.set_params(
-        As=A_s_fid,
-        ns=cosmo['n_s'])
+    if np.isfinite(cosmo["sigma8"]) or nonlin:
+        camb_res.calc_power_spectra()
```
damonge, line 173: It worries me that we're calling this function twice. How slow is this?

damonge:
So it's because of an update in gha?

tilmantroester:
I think because class is not forcing its cython dependence to be less than 3.0.0, the class installation now fails because cython updated to 3.0.0, which breaks class.

damonge:
OK, I say we wait a couple of days and see if class fix that. Otherwise we can try to force gha to use a previous cython version (disabling the class tests would be kind of tragic)

hsinfan1996:
classy 2.8.0 might work. Versions>2.8.0 will install Cython 3.0.0 when being installed.

hsinfan1996:
classy==2.8.0

damonge:
No luck, it seems...

hsinfan1996:
> No luck, it seems...

It doesn't work here.
https://github.com/LSSTDESC/CCL/actions/runs/5602134174/jobs/10246974180#step:8:185
pip still installs the latest Cython I guess.

hsinfan1996:
Does that mean only 2.9.4 and below are pip installable?
https://github.com/lesgourg/class_public/issues/527#issuecomment-1592711511

damonge:
@tilmantroester tests can now run, but some seem to be failing.

tilmantroester:
Can someone explain this test:
https://github.com/LSSTDESC/CCL/blob/master/pyccl/tests/test_power_mu_sigma_sigma8norm.py#L48-L61 ?

From this my understanding is that if sigma8 is given, then setting mu and Sigma should not have any effect because any rescaling of the power spectrum is undone by the sigma8 normalisation?

damonge:
Humm, I think it's just testing that initialising power spectra with A_s or its associated sigma8 gives the same result, even when you're passing MG parameters != 0

tilmantroester:
Ah, my confusion was that the input sigma8 is including MG. Should be fixed now. Looking at the MG implementation, I'm a bit worried though. It's quite opaque and the C-layer does things like creating new C cosmology structs. Since we don't support the C API any more, is that safe?
Would an API like the baryons, which modifies the power spectra (and growth rates in this case) make sense, instead of having mu and Sigma as first-class cosmology parameters? Especially since that only applies to linear scales, which are of restricted use for the LSS stuff CCL is for.

.github/environment.yml
```diff
@@ -1,4 +1,4 @@
-name: test  # default testing environment name from conda-incubator
+name: test  # default testingz environment name from conda-incubator
```
damonge, line 1: although this looks cool, perhaps remove the z

benchmarks/test_ptpk_lpt.py
```diff
@@ -12,6 +12,12 @@
                       T_CMB=2.725,
                       transfer_function='boltzmann_camb')
 
+# The accuracy of the CCL-internal sigmaR computation isn't that high
+SIGMA8_TOLERANCE = np.sqrt(
+    COSMO._accuracy_params["INTEGRATION_SIGMAR_EPSREL"]
+    * np.log(10)/(2*np.pi**2)
+)
+
```
damonge, line 20: What is this number roughly?

damonge: I'm kind of confused by why we now need this everywhere.

tilmantroester: After some digging, I think the CAMB estimate of sigma8 is quite rough, which was causing these issues. I think I can rewrite the CAMB interface to use the CCL sigma8 estimate though, which should resolve all these test accuracy problems.

damonge (approved):
LGTM!

PR #1109, opened 2023-07-20
Fix classy installation

hsinfan1996:
classy is not compatible with Cython 3.0.0.


hsinfan1996:
I don't get it...why specifying `==2.9.4` works but `<3` does not.

damonge (approved):
Awesome! Let's merge this.

PR #1111, opened 2023-07-20
Use default `Neff` in test_nonlin_camb_power

### LINKED ISSUE #964: Update `Neff = 3.044` (and revert #1024)
https://arxiv.org/abs/2008.01074
https://arxiv.org/abs/2012.02726

hsinfan1996:
close #964 and revert #1024

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -9,7 +9,6 @@ dependencies:
   - numpy
   - scipy
   - camb
-  #- isitgr
   - fast-pt
   - pytest
   - pytest-cov
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -8,12 +8,13 @@ dependencies:
   - pyyaml
   - numpy
   - scipy
-  - camb=1.5.0
-  - isitgr
+  - camb
+  #- isitgr
   - fast-pt
   - pytest
   - pytest-cov
   - pip:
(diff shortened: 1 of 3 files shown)


hsinfan1996:
With conda-installed `isitgr`, we will not be able to install the latest camb from conda. The solution is to move one or both of them to pip. Is this an acceptable solution?

.github/environment.yml
```diff
@@ -8,12 +8,13 @@ dependencies:
   - pyyaml
   - numpy
   - scipy
-  - camb=1.3.6
-  - isitgr
+  - camb
+  #- isitgr
```
damonge, line 12: Why did you leave this commented out?

damonge (commented):
Just one thing

damonge:
Fine to install isitgr from pip. Whatever works.

.github/environment.yml
hsinfan1996: I will remove it. I will also need to remove the -v arg in pip install before finalizing this PR.

damonge: OK great. I was not opposed to leaving the -v in, so up to you. I think we can merge after that, so I'll approve now so you don't have to wait for me.

damonge (approved):

PR #1113, opened 2023-07-24
Baryons objects as convenience input for Cosmology

damonge:
This implements, as discussed in the last CCL telecon, the possibility to pass baryonic effects objects as part of the cosmology initialisation. This is a convenience feature, so users can just keep using Cosmology objects as containers for the non-linear matter power spectrum, which now can include baryonic effects.

### DIFF
--- a/pyccl/cosmology.py
+++ b/pyccl/cosmology.py
@@ -17,7 +17,7 @@ import numpy as np
 from . import (
     CCLError, CCLObject, CCLParameters, CosmologyParams,
     DEFAULT_POWER_SPECTRUM, DefaultParams, Pk2D, check, lib,
-    unlock_instance, emulators)
+    unlock_instance, emulators, baryons)
 from . import physical_constants as const
 
 
@@ -174,6 +174,10 @@ class Cosmology(CCLObject):
             The transfer function to use. Defaults to 'boltzmann_camb'.
         matter_power_spectrum (:obj:`str` or :class:`~pyccl.emulators.emu_base.EmulatorPk`):
             The matter power spectrum to use. Defaults to 'halofit'.
+        baryonic_effects (:class:`~pyccl.baryons.baryons_base.Baryons`, `'bcm'` or `None`):
+            The baryonic effects model to use. Options are `None` (no baryonic effects),
+            `'bcm'` (the Schneider 2015 model with default parameters -- see :class:`~pyccl.baryons.schneider15.BaryonsSchneider15`),
+            or a baryonic effects object. Defaults to `None`.
         extra_parameters (:obj:`dict`): Dictionary holding extra
             parameters. Currently supports extra parameters for CAMB.
             Details described below. Defaults to None.
@@ -195,10 +199,15 @@ class Cosmology(CCLObject):
 
         extra_parameters = {"camb": {"halofit_version": "mead2020_feedback",
                                      "HMCode_logT_AGN": 7.8}}
+
+    .. note :: If using camb to compute the non-linear power spectrum with HMCode
+               to include baryonic effects, you should not include any extra
+               baryonic effects (i.e. set `baryonic_effects=None`).
     """ # noqa
     from ._core.repr_ import build_string_Cosmology as __repr__
     __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
-                    "_accuracy_params",)
+                    "_accuracy_params", "lin_pk_emu", 'nl_pk_emu',
+                    "baryons",)
 
     def __init__(
             self, *, Omega_c=None, Omega_b=None, h=None, n_s=None,
@@ -208,6 +217,7 @@ class Cosmology(CCLObject):
             mu_0=0, sigma_0=0, c1_mg=1, c2_mg=1, lambda_mg=0,
             transfer_function='boltzmann_camb',
             matter_power_spectrum='halofit',
+            baryonic_effects=None,
             extra_parameters=None,
             T_ncdm=DefaultParams.T_ncdm):
 
@@ -227,6 +237,12 @@ class Cosmology(CCLObject):
         if isinstance(matter_power_spectrum, emulators.EmulatorPk):
             self.nl_pk_emu = matter_power_spectrum
             matter_power_spectrum = 'emulator'
+
+        self.baryons = baryonic_effects
+        if not isinstance(self.baryons, baryons.Baryons):
+            if self.baryons not in (None, 'bcm'):
+                raise ValueError("`baryonic_effects` must be `None`, `'bcm'`, "
+                                 "or a `Baryons` instance.")
 
         # going to save these for later
         self._params_init_kwargs = dict(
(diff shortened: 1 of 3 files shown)


pyccl/cosmology.py
```diff
@@ -195,10 +199,15 @@ class Cosmology(CCLObject):
...
                                      "HMCode_logT_AGN": 7.8}}
+
+    .. note :: If using camb to compute the non-linear power spectrum with HMCode
+               to include baryonic effects, you should not include any extra
+               baryonic effects (i.e. set `baryonic_effects=None`).
     """ # noqa
     from ._core.repr_ import build_string_Cosmology as __repr__
     __eq_attrs__ = ("_params_init_kwargs", "_config_init_kwargs",
-                    "_accuracy_params",)
+                    "_accuracy_params", "lin_pk_emu", 'nl_pk_emu',
```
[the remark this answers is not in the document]
damonge: So this is not related to this PR, just trying to fix it quickly here.

readthedocs/api/pyccl.emulators.cosmicemu_MTIV_pk.rst
damonge, line 1: Apparently there was a stray rtd file that got erased when I recreated the docs.

pyccl/tests/test_baryons.py
```diff
@@ -41,3 +41,33 @@ def test_baryons_from_name():
...
+def test_baryons_in_cosmology():
+    # Test that applying baryons during cosmology creation works.
+    # 1. Outside of cosmo
+    cosmo_nb = ccl.CosmologyVanillaLCDM(
+        transfer_function='bbks', baryonic_effects=None)
+    pk_nb = cosmo_nb.get_nonlin_power()
+    pk_wb = bar.include_baryonic_effects(cosmo_nb, pk_nb)
+    # 2. In cosmo - default BCM model
+    cosmo_wb1 = ccl.CosmologyVanillaLCDM(
+        transfer_function='bbks', baryonic_effects='bcm')
```
elisachisari, line 55: So this takes the BCM fiducial values then? Can you pass the parameters here too? If so, could you add that to the test?

pyccl/cosmology.py
```diff
@@ -556,6 +572,12 @@ def _compute_nonlin_power(self):
         elif mps == 'emulator':
             pk = self.nl_pk_emu.get_pk2d(self)
 
+        # Include baryonic effects
+        if self.baryons is not None:
+            if self.baryons == 'bcm':  # Initialise convenience baryons model
+                self.baryons = baryons.BaryonsSchneider15(
+                    log10Mc=np.log10(1.2E14), eta_b=0.5, k_s=55.0)
```
elisachisari, line 579: Are these the fiducial values? Do we need to pass them explicitly?

pyccl/cosmology.py
```diff
@@ -228,6 +238,12 @@ def __init__(
             self.nl_pk_emu = matter_power_spectrum
             matter_power_spectrum = 'emulator'
 
+        self.baryons = baryonic_effects
+        if not isinstance(self.baryons, baryons.Baryons):
+            if self.baryons not in (None, 'bcm'):
```
elisachisari, line 243: Do we want to merge the Vd19 model first and add it to this PR?

pyccl/cosmology.py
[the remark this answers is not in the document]
elisachisari: Ok, I saw this. Are the tests passing now?

pyccl/tests/test_baryons.py
elisachisari, line 1: I would add another check that the suppression generated by calling the baryons in this way is the same as the one generated by the other way of calling it. I checked and it is, so this is working properly.

elisachisari (changes requested):
Working well for me. Some small suggestions.
I don't see files for the docs except the cosmicemu one - did you recompile the docs?

pyccl/tests/test_baryons.py
damonge: No, you can't pass parameters. The consensus seemed to be that we want to avoid polluting the Cosmology initialization function with so many non-cosmology parameters, so I'd rather do that...

pyccl/cosmology.py
damonge: yes, are there better values we should use as default? The idea is to provide 'bcm' as a quick option for someone who wants to check out what happens if you have baryons. Anyone who wants to properly deal with baryonic effects should create their own `Baryons` object and their associated parameters.

pyccl/cosmology.py
damonge: It wouldn't change anything in this PR, I think, right? I don't think we should support more than a single default string option ('bcm'). Anything else and the Cosmology initialiser will again become unwieldy.

pyccl/cosmology.py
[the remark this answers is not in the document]
damonge: yep

pyccl/cosmology.py
damonge: @tilmantroester ok, so you're proposing we even remove the "bcm" option? That'd be fine with me.

pyccl/tests/test_baryons.py
damonge: I think that's precisely what the test I made below does, right? Maybe I've misunderstood.

damonge:
@elisachisari @tilmantroester back to you. I think the main question mark is whether we keep `"bcm"` as a convenient option or just remove it completely (since it's not much harder than passing `ccl.baryons.BaryonsSchneider15()`.

marcpaterno:
While I’m somewhat late to the game, I think having a default-initialized `ccl.baryons.BaryonsSchneider15` object is clearer than having a string with a special value in the interface is clearer, and easier to search for with tools like an IDE or even `grep`.

damonge:
Thanks @marcpaterno , but to be clear, the default here will be `None` (meaning no baryonic effects correction)

pyccl/cosmology.py
elisachisari: I'd be fine with that as well

pyccl/cosmology.py
elisachisari: The change in the previous comment takes care of this

pyccl/tests/test_baryons.py
elisachisari: Yes, sorry, I realise that now

pyccl/tests/test_baryons.py
elisachisari: Ok

tilmantroester:
This might be a more general idea than what's covered in this PR but would it make sense to allow the `baryons` object to be more general than the `baryons.Baryons` API? What I have in mind is having something like a `HMCode2020Params` class that could be passed to the `baryons` argument to `Cosmology` and which internally sets the `extra_parameters` that get passed to CAMB. This `HMCode2020Params` wouldn't have the `include_baryonic_effects` API but would allow a consistent use of the `Cosmology` API for HMCode2020 baryons.

elisachisari (changes requested):
If we remove the 'bcm' option I think we are then done.

damonge:
OK, thanks all. The proposal is then:
a) the default value of `baryonic_effects` is `None`, which means "do not apply any baryonic effects".
b) the only other option is a `Baryons` object. No `'bcm'` option (users need to instantiate the corresponding Schneider15 class).

All happy?

marcpaterno:
I like this solution.

damonge:
Alright, done!

pyccl/cosmology.py
damonge: Done

elisachisari (approved):

PR #1115, opened 2023-07-27
Bocquet 2020 mass function emulator

### LINKED ISSUE #235: Add emulators for cluster mass function
This was a suggestion from the cluster experts, pointing out that the Tinker mass function currently implemented is not accurate enough for LSST. What do you think, @tmcclintock?

damonge:
Closes #235


pyccl/tests/test_hmfunc.py
```diff
@@ -153,3 +156,35 @@ def test_mass_function_mass_def_strict_always_raises():
...
+
+    nM1 = mf1(COSMO, Ms, 1.0)
+    nM2 = mf2(COSMO, Ms, 1.0)
+    assert np.allclose(nM1, nM2, atol=0, rtol=0.1)
+
+
+def test_nM_bocquet20_raises():
+    Ms = np.geomspace(1E12, 1E17, 128)
+
+    # Need A_s
```
carlosggarcia, line 176: If I understand it correctly, you need sigma8 for the emulator. So the comment would be "need sigma8"?

carlosggarcia (changes requested):
Few comments

benchmarks/test_bocquet20_hmf.py
```diff
@@ -0,0 +1,24 @@
+import numpy as np
+import pyccl as ccl
+
+
+def test_bocquet20_mf():
+    d = np.load("./benchmarks/data/mf_bocquet20mf.py.npz")
```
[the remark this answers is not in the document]
damonge: Well, just in case the package ever stops being maintained and we've reimplemented it. This also follows the same pattern we use for virtually all other benchmarks.

pyccl/tests/test_hmfunc.py
damonge: Yep, you're right. Done!

carlosggarcia (approved):
LGTM

PR #1131, opened 2023-10-26
Test CI

hsinfan1996:
Why CI is failing on macOS during unit tests but not on Ubuntu?


hsinfan1996:
Turns it is caused by pytest 7.4.3. Weird.

hsinfan1996:
The max relative difference of the elements in the two arrays is 1.15e-5. Not sure if pytest 7.4.2 did not catch this or the error is different for pytest 7.4.2 (seems unlikely to me).

carlosggarcia:
Hi @hsinfan1996. As a workaround, I increased the tolerance in #869 (see this commit: https://github.com/LSSTDESC/CCL/pull/869/commits/2b323eafe83dd989aafd067aac3bd788980c4889). With it both ubuntu and mac tests passed.

hsinfan1996:
> Hi @hsinfan1996. As a workaround, I increased the tolerance in #869 (see this commit: [2b323ea](https://github.com/LSSTDESC/CCL/commit/2b323eafe83dd989aafd067aac3bd788980c4889)). With it both ubuntu and mac tests passed.

Thanks. Yes, I think for now increasing the tolerance is the only way.

hsinfan1996:
I still don't know caused the tests on macOS to fail. Amongst the packages common to the two platforms, the differences I can spot are the compilers and libgfortran-related packages. macOS uses `libgfortran-devel_osx-64=12.2.0` while Ubuntu uses `libgfortran-ng=13.2.0`.

damonge:
Oh well... I'd be happy to just raise the tolerance and add a comment documented that this is because of mac on GHA (in case we can bring it back down in the future)

PR #1139, opened 2023-11-21
Bacco lbias

matteozennaro:
Implementation of the nonlinear galaxy bias prescription as in the hybrid Lagrangian bias expansion from the baccoemu emulator library. Followed closely the structure and conventions of the lpt module.


pyccl/nl_pt/bacco_lbias.py
```diff
@@ -0,0 +1,447 @@
...
+# All valid Pk pair labels and their aliases
+_PK_ALIAS = {
+    'm:m': 'm:m', 'm:b1': 'm:b1', 'm:b2': 'm:b2',
+    'm:b3nl': 'zero', 'm:bs': 'm:bs', 'm:bk2': 'm:bk2',
+    'b1:b1': 'b1:b1', 'b1:b2': 'b1:b2', 'b1:b3nl': 'zero',
+    'b1:bs': 'b1:bs', 'b1:bk2': 'b1:bk2', 'b2:b2': 'b2:b2',
+    'b2:b3nl': 'zero', 'b2:bs': 'b2:bs', 'b2:bk2': 'b2:bk2',
+    'b3nl:b3nl': 'zero', 'b3nl:bs': 'zero',
+    'b3nl:bk2': 'zero', 'bs:bs': 'bs:bs',
+    'bs:bk2': 'bs:bk2', 'bk2:bk2': 'bk2:bk2'}
```
damonge, line 18: Just a comment here: at some point we may want to consider making _PK_ALIAS common to the whole `nl_pt` module to avoid repeating it in different submodules. Can you just add a `TODO` comment here?

pyccl/nl_pt/bacco_lbias.py
```diff
@@ -0,0 +1,447 @@
...
+                          extrap_order_hik=2):
+        """Returns a :class:`~pyccl.pk2d.Pk2D` object containing
+        the power spectrum template for two of the PT operators. The
+        combination returned is determined by ``kind``, which must be
+        a string of the form ``'q1:q2'``, where ``q1`` and ``q2`` denote
+        the two operators whose power spectrum is sought. Valid
+        operator names are: ``'m'`` (matter overdensity), ``'b1'``
+        (first-order overdensity), ``'b2'`` (:math:`\\delta^2`
+        term in galaxy bias expansion), ``'bs'`` (:math:`s^2` term
+        in galaxy bias expansion), ``'b3nl'`` (:math:`\\psi_{nl}`
```
damonge, line 371: There is no b3nl in BACCO, right?

damonge (changes requested):
Awesome, thanks a lot @matteozennaro ! Only 2 small comments.

pyccl/nl_pt/bacco_lbias.py
matteozennaro: yep, will add!

pyccl/nl_pt/bacco_lbias.py
matteozennaro: True, I left it because I made sure that if you pass b3nl that term gets set to zero. But it's confusing, so now I left this behaviour unchanged (so if a b3nl gets passed it doesn't explode, just ignores it) but i changed the documentation so that b3nl doesn't appear as a legal parameter

matteozennaro: No, like in the lpt module the corresponding template is zero.

damonge: ok, great

damonge (approved):
LGTM!

damonge:
I approved, so feel free to merge, however, one extra comment in case you'd like to address it: to a non-expert, it won't be clear what this new class does that is different from the velocileptors-based calculator. You don't have to use HEFT ;-), but perhaps worth saying something about how this methodology is different from LPT?

PR #1144, opened 2023-12-09
BLD: update build system and use pyproject.toml

fangchenli:


setup.py
```diff
@@ -18,7 +15,7 @@ def _compile_ccl(debug=False):
            "-DPYTHON_VERSION=%d.%d.%d" % (v.major, v.minor, v.micro)]
     if debug:
         cmd += ["-DCMAKE_BUILD_TYPE=Debug"]
-    if call(cmd) != 0:
+    if call(cmd, shell=True) != 0:
```
hsinfan1996, line 18: Try without `shell=True` since it is added in python 3.12, but in GitHub Actions we use python 3.8.

damonge:
@fangchenli let me know if you'd like me or someone else to review this.

pyproject.toml
```diff
@@ -0,0 +1,49 @@
...
+[project.optional-dependencies]
+dev = [
+    "pytest",
+    "pytest-cov",
+]
+
+[tool.setuptools]
+packages = ["pyccl"]
+license-files = ["LICENSE", "LICENSE_COSMICEMU"]
+include-package-data = true
```
hsinfan1996, line 44: License files and package data will be included automatically by default and in setup.py we use `find_packages()`, so the four lines can be reduced to the following (to avoid a lot of warnings when `pip install . -v` is issued):

[tool.setuptools.packages.find]
include = ["pyccl*"]

damonge (approved):
LGTM

Issue #954, opened 2022-07-20
Implement lookback time


nikfilippas:
It would be nice to have this, and, along with #408 , this would complete the implementation of all time/distance measures in Hogg's reference manual [here](https://arxiv.org/pdf/astro-ph/9905116.pdf).

PR #1066, opened 2023-04-13
Complete background calculations

nikfilippas:
This PR adds the following background functions:
- `hubble_distance(cosmo, a)`
- `comoving_volume_element(cosmo, a)`
- `comoving_volume(cosmo, a, *, solid_angle=4*pi)`
- `lookback_time(cosmo, a)`
- `age_of_universe(cosmo, a)`

and with that closes #954 and closes #408 and now CCL will be able to calculate every distance quantity described in [Hogg's reference manual](https://arxiv.org/pdf/astro-ph/9905116.pdf) .

Issue #408, opened 2018-07-11
Add comoving volume element calculation to the background module


tmcclintock:
An essential quantity to calculate is the comoving volume element, dV/dOmega dz, or the comoving volume per unit redshift and solid angle. This can be accomplished by combining splines already present in CCL. For reference, see eq. 28 of [this paper](https://arxiv.org/abs/astro-ph/9905116) by David Hogg.

The equation to be implemented is
![codecogseqn](https://user-images.githubusercontent.com/10226392/42599940-9ee7c6d8-8515-11e8-97c8-fcf876bfcd76.png)

This issue is very easy, and could serve as a good hack day project at CMU.

c-d-leonard:
@tmcclintock I'm guessing you'd still like to see this?

nikosarcevic:
Hi @nikfilippas ,

if you want, feel free to use the code for the coming volume that was coded by me, Marco Bonici and Matthijs van der Wild.
I also have a similar code in my own pipeline (a DESC Project).

Here is the link to the coming volume function in our code for CosmOracle https://github.com/nikosarcevic/CosmOracle/blob/main/background.py 

Let me know if you need help with implementing it.

nikosarcevic:
> Hey Niko! I have already implemented it in a Python-only version of CCL which I'll submit for review in a few weeks (hopefully), after some of the other stuff gets pushed to the main branch. It is [here](https://github.com/nikfilippas/myccl/blob/main/myccl/background.py#L417), but for now it only lives in my own repo. It isn't a much requested feature and that's why I have refrained from opening a separate PR just for this, so that the PR list isn't cluttered.
> Having a quick look I think our implementations are equivalent, right?

I see! the only thing I would point out which is important is the omega_k part -- there is a good discussion around omega_k and David Hogg got good comments about it from Phillip Helbig back in the day. so we included that special case so have a look at our code, https://github.com/nikosarcevic/CosmOracle/blob/7dfdc63a127b7282dbbfa19db7b97789a611faca/background.py#L136

nikfilippas:
Brief follow-up after discussion in private: CCL defines `Cosmology` objects given `Ω_k`, and then calculates `Ω_Λ`, while `CosmOracle` calculates `Ω_k` from a given `Ω_Λ`. Which essentially means that CCL doesn't suffer from this truncation problem.

PR #1149, opened 2023-12-21
Background Calculations

hsinfan1996:
Merge distances_v3 branch into master (different from PR #1066 ). Closes issue #954 andd #408 .


pyccl/_core/parameters/parameters_base.py
```diff
@@ -134,6 +134,10 @@ class GSLParams(CCLParameters, instance=lib.cvar.user_gsl_params):
 class PhysicalConstants(CCLParameters, instance=lib.cvar.constants,
                         freeze=True):
     """Instances of this class hold the physical constants."""
+    # Mean solar day (s/day). [exact]
+    DAY = 86400.
+    # Sidereal year (days/yr). [IERS2014 in J2000.0]
+    YEAR = 365.256_363_004 * DAY
```
hsinfan1996, line 140: This part is ad-hoc. We can port all constants from C to Python level.

hsinfan1996:
@nikosarcevic I've done merging the master branch into the revived distances_v3. Now you can try the new functions and see if the calculations are correct and maybe modify the code if you want.

damonge:
@hsinfan1996 let me know when you'd like me to review this. Right now I see the tests are still failing, but happy to help.

hsinfan1996:
@damonge This is ready for review. Is there a way to fix the tests on macOS? I don't know why sometimes they passed but sometimes not.

hsinfan1996:
> @damonge This is ready for review. Is there a way to fix the tests on macOS? I don't know why sometimes they passed but sometimes not.

From my observation, faster test runs (typically < 25 min) on macOS will pass, slower ones (~35 min) will not. The weird thing is that the failed tests are not timed, but it is good to know that the problem is not likely caused by the code.

pyccl/background.py
```diff
@@ -197,6 +226,141 @@ def distance_modulus(cosmo, a):
...
+
+
+def comoving_volume(cosmo, a, *, solid_angle=4*np.pi):
+    r"""Comoving volume, in :math:`\rm Mpc^3`.
+
+    .. math::
+
+        V_{\rm C} = \int_{\Omega} \mathrm{{d}}\Omega \int_z \mathrm{d}z
+        D_{\rm H} \frac{(1+z)^2 D_{\mathrm{A}}^2}{E(z)}
+
```
damonge, line 287:
```suggestion
    See Eq. 29 in `Hogg 2000 <https://arxiv.org/abs/astro-ph/9905116>`_.
    
```

pyccl/cosmology.py
```diff
@@ -101,6 +103,12 @@ def _make_methods(cls=None, *, modules=_TOP_LEVEL_MODULES, name=None):
     return cls
 
 
+@dataclass
+class CosmologyData:
```
damonge, line 107: Can you add a comment here for future developers that the aim is for this class to be used to port other background calculators from C to python?

Also, we should make it a private class (`_CosmologyData`), and perhaps relabel it to `_CosmologyBackgroundData` to make it clearer what it's supposed to hold.

pyccl/tests/test_background.py
```diff
@@ -85,6 +85,29 @@ def test_background_a_interface(a, func):
...
+    # Check that integrating the comoving volume element gives comoving volume.
+    cosmo = ccl.CosmologyVanillaLCDM(Omega_k=Omega_k)
+    from scipy.integrate import quad
+    a_min = 0.4
+    V = cosmo.comoving_volume(a=a_min, solid_angle=1.0)
+    vdv = quad(cosmo.comoving_volume_element, a=a_min, b=1.0)[0]
+    rtol = 1e-7 if Omega_k == 0 else 1e-2
+    assert np.allclose(vdv, V, atol=0, rtol=rtol)
+
+
```
damonge, line 110: `comoving_volume_element` isn't being tested anywhere, right?

damonge (changes requested):
Thanks a lot @hsinfan1996 (and @nikfilippas for the initial implementation). Some comments below.

pyccl/tests/test_background.py
damonge: OK, yes, good point.

damonge (approved):
LGTM!

PR #1154, opened 2024-02-07
Test GHA

hsinfan1996:
Test GHA


hsinfan1996:
llvm-openmp is what I want to fix here. fftw was installed from source by by cmake, and now it is moved to conda.

tilmantroester:
Looks good to me but do you understand why removing llvm-openmp fixes the openmp errors?

hsinfan1996:
> Looks good to me but do you understand why removing llvm-openmp fixes the openmp errors?

I don't remember why it was installed separately, and only for macOS (maybe because on ubuntu the behavior was different). Now llvm-openmp is installed automatically along with other environment packages, so it is still there.

tilmantroester (approved):

Issue #1129, opened 2023-10-19
GSL integration errors in angular_cl


paulrogozenski:
TJPCov calls CCL function angular_cl to generate analytic covariance matrices. We're finding that we run into gsl integration errors in 3x2pt analyses and when increasing the l-binning scheme when calling angular_cl. In particular:

`gsl: qag.c:247: ERROR: roundoff error prevents tolerance from being achieved
Default GSL error handler invoked.
gsl: qag.c:247: ERROR: roundoff error prevents tolerance from being achieved
Aborted (core dumped)`

elisachisari:
Try increasing the CCL precision settings: https://ccl.readthedocs.io/en/latest/source/notation_and_other_cosmological_conventions.html 
That should help removing this error but it will take longer to compute

aguinot:
Hello,

Sorry to jump in.. (I can create a separate issue if more appropriate)
I have encounter a similar error while calling `angular_cl`. I join a code to reproduce the error.
I tried to change `pyccl.gsl_params["INTEGRATION_EPSREL"]` as suggested by @elisachisari but the error remains.
I also tried to enable/disable the spline for the integral but it has no effect.
The error raise for a particular combination of parameters. In the exemple if you set `b_ia=0` it works or if you change the value of `Ob` or `Om`.
(The value of the cosmological parameters are not "classic" because they come from a MCMC sampling)

I use CCL version `3.0.0` on a linux machine.

```python
import numpy as np
import pyccl as ccl


def get_model(
    theta,
    z,
    nz,
    z_clust,
    nz_clust,
    Omega_c,
    h,
    Omega_b,
    sig8,
    ns,
    b_ia,
    b_g,
):

    theta_deg = theta/60

    #ccl.gsl_params["INTEGRATION_EPSREL"] = 1e-10
    ccl.gsl_params.LENSING_KERNEL_SPLINE_INTEGRATION = False
    cosmo = ccl.Cosmology(
        Omega_c=Omega_c,
        Omega_b=Omega_b,
        h=h,
        n_s=ns,
        sigma8=sig8,

        transfer_function='boltzmann_camb',
        matter_power_spectrum='camb',
        extra_parameters={
            "camb": {
                "halofit_version": 'mead2016',  # 'mead2020',
                "kmax": 21,
                # We do this to avoid warnings in CAMB with (kmax < 5 or
                # 'kmax < 20 and z > 4') (set internally in ccl)
                # default: kmax=10
            }
        }
    )

[...]

PR #1155, opened 2024-02-09
Fix GSL integration abort

tilmantroester:
This is a quick fix for the aborts reported in #1129 by moving the call to disable the GSL error handling from `ccl_parameters_create` to `ccl_cosmology_create`. 

This uncovered a bunch of other issues though. From the python side, `ccl_parameters_create` isn't being used any more.
It is still used from within the C code though, mainly in halofit and the mu-Sigma MG stuff. So now there are two places in the code that deal with the non-trivial aspects of the `ccl_parameters` struct, namely the allocation of the massive neutrino arrays and MG arrays.

Speaking of these arrays, right now you can do some pretty nasty stuff by accessing arbitrary memory because `N_nu_mass` can be changed from python. I added a fix for one memory leak but there might be others.

```python
import pyccl as ccl
import numpy as np

print(f"{cosmo._params.N_nu_mass=}")
print(f"{cosmo._params.m_nu=}")

print("Set cosmo._params.m_nu")
cosmo._params.m_nu = np.array([0.1, 0.2, 0.3])

print(f"{cosmo._params.N_nu_mass=}")
print(f"{cosmo._params.m_nu=}")

print("Set cosmo._params.N_nu_mass")
cosmo._params.N_nu_mass = 10
print(f"{cosmo._params.N_nu_mass=}")
print(f"{cosmo._params.m_nu=}")
```
gives
```
cosmo._params.N_nu_mass=0
cosmo._params.m_nu=[]
Set cosmo._params.m_nu
cosmo._params.N_nu_mass=3
cosmo._params.m_nu=[0.1, 0.2, 0.3]
Set cosmo._params.N_nu_mass
cosmo._params.N_nu_mass=10
cosmo._params.m_nu=[0.1, 0.2, 0.3, 0.0, 5e-324, 1.6e-322, 2.1470999305e-314, 1.5e-323, 9e-323, 1.5543504411933145e-303]
```
 
I'm aware that `_params` is supposed to be private but it shouldn't be that easy to mess with C memory.

### DIFF
--- a/pyccl/_core/parameters/cosmology_params.py
+++ b/pyccl/_core/parameters/cosmology_params.py
@@ -16,5 +16,6 @@ class CosmologyParams(CCLParameters, factory=lib.parameters):
 
     def __setattr__(self, key, value):
         if key == "m_nu":
+            self._instance.N_nu_mass = len(value)
             return lib.parameters_m_nu_set_custom(self._instance, value)
         super().__setattr__(key, value)
--- a/pyccl/ccl_core.i
+++ b/pyccl/ccl_core.i
@@ -15,6 +15,9 @@
 
 %inline %{
 void parameters_m_nu_set_custom(ccl_parameters *params, double *mass, int num) {
+  if(params->m_nu != NULL) {
+    free(params->m_nu);
+  }
   params->m_nu = (double*) malloc(num*sizeof(double));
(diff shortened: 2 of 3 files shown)


damonge (approved):
LGTM

Issue #1159, opened 2024-02-20
`write_yaml` does not serialise completly


tilmantroester:
`write_yaml` currently ignores the `baryonic_effects` and `mg_parametrization` arguments to `Cosmology`, since it only uses what's in `self._params_init_kwargs` and `self._config_init_kwargs`. It also loses the information on `transfer_function` and `matter_power_spectrum` if they are emulators. That is, a roundtrip of `write_yaml` and `read_yaml` doesn't recover the same `Cosmology` object.

damonge:
@tilmantroester , should this have been solved by #1160 ?

tilmantroester:
Not exactly. Trying to use `write_yaml` with complex types for `Cosmology` arguments now raises an error instead of silently ignoring these arguments. But you still can't round-trip using `read_yaml` and `write_yaml` if an argument to `Cosmology` isn't something like a `float` or `str`.

damonge:
OK, I see. So I guess we'd want some kind of `_repr_` for the relevant CCL classes that contains enough information to recreate them from scratch.

Issue #1150, opened 2023-12-22
get_parameters_dict() method possibility 


nikosarcevic:
I was wondering, can we have an option to draw a param dict from the cosmology object?

cosmology = ccl.Cosmology(Omega_m=0.3, h=0.67...)

params = cosmology.get_parameters_dict()
params = {
            'Omega_c': cosmology['Omega_c'],
            'Omega_b': cosmology['Omega_b'],
            'h': cosmology['h'],
            'sigma8': cosmology['sigma8'],
            'n_s': cosmology['n_s'],
            # Add any other relevant parameters here
        }
        
    I would find this beyond useful as I work with dicts a lot
        
   @hsinfan1996

hsinfan1996:
For now you can retrieve the two private attributes `cosmo._params_init_kwargs` and `cosmo._config_init_kwargs`.

fjaviersanchez:
This kind of exists but it's ocult:
If `cosmo` is a `ccl.Cosmology` object you can do:
`cosmo.__dict__['_params_init_kwargs']`

fjaviersanchez:
Oh, @hsinfan1996 beat me to it :)

nikosarcevic:
@hsinfan1996  @fjaviersanchez brill! much obliged!!

tilmantroester:
`Cosmology.write_yaml` does that internally. We should probably factor that out into a `to_dict` method or similar.

nikosarcevic:
@tilmantroester ah I see you also prefer to have a method rather then calling cosmology._params_init_kwargs 

yeah actually makes sense (going back to my initial request)

tilmantroester:
You shouldn't use private APIs (eg anything with a leading underscore) outside of CCL, since they can change at any point. Better to add a proper API than hack something together.

nikosarcevic:
> You shouldn't use private APIs (eg anything with a leading underscore) outside of CCL, since they can change at any point. Better to add a proper API than hack something together.

Yeah totally makes sense, you are right. And this is not some elaborate code that requires 3 days (it's a few min in Python at least) and it is beyond useful to have. I have my own version of it and use it constantly.

tilmantroester:
We should look into how to make `to_dict` and `from_dict` methods that allow round trip generation of `Cosmology` objects. That's useful for serialisation (#1159), as well as for the use case where we want the same cosmology setup but with a few parameters changed, e.g. for blinding (@arthurmloureiro).
[...]

PR #1160, opened 2024-02-21
Add to_dict utility method

tilmantroester:
This adds a `Cosmology.to_dict` method and addresses #1159, as well as #1150. It allows round-tripping, even when the arguments are complex types:
```python
    cosmo = ccl.CosmologyVanillaLCDM(
        transfer_function=ccl.emulators.EmulatorPk(),
        matter_power_spectrum=ccl.emulators.CosmicemuMTIIPk(),
        baryonic_effects=ccl.baryons.BaryonsvanDaalen19(),
        mg_parametrization=ccl.modified_gravity.MuSigmaMG()
    )

    assert cosmo == ccl.Cosmology(**cosmo.to_dict())
```

Currently this dumps everything into the yaml file when calling `write_yaml`, which can be quite awkward for all the internal data of the emulators. The upside is that the write/read yaml works for round trips but I think throwing an error until we have a cleaner solution for yaml serialisation for complex types might be cleaner.

### DIFF
--- a/pyccl/cosmology.py
+++ b/pyccl/cosmology.py
@@ -235,13 +235,11 @@ class Cosmology(CCLObject):
         self.lin_pk_emu = None
         if isinstance(transfer_function, emulators.EmulatorPk):
             self.lin_pk_emu = transfer_function
-            transfer_function = 'emulator'
 
         # initialise nonlinear Pk emulators if needed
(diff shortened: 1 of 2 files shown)


damonge (approved):
LGTM

PR #1161, opened 2024-02-21
Nikosarcevic ccl v3 paper plots

nikosarcevic:


nikosarcevic:
this needs a review for the baryons halo model only. everything else has been checked

the review is needed as I am not sure if what comes out of the halo model notebook and then plotted together with the other baryon models makes sense

PR #1169, opened 2024-03-19
Throw error when initialising empty weak lensing or number counts tracer

### LINKED ISSUE #1133: Able to Define a ccl.WeakLensingTracer() object with has_shear = False and no ia_bias value.
Hello,

I was able to define a `ccl.WeakLensingTracer()` where the `has_shear = False`, but no `ia_bias` value was provided for example: [`ia_bias = (z,ia_bias_value)`]. The code did not raise an error. Finally, the error is raised when computing the angular power spectrum [doing an auto correlation (that tracer object with itself)]. The error is raised as follows: `CCLERROR: Error CCL_ERROR_INTEG: ccl_cls.c: ccl_angular_cls_limber(); integration error`.

I have attached the images of my code where I define the object and no error is raised and when i compute angular power spectra when the integration error is raised.

1)This Image shows defining the object with  `has_shear=False`, but no `ia_bias` value given.
 
Screenshot 2023-11-02 at 08 39 51

2) This Image shows the Integration error raised when using the object defined in 1) to compute the angular power spectrum.

Screenshot 2023-11-02 at 08 41 54

damonge:
Closes #1133

### DIFF
--- a/pyccl/tests/test_tracers.py
+++ b/pyccl/tests/test_tracers.py
@@ -426,6 +426,17 @@ def test_tracer_chi_min_max():
     assert tr.chi_max == tr._trc[1].chi_max
 
 
+def test_empty_wlnc_tracer():
+    z = np.linspace(0, 1.0, 32)
+    nz = np.exp(-0.5*((z-0.5)/0.05)**2)
+    with pytest.raises(ValueError):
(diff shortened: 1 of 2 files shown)


elisachisari (approved):
I checked locally that the warning is indeed raised and since the unit tests are passing through continuous integration as well, I think we can merge this one.

PR #1170, opened 2024-03-19
Outdated value of Neff in boltzmann.py

### LINKED ISSUE #1126: Discrepancy in default neutrino values?
It seems like the default Neff value (if you don't specify it) in the cosmology initiator has been updated to 3.044 but in boltzmann.py at line 89, 3.046 is hardcoded in calculating Delta Neff.

damonge:
Closes #1126


nikfilippas (approved):
lgtm

PR #1175, opened 2024-04-19
added clases dir with baryon halo model code

nikosarcevic:


damonge:
@nikosarcevic can I get some context for this? I understand it's an attempt at a halo model for baryons in CCL, but from a birds eye view, none of the code lives where I would have expected it to (in the halo model module!).

nikosarcevic:
> @nikosarcevic can I get some context for this? I understand it's an attempt at a halo model for baryons in CCL, but from a birds eye view, none of the code lives where I would have expected it to (in the halo model module!).

Hi, David

i can give an overview tomorrow in the ccl telecon.
Bottom line: this works but it is incredibly slow (and it needs some additional checking/testing). I will get to it after I am done with my thesis. So we can pause/remove any merge for now.
I am keeping the code on my branch.

damonge:
@nikosarcevic : ok, thanks for the explanation. I'll close the PR for now to avoid it taking space in our minds. Feel free to reopen whenever you think you'd like a review.

PR #1177, opened 2024-05-02
typo in docstring

svonhausegger:
Window function that is defined in docstring was incorrect (Factor of 3 within brackets instead of outside).  The window function actually used in the code is fine as far as I can tell.


damonge (approved):
Thanks @svonhausegger !

PR #1181, opened 2024-05-17
Function fixed: comoving_volume_element

pennalima:
Previous implementation was using the comoving_angular_distance. Fix to use the angular_diameter_diatance.

### DIFF
--- a/pyccl/background.py
+++ b/pyccl/background.py
@@ -271,7 +271,7 @@ def comoving_volume_element(cosmo, a):
     --------
     comoving_volume : integral of the comoving volume element
     """
-    Dm = comoving_angular_distance(cosmo, a)
+    Dm = angular_diameter_distance(cosmo, a)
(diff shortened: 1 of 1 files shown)


hsinfan1996:
What about this line?
https://github.com/LSSTDESC/CCL/blob/d762d995d9edb23522ff42818aa2343af3ec2d34/pyccl/background.py#L311

damonge:
@pennalima I think the function was correct as it was. Note that the function is meant to be `dV / (da dOmega)`, not `dV / (dz dOmega)`, which is maybe what you were thinking.

pennalima:
Hi  @hsinfan1996, 

Thanks a lot! 
Yes, I was reading it as dV/(dzdOmega). 
So, we can ignore it.

damonge:
Ok, thanks! I'll close this then.

Issue #1179, opened 2024-05-09
Costanzi et al. 2013 prescription on halo mass function with massive neutrinos


vittoriodx7:
In my recent cosmological analysis using pyCCL it came out during peer review that Costanzi et al. 2013 (https://ui.adsabs.harvard.edu/abs/2013JCAP...12..012C/abstract) prescription has to be used in the presence of massive neutrinos when computing the halo mass function.

I had edited a local version of pyCCL by:

in `halos/halo_model_base.py` at  line 196, `rho = (const.RHO_CRITICAL * cosmo['Omega_m'] * cosmo['h']**2)` should become:

```
        use_costanzi13=False
        try:
            use_costanzi13=cosmo["extra_parameters"]["use_costanzi13"]
        except (KeyError, TypeError):
            pass
        if not use_costanzi13:
            rho = (const.RHO_CRITICAL * cosmo['Omega_m'] * cosmo['h']**2)
        else:
            rho = (const.RHO_CRITICAL * (cosmo['Omega_c'] + cosmo['Omega_b']) * cosmo['h']**2)
```
and in `boltzmann.py` I suggest inserting at line 37:

```
use_costanzi13=False
    try:
        use_costanzi13=cosmo["extra_parameters"]["use_costanzi13"]
    except (KeyError, TypeError):
        pass
```

and then line 154, the CAMB call to get the power spectrum should become:

```
        if not use_costanzi13:
            k, z, pk = camb_res.get_linear_matter_power_spectrum(
            hubble_units=True, nonlinear=nonlin)
        else:
            k, z, pk = camb_res.get_linear_matter_power_spectrum(var1='delta_nonu', var2='delta_nonu', hubble_units=True, nonlinear=nonlin)
```

In this way if a user does not want to use it the code is unchanged, but if a user wants to use it they can pass it as an extra parameter in the call to pyccl.cosmology, e.g.:

` pyccl.Cosmology(..., extra_parameters={"use_costanzi13": True})
`
to allow this prescription to take place.

I therefore suggest this easy to do and backward compatible edit to the code to made available, if possible.

Thanks,
Vittorio

PR #1185, opened 2024-06-07
Adding prescription of Costanzi et al. 2013 for evaluating the halo mas function in nuCDM cosmologies 

DelonShen:
Neutrinos don’t cluster on halo scales and thus contribute negligibly to the collapse of massive dark matter halos. So, the halo mass function (HMF) should depend only on cold dark matter and baryons in nuCDM cosmologies. For cosmologies with massive neutrinos, it was found by [Costanzi et al. 2013](https://ui.adsabs.harvard.edu/abs/2013JCAP...12..012C/abstract) that using Plin_(CDM + Baryon) instead of Plin_matter when computing the HMF is important for the universality of the [Tinker et al. 2008](https://arxiv.org/abs/0803.2706) mass function in nuCDM cosmologies. This prescription is used in cluster cosmology analyses (e.g. the DES Y1 analysis in [To et al. 2021](https://ui.adsabs.harvard.edu/abs/2021PhRvL.126n1301T/abstract)).

For this PR, I just implemented the suggestion by  Vittorio Ghirardini in #1179 to add the option of using this prescription in CCL

### DIFF
--- a/pyccl/boltzmann.py
+++ b/pyccl/boltzmann.py
@@ -35,6 +35,15 @@ def get_camb_pk_lin(cosmo, *, nonlin=False):
     except (KeyError, TypeError):
         pass
 
+    # Get extra parameters to evaluate
+    # the halo mass function in cosmology with massive neutrinos
+    # using prescription by Costanzi et al. 2013(JCAP12(2013)012)
+    use_costanzi13 = False
+    try:
+        use_costanzi13 = cosmo["extra_parameters"]["use_costanzi13"]
+    except (KeyError, TypeError):
+        pass
+
     # z sampling from CCL parameters
     na = lib.get_pk_spline_na(cosmo.cosmo)
     status = 0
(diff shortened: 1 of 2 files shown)


damonge:
Thanks @DelonShen 

One issue with this is the following: a while ago we started implementing a more general way to deal with power spectra in CCL. The idea was that cosmology objects would be able to store a dictionary of Pks, which would include Pmm, but also Pmw, Pww (where `w` is the Weyl potential for lensing in non-LCDM models), and any other relevant power spectra. In practice, this dictionary currently only holds one power spectrum, labelled `"delta_matter:delta_matter"`.

The ideal way to implement what you're suggesting would be, instead, to have the CCL cosmology object read in both `delta_matter:delta_matter` and `delta_cb:delta_cb` from CAMB, and then only use the latter for halo model calculations. Implementing this, however, may involve more work than you might have expected to devote to this!

PR #1187, opened 2024-06-18
Scipy deprecations

damonge:
Just a small fix to how we use scipy's integrators to fix current deprecations.


elisachisari (approved):
Takes case of some deprecations due to updates on scipy and the broken HOD test. It seems the HOD test accuracy slightly depends on platform, so we might revisit this at some point. In any case the change is not large.
Checks are now passing.

damonge:
Thanks for documenting this better than me @elisachisari !

PR #1197, opened 2024-09-03
Bug in Despali16

### LINKED ISSUE #1195: Typo in the despali.16.py file
At line 44  the coefficients for the interpolation of the mass function parameters is
coeffs = [[A1, A0], [a2, a1, a0], [p2, p2, p0]]

and should be 
coeffs = [[A1, A0], [a2, a1, a0], [p2, p1, p0]]

damonge:
Closes #1195 
This bug was not caught by our tests because it would only show up when the chosen mass definition differs from Delta_vir.


damonge:
I've added a benchmark test that would have caught this bug. In doing so I regenerated all our benchmark files for halo mass function and halo biases using the latest version of colossus. The changes do not affect our benchmarks, which remain sufficiently accurate.

damonge:
Unrelated to this, there have been modifications to velocileptors which required me to update those benchmarks too.

Sorry for the many file modifications. Most of them are benchmark data, so any review should only focus on the code modifications (which are very small).

combet (approved):
I've looked at the changes in the code, cross-checking the coefficients with the Despali paper and the fix looks good. Also good to have an additional benchmark to catch anything going wrong if the mass definition is not in Delta_vir.

Issue #1191, opened 2024-07-12
Silent failure for P(k) rescaling in cosmologies scale-dependent MG


c-d-leonard:
CCL will allow you to define a cosmology with a scale-dependent mu / Sigma parameterization, and ask for a 'boltzmann_camb' or 'boltzmann_class' transfer function (rather than an 'isitgr') transfer function.  If you do this, CCL will re-scale the power spectrum for you using a growth function computed from solving the quasistatic growth equation with mu. But, it will do something where it seems to basically ignore the scale-dependence, silently (see the k=0 case in `ccl_mu_MG` in ccl_muSigma.c - this is how it is called in this scenario from within `growth_ode_system_muSig` in ccl_background.c.

It doesn't really make physical sense to do this rescaling option when you have a scale-dependent mu anyway - you need to solve the full Boltzmann equations. We should throw an error if somebody defines a cosmology with these parameters and transfer function combination then tries to get the matter power spectrum.

sankarshana16:
Hi, I've been looking at the code today. One thing that struck me is that even for the time-dependent case, you'd still want to use the isitgr transfer function, if you want to compute the weak-lensing Cls correctly right? It's true that you can compute the linear P(k) correctly by simply rescaling by the growth function. But I don't see any similar kind of re-scaling by \Sigma (if the mu_Sigma is called) in the angular_cl routine.  This means you'd have to do it manually, which can be ugly and probably the sort of thing we want to avoid anyway. So, I think it's just better to raise an error if any transfer function other than the isitgr one is called. Would you agree?

c-d-leonard:
Thanks for looking at this. I don't think I agree though. Two reasons:

- I am pretty sure that Sigma is indeed included in the projection integral when going down the rescaling route. For example, you can see this in tracers.py in the WeakLensingTracer class where _MG_add_tracer gets called and Sigma is properly projected over when sigma0 is nonzero. So I don't think there is a concern about having to do it manually.

- I think that given we have the ability to computer power spectra by rescaling by the growth factor, I don't see why we should abandon that. It's not fully correct at near-horizon scales, but often we don't want near-horizon scales, and for those cases this works pretty well. Also, being fully dependent on ISiTGR for any MG functionality has its own issues (as we saw, sometimes the connection breaks when external codes are updated etc). So I am on the side of maintaining our internal functionality since we have it. 

I'll tag @damonge as well for his thoughts?

sankarshana16:
[...]

PR #1202, opened 2024-09-26
Fix for issue #1191

sankarshana16:
Modified cosmology.py so that it complains when mu-sigma scale-dependent parameters are non-zero but transfer function isn't the isitgr one

### DIFF
--- a/pyccl/cosmology.py
+++ b/pyccl/cosmology.py
@@ -294,6 +294,16 @@ class Cosmology(CCLObject):
                 modified_gravity.MuSigmaMG):
             raise NotImplementedError("`mg_parametrization` only supports the "
                                       "mu-Sigma parametrization at this point")
+        """
+        if isinstance(
+                self.mg_parametrization,
+                modified_gravity.MuSigmaMG) and self.transfer_function_type is not 'boltzmann_isitgr':
+            raise ValueError("mu-Sigma parametrization is inconsistent with your transfer function choice (required to use isitgr)"
+                             )
+        """    
+
+        
+
 
         # going to save these for later
         self._params_init_kwargs = dict(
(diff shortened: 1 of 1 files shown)


pyccl/cosmology.py
```diff
@@ -471,6 +481,11 @@ def _build_parameters(
         c2_mg = self.mg_parametrization.c2_mg
         lambda_mg = self.mg_parametrization.lambda_mg
 
+        if(c1_mg != 0. or c2_mg !=0. or lambda_mg != 0.):
```
c-d-leonard, line 484: The default values of c1_mg and c2_mg are actually 1. GR is recovered when c1_mg = c2_mg=1 (regardless of lambda value) or when c1_mg=c2_mg=lambda_mg=0. So, need a slightly more complicated check here to cover both cases. This is why this branch is currently failing a bunch of unit tests I think.

c-d-leonard (changes requested):
This is generally a good fix, thank you @sankarshana16. There is one thing regarding the values of c1_mg and c2_mg which needs to be sorted out, this is what is causing the unit tests to fail.

Also just have a look at the flake8 results - you will need to format the new lines of code to pass those formatting requirements before merging. 

I wasn't able to test on my local machine yet because of some weird installation issues (which also happen on the master branch so I think it's me, not this branch). I'll try to fix that in the meantime while these fixes are implemented.

damonge:
@sankarshana16 are you able to address these?

pyccl/cosmology.py
sankarshana16: @c-d-leonard I've modified the code to only allow c1_mg = c2_mg = 1 or c1_mg = c2_mg = lambda_mg = 0., but complain otherwise with a better error message as you suggested. To be honest, I still had a few unit test problems, but from what I could see, they were not related to this (I think). Let me know if this is fine

PR #1209, opened 2024-11-08
Fix dynamic versioning

### LINKED ISSUE #1204: SWIG 4.3.0 is not compatible
SWIG 4.3.0 added a parameter to the function `SWIG_Python_AppendOutput`. A better solution is to use `SWIG_AppendOutput` instead. See https://www.swig.org/Release/CHANGES.current for more detail.

### LINKED ISSUE #1207: pyccl version in pypi and conda outdated
The version of pyccl in pypi is 3.0.2 instead of 3.1. Same in conda.

https://pypi.org/project/pyccl/

damonge:
Our latest releases haven't been pushed to pypi and conda because the automatic versioning was not working. This was broken in #1144 

Unfortunately fully testing this will require us to publish a new release and see if it pulls through. Hopefully we only need to do this once.

Closes #1207
Closes #1204

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -7,7 +7,7 @@ dependencies:
   - cmake
   - gsl
   - fftw
-  - swig==4.2.1
+  - swig
   - pyyaml
   - numpy
   # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - fftw
   - swig
   - pyyaml
-  - numpy<2
+  - numpy
   # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
   # We should remove the <1.14 flag as soon as this is fixed.
   - scipy<1.14
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -10,9 +10,10 @@ dependencies:
   - swig
   - pyyaml
   - numpy<2
-  # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
+  # The below is only because the current version of fast-pt uses deprecated scipy functions.
+  # DarkEmulator also uses an old scipy.
   # We should remove the <1.14 flag as soon as this is fixed.
-  - scipy
+  - scipy<1.14
   - camb
   - fast-pt
   - pytest
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - fftw
   - swig
   - pyyaml
-  - numpy
+  - numpy<2
   # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
   # We should remove the <1.14 flag as soon as this is fixed.
   - scipy<1.14
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -12,7 +12,7 @@ dependencies:
   - numpy<2
   # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
   # We should remove the <1.14 flag as soon as this is fixed.
-  - scipy<1.14
+  - scipy
   - camb
   - fast-pt
   - pytest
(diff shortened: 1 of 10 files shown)


pyccl/numpy.i
damonge, line 1: SWIG changed and this requires a new version of numpy.i (see [here](https://github.com/numpy/numpy/commit/850bec39d07ca4b001775c480846c7d7df8c722c))

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.5)
```
damonge, line 1: I am not sure why, but I wasn't able to fix dynamic versioning without this requirement.

setup.py
```diff
@@ -63,5 +63,6 @@ def run(self):
 
 setup(
     distclass=Distribution,
+    setup_requires=['setuptools_scm'],
```
damonge, line 66: This is the main thing that fixes dynamic versioning.

pyccl/halos/halo_model.py
```diff
@@ -186,7 +186,7 @@ def number_counts(self, cosmo, *, selection,
             mint[i] = self._integrator(
                 dVda[i] * self._mf[..., :] * _selm[..., :],
                 self._lmass
-            )
+            ).squeeze()
```
damonge, line 189: A few warnings were being triggered because of this (the result of the integral is an array of size 1 instead of a scalar, which numpy doesn't like).

.github/environment.yml
```diff
@@ -9,8 +9,9 @@ dependencies:
   - fftw
   - swig
   - pyyaml
-  - numpy
-  # The below is only because the currnt version of fast-pt uses deprecated scipy functions.
+  - numpy<2
```
carlosggarcia, line 12: Have you checked that this doesn't break other CCL stuff that might have been updated to numpy 2?

setup.py
carlosggarcia: why do you still require `setup.py` and not use just `pyproject.toml`?

carlosggarcia (changes requested):
Minor comments

.github/environment.yml
damonge: The opposite, actually. We cannot yet move to numpy 2 because many of our dependencies still don't use numpy 2. firecrown is in the same situation. This will be a tricky migration.

damonge:
@carlosggarcia not sure why I can't respond to your last comment. We do use pyproject.toml, but this is the only way I have found to make sure that the dynamic version numbers are assigned.

setup.py
damonge: @carlosggarcia not sure why I can't respond to your last comment. We do use pyproject.toml, but this is the only way I have found to make sure that the dynamic version numbers are assigned.

.github/environment.yml
carlosggarcia: You also want to add this `numpy<2` in the pyproject.toml dependencies.

setup.py
carlosggarcia: ok. weird. I think in TJPCov we have dynamic versioning and works fine with just `pyproject.toml`. I couldn't see any obvious difference that might cause this, though.

.github/environment.yml
damonge: Done

setup.py
damonge: Interesting. Well, if you can ever look into it, that'll be great. For now this is as far as I can get, and I think it is relatively urgent that we fix this! Our conda and pip releases are severely out of date.

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=64", "setuptools_scm>=8", "cmake", "swig"]
+requires = ["setuptools>=64", "setuptools_scm>=8", "cmake", "swig", "numpy<2"]
```
carlosggarcia, line 2: Would it go here or in the  `dependencies` section below?

damonge: sorry done

carlosggarcia (approved):
LGTM.

PR #1213, opened 2024-11-19
Warnings verbosity

damonge:
CCL often throws warnings about things that are relatively benign and should not plague the output of e.g. MCMC runs. This PR implements a granular way to control the verbosity of the warning messages thrown by CCL.

It uses some code by @nikfilippas .

While doing this I fixed a couple of bugs in the documentation.

Most of the changes are cosmetic. Anyone reviewing should start by looking at the changes in `error.py` (the rest is boilerplate)

### DIFF
--- a/pyccl/errors.py
+++ b/pyccl/errors.py
@@ -15,9 +15,9 @@ class warnings:
         category = kwargs.get('category')
         importance = _warning_importance[kwargs.pop('importance', 'low')]
 
-        if category == CCLWarning:
-            if importance < warnings._CCL_WARN_THRESHOLD:
-                return
+        if ((category is CCLWarning) and
+                (importance < warnings._CCL_WARN_THRESHOLD)):
+            return
 
         warnings_builtin.warn(*args, **kwargs)
 
--- a/pyccl/errors.py
+++ b/pyccl/errors.py
@@ -15,7 +15,7 @@ class warnings:
         category = kwargs.get('category')
         importance = _warning_importance[kwargs.pop('importance', 'low')]
 
-        if ((category is CCLWarning) and
+        if ((category in (CCLWarning, CCLDeprecationWarning)) and
                 (importance < warnings._CCL_WARN_THRESHOLD)):
             return
 
--- a/pyccl/errors.py
+++ b/pyccl/errors.py
@@ -9,7 +9,7 @@ _verbosity_thresholds = {'none': 100, 'low': 10, 'high': 1}
 
 
 class warnings:
-    _CCL_WARN_THRESHOLD = 1  # Equivalent to "low" verbosity
+    _CCL_WARN_THRESHOLD = 10  # Equivalent to "low" verbosity
 
     def warn(*args, **kwargs):
         category = kwargs.get('category')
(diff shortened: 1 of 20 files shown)


pyccl/errors.py
```diff
@@ -1,6 +1,50 @@
-__all__ = ("CCLError", "CCLWarning", "CCLDeprecationWarning",)
+__all__ = ("CCLError", "CCLWarning", "CCLDeprecationWarning",
```
damonge, line 1: Start reviewing here.

pyccl/errors.py
```diff
@@ -1,6 +1,50 @@
...
+class warnings:
+    _CCL_WARN_THRESHOLD = 10  # Equivalent to "low" verbosity
+
+    def warn(*args, **kwargs):
+        category = kwargs.get('category')
+        importance = _warning_importance[kwargs.pop('importance', 'low')]
+
+        if category == CCLWarning:
+            if importance < warnings._CCL_WARN_THRESHOLD:
+                return
```
nikfilippas, line 20: Just a minor thing which can cause breakage in the future if not treated right.

1. No need to double-nest.
2. Type equivalence check with `is`, otherwise it's unsafe.
```python
if category is CCLWarning and importance < warnings._CCL_WARN_THRESHOLD:
    return
```

pyccl/errors.py
```diff
@@ -40,8 +84,8 @@ def __hash__(self):
 
     @classmethod
     def enable(cls):
```
tilmantroester, line 86: Are the deprecation warnings covered by the verbosity filter as well?

damonge: I don't think so, but let me check

damonge: No they are not. Do you think they should?

tilmantroester: My gut feeling is yes - they're good to know while building a model but you probably don't want them in an MCMC log.

nikfilippas: If they mute them and CCL's deprecation cycle completes, their MCMC will simply stop working, and they won't be able to figure out why. I don't think it's a good idea muting those.

damonge: Well, if the deprecation cycle completes, they will get an error, which they should see, righ?

I'll mute those too @tilmantroester

nikfilippas: Indeed, but deprecation warnings have to do with CCL functionality whereas CCLWarnings mostly have to do with numerical stability, so they represent different things. At the very least I'd add a different flag to separate the verbosity level for deprecation-specific warnings. BTW now that v3 is out, is `CCLDeprecationWarning` used at all?

damonge: CCLDeprecationWarnings are not used right now. I'll leave them there for the future in any case.

tilmantroester: They seemed to be used here:
https://github.com/LSSTDESC/CCL/blob/master/pyccl/baryons/baccoemu_baryons.py#L221
and the decorators in
https://github.com/LSSTDESC/CCL/blob/master/pyccl/_core/deprecations.py
But the latter don't seem to be used.

damonge: As developers we can decide what the priority of the different warnings are. If we decide that specific deprecation warnings are sufficiently important, we can just tag them as "high", and they'll go through the filter.

@tilmantroester I've now added the CCLDeprecationWarnings to the list of warnings that can be suppressed if desired.

tilmantroester (approved):
LGTM!

tilmantroester:
We should document this somewhere though and make a point to check any model with `verbosity="high"`.

PR #1224, opened 2025-02-13
Allowing for cross-correlations in HOD power spectrum.

anicola:
This PR allows for cross-correlations with different HOD profiles. In the case in which the profiles are different, the code reverts to the usual Fourier-space two-point moment as implemented in Profile2pt. This holds under the assumption that the two tracers are disjoint.

### DIFF
--- a/pyccl/tests/test_profiles.py
+++ b/pyccl/tests/test_profiles.py
@@ -329,6 +329,7 @@ def test_hod_2pt():
     assert np.allclose(p2.fourier_2pt(COSMO, 1., 1e13, 1., pgood, prof2=None),
                        F0, rtol=0)
 
+    # raises for non-HOD profile
     with pytest.raises(TypeError):
         p2.fourier_2pt(COSMO, 1., 1E13, 1., pbad)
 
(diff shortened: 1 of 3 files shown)


pyccl/tests/test_profiles.py
```diff
@@ -337,11 +338,12 @@ def test_hod_2pt():
 
     p2.fourier_2pt(COSMO, 1., 1E13, 1., pgood, prof2=pgood_b)
 
-    with pytest.raises(ValueError):
-        pgood_b.update_parameters(log10M0_0=10.)
-        p2.fourier_2pt(COSMO, 1., 1E13, 1., pgood, prof2=pgood_b)
+    # doesn't raise for two different HOD profiles
+    pgood_b.update_parameters(log10M0_0=10.)
+    p2.fourier_2pt(COSMO, 1., 1E13, 1., pgood, prof2=pgood_b)
```
damonge, line 343: Can we check that this gives the same result as the standard profile2pt?

damonge (changes requested):
Thanks a lot @anicola . Just one small comment.

The checks are only not passing because coveralls is down for maintenance, but we can ignore that, so don't worry.

damonge (approved):
LGTM!

PR #1231, opened 2025-03-05
Fix the CMB lensing kernel to handle models with non-zero curvature.

vitenti:


src/ccl_tracers.c
```diff
@@ -559,7 +559,10 @@ void ccl_get_kappa_kernel(ccl_cosmology *cosmo, double chi_source,
   for (int ichi=0; ichi < nchi; ichi++) {
     double chi = chi_arr[ichi];
     double a = ccl_scale_factor_of_chi(cosmo, chi, status);
-    wchi[ichi] = lens_prefac*(ccl_sinn(cosmo,chi_source-chi,status))*chi/a;
+    if (chi != 0.0)
```
vitenti, line 562: @damonge This "if" is necessary to avoid 0/0 when chi == 0. If it is safe to assume that the chi_arr only contain 0.0 in its first element this can be simplified.

damonge (approved):
Looks good to me. Thanks a lot @vitenti

PR #1236, opened 2025-03-26
Supporting numpy2

damonge:


pyccl/emulators/cosmicemu_pk.py
```diff
@@ -124,7 +124,8 @@ def _initialize(self, data_path):
             # Add to diagonal
             for i, lw in enumerate(self.lamws[j]):
                 sigma_sim[i, :, :] += np.diag(np.full(m, 1./lw))
-            kb = np.linalg.solve(sigma_sim, self.w[j])
+            kb = np.linalg.solve(sigma_sim,
```
vitenti, line 127: This was an annoying change of behavior of numpy, it took me awhile to check if this is the best solution, and it looks like it is.

vitenti (approved):
Looks good to me.

pyccl/emulators/cosmicemu_pk.py
damonge: I know....
Thanks @vitenti !

Issue #1218, opened 2024-12-17
Missing factor in documentation for `halomod_Tk3D_SSC`


davidesciotti:
Hi, I think there may be a missing 1/3 factor in front of the term d log k^3 PL(k)/d log k in the documentation of the response functions [pyccl.halos.pk_4pt.halomod_Tk3D_SSC](https://ccl.readthedocs.io/en/latest/api/pyccl.halos.pk_4pt.html#pyccl.halos.pk_4pt.halomod_Tk3D_SSC) and [pyccl.halos.pk_4pt.halomod_Tk3D_SSC_linear_bias](https://ccl.readthedocs.io/en/latest/api/pyccl.halos.pk_4pt.html#pyccl.halos.pk_4pt.halomod_Tk3D_SSC_linear_bias) - see e.g. [2410.06962](https://arxiv.org/pdf/2410.06962). Note that the code, on the other hand, seems fine.

PR #1241, opened 2025-04-24
Update pk_4pt.py documentation

RyoTerasawa:
Fix issue #1218


carlosggarcia:
@RyoTerasawa thanks for addressing this! The lint test is not passing because the line 336 is too long. You can pass the `\\right)` to the next line

RyoTerasawa:
> @RyoTerasawa thanks for addressing this! The lint test is not passing because the line 336 is too long. You can pass the `\\right)` to the next line

Ah, I forgot to check the lint test. Now the lint test passed.

PR #1249, opened 2025-06-19
Adding large scale extrapolation of HEFT power spectra with biased li…

anicola:
This PR adds extrapolation of the BACCO HEFT power spectra at large scales with the appropriately biased linear power spectrum from baccoemu. This is to avoid extrapolation issues that happen when extrapolating outside the k range covered by the BACCO HEFT emulator.

### DIFF
--- a/pyccl/nl_pt/bacco_lbias.py
+++ b/pyccl/nl_pt/bacco_lbias.py
@@ -90,13 +90,15 @@ class BaccoLbiasCalculator(CCLAutoRepr):
 
     def __init__(self, *, cosmo=None,
                  log10k_min=-4, log10k_max=-0.47, nk_per_decade=20,
-                 a_arr=None, k_cutoff=None, n_exp_cutoff=4):
+                 a_arr=None, k_cutoff=None, n_exp_cutoff=4, extrap_lin=True):
         # Load emulator
         import warnings
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', category=UserWarning)
             import baccoemu
             self.emu = baccoemu.Lbias_expansion()
+            # Load linear matter power spectrum emulator
+            self.emu_pklin = baccoemu.Matter_powerspectrum()
 
         # k sampling
         nk_total = int((log10k_max - log10k_min) * nk_per_decade)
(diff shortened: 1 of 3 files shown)


pyccl/nl_pt/bacco_lbias.py
```diff
@@ -90,13 +90,15 @@ class BaccoLbiasCalculator(CCLAutoRepr):
 
     def __init__(self, *, cosmo=None,
                  log10k_min=-4, log10k_max=-0.47, nk_per_decade=20,
-                 a_arr=None, k_cutoff=None, n_exp_cutoff=4):
+                 a_arr=None, k_cutoff=None, n_exp_cutoff=4, extrap_lin=True):
```
damonge, line 93: The new argument needs documenting

pyccl/nl_pt/bacco_lbias.py
```diff
@@ -90,13 +90,15 @@ class BaccoLbiasCalculator(CCLAutoRepr):
...
         # Load emulator
         import warnings
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', category=UserWarning)
             import baccoemu
             self.emu = baccoemu.Lbias_expansion()
+            # Load linear matter power spectrum emulator
+            self.emu_pklin = baccoemu.Matter_powerspectrum()
```
damonge, line 101: Should we encase this in `if extrap_lin:`?

pyccl/nl_pt/bacco_lbias.py
```diff
@@ -181,6 +191,12 @@ def update_ingredients(self, cosmo):
         # save templates in a table
         self.lpt_table = lpt_table
 
+        # Get linear power spectrum from baccoemu
```
damonge, line 194: as before, should this be encased in `if extrap_lin`? If I follow the code logic, if `extrap_lin` was `False` you'd still calculate this but then multiply it by zero, right? I guess this is fine if all emulator evaluations are super fast. Just want to make sure this is what you intended! (in which case, fine not to add the if checks)

damonge (changes requested):
Looks good! Just minor comments below, and a less minor one here: could we add a unit test checking that the extrapolation is doing the right thing? I.e. call bacco without extrapolation and then check that at k=1E-4 or whatever it matches the linear prediction.

PR #1252, opened 2025-07-07
Generic module to include baryonic effects

sambit-giri:
This pull request contains a custom class (`BaryonCustom`) to add baryonic effects from any framework. This class takes the suppression or boost data and creates an interpolator.


damonge (changes requested):
With many apologies for the delay, here's a review.

In addition to the comments below, flake8 is currently failing.

damonge:
@sambit-giri looks great! Could we add one unit test to check that this is doing the right thing?
You can see a few examples here: https://github.com/LSSTDESC/CCL/blob/master/pyccl/tests/test_baryons_vd19.py

In this case this could be extremely simple (just checking that you recover the arrays you put in when you query the boost factor, for example).

sambit-giri:
Thanks @damonge! I've added unit tests in `pyccl/tests/test_baryons_custom.py` (commit 39447aa). They check that the input boost array is recovered when querying `boost_factor` on the input grid, plus output shapes for scalar/array `k` and `a`, the default boost of 1 outside the input range, `include_baryonic_effects` against the boost factor at the spline nodes, `update_parameters`, and `from_name`. All pass locally and flake8 is clean.

PR #1257, opened 2025-10-31
Change GR flag from 1 to 0 in boltzmann.py

bastiencarreres:
Change the GR flag from 1 to 0 to activate modified gravity

### DIFF
--- a/pyccl/boltzmann.py
+++ b/pyccl/boltzmann.py
@@ -271,7 +271,7 @@ def get_isitgr_pk_lin(cosmo):
     cp.ombh2 = cosmo['Omega_b'] * h2
     cp.omch2 = cosmo['Omega_c'] * h2
     cp.omk = cosmo['Omega_k']
-    cp.GR = 1  # means GR modified!
+    cp.GR = 0  # means GR modified!
     cp.ISiTGR_muSigma = True
     cp.mu0 = cosmo.mg_parametrization.mu_0
     cp.Sigma0 = cosmo.mg_parametrization.sigma_0


nikosarcevic (commented):
Hi, Bastien!

Thank you for this.

Can you please submit some plots that show how this fixes the problem?

bastiencarreres:
This plot show the power spectrum for various value of mu0 when setting GR=0 or 1
image

Lhior (approved):
Only potential improvement is to add a check if there's any MG parameters actually being used in order to enable MG, e.g. 
`    is_mg = (abs(cosmo.mg_parametrization.mu_0) > 0 or
             abs(cosmo.mg_parametrization.sigma_0) > 0 or
             abs(cosmo.mg_parametrization.lambda_mg) > 0)
    cp.GR = 0`
 `if is_mg else 1  # GR==0 to allow MG in ISiTGR`
   But you can very easily convince me that the fact that someone is actively calling isitgr is enough to assume they're using MG, so the current modifications look good to me as is!

nikosarcevic:
> Only potential improvement is to add a check if there's any MG parameters actually being used in order to enable MG, e.g. ` is_mg = (abs(cosmo.mg_parametrization.mu_0) > 0 or abs(cosmo.mg_parametrization.sigma_0) > 0 or abs(cosmo.mg_parametrization.lambda_mg) > 0) cp.GR = 0` `if is_mg else 1 # GR==0 to allow MG in ISiTGR` But you can very easily convince me that the fact that someone is actively calling isitgr is enough to assume they're using MG, so the current modifications look good to me as is!

I think this check is still important to do. Before anyone approving this merge can we ask MG people to check of if @bastiencarreres can do it quickly?

bastiencarreres:
@Lhior @nikosarcevic 
I think that default values of MG parameters are GR values thus even if GR=0 having the MG parameters set to GR values will return the GR power spectrum!

nikosarcevic:
> There are test that are failing. They are related to the correlations functions. Can you check what's going on?

Currently the CI is failing on main. So every time you branch out to fix something you inherit the CI fails (I tihnk for me it is in bacco tests). Can you confirm that?
When I was working on my PRs I stated that. I checked my tests only on the files I have changed but not the stuff that is currently killing the CI from main.

also there is this circular import issue aside from the bacco test #1259

bastiencarreres:
@carlosggarcia It seems that some of the failing tests are related to MG models, if the values the tests are using as baseline were computed using the GR = 1 flag (?), I guess that it cannot match.

nikosarcevic:
> @carlosggarcia It seems that some of the failing tests are related to MG models, if the values the tests are using as baseline were computed using the GR = 1 flag (?), I guess that it cannot match.

can we tag @Lhior to check this and help out?

Lhior:
I believe what @carlosggarcia pointed out is exactly right, specifically the tests in `test_power_mg2.py` are failing because the reference values where computed with the wrong flag. All other MG tests (e.g. `OMP_NUM_THREADS=2 python -m pytest -vv benchmarks/test_*MG*.py benchmarks/test_*mg*.py` pass on my laptop even when manually lowering the tolerance to `1e-6`. 

After further investigation though, I have some questions... why is [the tolerance in the classy implementation set to 1e-2](https://github.com/LSSTDESC/CCL/blob/fbe98b21d1090173910c87ec6aa606f3872ef7b5/benchmarks/test_power_mg.py#L7) while [the tolerance in the isitgr implementation set to 1e-4](https://github.com/LSSTDESC/CCL/blob/fbe98b21d1090173910c87ec6aa606f3872ef7b5/benchmarks/test_power_mg2.py#L7)? If I set the classy tolerance to 1e-4, all `test_power_mg.py` tests fail. On the other side, if I set the tolerance in `test_power_mg2.py` to 1e-2 all tests pass. 
I found this while trying to do a temporal workaround of benchmarking isitgr against classy evaluates values on the fly instead of using the current reference ones, even in that case the tolerance needs to be set to 1e-2 for the tests to pass and fails when set to 1e-4.

Issue #1259, opened 2025-11-12
ImportError: cannot import name ccllib from partially initialized module pyccl in macOS CI


nikosarcevic:
CI on macOS is currently failing during test collection with:

```
pyccl/__init__.py:22: in <module>
    from . import ccllib as lib
ImportError: cannot import name 'ccllib' from partially initialized module 'pyccl'
```

This happens before any tests run:
`OMP_NUM_THREADS=2 pytest -vv pyccl --cov=pyccl`

The failure also occurs on a clean checkout of main, so this appears to be inherited from a recent merge (likely after PR #1258 / issue #2413). [My branch](https://github.com/LSSTDESC/CCL/pull/1262/commits) only adds FKEM code and a one-line hook in __init__.py, but the error persists even if that line is removed.

Looks like the macOS build isn’t producing or loading the ccllib Python shim correctly, leading to the “partially initialized module” state.
Could someone familiar with the build / SWIG setup advise?

PR #1266, opened 2025-11-28
Refactor & Expand Baryon Model Tests + Add Benchmarks

### LINKED ISSUE #1265: Refactor & Expand Baryon Model Tests + Add Benchmarks
The current baryonic-effects tests in pyccl/tests/ cover all three explicit baryon models (Schneider15, Baccoemu, vanDaalen19) and partially cover baryon-related nonlinear power-spectrum options (CosmicEmu MTIV, HMCode Mead2020 feedback).
However, the coverage is uneven, some tests are bundled together, and there are no structured performance/consistency benchmarks across models.
@elisachisari and I have discussed this and we propose a full cleanup + expansion of the baryon test suite.

### Audit existing baryon tests

- Review:
  - test_baryons.py
  - test_barryons_vd19.py
  - test_baccoemu.py
  - test_cosmicemu.py
  - test_nonlin_camb_power.py (HMCode + AGN feedback)
  -  test_cosmology.py (extra_parameters / halofit plumbing)
- Identify:
  - duplicated logic
  - missing edge cases
  - inconsistent patterns across models

1### Restructure the test layout

Introduce “one model per test file” pattern:

- test_baryons_base.py (from_name, cosmology plumbing, basically tests calls)
- test_baryons_schneider15.py
- test_baryons_baccoemu.py
- test_baryons_vandaalen19.py
- test_baryons_cosmicemu_mtiv.py
- test_baryons_hmcode_mead2020.py

This improves readabiltiy and makes it easier to expand bayron support later when we add more models

###  Add a small benchmark suite

Create a benchmarks/baryons/ folder with scripts that:
- generate P(k) ratios (baryons on/off) at multiple redshifts
- evaluate models on a standard k–a grid
- optionally compare performance vs reference literature curves
- serve as manual validation for FKEM + baryons integrations
- This will help detect regressions + inconsisetncies

nikosarcevic:
Closes #1265

This PR cleans up and expadns the baryon test suite:

- I have spli tbaryon unit tests into model-specific files (Schneider15, Arico21, vanDaalen19, Mead20)
- added benchmrak tests for Aricò21 (lin and nlin cases, sigma8 and A_s cases)
- Added new test_cells_baryons.py covering WL, clustering, and GGL, including consistency checks between p_of_k_a and CosmologyCalculator
- Cleaned E501 lint in new files

ETA: added Van Daalen benchmarks (benchmark files sent by Marcel to Niko directly)

All baryon tests and benchmarks now pass. Linting healthy.

CI is failing now only for MacOS build and this s inherited from main. See this issue I explained https://github.com/LSSTDESC/CCL/issues/1259.


benchmarks/data/baryons_vd19_fk_500c.txt
elisachisari: See my other comment, I wonder why we cannot produce this on the fly during the test

benchmarks/test_baryons_arico21_benchmark.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def test_baccoemu_baryons_a_s():
+    """Tests that BACCOemu baryons match reference fk(k) for As cosmology."""
+    baryons = ccl.BaryonsBaccoemu()
+    cosmo = ccl.Cosmology(
+        Omega_c=0.27,
+        Omega_b=0.05,
+        h=0.67,
+        A_s=2.2194e-9,
```
elisachisari: What happens to the sigma_8 which corresponds to A_s after the boost factor is applied?

benchmarks/test_baryons_arico21_nlin_benchmark.py
elisachisari: Just renaming/moving code around?

benchmarks/test_baryons_vandaalen19_benchmark.py
```diff
@@ -0,0 +1,53 @@
+"""Benchmark tests for the van Daalen+2019 baryonic model."""
+
+from pathlib import Path
+
+import numpy as np
+
+import pyccl as ccl
+
+
+VD19_TOLERANCE = 1e-8
```
elisachisari: Aren't we testing CCL against itself here? What is the rationale for this test and tolerance?

pyccl/tests/test_baryons_mead20.py
```diff
@@ -0,0 +1,136 @@
...
+import pytest
+
+import numpy as np
+
+import pyccl as ccl
+
+
+camb = pytest.importorskip("camb")
+
+TOL_MEAD20 = 3e-5
```
elisachisari: Why is this so high? Shouldn't we be getting much better agreement given that we should be calling camb in exactly the same way?

benchmarks/test_baryons_arico21_benchmark.py
```diff
@@ -0,0 +1,65 @@
...
+        Neff=3.046,
+        mass_split="normal",
+        m_nu=0.1,
+        Omega_g=0.0,
+        Omega_k=0.0,
+        w0=-1.0,
+        wa=0.0,
+    )
+
+    _check_baccoemu_baryons(cosmo, baryons)
```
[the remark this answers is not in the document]
nikosarcevic: The boost factor does not modify sigma8 itself.
In CCL the baryonic model multiplies the matter power spectrum by f(k,a), which changes the amplitude on nonlin scales but leaves the underlying linear normalization (either sigma8 or A_s  in the input cosmology) untouched.

So sigma8 remains whatever was set in the cosmology -- only the shape and small-scale amplitude of P(k,a) change after applying the baryonic correction.

benchmarks/test_baryons_arico21_benchmark.py
nikosarcevic: The boost factor doesn’t change sigma8 at all.
In CCL, sigma8 is fixed when the cosmology is initialized (even when using A_s), and the baryonic model only rescales Pk not the normalization.

I’m planning to add a small CCLX notebook to document this behaviour clearly, since it’s not widely known.

benchmarks/test_baryons_vandaalen19_benchmark.py
nikosarcevic: Yes — this is intentionally CCL vs CCL.

Since Van Daalen+19 doesn’t provide public code or tables, there’s no true external benchmark to compare against. The txt file is a frozen regression target generated once with a fixed cosmology, so this test simply checks that the implementation stays numerically stable over time.
I added a note at the top of the file explaining this clearly so it’s not mistaken for a physical benchmark in 175ce5a773734bd2226114c7ba3e59f257707b66

pyccl/tests/test_baryons_mead20.py
nikosarcevic: This tolerance is inherited from the existing test_nonlin_camb_power test, which already used rtol=3e-5 for CAMB+Mead2020 vs CCL

In local runs the max relative difference between CAMB’s nonlinear P(k) and CCL’s matter_power_spectrum="camb" with halofit_version="mead2020_feedback" is at the ~1e-5 level once we restrict to a safe k-range, so 3 times that  is just a small safety margin for interpoltaion / platform differences

I can try tightening it to 1e-5 and see if CI is stable; if that passes I’m happy to lower the tolerance

carlosggarcia:
@nikosarcevic I was having a look at this PR. Nothing thorough. I've got a question and a comment ;) 

Question: Are most of the changes just change of names? If so, do we really need them? Each change of code can introduce bugs so I would avoid messing around with things that work unless necessary.

Comment: Re the rebranding of baccoemu tests are wrong. The non-lin and lin modules are Pk dark matter only so they should keep the old names. Then, personally, I would keep the old name for the baryon one too, that points out easily at the emulator that is being used. (One can think of ordering them by baryons or by different emulators).

PR #1289, opened 2026-03-28
quality of life fix for pip users

James11222:
This pull request adds quality-of-life support for pip users by adding optional dependency groups in `pyccl`, making it easier for users to install only the packages they need for specific functionalities (e.g., Boltzmann codes, perturbation theory, emulators). It updates the documentation to explain these new installation options, improves error handling for missing optional dependencies, and sets the default transfer function based on installed packages.

**Existing Issue this Fixes:** when you install the base package via pip, Boltzmann codes are not installed but the default `transfer_function` in `ccl.Cosmology()` is `'boltzmann_camb'`. This creates a problem when you go to use the cosmology object in components of the halo model for example. Now with these changes, I've made the default behavior be that if the Boltzmann codes are not installed, `'eisenstein_hu'` is the default, otherwise camb is the default (I believe the default behavior in the conda installation). 

**Dependency management and installation:**
- Added optional dependency groups (`boltzmann`, `pt`, `emulators`, `full`) to `pyproject.toml` for easier installation of extra features.
- Updated installation instructions in both `README.md` and `readthedocs/source/installation.rst` to document the new extras and clarify default behaviors depending on installed packages. [[1]](diffhunk://#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R66-R79) [[2]](diffhunk://#diff-a45d557f15214f523edf28a824b5395ab83ef7f0c3edd35b31dec0832e726c3eR46-R62) [[3]](diffhunk://#diff-a45d557f15214f523edf28a824b5395ab83ef7f0c3edd35b31dec0832e726c3eR81-R82) [[4]](diffhunk://#diff-a45d557f15214f523edf28a824b5395ab83ef7f0c3edd35b31dec0832e726c3eR134-R135) [[5]](diffhunk://#diff-a45d557f15214f523edf28a824b5395ab83ef7f0c3edd35b31dec0832e726c3eR166-R167)
```bash
pip install pyccl[boltzmann]   # Boltzmann codes: CAMB, CLASS (classy), ISiTGR
pip install pyccl[pt]          # Perturbation theory: FAST-PT, velocileptors
pip install pyccl[emulators]   # Emulators: BaccoEmu, MiraTitan, Dark Emulator
pip install pyccl[full]        # All of the above
```

**Default behavior and configuration:**
[...]

PR #1290, opened 2026-04-14
Automatic wheels building for PyPI

ggalloni:
This PR adds automatic building and publishing of wheels for Ubuntu and MacOS over the supported range of Python.

The reason for this PR is found in SOLikeT, where we depend on the PyPI version of `pyccl`, which, at this point, is dragging us down in terms of `numpy` (<2) and Python versions.

Of course, this PR also brings automatic updates on PyPI as a bonus feature.

The workflow to produce wheels took ~13 minutes and should hopefully cover the majority of needed builds.
If more are needed, they can eventually be added.

Off-topic: I've also added to the optional dependencies all the missing packages that were breaking (some) tests if not present.

EDIT: I've removed the extra dependencies just mentioned, as that is fixed in a better way by #1289.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,35 +1,85 @@
-name: Upload to PyPI
+name: Build and publish to PyPI
 
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 jobs:
-  Upload:
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      id-token: write  # Required for trusted publishing
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            cibw_archs: x86_64
+          - os: macos-latest
+            cibw_archs: arm64
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v4
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
 
-    - name: Setup python
-      uses: actions/setup-python@v5
-      with:
-        python-version: '3.10'
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v3.3.1
+        env:
+          CIBW_ARCHS: ${{ matrix.cibw_archs }}
 
-    - name: Install build dependencies
-      run: |
-        python -m pip install --upgrade pip
-        python -m pip install build
+      - name: Upload wheels
+        uses: actions/upload-artifact@v4
+        with:
+          name: cibw-wheels-${{ matrix.os }}-${{ matrix.cibw_archs }}
+          path: ./wheelhouse/*.whl
 
-    - name: Build source tarball only
-      run: |
-        python -m build --sdist
+  build_sdist:
+    name: Build source distribution
+    runs-on: ubuntu-latest
 
-    - name: Publish to PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.PYPI_PYCCL_UPLOAD }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Install build dependencies
+        run: python -m pip install --upgrade pip build
+
+      - name: Build sdist
+        run: python -m build --sdist
+
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: cibw-sdist
+          path: dist/*.tar.gz
+
+  publish:
+    name: Publish to PyPI
+    needs: [build_wheels, build_sdist]
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release'
+    permissions:
+      contents: read
+      id-token: write
+
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+        with:
+          pattern: cibw-*
+          path: dist
+          merge-multiple: true
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_PYCCL_UPLOAD }}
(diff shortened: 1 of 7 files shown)


.github/workflows/publish.yml
```diff
@@ -1,35 +1,85 @@
...
+
+      - name: Build sdist
+        run: python -m build --sdist
+
+      - name: Upload sdist
+        uses: actions/upload-artifact@v4
+        with:
+          name: cibw-sdist
+          path: dist/*.tar.gz
+
```
carlosggarcia, line 64: Can you add a block for test.pypi? Just to make sure the release doesn't break the official packages.

This was suggested by copilot.

```
  publish-testpypi:
    name: Publish to TestPyPI
    needs: [build_wheels, build_sdist]
    runs-on: ubuntu-latest
    if: github.event_name == 'workflow_dispatch'   # manual trigger only
    environment:
      name: testpypi
      url: https://test.pypi.org/p/pyccl
    permissions:
      contents: read
      id-token: write

    steps:
      - name: Download all artifacts
        uses: actions/download-artifact@v4
        with:
          pattern: cibw-*
          path: dist
          merge-multiple: true

      - name: Publish to TestPyPI
        uses: pypa/gh-action-pypi-publish@release/v1
        with:
          repository-url: https://test.pypi.org/legacy/
```

@damonge, @c-d-leonard I think we need your intervention to set up the identity provider in test-pypi https://test.pypi.org/project/pyccl/. Also, it might be good if you add desc-ccl as maintainer (which I believe is what we use to publish to pypi?).

.github/workflows/publish.yml
```diff
@@ -1,35 +1,85 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: cibw-*
+          path: dist
+          merge-multiple: true
+
+      - name: Publish to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          password: ${{ secrets.PYPI_PYCCL_UPLOAD }}
```
carlosggarcia: I believe this is not needed if we use the trusted connection https://docs.github.com/en/actions/how-tos/secure-your-work/security-harden-deployments/oidc-in-pypi.

.github/workflows/publish.yml
```diff
@@ -1,35 +1,85 @@
...
+
+  publish:
+    name: Publish to PyPI
+    needs: [build_wheels, build_sdist]
     runs-on: ubuntu-latest
+    if: github.event_name == 'release'
     permissions:
       contents: read
-      id-token: write  # Required for trusted publishing
+      id-token: write
```
carlosggarcia, line 101: Here, copilot suggested to add an `environment` section. I am not familiar enough with pypi publishing to know if we should pay attention to it or not.
```
  environment:
    name: pypi
    url: https://pypi.org/p/pyccl
```

pyproject.toml
```diff
@@ -25,7 +25,7 @@ classifiers = [
     "Topic :: Scientific/Engineering :: Physics",
 ]
 dependencies = [
-    "numpy",
+    "numpy>=2.0",
```
carlosggarcia: Do you know what package is pulling `numpy<2`? Because we cannot enforce it if we keep compatibility with all the packages we say we support.

carlosggarcia (changes requested):
Thanks for the PR! I think it's a great addition.

I left some comments. In particular, before using the `publish` action in this PR, I'd like to test it in test.pypi to avoid issues like the one in #1284.

.github/workflows/publish.yml
ggalloni: Added, as you say, also in the next comments, I am assuming that trusted publishing is set up. So maybe some work is required by maintainers to fix that, if that is not true.

.github/workflows/publish.yml
ggalloni: Removed :+1:

.github/workflows/publish.yml
ggalloni: Yeah, that is required by trusted publishing, I've added that.

pyproject.toml
ggalloni: Mmm, I think this is a leftover of an early test I was doing to ensure wheels were built against numpy 2. I don't think that is necessary now, so I removed the pin.

carlosggarcia:
@ggalloni  Thanks for the changes! I got access to the pypi and test.pypi and started working on setting up the trusted publisher. Doing so, I've discovered I have to set up the environment to publish, which I am not very sure how to do correctly right now. I've discovered this morning that the testpypi API credentials were already in GH (this `TEST_PYPI_PYCCL_UPLOAD`), so it might be faster if we use the old API credentials approach.

Would you mind reverting the trusted publisher set up and go back to using passwords? Sorry about the back and forth.

ggalloni:
Hi @carlosggarcia,  before reverting, let's give trusted publishing a try, as it should be much easier than anything else, and the standard procedure for PyPI publishing.

I did it yesterday for another repo, so the procedure is clearer to me too: on the side of pypi/testpypi, it is sufficient to specify the metadata of the repo and most importantly the name of the workflow used (publish.yml), and an environment name matching the one set by the workflow itself (pypi and testpypi, respectively). That's it, below I attach the screenshot for the testpypi case.

image


On the side of the repo, you just need to enter Settings > Environments and create said environments (pypi/testpypi). No need to set and manage secrets and such. Screenshot of the view on my repo below.

image

carlosggarcia (approved):
Okay. I've configured the environments and pypi/test.pypi. Let's see if it works.

carlosggarcia:
Ok. The problem now is that `velocileptors` is pulling `scipy-1.17`. I believe if you copy `scipy<1.14` in `.github/environment.yml` to the pip section, it will respect the constraint. Could you do that? I don't think I can push to your fork.

If that doesn't work, copilot suggests creating a "constraint" file and modify the `ci.yml` with:
```
- name: Install CCL
  run: |
    PIP_CONSTRAINT=.github/test-constraints.txt pip install -v -e .
```

carlosggarcia:
@ggalloni thanks for the work! I was going to do a PR for dark emulator and it seems that they already fixed this issue in master. Instead of using old scipy, I believe it'd be better to install, for the test, the `main` branch.

This is the issue: https://github.com/DarkQuestCosmology/dark_emulator_public/issues/28

carlosggarcia:
I'll merge this and solve the issue with `dark_emulator` in another PR. Thanks for the work!!

ggalloni:
You're welcome and thank you for the support! 😊

PR #1292, opened 2026-05-11
Cell memory fix

rreischke:
I think there are some potential memory leakages, especially if called in parallel. In particular I fixed:
- `a_tmp` and `b_tmp` were declared without initialisation 
-  `forward_plan` and `reverse_plan` were created unconditionally, they are also destroyed unconditionally now
-  If `ccl_sigma8` fails, the first `psp_out` was never freed.


carlosggarcia (approved):
LGTM!

PR #1293, opened 2026-05-11
Non-Limber memory fix

paulrogozenski:
Tracer collections allocated wihtin the FKEM non-Limber code are not freed, leading to out-of-memory errors when running chains. This patch adds the necessary calls to free these structures.


carlosggarcia (approved):
LGTM. When the last test passes, feel free to merge! Thanks for the work!

carlosggarcia:
@paulrogozenski can you remove this constraint? https://github.com/LSSTDESC/CCL/blob/b9aa723431f921777fdd031d1bc3546b935b6152/pyproject.toml#L30 so the line reads `scipy` instead of `scipy<1.14`. I don't think we should enforce it at CCL level, just for the tests.

PR #1296, opened 2026-05-20
Curved-sky kernels and bin-averaging for angular correlation functions

DhayaaAnbajagane:
CCL currently provides FFTLog, Legendre, and Bessel routines to go from C_ells to two-point angular correlations for spin-0 and spin-2 fields. The Legendre method is the only accurate one for wide-angle curved-sky predictions, where the latter very much matters for DESC analyses. However, this method is currently not implemented for weak lensing, and the one implemented for GG-lensing is quite slow.

This PR does the following:

1. Vectorize the legendre polynomial calculation in the GGL curved-sky kernel (the previous run was drastically slower due to GSL's recurvsive calculation of the polynomials) by using the array method we'd already been using for wtheta.
2. Implement the curved-sky kernel for weak lensing (xi_plus and xi_minus) based on: https://arxiv.org/abs/astro-ph/9609149
3. Implement bin-averaging by integrating the polynomials, following https://arxiv.org/abs/2012.08568. This is done for wtheta, GGL, and cosmic shear (NN, NG, and GG+/GG- in CCL's language).

==============

The API needed some thinking since now we have a theta_min and theta_max. In order to be as non-intrusive as possible, we currently keep the `correlation()` function but simply add a new input `theta_max = None`. When it is not None, it is treated as the upper edge of the theta bin, and the existing `theta` arg is treated as the lower edge of the bin. If `theta_max` is not passed, then everything works as usual.

The unit tests are still in prep with help from @junzhou502 , but wanted to start this PR so I can at least check if the API is acceptable before I move further.


DhayaaAnbajagane:
With significant help from @junzhou502 , I've now included a benchmark against COCOA, which shows that our curved sky implementation agrees within 1% of COCOA's for various calculations of cosmic shear, GGL, and wtheta. This is the main benchmark we had in mind, and has now been completed on the CCL side. All other standard tests have passed as well.

From my vantage, this PR is now ready for review! @carlosggarcia

PR #1298, opened 2026-06-22
[WIP] Completing pgi and pii to one loop order

CarterDW:
We add relevant second order contributions to pii as well as all necessary nonlinear galaxy bias terms to pgi in ept.py to complete both tracers to one loop order. These terms are calculated in FAST-PT and use the most up to date version of that code (Note: as of the time of writing, fastpt currently has a bug due to a deprecated feature that should be resolved shortly).

This PR Adds:

- in ept.py, updated pgi and pii to be complete to one loop order
- in tracers.py, added additional ak and at bias prefactors for higher order IA contributions (the derivative term and velocity shear contributions)
- updating unit tests in test_ept_power.py to properly validate 
- updated benchmark code to properly account for new terms (regenerated benchmarks), set T_CMB to the same in both files to ensure same underlying cosmology for benchmark
- Passes flake8 on edited files

Need to do before pulled in:

- currently passes pytest, need ensure it passes all other required merge tests


pyccl/nl_pt/ept.py
```diff
@@ -45,7 +51,7 @@ class EulerianPTCalculator(CCLAutoRepr):
 
     .. math::
         s^I_{ij}=c_1\\,s_{ij}+c_2(s_{ik}s_{jk}-s^2\\delta_{ik}/3)
-        +c_\\delta\\,\\delta\\,s_{ij}
+        +c_\\delta\\,\\delta\\,s_{ij + c_k\\k^2\\,s_{ij}}
```
nikosarcevic: Could we fix the IA expansion in the class docstring? The new term seems to have broken the LaTeX:
+c_\delta\,\delta\,s_{ij + c_k\k^2\,s_{ij}}
I think this should close the s_{ij} term before adding the c_k term. Also, since the implementation now includes both ck and ct, it would be helpful for the docstring to explicitly define both operators

pyccl/nl_pt/ept.py
```diff
@@ -189,8 +218,11 @@ def __init__(self, *, with_NC=False, with_IA=False,
             raise ValueError(f"Unknown P(k) prescription {b1_pk_kind}")
         if bk2_pk_kind not in ['linear', 'nonlinear', 'pt']:
             raise ValueError(f"Unknown P(k) prescription {bk2_pk_kind}")
+        if ak2_pk_kind not in ['linear', 'nonlinear', 'pt']:
+            raise ValueError(f"Unknown P(k) prescription in {ak2_pk_kind}")
         self.b1_pk_kind = b1_pk_kind
         self.bk2_pk_kind = bk2_pk_kind
+        self.ak2_pk_kind = ak2_pk_kind
         if (self.b1_pk_kind == 'pt') or (self.bk2_pk_kind == 'pt'):
```
nikosarcevic: i thnik a#`ak2_pk_kind == "pt" `should also set `self.with_matter_1loop = True `

right now only `b1_pk_kind == "pt"` or `bk2_pk_kind == "pt" `do this, but the ak2 PT branch later uses
`pka += self._g4T * self.one_loop_dd[0]`

so basically if someone sets with_matter_1loop=False and ak2_pk_kind="pt", self.one_loop_dd may not exist.
perhaps do smth like:
```
if (
    self.b1_pk_kind == "pt"
    or self.bk2_pk_kind == "pt"
    or self.ak2_pk_kind == "pt"
):
    self.with_matter_1loop = True
```

nikosarcevic: also check for this downstream  and also upstream

pyccl/nl_pt/ept.py
```diff
@@ -177,7 +189,24 @@ def __init__(self, *, with_NC=False, with_IA=False,
...
+            raise ImportError("Your attempted import of FAST-PT has failed. "
+                              "You either dont have fast-pt installed"
+                              "or have the wrong version."
+                              "Try running pip install fast-pt or conda "
+                              " install fast-pt, then try again")
+
+        # Verify the installed FAST-PT exposes the routines CCL needs.
+        required = ['one_loop_dd', 'one_loop_dd_bias_b3nl', 'IA_ta', 'IA_tt',
+                    'IA_mix', 'IA_ct', 'gI_ta', 'gI_tt', 'gI_ct']
+        missing = [m for m in required if not hasattr(fpt.FASTPT, m)]
```
nikosarcevic, line 203: if usefptk=True, the code later calls self.pt.IA_der(**kw), but IA_der is not included in the required FAST-PT method list. 

Could we add it conditionally when usefptk=True? Otherwise an incompatible FAST-PT version would pass the upfront check and fail later with an AttributeError.

my suggeston is smth like
```
required = [
    "one_loop_dd",
    "one_loop_dd_bias_b3nl",
    "IA_ta",
    "IA_tt",
    "IA_mix",
    "IA_ct",
    "gI_ta",
    "gI_tt",
    "gI_ct",
]
if usefptk:
    required.append("IA_der")
```

pyccl/nl_pt/ept.py
```diff
@@ -258,6 +290,21 @@ def reshape_fastpt(tupl):
             reshape_fastpt(self.ia_tt)
             self.ia_mix = self.pt.IA_mix(**kw)
             reshape_fastpt(self.ia_mix)
+            self.ia_ct = self.pt.IA_ct(**kw)
+            reshape_fastpt(self.ia_ct)
+            if (self.ufpt):
+                self.ia_der = self.pt.IA_der(**kw)
+                reshape_fastpt(self.ia_der)
+            self.ia_ct = self.pt.IA_ct(**kw)
+            reshape_fastpt(self.ia_ct)
```
nikosarcevic: hmmmm looks like IA_ct is computed twice here=? 

```
self.ia_ct = self.pt.IA_ct(**kw)
reshape_fastpt(self.ia_ct)
...
self.ia_ct = self.pt.IA_ct(**kw)
reshape_fastpt(self.ia_ct)
```
we can remove one of these calls

pyccl/nl_pt/ept.py
```diff
@@ -129,6 +138,8 @@ class EulerianPTCalculator(CCLAutoRepr):
              documentation for more details.
         sub_lowk (:obj:`bool`): if ``True``, the small-scale white noise
              contribution to some of the terms will be subtracted.
+        usefptk (::obj:`bool`):) if `True``, will use the FASTPT IA k2
+        term instead of the CCL IA k2 term
     """
     __repr_attrs__ = __eq_attrs__ = ('with_NC', 'with_IA', 'with_matter_1loop',
                                      'k_s', 'a_s', 'exp_cutoff', 'b1_pk_kind',
```
nikosarcevic, line 144: snice this adds new behaviour changing options, should ak2_pk_kind and the usefptk choice be included in `__repr_attrs__ / __eq_attrs__`?

 Otherwise two calculators with different deriv term prescriptions could display or compare as if they are equivalent
 so smth like
```
    "with_NC",
    "with_IA",
    "with_matter_1loop",
    "k_s",
    "a_s",
    "exp_cutoff",
    "b1_pk_kind",
    "bk2_pk_kind",
    "ak2_pk_kind",
    "ufpt",
    "fastpt_par",
)
```

pyccl/nl_pt/ept.py
```diff
@@ -129,6 +138,8 @@ class EulerianPTCalculator(CCLAutoRepr):
              documentation for more details.
         sub_lowk (:obj:`bool`): if ``True``, the small-scale white noise
              contribution to some of the terms will be subtracted.
+        usefptk (::obj:`bool`):) if `True``, will use the FASTPT IA k2
+        term instead of the CCL IA k2 term
```
nikosarcevic: tiny naming suggestion: could we use a more explicit name than usefptk / ufpt? Something like use_fastpt_ia_derivative or use_fastpt_ia_k2 would make the API easier to understand, especially because this is a new public option

pyccl/nl_pt/ept.py
```diff
@@ -11,19 +11,25 @@
...
-    'bs:bk2': 'zero', 'bs:c1': 'zero', 'bs:c2': 'zero',
-    'bs:cdelta': 'zero', 'bk2:bk2': 'zero', 'bk2:c1': 'zero',
+    'bs:bk2': 'zero', 'bs:c1': 'bs:c1', 'bs:c2': 'bs:c2',
+    'bs:cdelta': 'bs:cdelta', 'bk2:bk2': 'zero', 'bk2:c1': 'm:bk2',
     'bk2:c2': 'zero', 'bk2:cdelta': 'zero', 'c1:c1': 'm:m',
-    'c1:c2': 'm:c2', 'c1:cdelta': 'm:cdelta', 'c2:c2': 'c2:c2',
-    'c2:cdelta': 'c2:cdelta', 'cdelta:cdelta': 'cdelta:cdelta'}
+    'c1:c2': 'm:c2', 'c1:cdelta': 'm:cdelta', 'c1:ck': 'm:ck',
+    'c2:c2': 'c2:c2', 'c2:cdelta': 'c2:cdelta',
+    'cdelta:cdelta': 'cdelta:cdelta', 'ck:ck': 'zero', 'm:ct': 'm:ct',
```
nikosarcevic, line 27: the alias table now promotes several number-counts × IA combinations that were previously zero, for example b2:c1, bs:c1, b2:c2, b2:cdelta, and the new ct/ck combinations. 

Could we add targeted tests for these aliases through get_pk2d_template? That would make it safer to verify that each new alias maps to the intended FAST-PT template and normalization factor

pyccl/nl_pt/ept.py
```diff
@@ -377,10 +454,25 @@ def _get_pgi(self, trg, tri):
...
         cd = tri.cdelta(self.z_s)
+        ck = tri.ck(self.z_s)
+        ct = tri.ct(self.z_s)
+
+        pgi = (b1[:, None]*(c1[:, None] * Pd1d1 +
+                            (self._g4*cd)[:, None] * (a00e + c00e) +
+                            (self._g4*c2)[:, None] * (a0e2 + b0e2) +
+                            (self._g4*ct)[:, None] * d0te +
+                            ck[:, None] * Pak2) +
+               0.5*b2[:, None]*((self._g4*c1)[:, None]*d2e +
```
nikosarcevic, line 455: the warning still says this assumes linear galaxy bias for galaxy-IA cross-correlations, but _get_pgi now includes terms proportional to b2, bs, b3nl, and bk2. 

i think the warning be updated to describe the new behavior more precisely. rn it sounds like those terms are ignored, but they are included here

pyccl/nl_pt/tracers.py
```diff
@@ -10,8 +10,8 @@
 from ..pyutils import _check_array_params
 
 
-def translate_IA_norm(cosmo, *, z, a1=1.0, a1delta=None, a2=None,
-                      Om_m2_for_c2=False, Om_m_fid=0.3):
+def translate_IA_norm(cosmo, *, z, a1=1.0, a1delta=None, a2=None, ak=None,
```
nikosarcevic, line 13: Could we add `ak` and `at` to the translate_IA_norm docstring args? they are now public inputs in the func signature, but the docstring still only documents a1, a1delta, and a2

so add smth like:
```
        ak (:obj:`float` or `array`): IA derivative amplitude at input z values.
        at (:obj:`float` or `array`): IA tensor/source amplitude at input z values.
```
(im not sure if this is completely correct wording)

pyccl/nl_pt/tracers.py
nikosarcevic, line 36: the return signature changed from 3 values to 5 values, but the docstring still documents only c1, c1delta, and c2.  we need to update the `Returns` section to include ck and ct?


so smth like
```
    Returns:
        Tuple of IA bias parameters

        - c1 (:obj:`float` or `array`): IA :math:`C_1` at input z values.
        - c1delta (:obj:`float` or `array`): IA :math:`C_{1\delta}` at
          input z values.
        - c2 (:obj:`float` or `array`): IA :math:`C_2` at input z values.
        - ck (:obj:`float` or `array`): IA derivative bias at input z values.
        - ct (:obj:`float` or `array`): Additional IA bias at input z values.
```

pyccl/nl_pt/tracers.py
```diff
@@ -54,8 +55,12 @@ def translate_IA_norm(cosmo, *, z, a1=1.0, a1delta=None, a2=None,
             c2 = a2*5*5e-14*rho_crit*Om_m**2/(Om_m_fid*gz**2)
         else:  # DES convention
             c2 = a2*5*5e-14*rho_crit*Om_m/(gz**2)
+    if ak is not None:
+        ck = -1*ak*(knorm**2)*5e-14*rho_crit*Om_m/gz
+    if at is not None:
+        ct = at*5e-14*rho_crit*Om_m/gz
 
-    return c1, c1delta, c2
+    return c1, c1delta, c2, ck, ct
```
nikosarcevic, line 70: this changes translate_IA_norm from returning 3 values to returning 5 values

so `c1, cdelta, c2 = translate_IA_norm(...)`
will now fail with ValueError: too many values to unpack i think

 If we need to preserve compatibility, maybe we should either add a flag for extended outputs or introduce a new helper for the extended IA normalization? need to consult @carlosggarcia  et al.

pyccl/nl_pt/tracers.py
```diff
@@ -40,8 +40,9 @@ def translate_IA_norm(cosmo, *, z, a1=1.0, a1delta=None, a2=None,
 
     Om_m = cosmo['Omega_m']
     rho_crit = physical_constants.RHO_CRITICAL
-    c1 = c1delta = c2 = None
+    c1 = c1delta = c2 = ck = ct = None
     gz = cosmo.growth_factor(1./(1+z))
+    knorm = 1  # Units of Mpc/h, normalizes units out of ck term
```
nikosarcevic: can we clarify the normalization of the ck term? Right now knorm = 1 is hard-coded with a units comment, but it is not exposed as an argument and it is not documented. Since ck multiplies a k^2 contribution later, it would be good to make the convention explicit: is this assuming k in h/Mpc, 1/Mpc, or matching FAST-PT’s internal units? CCL is always in Mpc never in Mpc/h. @carlosggarcia @elisachisari et al.

pyccl/nl_pt/tracers.py
nikosarcevic: also the sign and normalization conventions for ak and at? ak gets the same negative sign as a1, while at doesnt:
```
ck = -1*ak*(knorm**2)*5e-14*rho_crit*Om_m/gz
ct = at*5e-14*rho_crit*Om_m/gz
```
that may be correct, but it would be useful to state which convention this follows, especially because `translate_IA_norm `already references the Blazek/DES conventions.

pyccl/nl_pt/tracers.py
```diff
@@ -212,7 +217,7 @@ class PTIntrinsicAlignmentTracer(PTTracer):
     """
```
nikosarcevic, line 229: The PTIntrinsicAlignmentTracer docstring still describes only c1, c2, and cdelta, but the constructor now accepts ck and ct. Could we update the class description and Args section so the new IA terms are discoverable?
 so smth like 
```
ck (:obj:`float` or :obj:`tuple`): as above for the derivative
            alignment bias.
        ct (:obj:`float` or:obj:`tuple`): as above for the additional
            IA bias term.
```

pyccl/nl_pt/tracers.py
```diff
@@ -240,3 +249,15 @@ def cdelta(self):
...
+    def ck(self):
+        """Internal derivative bias function
+        """
+        return self.biases['ck']
+
+    @property
+    def ct(self):
+        """Internal velocity bias function
+        """
+        return self.biases['ct']
```
nikosarcevic, line 275: we sohuld make the ck and ct property docstrings more explicit and consistent with the terminology used in the calculator/docstring? In patricular, “velocity bias” for ct may be confusing unless that is the intended physical interpretation??
 smth like this
```
    @property
    def ck(self):
        """Internal derivative IA bias function."""
        return self.biases['ck']

    @property
    def ct(self):
        """Internal additional IA bias function."""
        return self.biases['ct']
```

nikosarcevic (commented):
hey , Carter, many thanks for starting this. I have some feedback for you so you can work on this when you find the time. Cheerio!

PR #1304, opened 2026-07-24
Fixes Issue1303: allows camb v2

carlosggarcia:
This PR fixes the bug introduced by the changes introduced by CAMB v2 in `set_for_lmax`. In order to keep the Pk linear, we need to not request lensing, which is now computed by default, by setting `lens_potential_accuracy=0`, as in CAMB v1. See [docs](https://camb.readthedocs.io/en/latest/model.html#camb.model.CAMBparams.set_for_lmax) and [GH diff](https://github.com/cmbant/CAMB/compare/1.6.6...2.0.0#diff-3bb9b02c0437dfe01d523940691a92ca0623718f7bbf43c80849c94a30d87d82R1032)

### DIFF
--- a/.github/environment.yml
+++ b/.github/environment.yml
@@ -15,7 +15,7 @@ dependencies:
   - pyyaml
   - numpy < 2.4  # fast-pt uses np.trapz, which is deleted in numpy 2.4
   - scipy < 1.14 # velocileptors uses functions removed in recent scipy versions
-  - camb < 2
+  - camb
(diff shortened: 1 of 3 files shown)


damonge (approved):

carlosggarcia:
I also had to set the general `AccuracyTarget = 0` to get the same level of accuracy as CAMB v1. This should make it pass the benchmarks

PR #1305, opened 2026-07-27
Bug in EPT calculator when assigning linear and SPT templates

damonge:
This is a bug reported originally by Alex Roskill (@AlexRoskill)

When the EPT calculator object is requested to use the linear power spectrum for the b1 term and the SPT power spectrum for the bk2 term, or vice-versa, an error in how arrays are handled leads to both power spectra being the SPT power spectrum. This PR fixes this.


carlosggarcia (approved):
LGTM

